Issue #2044, opened 2014-10-07
Deis publisher improvements


gust1n:
Hey! 
I thought I would create an issue before a PR so that we can discuss if you want this change or not. I've been keeping track of the new publisher service since it was first discussed and I've build one of my own for my company. Two things that bugs me with your approach are:
- The inability to unregister a service immediately when the container stops.
- Not handling multiple ports

Luckily I have solved both issues for me and I thought I would contribute my code if you want to:

The first issue is solved by keeping a simple in memory map of container id -> published keys. Since the only thing we get when container dies or stops it it's ID we can still safely remove all published keys for that container.

The second issue is solved by using the containerport as a sub key so the keys looks something like this:

`/deis/services/userservice/userservice-1/80 "1.2.3.4:49872"`
`/deis/services/userservice/userservice-1/3000 "1.2.3.4:49873"`

which for us worked really well both in our heads and in code. This works kind of the same way as docker env vars where you also state the port (like DB_PORT_5432_TCP_ADDR)

I could of course split this up in separate issues and PRs, but I wanted to know if this is interesting before I start working on it (my solution is not as generic as it has to be in a project like deis). Let me know what you think!

bacongobbler:
considering that #3069 covers the "support multiple ports" part of this and we have a TTL in place for when a container is stopped I'm going to create a new issue for the "instant stop" feature and close this one out. Thanks @gust1n :)

PR #2095, opened 2014-10-09
feat(Publisher): Adding support for publishing multiple service ports

gust1n:
As discussed in #2044.

This changed the behavior of the publisher into using the interal port as part of the query. Since we want to support multiple ports we somehow need to differentiate the ports but I'd rather not switch to posting JSON or something to keep the keys simple (as they are now).

This has worked great for us and actually makes sense. The query (in pseudo) for the clients will be for example "give me port 5432 of an instance of postgres" and the answer will be "1.2.3.4:49123" or whatever.

I hope I made my point clear about this approach. Will follow up with another PR that cleans up the logging and also deletes etcd keys on docker events when this is merged/rejected.

### DIFF
--- a/publisher/publisher.go
+++ b/publisher/publisher.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"log"
 	"os"
+	"path"
 	"regexp"
 	"strconv"
 	"time"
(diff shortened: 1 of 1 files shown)


deis-admin:

carmstrong:
closing, as this isn't mergeable as-is. I've filed a proposal in #3069.

Thanks for the start on this, @gust1n!

Issue #18, opened 2013-07-24
Application SSL support


gabrtv:
The controller currently supports HTTP and HTTPS cluster-wide by installing an SSL cert on the load balancer fronting Deis.  We need to provide a way of distributing SSL keys to the router for applications.

jparkerCAA:
Sorry, I've been off the grid for a long stretch.
@nathansamson let me explain my approach to some of these points.
 
- Wildcard certificates can be used by multiple applications
  perhaps I am oversimplifying but I figured that for the splat cert use case one would simply add the    same cert to each app that requires it. Perhaps I'm oversimplifying this but this actually happens to match my use case (the reason I'm adding this feature) and I'm ok with it. (at least as a first step towards some more sophisticated solution)

- mulit-domain same as above

- not every domain will have a cert. My assumption (perhaps a foolish one) was that a mult-domain ssl app would employ a single multi-domain certificate. I have not tested this like I have tested the wildcard but seems like it should work. Can you describe the scenario where one domain is secure and another is not for a single app? I suppose it shouldn't be too difficult to attach the cert to an app domain but I'd prefer to keep it as simple as possible.

jparkerCAA:
@nathansamson ah, I see your use case on the other issue

nathansamson:
* Wildcard certificates can be used by multiple applications.
* Multidomain. 
 Yes I agree in both cases, it should just be added by the different applications for their respective domains.
I just believe it is worth nothing in combination with the other requirements (mostly that not applications should be linked to certs, but rather domains)

I think the main thing that needs to change from your existing implementation in #2799 (although I didn't take an exact look) is to change the nginx configuration.
So rather than just adding all domains in one server block add them one by one... But I think this might get rather complicated in the case of multidomain & wildcard certificates.
Also it needs to be taken into account that some domains of an app won't have the certificate so should still listen on HTTP port 80 without redirect to HTTPs

Nathan

bacongobbler:
I've created an initial proposal at #2911 which follows https://devcenter.heroku.com/articles/ssl-endpoint. I'd really appreciate your input there!

PR #2570, opened 2014-11-23
feat(controller+router+client): application SSL support

ghost:
Adds Heroku inspired workflow for adding a certificate to an application.  ( Addresses https://github.com/deis/deis/issues/18 )
Please review and see what you think.

Notes:
For the router, I thought that the simplest thing for getting the certs to disk was to use a shell script to write confd templates for each cert pair found in etcd. It's a little bit meta that the confd generator is itself kicked off by confd but it's pretty straightforward.

With regard to the nginx server directives, I have apps with certs listen on both https and http. Would it make sense to have an option to 301 to ssl for cert enabled sites? Perhaps this is an app concern.

From the controller I'm only allowing one cert per app but I kept the Heroku command line convention of removing certs by CN. so `deis certs:remove mydomain.com` instead of `deis certs:remove`. I figured support for multiple certs are inevitable.


jparkerCAA:
@bacongobbler sorry, I was off the grid for a long while. Thanks.

PR #2607, opened 2014-11-28
fix(deisctl): prevent build on windows

Xe:
On windows the deisctl client will build, but it runs into a lot of problems at runtime due to limitations in fleetctl. It depends on the ssh-agent domain socket which cannot exist in windows as windows does not support the idea of domain sockets.

This will prevent issues at runtime at the expense of being able to build and run deisctl in windows.

### DIFF
--- a/deisctl/deisctl.go
+++ b/deisctl/deisctl.go
@@ -1,3 +1,5 @@
+// +build !windows
+
 package main
 
 import (


withinboredom:
Ha! I was about to open an issue about it not building in a cygwin environment, which does support domain sockets I believe...

Xe:
It doesn't matter. Even in Cygwin it doesn't support domain sockets purely. They are implemented as TCP sockets IIRC.

PR #2609, opened 2014-11-29
Proposal: Containerized jenkins-slaves

johanneswuerbach:
Containerized the jenkins slave to allow moving slaves easily between clouds / environments.

As I don't have NODE_NAME and NODE_SECRET I could only test whether the container builds and not the actual test running. Most steps are copied from https://github.com/deis/deis/blob/master/tests/bin/setup-node.sh

@mboersma I'm not sure how you imagined the AWS CI setup, but I thought having a CoreOS cluster and jenkins-slaves deployed using fleet is the best way. If you like this approach I could look into the missing components.

TODO:
- [ ] Link postgres or leave it inside the container?
- [ ] Start dev-registry or link one from outside?
- [ ] Move docker / jenkins / vagrant data from copy on write

### DIFF
--- a/tests/jenkins-slave/Dockerfile
+++ b/tests/jenkins-slave/Dockerfile
@@ -1,0 +1,53 @@
+FROM ubuntu-debootstrap:14.04
+
+ENV DEBIAN_FRONTEND noninteractive
+
+# configure locale
+RUN locale-gen en_US.UTF-8 && \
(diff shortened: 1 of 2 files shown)


carmstrong:
Closing, as this isn't mergeable as-is. Tracking the proposal in #3071.

PR #2647, opened 2014-12-04
WIP: ref(store): remove docker volumes for ceph components

aledbf:
This removes the use of docker volumes for ceph components adding a new drive dedicated to ceph.
It will format the new drive with `btrfs` and mount it in `/media/ceph`

- [X] provision new drive
- [ ] remove drive on vagrant destroy
- [ ] benchmark docker volumes vs dedicated drive


carmstrong:
Thanks for this research, @aledbf! I think the solution here is to enable NOCOW generally in our bin/boot scripts for store. I filed #2972 for this, and will get to it for the next release.

PR #2717, opened 2014-12-10
feat(router): add PROXY protocol support

Blystad:
This pull requests adds support for the PROXY Protocol on port 8080, if enabled through etcd. It it an alternative to #2700.

The [PROXY Protocol](http://www.haproxy.org/download/1.5/doc/proxy-protocol.txt) provides a way to get the clients IP address through a TCP only proxy. For my use case, Amazon ELB was [lacking websockets support](https://forums.aws.amazon.com/thread.jspa?threadID=84606) and it doesn't look like it will get it in any near future. Amazon ELB, however, supports the Proxy Protocol, and by setting the listener into TCP mode, Websockets will work fine.

The reason why I exposed port 8080 as an additional port, was because of how Amazon ELB health checks work. When sending a request to a HTTP server with Proxy protocol enabled, it first expects to get:
`PROXY TCP4 255.255.255.255. ....\r\n`, and then the request (`GET /health-check HTTP1.1\r...`). 
If it doesn't get this, it will return a `400 Client error`.

## Steps to properly test it
To test this with Amazon ELB, a couple of extra steps must be done (`X` will vary, names can vary):
**1.** [Turn on proxy protocol support through the AWS CLI](http://docs.aws.amazon.com/ElasticLoadBalancing/latest/DeveloperGuide/enable-proxy-protocol.html) (No GUI support):
**1.1** Create custom LoadBalancerPolicy: 
```
aws elb create-load-balancer-policy \
--load-balancer-name deis-DeisWebE-X \
--policy-name EnableProxyProtocol \
--policy-type-name ProxyProtocolPolicyType \
--policy-attributes AttributeName=ProxyProtocol,AttributeValue=true
```

**1.2** Associate this with instance port 8080: 
```
aws elb set-load-balancer-policies-for-backend-server \
--load-balancer-name deis-DeisWebEL-X \
--instance-port 8080 \
--policy-names EnableProxyProtocol
```

**2.** Enable ProxyProtocol through ETCD: `etcdctl set /deis/router/proxyProtocol 1`
**3.** Change Aws ELB *Load Balancer Protocol* from `HTTP` to `TCP`
**4.** Change Aws ELB *Health Check* from `HTTP` to `TCP`
**5.** Test that everything works.

To test that everything works, I've created a simple HTTP server in Go which just writes out the HTTP Headers, like so:
```
GET / HTTP/1.1
Host: http-headers.dev.mydeiscluster.com
User-Agent: Mozilla/5.0 (Windows NT 6.1; WOW64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/39.0.2171.71 Safari/537.36
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,*/*;q=0.8
Accept-Encoding: gzip, deflate, sdch
Accept-Language: no,en-US;q=0.8,en;q=0.6,nb;q=0.4,en-GB;q=0.2
Connection: close
X-Forwarded-For: 85.1.11.11
```
It's available here: https://github.com/Blystad/http-server-header-dump or as a `docker/deis pull blystad/http-server-header-dump`

With PROXY Protocol support, X-Forwarded-For will show the correct IP. Without it, you'll see a internal IP (10.XX or similar).


Blystad:
@bacongobbler Had any chance to test this yet?

I've rebased it on top of master now, for easier merging, as I saw there was some changes to router building.

Blystad:
I've added a script now to make testing (and using) this feature easier. As explained in the commit message, I've created a script instead of doing this through CloudFormation since it also needs to set a property through `deisctl` or `deis-router` won't accept incoming requests from Amazon ELB

contrib/ec2/setup-proxy-protocol.sh
```diff
@@ -0,0 +1,62 @@
+#!/bin/bash -e
+#
+# Usage: setup-proxy-protocol <cloudformation stack name>
```
bacongobbler: Angle brackets implies that its mandatory. Could you change it to `setup-proxy-protocol.sh [STACK]`, please?

contrib/ec2/setup-proxy-protocol.sh
```diff
@@ -0,0 +1,62 @@
...
+
+aws elb set-load-balancer-policies-for-backend-server \
+	--load-balancer-name ${LOAD_BALANCER_NAME} \
+	--instance-port 80 \
+	--policy-names EnableProxyProtocol
+
+echo "Setup TCP Health check"
+aws elb configure-health-check \
+	--load-balancer-name ${LOAD_BALANCER_NAME} \
+	--health-check Target=TCP:80,Interval=15,Timeout=5,UnhealthyThreshold=2,HealthyThreshold=4
```
bacongobbler: missing newline

contrib/ec2/setup-proxy-protocol.sh
```diff
@@ -0,0 +1,62 @@
...
+echo "Changing port 80 from HTTP -> TCP"
+aws elb delete-load-balancer-listeners \
+	--load-balancer-name ${LOAD_BALANCER_NAME} \
+	--load-balancer-ports 80
+
+aws elb create-load-balancer-listeners \
+	--load-balancer-name ${LOAD_BALANCER_NAME} \
+	--listeners Protocol=TCP,LoadBalancerPort=80,InstanceProtocol=TCP,InstancePort=80
+
+SSL_CERTIFICATE=$(aws elb describe-load-balancers --load-balancer-names ${LOAD_BALANCER_NAME} | grep -m 1 "SSLCertificateId" | sed 's/ *//g' | cut -d ':' -f2-10 | cut -d '"' -f2)
```
bacongobbler: wouldn't it be easier to check for a https load balancer listener than checking for the certificate?

docs/using_deis/index.rst
```diff
@@ -19,5 +19,6 @@ Using Deis
     using-buildpacks
     using-dockerfiles
     using-docker-images
+    using-websockets
```
bacongobbler: This should be under "Managing Deis". "Using Deis" is reserved for using the platform.

contrib/ec2/setup-proxy-protocol.sh
Blystad: Maybe, but I still need the SSL_CERTIFICATE in the next part.

It's not the prettiest solution. The script I created for the company uses [jq](https://github.com/stedolan/jq) but I didn't want to add extra dependencies to Deis.

contrib/ec2/setup-proxy-protocol.sh
Blystad: Fixed

bacongobbler:
I can see the merits for this, and since this is now on the default ports for the load balancer/all covered in `awscli` this looks much better. 

You may need to fix up c999975's commit message. It's missing a colon :)

code LGTM! :blue_heart:

Blystad:
Fixed commit message.

What do you mean by "everything still work for HTTP-based applications"? 

I've got a router built on this running on our Deis production cluster and both our HTTP and Websocket applications work perfectly fine.

bacongobbler:
> I've got a router built on this running on our Deis production cluster and both our HTTP and Websocket applications work perfectly fine.

That's what I meant. I was just concerned that if users enabled proxy protocol support for their websocket applications, HTTP apps would no longer work.

cc @carmstrong @mboersma does this approach look good to you? If so then I can take a look at trying this out for a spin.

Blystad:
So, yesterday I started to think a bit more about this PR after seeing #2976 (Add enforceHttp support), and I found a problem with this PR...

```
GET / HTTP/1.1
Host: http-headers.dev.deiscluster.com
User-Agent: Mozilla/5.0 (Windows NT 6.1; WOW64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/40.0.2214.93 Safari/537.36
Connection: close
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,*/*;q=0.8
Accept-Encoding: gzip, deflate, sdch
Accept-Language: no,en-US;q=0.8,en;q=0.6,nb;q=0.4,en-GB;q=0.2
Cache-Control: max-age=0
Connection: close
Cookie: [Cookies]
X-Forwarded-For: 84.213.0.0, 10.0.0.83
X-Forwarded-Port: 443
X-Forwarded-Proto: https
```

This is a domain running on https, and as we see, `X-Forwarded-Proto` is exposed saying that the connection is over https, which is awesome.

However, with Proxy protocol enabled, Nginx has no way of actually knowing that the original request was over HTTPS (since that part isn't exposed in the Proxy protocol). Some frameworks (Spring for instance) include ways to redirect HTTP -> HTTPS, and detect X-Forwarded-Proto and use that to enable HSTS, activate secure cookies and so forth.

Now, a workaround to that would be to expose a extra port, or to make nginx listen to 443 but with no SSL certificates (since I want that to terminate at the AWS ELB) and then make nginx send the `X-Forwarded-Proto` headers. I fancy neither of these solutions to be honest.

I think my preferred workaround for the whole websocket issue is to create a seperate ELB for the websocket endpoint, expose the websocket endpoint on a seperate subdomain and then keep on dreaming Amazon fixes ELB WebSocket support instead of releasing new features every month. 

To make this ELB follow the autoscaling it needs to be created at the same time the autoscaling group is created. I.e. as part of the `cloudformation json`.

This is how the **ELB feature matrix** looks: 


Now, based on the Matrix one might think - _the best solution for now is two ELBs, one in HTTP mode and one in TCP mode with Proxy protocol enabled,_ but alas it's not that easy...If NGINX is set to serve Proxy requests then it will do that. Any non-proxy requests would be denied.

This too can be solved, of course, but it will increase the complexity of the router and require exposing more ports.

My **recommendation** is to drop this PR and then I'll create a new one with a `cloudformation.json` with two ELBs (configurable of course), one in TCP mode and one in HTTP mode, as well as documentation on managing websockets.

johanneswuerbach:
According to http://www.haproxy.org/download/1.5/doc/proxy-protocol.txt, the PROXY Header also includes the port the client initially connected to? Wouldn't this work to determine the whether he was using 80 (http) or 443 (https)?

Blystad:
You're right....Hmm, I wonder why nginx didn't expose that then...

Ah, found it. If we look at the commit when `proxy_protocol` was introduced in nginx: [http://trac.nginx.org/nginx/changeset/3a72b1805c52c84a5567dcfa4562f44722bede40/nginx](http://trac.nginx.org/nginx/changeset/3a72b1805c52c84a5567dcfa4562f44722bede40/nginx) Then we see no reference to any port number.

If we look at the way the proxy protocol is parsed: [http://lxr.nginx.org/source/src/core/ngx_proxy_protocol.c](http://lxr.nginx.org/source/src/core/ngx_proxy_protocol.c)

We see that it reads `PROXY`, and then it expects `TCP[4|6]` as the next value, and then it reads until the next space again. (`if (ch == ' ') { break; }`). E.g. with the following header:
`    PROXY TCP4 192.168.0.1 192.168.0.11 56324 443\r\n`
It stops reading after:
`    PROXY TCP4 192.168.0.1 `

And discards the rest, as it only cares about the first IP address.

Damn :confused:

johanneswuerbach:
Wouldn't [terminating SSL on the router](http://docs.deis.io/en/latest/managing_deis/ssl-endpoints/#installing-ssl-on-the-deis-routers) work when the ELB is configured to forward 80 & 443 in TCP mode without terminating SSL.

Blystad:
Probably yes. I can try to test that during the weekend.

bacongobbler:
> I think my preferred workaround for the whole websocket issue is to create a seperate ELB for the websocket endpoint, expose the websocket endpoint on a seperate subdomain and then keep on dreaming Amazon fixes ELB WebSocket support instead of releasing new features every month.

Yeah unfortunately this is partially out of our hands. The router fully supports websockets; it's just Amazon's ELB that does not (in a clean way, at least).

> This too can be solved, of course, but it will increase the complexity of the router and require exposing more ports.

Personally I'd rather not waste everyone's time by working on a solution like proxy_protocol which would become obsolete down the road, should Amazon's ELB eventually supports websockets. I'd also prefer not to drop in code to the router which only works on one provider and/or it's only used in small use cases (i.e. only when you're using ELB). I think that...

> My recommendation is to drop this PR and then I'll create a new one with a cloudformation.json with two ELBs (configurable of course), one in TCP mode and one in HTTP mode, as well as documentation on managing websockets.

... would be the most agreeable approach here.

carmstrong:
> ... would be the most agreeable approach here.

Agreed. @Blystad thank you so much for taking the time to work on this, and for writing such excellent documentation (with screenshots)! We look forward to reviewing the new PR.

PR #2772, opened 2014-12-18
DEIS-DNS - Act 1

PierreKircher:
pr 1 of 3

base system for dns using "X.deis.local" as localdomain

deisctl > 
changes to be installed as part of the platform

side note : jenkins-test probably fail due the unit-file not in deis/deis yet


carmstrong:
Thanks for this, @PierreKircher! We're going to open up a discussion about internal DNS discovery over in #3072 before we come back to the implementation.

PR #2774, opened 2014-12-18
feat(dns): provide an internal dns service (skydns) for apps and deis components

aledbf:
This PR provides:

**default domain**
- [x] uses /deis/platform/domain by default
- [x] fallback to deis.local

**custom dns server**
- [x] allow custom dns servers (/deis/platform/nameServers)
- [x] fallback to 8.8.8.8 and 8.8.4.4
  
**custom domain name**
`etcdctl set /deis/platform/domainName my-custom.domain.io`

**register every node in the cluster**
- [x] register the node at start
- [ ] remove the A record if the node is not available (I don't know if this is desirable)
- [x] replace /etc/resolv.conf
- [x] add the local node as dns server
- [x] use the local domain name in resolv.conf

**register the apps as hosts using A records**
- [x] change deis-publisher to register the app
- [x] support multiple instances of an app

**Notes:**
- the use of /deis/platform/domain is just for convenience
- skydns only supports 1 domain https://github.com/skynetservices/skydns/issues/47
- the access to the port 53 in every node should be protected (only the cluster members can use the dns service)
- using the firewall included in contrib the access to the port 53 is disabled

complements #2772


Blystad:
Any updates on this @aledbf? I'm really interested in this feature on our Deis deployment, as I have some services I don't want to expose to the world unless necessary, and also it would be faster connecting directly from `server -> server` then todays solution with `server -> aws elb -> deis router -> server`.

If you need any help testing I'll be glad to help.

aledbf:
@Blystad this is working. I'm waiting review and feedback from the deis team

aledbf:
ping @bacongobbler @mboersma @carmstrong

mboersma:
Thanks for the nudge @aledbf. We're testing a new Deis release today but we'll look at this as soon as possible.

carmstrong:
Thanks for the great work on this, @aledbf! We're going to open up a discussion about internal DNS discovery over in #3072 before we come back to the implementation.

PR #2799, opened 2014-12-23
feat(controller): teach controller about application certificates

bacongobbler:
supercedes #2570, rebased and all comments addressed.


nathansamson:
By looking at the API & Client this might be exactly what Heroku is doing and might actually support my use case...

The only thing I am not sure of, but might actually work if multiple certs (especially if not all domains have a cert) actually works in the router.

I don't really have the resources (both test cluster as time) wise to actually test this at the moment, but if @jparkerCAA or any of the maintainers could give this a go I think this might be ready to merge (or at least support my test case)


To outline my test case
~~~
deis certs:add <my_cert_for_common_name *.example.org>
deis domains:add www.example.org
deis domains:add www2.example.org

deis certs:add <my_cert_for_common_name something.example.info>
deis domains:add something.example.info

deis domains:add notssl.example.com
~~~

So when accessing www.example.org,  www2.example.org & something.example.info SSL should work. I believe it should be the apps responsability (by default) to redirect to HTTPs if needed.
When accessing notssl.example.com normal HTTP needs to be used.

Maybe later an option can be created that will (on a per app basis) redirect all HTTP to HTTPs traffic in the router to offload the app.

bacongobbler:
@nathansamson thanks for this. In the future, let's keep the ball rolling in #18 for this discussion.

PR #2822, opened 2014-12-30
feat(contrib/coreos): disable transparent huge pages

aledbf:
To test the execution, this should be the output:
```
$ more /sys/kernel/mm/transparent_hugepage/enabled
always madvise [never]
```


aledbf:
Jenkins, test this please.

carmstrong:
@aledbf Thank you for this! We discussed this and aren't convinced the change is necessary. Additionally, we're hesitant to make a system-wide tweak for a single Deis component.

PR #2824, opened 2014-12-30
add azure deis cluster creation tools and documentation

timfpark:


timfpark:
Hi Chris, 

I'm planning on doing something like contrib/gce/create-gce-user-data and doing all of the Azure related changes to user-data as part of that.  Does that sound right?

Tim

carmstrong:
> Does that sound right?

Yep! Exactly. Note that we'll also need to override one of the `write_files` entries for Azure:
```
  - path: /etc/systemd/system/docker.service.d/50-insecure-registry.conf
    content: |
        [Service]
        Environment="DOCKER_OPTS=--insecure-registry 10.0.0.0/8 --insecure-registry 172.16.0.0/12 --insecure-registry 192.168.0.0/16"
```

For Azure, it'll be:
```
  - path: /etc/systemd/system/docker.service.d/50-insecure-registry.conf
    content: |
        [Service]
        Environment="DOCKER_OPTS=--insecure-registry 0.0.0.0/0"
```

You should be able to override this once it's been parsed into an object, but you may need to iterate through the elements to find the right one.

timfpark:
@carmstrong: ok - done.

contrib/azure/azure-user-data-template
```diff
@@ -0,0 +1,25 @@
+#cloud-config
+coreos:
+  units:
+    - name: format-ephemeral.service
```
carmstrong: The cloud-config is parsed at boot, so this will wipe the filesystem every time the host is rebooted. You'll need to add a check to see if the filesystem has been created first. See https://github.com/deis/deis/blob/master/contrib/gce/gce-user-data-template#L9,L15

contrib/azure/azure-coreos-cluster
```diff
@@ -0,0 +1,248 @@
...
+  print 'Generated SSH certificate with thumbprint ' + args.ssh_thumb
+
+# Get Deis config
+if args.deis and args.custom_data:
+  print 'Conflicting options --deis and --custom-data provided, using --custom-data'
+if args.deis and not args.custom_data:
+  print 'Getting deis recommended CoreOS configuration'
+  discovery_url_fetch = urllib2.urlopen('https://discovery.etcd.io/new')
+  discovery_url = discovery_url_fetch.read()
+  deis_config_master = urllib2.urlopen('https://raw.githubusercontent.com/deis/deis/master/contrib/coreos/user-data.example')
```
carmstrong: Can we get rid of this codepath entirely? It makes me nervous (and is confusing for users). We should never be pulling the user-data from GitHub - it's in our local repo.

contrib/azure/azure-coreos-cluster
```diff
@@ -0,0 +1,248 @@
...
+parser.add_argument('--discovery-service-url',
+                   help='optional, url for an existing cluster discovery service. Else we will generate one.')
+parser.add_argument('--pip', action='store_true',
+                   help='optional, assigns public instance ip addresses to each VM')
+parser.add_argument('--deis', action='store_true',
+                   help='optional, automatically gets deis recommended CoreOS configuration')
+parser.add_argument('--data-disk', action='store_true',
+                   help='optional, attaches a data disk to each VM')
+
+cloud_init_template = """#cloud-config
```
carmstrong: Let's remove this, too. We should also be using the local cloud-config. We should make this script as specific to Deis (and thus as simple) as possible.

timfpark:
I can do that - I was copying Patrick's tool en mass into the directory but I will slim it down to just the Deis use case.

chanezon:
Chatting with Tim:
1. is it OK to keep the default CoreOS template as a here document in the script? This is not useful for deis, but for creating a quick coreos cluster, it removes the need for 1 parameter and some manual work. I'd prefer keeping the tool generic enough.
2. Chris you mentioned adding a --data-disk-url param https://github.com/chanezon/azure-linux/issues/13 Do you still want that? If so, I'll work on it tomorrow.

carmstrong:
> 1. is it OK to keep the default CoreOS template as a here document in the script? This is not useful for deis, but for creating a quick coreos cluster, it removes the need for 1 parameter and some manual work. I'd prefer keeping the tool generic enough.

Since the provision scripts live within the Deis repository, they are pretty minimalistic and simplistic so they are easy to customize/hack on for our users. However, it's not necessarily a hard-and-fast rule, so if the general CoreOS script could conceivably be useful to them, I think that's okay (as long as it is easy/straightforward to deploy Deis).

> 2. Chris you mentioned adding a --data-disk-url param chanezon/azure-linux#13 Do you still want that? If so, I'll work on it tomorrow.

It's not necessary - Jim asked to build a VHD with Deis preinstalled to put in the Azure marketplace. If we're going to do that for each release, we may as well make use of it to speed up Deis provisioning. But that can always be added later - the only thing necessary for Azure to work is for us to publish a custom builder with the `--insecure-registry` changes for Azure. I would connect with Jim on that one.

chanezon:
Thanks Chris. I'll merge the parts of Tim's PR that are relevant and we should be good to go then.

timfpark:
I merged @chanezon 's latest changes to the azure-coreos-cluster script.  we should be good to go now @carmstrong

docs/installing_deis/azure.rst
```diff
@@ -0,0 +1,116 @@
...
+
+    $ ./azure-coreos-cluster [cloud service name]
+         --subscription [subscription id]
+         --azure-cert azure-cert.pem 
+         --num-nodes 3
+         --location "West US"    
+         --vm-size Large  
+         --pip
+         --deis
+         --blob-container-url https://[blob container].blob.core.windows.net/vhds/
```
carmstrong: We should probably provide instructions on how to create a storage account.

docs/installing_deis/azure.rst
```diff
@@ -0,0 +1,116 @@
...
+
+* Create a container called 'vhds' within a storage account in the same region as your cluster using the Azure portal. Note the URL of the container for the cluster creation script below.
+* Choose a cloud service name for your Deis cluster for the script below. The script will automatically create this cloud service for you.
+
+With that, let's run the azure-coreos-cluster script which will create the CoreOS cluster. Fill in the bracketed values with the values for your deployment you created above.
+
+.. code-block:: console
+
+    $ ./azure-coreos-cluster [cloud service name]
+         --subscription [subscription id]
```
carmstrong: We should show how to find one's subscription ID.

timfpark:
Ok - done @carmstrong.

carmstrong:
@timfpark I'm getting this:

```
created service charmstrong
Succeeded
upload cert took:0s
  File "/Users/carmstrong/code/deis/contrib/azure/azure-coreos-cluster", line 207, in <module>
    role_size=args.vm_size, network_config=network, data_virtual_hard_disks=data_disks)
  File "/Library/Python/2.7/site-packages/azure/servicemanagement/servicemanagementservice.py", line 1260, in create_virtual_machine_deployment
    async=True)
  File "/Library/Python/2.7/site-packages/azure/servicemanagement/servicemanagementclient.py", line 235, in _perform_post
    response = self.perform_post(path, body, x_ms_version)
  File "/Library/Python/2.7/site-packages/azure/servicemanagement/servicemanagementclient.py", line 162, in perform_post
    response = self._perform_request(request)
  File "/Library/Python/2.7/site-packages/azure/servicemanagement/servicemanagementclient.py", line 192, in _perform_request
    return _management_error_handler(ex)
  File "/Library/Python/2.7/site-packages/azure/servicemanagement/__init__.py", line 1811, in _management_error_handler
    return _general_error_handler(http_error)
  File "/Library/Python/2.7/site-packages/azure/__init__.py", line 929, in _general_error_handler
    http_error.respbody.decode('utf-8'))
azure.WindowsAzureError: Unknown error (Forbidden)
<Error xmlns="http://schemas.microsoft.com/windowsazure" xmlns:i="http://www.w3.org/2001/XMLSchema-instance"><Code>ForbiddenError</Code><Message>You don't have permission to use this image.</Message></Error>
```

I'm guessing with `You don't have permission to use this image` the CoreOS image on Azure has been updated? Any ideas on how we can accommodate this?

Googling I found http://blogs.technet.com/b/windowshpc/archive/2014/06/24/how-to-make-sure-you-always-have-the-latest-os-image-in-your-azure-deployment-scripts.aspx

carmstrong:
Looks like the latest image is `2b171e93f07c4903bcad35bda10acf22__CoreOS-Stable-522.4.0`. We'll probably just have to keep this up-to-date.

carmstrong:
We should also get rid of the debug output: `{'status': u'InProgress', 'error': None, 'id': u'', 'http_status_code': u''}`.

carmstrong:
@timfpark I'll probably take over this PR and clean it up a bit.

timfpark:
Ok - let me know if I can help.

carmstrong:
@timfpark See #2893.

chanezon:
To determine the latest, you can do it manually with:
azure vm image list | grep CoreOS

Or automate it in Python with something like:
from azure import *
from azure.servicemanagement import *
import re

subscription_id = '9b5910a1-...d5ea2841'
cert_file = '9b5910a1-...d5ea2841.pem'
sms = ServiceManagementService(subscription_id, cert_file)

result = sms.list_os_images()
regex = re.compile('.*CoreOS-Stable-(\d+).*')
last = sorted([(int(m.group(1)), m.group(0)) for m in [regex.match(image.name) for image in result] if m], key=lambda x: x[0], reverse=True)[0][1]
print(last)

I can add that to the cluster creation script if you find it useful.

PR #2825, opened 2014-12-30
feat(builder+docker+user-data): proxy settings

### LINKED ISSUE #1734: Slugbuilder fails behind proxy
I have a deis cluster (one node) on a bare metal instance sitting behind a corporate proxy.
It was a little challenging but pushing a project with a Dockerfile is working.
However following after https://github.com/deis/deis/issues/1421, pushing the deis/example-php project or any other projects that use the slugbuilder doesn't work.

From the example-php directory
```
$ git push deis master
Counting objects: 198, done.
Compressing objects: 100% (145/145), done.
Writing objects: 100% (198/198), 360.78 KiB | 0 bytes/s, done.
Total 198 (delta 47), reused 198 (delta 47)
remote: -----> PHP app detected
remote:
remote:        !     WARNING: No composer.json found.
remote:        Using index.php to declare PHP applications is considered legacy
remote:        functionality and may lead to unexpected behavior.
remote:
remote:
remote:        !     ERROR: Failed to resolve runtime requirements.
remote:        If the issue persists, please contact support.
remote:
remote: Slugbuilder returned error code
To ssh://git@deis.local.deisapp.com:2222/example-php.git
 * [new branch]      master -> master
```

So i logged into the coreos server that host the deis-builder and the error is the same : 

```
core@flo-deis-01 ~ $ nse deis-builder
root@2ad55a2a747a:/home/git/example-php.git# git archive master | docker run -i -a stdin -a stdout deis/slugbuilder
-----> PHP app detected

       !     WARNING: No composer.json found.
       Using index.php to declare PHP applications is considered legacy
       functionality and may lead to unexpected behavior.


       !     ERROR: Failed to resolve runtime requirements.
       If the issue persists, please contact support.
```

I think it fails installing buildpacks during boot from https://github.com/deis/deis/blob/master/builder/slugbuilder/builder/install-buildpacks because the proxy isnt configured.

I'll try building a custom slugbuilder to make it works, but i believe there should be an easier way. Also a better error message :D.

lorieri:
Proxy settings for coreos, builder, slugrunner, slugbuilder and docker

originally using /etc/environment, changed to /etc/environment_proxy to avoid current automations failures

closes #1734 -- edited by @bacongobbler

### DIFF
--- a/builder/tests/builder_test.go
+++ b/builder/tests/builder_test.go
@@ -8,10 +8,14 @@ import (
 	"github.com/deis/deis/tests/dockercli"
 	"github.com/deis/deis/tests/etcdutils"
 	"github.com/deis/deis/tests/utils"
+
+	"io/ioutil"
+	"os"
 )
 
 func TestBuilder(t *testing.T) {
 	var err error
+	var errfile error
 	setkeys := []string{
 		"/deis/registry/protocol",
 		"/deis/registry/host",
@@ -31,6 +35,15 @@ func TestBuilder(t *testing.T) {
 		"/deis/domains",
 		"/deis/services",
 	}
+	setproxy := []byte("HTTP_PROXY=\nhttp_proxy=\n")
+
+	tmpfile, errfile := ioutil.TempFile("/tmp", "deis-test-")
+	if errfile != nil {
+		t.Fatal(errfile)
+	}
+	ioutil.WriteFile(tmpfile.Name(), setproxy, 0644)
+	defer os.Remove(tmpfile.Name())
+
 	tag, etcdPort := utils.BuildTag(), utils.RandomPort()
 	etcdName := "deis-etcd-" + tag
 	cli, stdout, stdoutPipe := dockercli.NewClient()
@@ -52,7 +65,9 @@ func TestBuilder(t *testing.T) {
 			"-e", "HOST="+host,
 			"-e", "ETCD_PORT="+etcdPort,
 			"-e", "EXTERNAL_PORT="+port,
-			"--privileged", "deis/builder:"+tag)
+			"--privileged",
+			"-v", tmpfile.Name()+":/etc/environment_proxy",
+			"deis/builder:"+tag)
 	}()
 	dockercli.PrintToStdout(t, stdout, stdoutPipe, "deis-builder running")
 	if err != nil {
--- a/contrib/coreos/user-data.example
+++ b/contrib/coreos/user-data.example
@@ -89,6 +89,7 @@ write_files:
   - path: /etc/systemd/system/docker.service.d/50-insecure-registry.conf
     content: |
         [Service]
+        EnvironmentFile=/etc/environment_proxy
         Environment="DOCKER_OPTS=--insecure-registry 10.0.0.0/8 --insecure-registry 172.16.0.0/12 --insecure-registry 192.168.0.0/16 --insecure-registry 100.64.0.0/10"
   - path: /run/deis/bin/get_image
     permissions: '0755'
@@ -141,3 +142,14 @@ write_files:
       TOOLBOX_DOCKER_IMAGE=ubuntu-debootstrap
       TOOLBOX_DOCKER_TAG=14.04
       TOOLBOX_USER=root
+  - path: /etc/environment_proxy
+    owner: core
+    content: |
+      HTTP_PROXY=
+      HTTPS_PROXY=
+      ALL_PROXY=
+      NO_PROXY=
+      http_proxy=
+      https_proxy=
+      all_proxy=
+      no_proxy=
--- a/builder/image/bin/boot
+++ b/builder/image/bin/boot
@@ -6,6 +6,8 @@
 # fail hard and fast even on pipelines
 set -eo pipefail
 
+source /etc/environment_proxy
+
 # set debug based on envvar
 [[ $DEBUG ]] && set -x
 
--- a/builder/image/bin/entry
+++ b/builder/image/bin/entry
@@ -1,6 +1,8 @@
 #!/bin/bash
 set -eo pipefail
 
+source /etc/environment_proxy
+
 # START jpetazzo/dind wrapper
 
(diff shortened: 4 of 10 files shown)


bacongobbler:
code LGTM, though let's squash these commits again and change the commit message to

```
feat(builder): add proxy support
```

As well as an explanation/justification of the problem as noted in http://docs.deis.io/en/latest/contributing/standards/#commit-style so we can better understand the changes involved here.

carmstrong:
@lorieri Can you add documentation around using this? It's a helpful change, but a user wouldn't know it's there without documentation.

It also looks like you need to run `gofmt` on the source.

contrib/coreos/user-data.example
```diff
@@ -133,3 +134,14 @@ write_files:
...
+    owner: core
+    content: |
+      HTTP_PROXY=
+      HTTPS_PROXY=
+      ALL_PROXY=
+      NO_PROXY=
+      http_proxy=
+      https_proxy=
+      all_proxy=
+      no_proxy=
```
ianblenke: Why both uppercase and lowercase? I'm familiar with the lowercase form for many unix commands, but not the uppercase form as a shell environment variable (outside of CGI which is a different use context and scope).

lorieri: @ianblenke builder runs custom buildpacks and we don't know what kind of program it will run, some programs use uppercase and others lowercase

lorieri: https://wiki.archlinux.org/index.php/proxy_settings#Environment_variables

carmstrong:
@lorieri I took your base docs and fleshed them out a bit. Feel free to make changes as necessary.

carmstrong:
ping @bacongobbler @mboersma for docs review - otherwise, I think this is ready to ship.

lorieri:
@carmstrong thanks, I have nothing to add

builder/tests/builder_test.go
```diff
@@ -31,6 +35,14 @@ func TestBuilder(t *testing.T) {
 		"/deis/domains",
 		"/deis/services",
 	}
+	setproxy := []byte("HTTP_PROXY=\nhttp_proxy=\n")
+
+        tmpfile,errfile := ioutil.TempFile("/tmp", "deis-test-")
```
mboersma: Indentation here isn't consistent: we need to `gofmt` this.

lorieri: I run gofmt but I was waiting @carmstrong to help me with the docs, let me do this now

lorieri:
@carmstrong I've lost your docs commit, could commit it again ?

builder/tests/builder_test.go
```diff
@@ -65,4 +79,6 @@ func TestBuilder(t *testing.T) {
 	dockercli.DeisServiceTest(t, name, port, "tcp")
 	etcdutils.VerifyEtcdValue(t, "/deis/builder/host", host, etcdPort)
 	etcdutils.VerifyEtcdValue(t, "/deis/builder/port", port, etcdPort)
+
+	defer os.Remove(tmpfile.Name())
```
bacongobbler: This should be moved up to line 45. That way, the next person reading this won't be wondering where the call to `defer` is hiding.

contrib/coreos/user-data.example
```diff
@@ -133,3 +134,14 @@ write_files:
       TOOLBOX_DOCKER_IMAGE=ubuntu-debootstrap
       TOOLBOX_DOCKER_TAG=14.04
       TOOLBOX_USER=root
+  - path: /etc/environment_proxy
```
bacongobbler: Hmm, I'm wondering if it'd be better to throw settings like this into a deis-specific directory like `/etc/deis/environment` and call it the official place to store host-level envvars.

lorieri: In my installation I use /etc/environment because it does make sense to use for the whole system

but it is up to you

bacongobbler: That's a possibility, too. We'd just need to write out COREOS_PRIVATE_IPV4 and COREOS_PUBLIC_IPV4. That may also remove the logic to source /etc/environment_proxy as well.

```
- path: /etc/environment
  content: |
    COREOS_PUBLIC_IPV4=$public_ipv4
    COREOS_PRIVATE_IPV4=$private_ipv4
    HTTP_PROXY=
    [...]
```

carmstrong: I'd rather not overwrite /etc/environment in our logic, as we would potentially be clobbering changes made upstream unless we account for them ourselves. I don't think `/etc/deis/environment` is appropriate either, as the changes aren't specific to Deis (Docker also reads the file). So, I'm not in love with `/etc/environment/proxy`, but not sure if there's a better candidate.

builder/tests/builder_test.go
lorieri: sure, at that time I was not sure I could do that :)

carmstrong:
@lorieri Could you rebase this off master? We'll review again and try to get it in v1.4.0. Thanks!

lorieri:
done

carmstrong:
@lorieri You'll have to rebase one last time - we just fixed a bug in the tests in master. Sorry!

lorieri:
done :+1:

builder/tests/builder_test.go
```diff
@@ -31,6 +35,15 @@ func TestBuilder(t *testing.T) {
 		"/deis/domains",
 		"/deis/services",
 	}
+	setproxy := []byte("HTTP_PROXY=\nhttp_proxy=\n")
+
+	tmpfile, errfile := ioutil.TempFile("/tmp", "deis-test-")
+	if errfile != nil {
+		panic(errfile)
```
mboersma: We should call `t.Fatal` here instead of `panic`.

bacongobbler: :+1:

lorieri:
@mboersma done

mboersma:
These proxy settings changes and related docs LGTM.

Issue #2811, opened 2014-12-29
Deis pull does not respect permissions.


glogiotatidis:
Any user can `deis pull` into any app without having the needed permissions. Steps to reproduce:

```
 - deis login --username user1
 - deis create
 - deis pull postgres
   # creates release v2
 - deis login --username user2
 - deis pull postgres
   # creates release v3
```

The last step should fail. Other commands like `deis info` or `deis config` fail as excepted because `user2` does not have permissions on this app.

PR #2826, opened 2014-12-30
fix(controller): disallow unauthorized users from modifying apps

### LINKED ISSUE #2820: Any user can add a domain to any app
See https://github.com/deis/deis/pull/2818#issuecomment-68319676

```console
$ deis register ...
$ deis create
$ git push deis master
$ deis logout
$ deis register ...
$ deis domains:add test.lol.com
Adding test.lol.com to quinoa-icehouse... done
```

bacongobbler:
Authenticated users who were not given explicit permission
to view or modify an app were still allowed to create new releases,
run commands, retrieve logs etc. This makes those requests return a 404
if the user should not be able to see the application.

Please let me know if I missed any loopholes and I will happily correct them.

closes #2820, refs #2811

### DIFF
--- a/docs/reference/api-v1.1.rst
+++ b/docs/reference/api-v1.1.rst
@@ -6,7 +6,7 @@
 Controller API v1.1
 ===================
 
-This is the v1.0 REST API for the :ref:`Controller`.
+This is the v1.1 REST API for the :ref:`Controller`.
 
 
 What's New
@@ -15,6 +15,8 @@ What's New
 **New!** All controller responses now return the ``X_DEIS_API_VERSION`` header.
 
 **New!** All controller responses now return the ``X_DEIS_PLATFORM_VERSION`` header.
+
+**New!** Users see a 404 response when modifying applications they are unauthorized to see.
 
 
 Authentication
--- a/controller/api/__init__.py
+++ b/controller/api/__init__.py
@@ -2,4 +2,4 @@
 The **api** Django app presents a RESTful web API for interacting with the **deis** system.
 """
 
-__version__ = '1.1.1'
+__version__ = '1.1.2'
--- a/controller/api/tests/test_app.py
+++ b/controller/api/tests/test_app.py
@@ -253,6 +253,39 @@ class AppTest(TestCase):
         self.assertEqual(response.data, "No build associated with this release "
                                         "to run this command")
 
+    def test_unauthorized_user_cannot_see_app(self):
+        """
+        An unauthorized user should not be able to access an app's resources.
+
+        Since an unauthorized user should not know about the application at all, these
+        requests should return a 404.
+        """
+        app_id = 'autotest'
+        base_url = '/v1/apps'
+        body = {'id': app_id}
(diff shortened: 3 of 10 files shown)


carmstrong:
```
/home/jenkins/workspace/test-integration/src/github.com/deis/deis/tests
apps:run echo Hello, 世界
Running `echo Hello, 世界`...
Hello, 世界
--- FAIL: TestApps (137.27s)
	itutils.go:272: <nil>
FAIL
exit status 1
FAIL	github.com/deis/deis/tests	137.273s
godep: go exit status 1
make[1]: Leaving directory `/home/jenkins/workspace/test-integration/src/github.com/deis/deis/tests'
```

Weird, considering #2813 passed Jenkins. Maybe a red herring? I'll run Jenkins again.

carmstrong:
Failed again in exactly the same spot. Did this PR modify anything that would affect that?

carmstrong:
@bacongobbler did this pass integration tests for you locally? That would rule out this PR.

carmstrong:
Code LGTM, although I didn't comb through the code to see if we missed anything. @mboersma does it look like we missed anything, at first glance?

mboersma:
I looked through the models.py and I couldn't think of similar security issues; this seems to cover it.

@bacongobbler pointed out that since this changes 403 to 404 responses, it should probably require an API version bump. Also we may want to reword this to use "unauthorized" instead.

LGTM, pending resolution of those two items.

bacongobbler:
bumped the controller API version to v1.1.2 in response to the changes as well as an added note in the API docs. Renamed all references of "evil" to "unauthorized" or "foo/bar" as well. Please take another look :)

docs/reference/api-v1.1.rst
```diff
@@ -16,6 +16,8 @@ What's New
 
 **New!** All controller responses now return the ``X_DEIS_PLATFORM_VERSION`` header.
 
+**New!** Users see a 404 response when modifying applications they are unauthorized to see.
+
```
mboersma: Should we also update all the example requests in this doc to say `X_DEIS_API_VERSION: 1.1.2`? Looks like we didn't do that for the 1.1.1 API change, but I think that would be consistent. I've been updating the `X_DEIS_PLATFORM_VERSION:` in the examples with each release.

bacongobbler: yes, good point.

bacongobbler: done!

bacongobbler: d'oh, you meant X_DEIS_API_VERSION. In the response, the patch version gets cut off (clients should not care what happens in patch releases as they're just patches), so `1.1` is intended: https://github.com/deis/deis/blob/6f00ed544d8d008ae8f5769b2581e7e1897f9249/controller/api/middleware.py#L32

No change necessary, then :)

mboersma: Oh, ok, that makes sense. Sorry for my confusion. LGTM again.

carmstrong:
:sheep:

PR #2829, opened 2015-01-02
feat(client): Add controller to whoami command.

glogiotatidis:
Whoami command needs to display both username and controller. Useful
when using multiple deis clusters.

### DIFF
--- a/client/deis.py
+++ b/client/deis.py
@@ -869,7 +869,8 @@ class DeisClient(object):
         """
         user = self._settings.get('username')
         if user:
-            self._logger.info(user)
+            self._logger.info(
+                'You are {} at {}'.format(user, self._settings['controller']))
         else:
             self._logger.info(
                 'Not logged in. Use `deis login` or `deis register` to get started.')


deis-admin:

bacongobbler:
I can see the usefulness in this, but personally I'd rather that `deis whoami` would stick to displaying only the username. Maybe we could stick this into a command like `deis info`?

glogiotatidis:
I see what you mean and I agree. `deis info` needs an app name so maybe not the perfect place to put it.

Other suggestions:
 - `deis whoami` returns `username@deis.example.com` so it's still one thing but it includes all information.
 - `whoami --full` flag to display controller information.

Thoughts?

bacongobbler:
Oh right. We currently have `deis info` as a shortcut for `deis apps:info`. `username@deis.example.com` works for me.

glogiotatidis:
@bacongobbler updated to display `username@netloc`

bacongobbler:
code LGTM

mboersma:
Code LGTM.

carmstrong:
@glogiotatidis Could you change this to something like: `You are carmstrong at https://deis.gingerweather.com`?

This shows explicitly the user, the protocol, and the controller host.

glogiotatidis:
Updated based on the latest comment.

client/deis.py
```diff
@@ -869,7 +869,8 @@ def auth_whoami(self, args):
         """
         user = self._settings.get('username')
         if user:
-            self._logger.info(user)
+            self._logger.info(
+                'You are {user} at {controller}'.format(user=user, controller=self._settings['controller']))
```
bacongobbler: Since we only support python 2.7, the positional arguments can be omitted. You can greatly simplify the logic here with:

```
'You are {} at {}'.format(user, self._settings['controller']))
```

Which also follows the style with the rest of the codebase.

glogiotatidis:
OK updated with the latest `format()` format

carmstrong:
Code LGTM.

bacongobbler:
Jenkins, add to whitelist.

mboersma:
```console
$ deis --version
1.3.0-dev
$ deis whoami
You are test at http://deis.deis.works
```
LGTM.

PR #2832, opened 2015-01-02
feat(docker): add env DEIS_RELEASE

### LINKED ISSUE #2199: Add DEIS_RELEASE to deis images
The tag from a component should not be used to know to which release belong

### LINKED ISSUE #2568: Modify Dockerfiles with each release
Currently when we publish `v1.0.0` and `v1.0.1` tags of a component to the Docker Hub where the layers haven't changed, someone who provisions Deis may see either tag in the output of `docker images`. It's confusing--it makes it appear that you haven't successfully upgraded when really it's just a cosmetic issue.

We could perhaps work around this by changing a trivial line in each `Dockerfile` as part of the release process.

aledbf:
closes #2199
closes #2568

### DIFF
--- a/builder/image/Dockerfile
+++ b/builder/image/Dockerfile
@@ -60,3 +60,6 @@ ADD templates/shim.dockerfile /home/git/
 ADD etc /etc
 ADD . /app
 RUN chown -R root:root /app
+
+ENV DEIS_RELEASE 1.2.0-dev
+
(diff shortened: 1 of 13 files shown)


carmstrong:
Thanks for taking the time to do this, @aledbf! My only question is whether we should be using `1.2.0-dev` in the Dockerfiles, or some other version string - I would defer to @mboersma on that. Otherwise, code LGTM!

mboersma:
I think `1.2.0-dev` is good in this case. Too bad we have to repeat ourselves so much, but I don't see a simple alternative, and it's not much of a headache to update a few more version strings every so often. It also makes the Dockerfiles a bit more self-documenting. Thanks @aledbf.

Code LGTM.

PR #2837, opened 2015-01-03
feat(controller): add hostname exposing to units

### LINKED ISSUE #2835: Apps should be published with an descriptive hostname
Hello,

I'm currently testing out New Relic for monitoring applications we have running on Deis, and all the instances (JVMs) show up with names like this: `1ecbec0cb893`, since that is the Docker hostname.

What I would rather prefer, is a hostname like`my-application.cmd.1.deis.example.com` or  `my-application.cmd.1`, since this would make looking up the actual instance in New Relic easier.

I think the solution would be to use Dockers `--hostname=` parameter in `docker run` ( https://docs.docker.com/articles/networking/#dns ).

>`-h HOSTNAME` or `--hostname=HOSTNAME` — sets the hostname by which the container knows itself. This is written into `/etc/hostname`, into `/etc/hosts` as the name of the container's host-facing IP address, and is the name that `/bin/bash` inside the container will display inside its prompt. But the hostname is not easy to see from outside the container. It will not appear in docker ps nor in the `/etc/hosts` file of any other container.

Any thoughts?

Blystad:
This exposes a configurable hostname type to Docker container through
fleet unit files.

Three hostname types are supported:
- default     : standard Docker hostname
- application : application hostname (e.g. my-app.v54.web.1)
- server      : Core OS hostname

This fixes #2835 and is an alternative to #2428, which got rejected.

It is necessary when using a monitoring solution like New Relic.

### DIFF
--- a/controller/scheduler/fleet.py
+++ b/controller/scheduler/fleet.py
@@ -7,6 +7,7 @@ import paramiko
 import socket
 import re
 import time
+from django.conf import settings
 
 
 MATCH = re.compile(
@@ -112,6 +113,8 @@ class FleetHTTPClient(object):
             l.update({'cpu': '-c {}'.format(cpu)})
         else:
             l.update({'cpu': ''})
+        # set unit hostname
+        l.update({'hostname': self._get_hostname(name)})
         # should a special entrypoint be used
         entrypoint = kwargs.get('entrypoint')
         if entrypoint:
@@ -136,6 +139,19 @@ class FleetHTTPClient(object):
             except:
                 if attempt == (RETRIES - 1):  # account for 0 indexing
                     raise
+
+    def _get_hostname(self, application_name):
+        hostname = settings.UNIT_HOSTNAME
+        if hostname == "default":
+            return ''
+        elif hostname == "application":
+            # replace underscore with dots, since underscore is not valid in DNS hostnames
+            dns_name = application_name.replace("_", ".")
+            return '-h ' + dns_name
+        elif hostname == "server":
+            return '-h %H'
+        else:
+            raise RuntimeError('Unsupported hostname: ' + hostname)
 
     def start(self, name):
         """Start a container"""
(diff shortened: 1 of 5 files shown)


deis-admin:

carmstrong:
Jenkins, add to whitelist.

carmstrong:
I think this is a great compromise for hostnames! My only comment would be that we typically don't include notes like these in the Customizing Deis section, but I honestly can't think of a better location for the information in this PR.

I would also add an example for setting this with ``deisctl``, but otherwise this looks great!

carmstrong:
@Blystad Jenkins is failing with a style error in `fleet.py`. I confirmed that this isn't a problem on master:

```console
➜  controller git:(pr-2837) venv/bin/flake8
./scheduler/fleet.py:101:1: C901 'FleetHTTPClient._create_container' is too complex (14)
➜  controller git:(pr-2837) git checkout master
Switched to branch 'master'
Your branch is up-to-date with 'origin/master'.
➜  controller git:(master) git pull
Already up-to-date.
➜  controller git:(master) venv/bin/flake8
➜  controller git:(master)
```

controller/scheduler/fleet.py
```diff
@@ -112,6 +113,18 @@ def _create_container(self, name, image, command, unit, **kwargs):
             l.update({'cpu': '-c {}'.format(cpu)})
         else:
             l.update({'cpu': ''})
+        # set unit hostname
+        hostname = settings.UNIT_HOSTNAME
+        if hostname == "default":
+            l.update({'hostname': ''})
+        elif hostname == "application":
+            # replace underscore with dots, since underscore is not valid in DNS hostnames
+            dns_name = name.replace("_", ".")
```
carmstrong: Maybe we should replace underscores with hyphens instead? I'm thinking dots typically imply subdomains, which may cause issues with some applications. Or maybe I'm entirely off-base.

Blystad: I considered hyphens, and I considered stripping it out entierly. I can't see how that could cause issues with some applications, and also it would make filtering easier as it will be easier to write filters (when searching through logs which includes the hostname) where `my-application.*.web.*` is found.

Blystad:
Added `deisctl` example and fixed the complexity by moving the hostname functionality into it's own function, which is better anyways. Now I'll just wait for Jenkins to pass.

Blystad:
It passed, and now I see a "Confirm Merge" button. Should I click that?

carmstrong:
Code LGTM! We're shipping Deis v1.2.0 on Tuesday, so this will likely merge shortly thereafter.

bacongobbler:
this is really well designed. Nice job @Blystad!

docs/customizing_deis/controller_settings.rst
```diff
@@ -73,3 +74,38 @@ Be sure that your custom image functions in the same way as the `stock controlle
...
+Deis supports configuring hostname assignment through the ``unitHostname`` setting.
+You can change the assignment solution using the following command:
+
+.. code-block:: console
+
+    $ deisctl config controller set unitHostname=application
+
+The valid ``unitHostname`` alternatives are:
+
+default
```
bacongobbler: Could we please make this a [definition list](http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#definition-lists) instead?

Blystad:
@bacongobbler Thanks for the kind words. I've changed the documentation into a definition list now.

bacongobbler:
code LGTM. :shipit:

PR #2840, opened 2015-01-05
feat(store): add status dashboard

aledbf:
This PR provides a dashboard to show the status of the ceph cluster using [ceph-dash](https://github.com/Crapworks/ceph-dash)
*ceph-dash* it's a wsgi application that uses librados (for python) to retrieve the status of the cluster. 
There is no persistence or any dependency to run the component besides be able to connect to the ceph cluster.

*This is a screenshot of the dashboard in a new deis cluster:*
ceph-dashboard

*And this one of a cluster with problems*
ceph-dashboard-2


carmstrong:
Awesome, @aledbf! I'll give this a shot.

As this is an optional component, we should add instructions for starting the component.

aledbf:
@carmstrong it's required add the service in deisctl? 
There is no other "optional" component, so please tell me where's the place for the instructions.
The instructions are simple: `fleetctl submit deis-store-dashboard.service` and `fleetctl start deis-store-dashboard.service`

carmstrong:
> The instructions are simple:

Sure, but the `.service` file has to live somewhere, and I agree it probably shouldn't be in deisctl. Maybe put it in `store/dashboard`?

> There is no other "optional" component, so please tell me where's the place for the instructions.

Honestly, I'm not sure. It could live in Managing Deis, but I kind of think this should live in Customizing Store like the changes in #2837.

How featured is the dashboard, @aledbf? Does it help people to recover their cluster, or is it only for an informational display?

aledbf:
> How featured is the dashboard?

It is only an informational display.

There's only 2 alternatives to manage the ceph cluster: [calamari](https://github.com/ceph/calamari) and [inkscope](https://github.com/inkscope/inkscope). 
**Calamari** is the first option, only requires postgres (deis-database could provide the service) but as you know the current installation process is not easy. Watching [CDS Hammer](https://www.youtube.com/watch?v=bGB8XzGNC1o&feature=youtu.be) they mentioned that in the next ceph release calamari will provide installation packages.
**Inkscope** it's really nice but it requires mongodb so I don't see that as an option.

In the mean time I think that this could provide a status of the cluster without using `nse` or any ceph knowledge

carmstrong:
> Watching CDS Hammer they mentioned that in the next ceph release calamari will provide installation packages.

Okay. I'll give this dashboard a go. I'm wondering if it's worth waiting to implement a store dashboard until we have one that can actually change things (like Calamari), especially if they're working on making it easier to install.

carmstrong:
@aledbf We've discussed this internally and have decided that we don't want to bring a Ceph dashboard into Deis core -- exposing a Ceph dashboard would leak the `deis-store` abstraction more than we're comfortable with.

That being said, we'd love to build a page linking to these sorts of tools so Deis users can find and use them: #2952.

PR #2844, opened 2015-01-06
fix(contrib/coreos): fix debug-etcd service

carmstrong:
The debug-etcd service added in #2779 has an issue - the `command: start`
is invalid.

This doesn't *appear* to break things, but the service that handles
cloud-config parsing will error out, which can result in services not starting.

### DIFF
--- a/contrib/coreos/user-data.example
+++ b/contrib/coreos/user-data.example
@@ -57,7 +57,6 @@ coreos:
       RemainAfterExit=yes
       Type=oneshot
   - name: debug-etcd.service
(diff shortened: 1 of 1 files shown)


mboersma:
I had to re-read https://coreos.com/docs/cluster-management/setup/cloudinit-cloud-config/. I'm surprised the invalid command was ignored at all, but good catch. Code LGTM.

bacongobbler:
code LGTM. I missed this while reviewing and should have picked up on it. We have seen this in the past.

PR #2847, opened 2015-01-06
feat(store): add S3 mock gateway

aledbf:


mboersma:
@aledbf this is so great. I'm going to see if I can hook it up to the functional tests.

Issue #2793, opened 2014-12-22
fix: Move /var/lib/docker to separate volume, prevents all units from crashing on a single machine


carmstrong:
When debugging performance issues with Deis on Azure, we discovered that machines became unresponsive whenever we had high activity (such as when doing a `docker pull` when components start). We actually witnessed systemd timing out and issuing a sigterm to all of its processes(!), effectively emulating a system reboot. This of course causes issues for etcd/fleet, as well as Deis and anything else on the machine. We suspect this is the cause of all of the units on a machine showing as dead.

In attempting to resolve this issue, we tried moving Docker's `/var/lib/docker` directory to a separate volume. The difference is night and day - no more performance issues, and Deis came up fine on the first try.

We should try this on EC2 as well, as we see what appears to be this same issue there (see #2506). We should also generally document this as a best practice.

Issue #2506, opened 2014-11-14
Performance issues on EC2 (etcd/fleet timeout, builder-caused instance crashes)


carmstrong:
Working with @cap10morgan in IRC today, he provided me with some fleet and etcd logs and reported that he couldn't get the cluster up and running due to fleet connection timeout issues. His logs reflect bad latency between nodes:

```
Nov 13 17:55:12 ip-10-36-16-122.us-west-2.compute.internal etcd[3465]: [etcd] Nov 13 17:55:12.130 INFO      | d0912b5c56ba4a5c93f531bc52dc0850 starting in peer mode
Nov 13 17:55:12 ip-10-36-16-122.us-west-2.compute.internal etcd[3465]: [etcd] Nov 13 17:55:12.130 INFO      | d0912b5c56ba4a5c93f531bc52dc0850: state changed from 'initialized' to 'follower'.
Nov 13 17:55:12 ip-10-36-16-122.us-west-2.compute.internal etcd[3465]: [etcd] Nov 13 17:55:12.130 INFO      | d0912b5c56ba4a5c93f531bc52dc0850: state changed from 'follower' to 'leader'.
Nov 13 17:55:12 ip-10-36-16-122.us-west-2.compute.internal etcd[3465]: [etcd] Nov 13 17:55:12.130 INFO      | d0912b5c56ba4a5c93f531bc52dc0850: leader changed from '' to 'd0912b5c56ba4a5c93f531bc52dc0850'.
Nov 13 17:55:12 ip-10-36-16-122.us-west-2.compute.internal etcd[3465]: [etcd] Nov 13 17:55:12.563 INFO      | d0912b5c56ba4a5c93f531bc52dc0850: peer added: 'd20c42216aab422988c04fd637c5f15d'
Nov 13 17:55:18 ip-10-36-16-122.us-west-2.compute.internal etcd[3465]: [etcd] Nov 13 17:55:18.573 INFO      | d0912b5c56ba4a5c93f531bc52dc0850: peer added: 'fda33bb22d444f148c5aee273ae6c904'
Nov 13 17:59:09 ip-10-36-16-122.us-west-2.compute.internal etcd[3465]: [etcd] Nov 13 17:59:09.073 INFO      | d0912b5c56ba4a5c93f531bc52dc0850: warning: heartbeat time out peer="fda33bb22d444f148c5aee273ae6c904" missed=1 backoff="2s"
[...]

PR #2850, opened 2015-01-06
feat(contrib/ec2): use ephemeral drive for etcd journal

carmstrong:
The changes in #2803 unfortunately don't address all of the etcd-related
performance issues. Provisioning Deis can still cause etcd timeouts.
Moving the etcd journal to the unused ephemeral disk allows etcd to
write to a very fast disk, resulting in fewer etcd/fleet issues.

refs #2793
refs #2506

### DIFF
--- a/contrib/ec2/gen-json.py
+++ b/contrib/ec2/gen-json.py
@@ -6,9 +6,42 @@ import yaml
 CURR_DIR = os.path.dirname(os.path.realpath(__file__))
 
 # Add EC2-specific units to the shared user-data
-FORMAT_EPHEMERAL = '''
+FORMAT_EPHEMERAL_VOLUME = '''
   [Unit]
-  Description=Formats the ephemeral drive
+  Description=Formats the ephemeral volume
+  ConditionPathExists=!/etc/ephemeral-volume-formatted
+  [Service]
+  Type=oneshot
+  RemainAfterExit=yes
+  ExecStart=/usr/sbin/wipefs -f /dev/xvdb
+  ExecStart=/usr/sbin/mkfs.ext4 /dev/xvdb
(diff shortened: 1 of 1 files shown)


carmstrong:
Note that I stopped Jenkins since he wouldn't test this.

carmstrong:
I manually tested this with both PV and HVM instance types and scaled to 10 containers on the first try with no problems, which I was unable to do earlier today without fleet errors.

bacongobbler:
if this removes the etcd errors that I was seeing yesterday, then code LGTM

mboersma:
> I manually tested this with both PV and HVM instance types...

Great, that answers my only question. These changes to mount an ephemeral drive on AWS LGTM.

carmstrong:
> if this removes the etcd errors that I was seeing yesterday

Unfortunately this is only for EC2, but #2851 tracks doing this for other providers.

ianblenke:
I'm not sure I'm seeing any problems with gp2 SSD volumes... should I really be considering using ephemeral volumes for the etcd journal?

carmstrong:
> I'm not sure I'm seeing any problems with gp2 SSD volumes... should I really be considering using ephemeral volumes for the etcd journal?

We saw tons of etcd leader election thrashing before, and we remedied this by moving the etcd data volume to the fast local disk and using a separate EBS for Docker. We suspect that the IO to the Docker volume was preventing etcd from syncing to disk often enough. 

If you're not seeing issues, I wouldn't be too concerned. This was all before we switched the EBSes over to gp2, so that could alleviate this as well (although separating etcd does seem to be a best practice).

I'd hit the etcd stats endpoint to see if it reports hitting any heartbeat timeouts on its peers.

PR #2853, opened 2015-01-06
fix(controller): disallow unauthorized users from deleting apps

### LINKED ISSUE #2845: "deis destroy" works for an unauthorized user
Testing interactively with current master and attempting to verify #2826 shows there is more work to do.

### LINKED ISSUE #2846: "deis rollback" works for an unauthorized user
Testing interactively with current master and attempting to verify #2826 shows there is more work to do.

```console
$ deis register http://deis.deis.rocks
username: badguy
...
$ deis rollback
Rolling back one release... done, v5
$ deis info
404 NOT FOUND
Detail:
Not found
```

bacongobbler:
Authenticated users who were not given explicit permission to view
or modify an application were still allowed to delete other user's
applications, should the unauthorized user know what the ID of the
application would be. This returns a 404 message if the user is
unauthorized to see the application.

Because we were calling get_object_or_404(), we were not performing
any authorization against the API call. Calling self.get_object() is
the preferred method to retrieve an object since it authenticates
against the classes' permissions classes to ensure that the user is
authorized to see the application.

This also fixes up an issue where administrators were not explicitly
given permission to an app. They fell through the same "unauthorized"
issue as above.

closes #2845
closes #2846

### DIFF
--- a/controller/api/permissions.py
+++ b/controller/api/permissions.py
@@ -36,6 +36,8 @@ class IsAppUser(permissions.BasePermission):
     an app-related model.
     """
     def has_object_permission(self, request, view, obj):
+        if request.user.is_superuser:
+            return True
         if isinstance(obj, models.App) and obj.owner == request.user:
             return True
         elif hasattr(obj, 'app') and obj.app.owner == request.user:
--- a/controller/api/tests/test_app.py
+++ b/controller/api/tests/test_app.py
@@ -275,6 +275,10 @@ class AppTest(TestCase):
         url = '{}/{}/logs'.format(base_url, app_id)
         response = self.client.get(url, HTTP_AUTHORIZATION='token {}'.format(unauthorized_token))
         self.assertEqual(response.status_code, 404)
+        url = '{}/{}'.format(base_url, app_id)
+        response = self.client.delete(url,
+                                      HTTP_AUTHORIZATION='token {}'.format(unauthorized_token))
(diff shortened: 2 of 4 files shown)


mboersma:
`get_object_or_404` is a standard Django view func, but `self.get_object` is a DRF method, so I see why the latter is preferable to invoke the intended auth. This code LGTM, but I'm going to test it by hand for a while as well.

bacongobbler:
also fixed up #2846 with this one. More funny view logic.

```
><> deis rollback -a go
Rolling back one release... 403 FORBIDDEN
Detail:
You do not have permission to perform this action.
```

carmstrong:
Do we know why the rollback test didn't catch this?

bacongobbler:
> Do we know why the rollback test didn't catch this?

I'm not quite sure. I'm still investigating the issue there, but this patched up the issue :)

mboersma:
I've tested this by hand and can't find any gaps now. After discussing this, we still want to revert some of the 404s to 403s, but we can do that in a subsequent PR.

LGTM.

carmstrong:
Code LGTM.

bacongobbler:
unit tests passed with the most recent commit. No code changes were made; it was just fixing up the issue @carmstrong mentioned in https://github.com/deis/deis/pull/2853#issuecomment-68910935. Merging!

Issue #2854, opened 2015-01-06
ref(controller): API views need refactoring


bacongobbler:
With Django Rest Framework v3, a lot of our Class views can be optimized, refactored, and otherwise hardened such that we no longer need to update function calls in multiple classes. This will really help us with future security patches, such as the work being done at #2853.

bacongobbler:
related: #2843

Issue #2843, opened 2015-01-05
REST errors question


aledbf:
I'm using the deis API and I see one "annoying" thing: the error formats.
In case of an error (not relevant) the error returns a JSON object where the key changes and the value can be an array.
It's possible to define 1 format and wrap the errors just indicating the first error? Like:
```
{
  "message": "invalid credentials"
}
```

**Some errors responses from the API:**
```
{"id":["App with this Id already exists."]}
{"username":["User with this Username already exists."]}
```

bacongobbler:
I think this has to do with how we're handling these errors through Django Rest Framework's Serializer class. How this format works is the following:

```
{"<invalid_field_in_api_request>": [<reasons why>]}
```

Which makes sense, considering that multiple validation rules could be broken with a specific app ID (for example, using capital letters and making the name shorter than 4 characters in length).

The first step to standardizing these errors would be to identify all of the error cases and document them. If we document and throw these into the [API documentation](http://docs.deis.io/en/latest/reference/api-v1.1/) then we can take a closer look at all of the responses and standardize them accordingly.

PR #2856, opened 2015-01-06
ref(controller): return 403s for unauthorized endpoints

mboersma:
A 403 response doesn't necessarily acknowledge the existence of a resource, but instead indicates that the server understood the request and refuses to fulfill it.
- https://en.wikipedia.org/wiki/HTTP_403
- http://programmers.stackexchange.com/questions/82371/should-mvc-rest-return-a-403-or-404-for-resources-belonging-to-other-users

Some of our endpoints that were changed to return 404s are probably being over-cautious. This isn't set in stone for REST APIs, but after discussing with our BDFL this morning, we think 403s are more correct here.

This changes unauthorized responses to 403s as they were before, and rolls the API version back to 1.1.1 since (I think) this is previous behavior. It was too tricky to change "apps:run" and "apps:logs" off of 404s, however, so those tests are marked as `expectedFailure`.


gabrtv:
Thanks for this @mboersma.  We can tackle the response codes for `apps:run` and `apps:logs` separately.  Code LGTM.

controller/api/views.py
```diff
@@ -418,35 +389,21 @@ class DomainViewSet(OwnerViewSet):
...
+        self.check_object_permissions(self.request, app)
         request._data = request.DATA.copy()
         request.DATA['app'] = app
         return super(DomainViewSet, self).create(request, *args, **kwargs)
 
     def get_queryset(self, **kwargs):
         app = get_object_or_404(models.App, id=self.kwargs['id'])
-        self.check_obj_permissions(app)
-        return self.model.objects.filter(app=app)
+        qs = self.model.objects.filter(app=app)
```
bacongobbler: nitpick: let's turn this into a one-liner

```
return self.model.objects.filter(app=app)
```

controller/api/views.py
```diff
@@ -418,35 +389,21 @@ class DomainViewSet(OwnerViewSet):
...
         app = get_object_or_404(models.App, id=self.kwargs['id'])
-        self.check_obj_permissions(app)
+        self.check_object_permissions(self.request, app)
         request._data = request.DATA.copy()
         request.DATA['app'] = app
         return super(DomainViewSet, self).create(request, *args, **kwargs)
 
     def get_queryset(self, **kwargs):
         app = get_object_or_404(models.App, id=self.kwargs['id'])
-        self.check_obj_permissions(app)
```
bacongobbler: how come we're not replacing the checks here?

bacongobbler:
The apps:run issues with returning a 404 probably stems from the fact that AppViewSet is not a child of AppBaseViewSet. Breathing new life into the views via #2854 will help us solve these issues as well as help find any other issues that we may have missed.

Also related: #2843

controller/api/views.py
mboersma: We are now.

PR #2857, opened 2015-01-07
fix(client): remove fetch call from apps:create

bacongobbler:
If you have a violating key in your known_hosts file, it will cause
you to fail adding the git remote, even though the app is already
created on the controller's side. Validating that the server's key
is correct should happen at git push to match Heroku.


carmstrong:
Code LGTM.

mboersma:
Code LGTM.

PR #2859, opened 2015-01-07
fix(contrib/ec2): use ext3 for etcd data volume

carmstrong:
We saw etcd panics when using an ext4 filesystem. We cannot replicate
these crashes with ext3.


mboersma:
Code LGTM. See http://www.pointsoftware.ch/en/4-ext4-vs-ext3-filesystem-and-why-delayed-allocation-is-bad/ for an explanation consistent with what testing showed.

carmstrong:
It turns out this isn't the fix - still seeing etcd panics :(

PR #2863, opened 2015-01-07
Revert ec2 fixes

carmstrong:


mgartner:
Were the issues seen in #2506 resolved in another way? What is the motivation for backing this out?

carmstrong:
> Were the issues seen in #2506 resolved in another way? What is the motivation for backing this out?

Having the etcd data directory on another mount results in a nasty panic and etcd crash when another node joins the cluster: https://github.com/coreos/etcd/issues/2050

We're working with the CoreOS team to investigate, so this PR will only get merged as a last resort. v1.2.0 is unshippable with this crash.

carmstrong:
This is thankfully no longer needed as of #2872.

PR #2867, opened 2015-01-08
test(controller): add expected unit test failures for response codes

bacongobbler:
When testing the security holes found recently (#2826), we found a few response
codes that were not what was to be expected. Adding test coverage around
this such that we can address these tests in future refactors.


mboersma:
I looked through the tests and compared it to our v1.2.0 security issues testing spreadsheet, and with this addition I think we're covered now. Nice work on this and all the other tests @bacongobbler. Refactor is go.

LGTM.

carmstrong:
:ship:

PR #2868, opened 2015-01-08
Use store-mock container in functional tests

mboersma:
This adds on to #2847 to use store-mock in functional tests, rather than the set of real Ceph containers. This is more correct for testing, removes a potential point of failure, and shaves 1.5 minutes off a test run for me (`real 6m4.984s` versus `real 7m34.428s`).

Questions:
- ~~Should we build `store-mock` by default in `make build`?~~ no
- ~~We probably don't need to push `store-mock` to the registry, ever.~~ correct
- ~~Should we move the `mock` directory under `tests/` somewhere instead of under `store/`?~~ yes

Closes #2847.


bacongobbler:
I think `make build` is reserved for building component images. Maybe a `make test/build` target would be appropriate here?

bacongobbler:
> We probably don't need to push store-mock to the registry, ever.

Certainly not. You don't see us pushing mock-etcd around, either!

> Should we move the mock directory under tests/ somewhere instead of under store/

Hmm, good point. It's not necessarily a store component; more of a mock of store. Maybe under `tests/fixtures/mock-store`?

mboersma:
I concur with all your responses @bacongobbler. I'll make it so.

mboersma:
@aledbf I renamed it to "mock-store" and moved it under tests/fixtures/mock-store. Does this look ok to you?

aledbf:
@mboersma sure. The only reason why I used store/mock was the relation with the other components :)

carmstrong:
Code LGTM.

bacongobbler:
code LGTM!

PR #2870, opened 2015-01-09
feat(registry): add authentication for registries (including privates)

aledbf:
This PR allows the authentication against docker registries that require authentication:

1. use `docker login` to add the account/s (is possible to use multiple registries)
2. this create a `.dockercfg` file in $HOME
2. run the utility `read-dockercfg` included in this PR to populate etcd from the `.dockercfg` file
3. `deis-registry` will use the etcd content to create a file (/docker-registry/config/auth.json)
4. in a `deis pull` operation:

 a. the registry checks if in the auth.json exists the registry from where is pulling
 b. in case that exists it will add the authentication headers to the request
 c. make the request with authentication

requires https://github.com/deis/docker-registry/pull/8

closes 2124
closes 2232
closes 2680


aledbf:
ping @bacongobbler @gabrtv

aledbf:
Jenkins, test this please.

bacongobbler:
I don't think this meets the requirements on how we wanted to tackle this problem... Users of the platform should not be allowed to access etcd, only platform administrators. If you could make `read-dockercfg` be a deis-native command then we may be in business here.

aledbf:
@bacongobbler great feedback

> A user who attaches a dockercfg to his account should be available for his or her use only

this is not possible unless the post to deis-registry contains the deis user (currently only src is in the post)

> Config (like keys) should be stored in the database

ok but how deis-registry access that? (without using deis-database as dependency)

> The request should go through the controller

you are right, but as I mentioned in other PR I can't do that in python, sorry about that.

> There should be unit tests available

Agree, but of what?

the "workflow" requires a valid `.dockercfg` file generated by `docker login` (if you pass an invalid file it will just exit) and the go code uses exported functions from docker that already have test

aledbf:
The comments from @bacongobbler are correct. 
This PR is not  "one size fits all" and is not using the expected user interfaces (deis cli and controller). 
But if any user needs authentication knowing the limitations this could provide a temporary solution.

aledbf:
One interesting thing after a couple days of using this PR:
Using the generated .dockercfg in the /root (root user) is not enough to use a registry with authentication.
It's required to run `docker login` with the credentials.

aledbf:
Another thing is that deis-builder also requires the use of `.dockercfg` to push the custom apps

aledbf:
Adding support for `.dockercfg` in the builder is not enough. If the builder script do not call `docker login` before the push it fails.

aledbf:
I found the issue related to .dockercfg https://github.com/coreos/bugs/issues/236

aledbf:
In the controller the method `def _api_call....` is the place to add the registry authentication headers

aledbf:
To be able to integrate docker with a private registry the `docker.service` must contain something like
```
Environment=HOME=/root
ExecStartPost=/usr/bin/sh -c 'docker login -u <username> <repo>'
```

without HOME it will not find the `.dockercfg` file and the login must be called before the access to the registry

aledbf:
In deis-builder the builder template must include the `docker login` before the push

aledbf:
Is required to add `HOME=/root` to `/etc/environment`. Without that any invocation from a service (any deis component or app) to `docker pull/push` will not find the file `.dockercfg`

PR #2872, opened 2015-01-09
fix(contrib/ec2): prepare etcd before starting etcd

carmstrong:
coreos-cloudinit processes units as they are listed in the cloud-config
file, so even though `prepare-etcd-data-directory.service` asks to be
started before etcd with `Before=etcd.service`, etcd has already
been started by the time this unit is encountered.

This commit prepends the Docker volume and etcd data volme preparation
to the main unit list, so they are started before etcd/fleet.

TESTING: Provision an EC2 cluster. You should notice that
`/media/ephemeral/etcd` has `log` and `conf` files, as well as a
`snapshots` directory. The lack of a snapshots directory is what
resulted in a panic when etcd tried to write a snapshot to disk.

### DIFF
--- a/contrib/ec2/gen-json.py
+++ b/contrib/ec2/gen-json.py
@@ -37,7 +37,7 @@ PREPARE_ETCD_DATA_DIRECTORY = '''
   Type=oneshot
   RemainAfterExit=yes
   ExecStart=/usr/bin/mkdir -p /media/ephemeral/etcd
-  ExecStart=/usr/bin/chown -R etcd /media/ephemeral/etcd
+  ExecStart=/usr/bin/chown -R etcd:etcd /media/ephemeral/etcd
 '''
 FORMAT_DOCKER_VOLUME = '''
   [Unit]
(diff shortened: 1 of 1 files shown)


contrib/ec2/gen-json.py
```diff
@@ -62,12 +62,19 @@
...
 data = yaml.load(file(os.path.join(CURR_DIR, '..', 'coreos', 'user-data'), 'r'))
-data['coreos']['units'].append(dict({'name': 'format-ephemeral-volume.service', 'command': 'start', 'content': FORMAT_EPHEMERAL_VOLUME}))
-data['coreos']['units'].append(dict({'name': 'media-ephemeral.mount', 'command': 'start', 'content': MOUNT_EPHEMERAL_VOLUME}))
-data['coreos']['units'].append(dict({'name': 'prepare-etcd-data-directory.service', 'command': 'start', 'content': PREPARE_ETCD_DATA_DIRECTORY}))
-data['coreos']['units'].append(dict({'name': 'format-docker-volume.service', 'command': 'start', 'content': FORMAT_DOCKER_VOLUME}))
-data['coreos']['units'].append(dict({'name': 'var-lib-docker.mount', 'command': 'start', 'content': MOUNT_DOCKER_VOLUME}))
+
+# coreos-cloudinit will start the units in order, so we want these to be processed before etcd/fleet
+# are started
+data['coreos']['units'] = new_units + data['coreos']['units']
```
bacongobbler: A neat trick here, for future reference:

```
data['coreos']['units'][:0] = new_units
```

Both work the same here so I won't ask you to fix this up, but it helps with code duplication at the microscopic level :)

carmstrong: Yeah I actually found that in my Googling, but I like the explicit array merge for folks like me without Python-focused eyes :)

bacongobbler:
code LGTM. Makes sense if coreos-cloudinit parses and starts each unit in order. etcd would've been started but the mountpoint hadn't yet been loaded. Nice catch

bacongobbler:
To avoid this in the future, maybe we can contribute a fix upstream so coreos-cloudinit loads all units before starting them. Not sure if that's the real fix here but it's a thought.

carmstrong:
> Not sure if that's the real fix here but it's a thought.

Yeah, I'm not sure what's intentional behavior. One could argue it's us since we're explicitly asking etcd to be started: https://github.com/deis/deis/blob/18cf1a99d333bde630f1bd3bad8029752d1799a9/contrib/coreos/user-data.example#L22

One could argue pretty easily either way. /me shrugs

mboersma:
Tested on EC2 with the release-1.2.0 branch, works as advertised. :smiley: 

LGTM.

PR #2875, opened 2015-01-11
ref(builder): Remove use of root in gitreceive

### LINKED ISSUE #1843: git push: insufficient permission for adding an object to repository database
This just happened out of nowhere. I was able to push to my deis master just fine until this happened.

```
> git push deis master                                                                                                                                                                            
Counting objects: 9, done.                                                                                                                                                                        
Compressing objects: 100% (5/5), done.
Writing objects: 100% (5/5), 477 bytes | 0 bytes/s, done.
Total 5 (delta 3), reused 0 (delta 0)
error: insufficient permission for adding an object to repository database ./objects

fatal: failed to write object
error: unpack failed: unpack-objects abnormal exit
To ssh://git@deis.staging.asia.gce.redacted.com:2222/redacted-website.git
 ! [remote rejected] master -> master (n/a (unpacker error))
error: failed to push some refs to 'ssh://git@deis.staging.asia.gce.redacted.com:2222/redacted-website.git'
```

Edit:
Permissions log: http://pastebin.com/zjsER1e8

Edit 2:
For anyone else that is experiencing this issue, the fix was we deleted the repository from the builder. It works but it's not ideal.

Blystad:
Changed the builder from running an entire script as root through sudo, to just a single command.

This also fixes #1843, which happened since the builder script was ran as root, and we executed "git gc" as root, causing files and folders to "randomly" be owned by root:root instead of git:git

### DIFF
--- a/builder/image/Dockerfile
+++ b/builder/image/Dockerfile
@@ -38,9 +38,6 @@ RUN useradd -d $GITHOME $GITUSER
 RUN mkdir -p $GITHOME/.ssh && chown git:git $GITHOME/.ssh
 RUN chown -R $GITUSER:$GITUSER $GITHOME
 
-# let the git user run `sudo /home/git/builder` (not writeable)
-RUN echo "%git    ALL=(ALL:ALL) NOPASSWD:/home/git/builder" >> /etc/sudoers
-
 # HACK: import progrium/cedarish as a tarball
 # see https://github.com/deis/deis/issues/1027
 RUN curl -#SL -o /progrium_cedarish.tar.gz \
@@ -55,6 +52,11 @@ ENTRYPOINT ["/app/bin/entry"]
 CMD ["/app/bin/boot"]
 EXPOSE 22
 RUN addgroup --quiet --gid 2000 slug && useradd slug --uid=2000 --gid=2000
+
+# $GITUSER is added to docker group to use docker without sudo and to slug
+# group in order to share resources with the slug user
+RUN usermod -a -G docker $GITUSER
+RUN usermod -a -G slug $GITUSER
 
 ADD templates/shim.dockerfile /home/git/
 ADD etc /etc
--- a/builder/image/templates/builder
+++ b/builder/image/templates/builder
@@ -102,8 +102,13 @@ if [ ! -f Dockerfile ]; then
     BUILD_OPTS+=' -v '
     BUILD_OPTS+=$(echo $CACHE_DIR)
     BUILD_OPTS+=':/tmp/cache:rw'
-    # give non-root slugbuilder user R/W perms for docker volumes
-    chown -R 2000:2000 $TMP_DIR $CACHE_DIR
+    # give slug group ownership of TMP_DIR and CACHE_DIR.
+    chown -R :2000 $TMP_DIR
+    chown :2000 $CACHE_DIR
+    # TMP_DIR is created using mktemp, which sets permissions to 700. Since
+    # we share this with the slug group, the slug group needs to be able to
+    # work with it.
+    chmod g+rwx $TMP_DIR
 
     BUILD_OPTS+=' deis/slugbuilder'
 
--- a/builder/image/templates/builder
+++ b/builder/image/templates/builder
@@ -94,12 +94,6 @@ BUILD_OPTS+=$(echo $RESPONSE | /app/bin/get-app-values)
 # if no Dockerfile is present, use slugbuilder to compile a heroku slug
 # and write out a Dockerfile to use that slug
 if [ ! -f Dockerfile ]; then
-    if [ -f /buildpacks ]; then
-        BUILD_OPTS+=' -v /buildpacks:/tmp/buildpacks:rw'
-        # give non-root slugbuilder user R/W perms for docker volumes
-        chown -R 2000:2000 /buildpacks
-    fi
-
     # run in the background, we'll attach to it to retrieve logs
     BUILD_OPTS+=' -d'
     BUILD_OPTS+=' -v '
--- a/builder/image/templates/gitreceive
+++ b/builder/image/templates/gitreceive
@@ -57,8 +57,7 @@ EOF
         cd $GITHOME
         # if we're processing a receive-pack on an existing repo, run a build
         if [[ $SSH_ORIGINAL_COMMAND == git-receive-pack* ]]; then
-          # SECURITY: git user runs the builder as root (for docker access)
-          sudo $GITHOME/builder $RECEIVE_USER $RECEIVE_REPO $newrev 2>&1 | strip_remote_prefix
+          $GITHOME/builder $RECEIVE_USER $RECEIVE_REPO $newrev 2>&1 | strip_remote_prefix
         fi
 
         rm -f "$LOCKFILE"


builder/image/templates/builder
```diff
@@ -112,8 +112,10 @@ if [ ! -f Dockerfile ]; then
     BUILD_OPTS+=' -v '
     BUILD_OPTS+=$(echo $CACHE_DIR)
     BUILD_OPTS+=':/tmp/cache:rw'
-    # give non-root slugbuilder user R/W perms for docker volumes
-    chown -R 2000:2000 $TMP_DIR $CACHE_DIR
+    # give slug group ownership of TMP_DIR and CACHE_DIR. This is used by the
+    # slug builder. Contents in these folders will be owned by slug:slug.
+    chown :2000 $TMP_DIR $CACHE_DIR
+    chmod g+rwx $TMP_DIR $CACHE_DIR
```
johanneswuerbach: Why are you changing the file permission here (make all executable)?

Blystad: It's not recursive, so it only applies to the folder. [+x on a folder is different to +x on a file, and is necessary to access files inside the folder](https://wiki.archlinux.org/index.php/File_permissions_and_attributes#Folders)

builder/image/templates/builder
```diff
@@ -101,7 +101,7 @@ if [ ! -f Dockerfile ]; then
     if [ -f /buildpacks ]; then
         BUILD_OPTS+=' -v /buildpacks:/tmp/buildpacks:rw'
         # give non-root slugbuilder user R/W perms for docker volumes
-        chown -R 2000:2000 /buildpacks
+        sudo chown -R 2000:2000 /buildpacks
```
johanneswuerbach: Didn't you remove the sudo permission for `git` above?

Blystad: I changed it to allow this single command, since it's not possible for the git user to do this without sudo. See: [https://superuser.com/questions/697608/chown-operation-not-permitted](https://superuser.com/questions/697608/chown-operation-not-permitted)

`+RUN echo "%git    ALL=(ALL:ALL) NOPASSWD:/bin/chmod -R 2000\:2000 /buildpacks" >> /etc/sudoers`

johanneswuerbach: Thanks, now I got it.

johanneswuerbach: Can't we move this chown into the slugbuilder? https://github.com/deis/deis/blob/master/builder/image/slugbuilder/Dockerfile#L7

Blystad: It has to be done at container startup (i think?), so it would require changing the entrypoint in slugbuilder to a new script - executed as root. This script would `chown` and then execute the current `builder/build.sh` script then as the `slug` user.

johanneswuerbach: The docker daemon is running as root (thats why sudo was required) and all commands before https://github.com/deis/deis/blob/master/builder/image/slugbuilder/Dockerfile#L8 are executed as root, so chown-ing the buildpacks dir before switching the user should work (if it was mounted).

Blystad:
I've refactored the solution a bit after feedback from @johanneswuerbach . Turns out the `/buildpacks` override didn't work in the first place, since we checked if `/buildpacks` was a file, not a folder.

In any case I am wondering...is it worth keeping the `/buildpacks` override? Users already have the option of creating a custom builder, and referencing custom buildpacks in the BUILDPACK_URL configuration variable, and through the multi-buildpack buildpack.

builder/image/Dockerfile
```diff
@@ -38,9 +38,6 @@ RUN useradd -d $GITHOME $GITUSER
 RUN mkdir -p $GITHOME/.ssh && chown git:git $GITHOME/.ssh
 RUN chown -R $GITUSER:$GITUSER $GITHOME
 
-# let the git user run `sudo /home/git/builder` (not writeable)
-RUN echo "%git    ALL=(ALL:ALL) NOPASSWD:/home/git/builder" >> /etc/sudoers
```
johanneswuerbach: I guess you can also remove the `sudo` package now: https://github.com/Blystad/deis/blob/builder_non-root_docker/builder/image/Dockerfile#L6

johanneswuerbach:
Awesome! As filling the `/buildpacks` dir would require a custom builder image and it never worked in the first place, I think it's safe to remove this option. What do you think @bacongobbler?

builder/image/Dockerfile
Blystad: I've fixed this now. Good suggestion.

builder/image/templates/builder
bacongobbler: Is there something wrong with the old `chown -R 2000:2000`?

builder/image/Dockerfile
```diff
@@ -3,7 +3,7 @@ FROM ubuntu-debootstrap:14.04
 ENV DEBIAN_FRONTEND noninteractive
 
 # install common packages
-RUN apt-get update && apt-get install -y curl net-tools sudo
+RUN apt-get update && apt-get install -y curl net-tools
```
bacongobbler: let's keep `sudo` in here. We [do this intentionally](https://github.com/deis/deis/blob/master/controller/Dockerfile#L6) in other components such that this layer is shared across all components.

builder/image/templates/builder
Blystad: We're not allowed to do that without root. 
See: https://superuser.com/questions/697608/chown-operation-not-permitted

builder/image/slugbuilder/builder/build.sh
```diff
@@ -19,7 +19,6 @@ buildpack_root=/tmp/buildpacks
 
 mkdir -p $app_dir
 mkdir -p $cache_root
-mkdir -p $buildpack_root
```
bacongobbler: how come this needs to move into the dockerfile?

builder/image/Dockerfile
Blystad: Ack

Blystad:
I'll fix the sudo thing, but what do you think about removing the `/buildpacks` override @bacongobbler ? (See above two posts from me and Johannes).

bacongobbler:
Considering that users can just fork the builder to supply their own list of buildpacks, I don't see a big enough reason to keep it in if it's preventing this fix from being merged. However, I'd rather if we kept it in if it's a non-blocker.

ping @gabrtv

Blystad:
It does work fine with this, I even tested it (which is why I discovered the `/buildpacks` file vs directory check bug in the first place).

These are my motivations for removing it:
1. Seeing as no one has reported it as an issue, it doesn't seem like a very much used feature
2. There are multiple ways to add custom buildpacks
3. Avoid the `/buildpacks` related changes in the PR.

bacongobbler:
I talked to @gabrtv offline about this and he's okay with removing the /buildpacks feature.

Blystad:
Alright. I'll remove `/buildpacks` and re-add `sudo` tomorrow then.

Blystad:
pinging @bacongobbler (in case you havn't seen that I've updated the commits :) )

builder/image/templates/builder
```diff
@@ -112,8 +106,10 @@ if [ ! -f Dockerfile ]; then
     BUILD_OPTS+=' -v '
     BUILD_OPTS+=$(echo $CACHE_DIR)
     BUILD_OPTS+=':/tmp/cache:rw'
-    # give non-root slugbuilder user R/W perms for docker volumes
-    chown -R 2000:2000 $TMP_DIR $CACHE_DIR
+    # give slug group ownership of TMP_DIR and CACHE_DIR. This is used by the
+    # slug builder. Contents in these folders will be owned by slug:slug.
```
bacongobbler: The last comment here is misleading. The contents of these folders are shared by users of the `slug` group, but the contents are still owned by root, I'd assume

bacongobbler:
code LGTM, though this should be tested manually due to the file permission changes.

builder/image/templates/builder
Blystad: So, I checked this a bit further...

`TMP_DIR` is owned by `git:slug`, but the contents inside the directory is owned by `git:git`.
`CACHE_DIR` is owned by git:slug, and the contents (`.m2` / `.maven`) is owned by `slug:slug`.

I'll update the commit once some manual testing has been done, in case you're able to uncover any issues I havn't solved so far.

bacongobbler: Yes, that seems to be the correct behaviour. `TMP_DIR` is synonymous with a buildpack's `$BUILD_DIR` (IOW, the application's source code), which is copied in. `CACHE_DIR` is used by the buildpack to cache assets between builds, so anything built inside there would be owned by `slug:slug`.

> TMP_DIR is owned by git:slug, but the contents inside the directory is owned by git:git.

That won't be good for some buildpacks. IIRC some require write access to some application's source subdirectories for vendoring in assets like gems.  You'll probably want to make the group permissions recursive.

builder/image/templates/builder
bacongobbler: @johanneswuerbach does bring up a good point though. We didn't change file permissions on these folders before, so why do we have to change them now?

builder/image/templates/builder
Blystad: `chown -R :2000 ..` seems reasonable. I didn't encounter issues with this on my Maven builds, but, the folders would then be created in slugbuilders /app folder and, since they weren't part of the git repo, there wouldn't be any problem.

builder/image/templates/builder
Blystad: Right, default permission level for folders is u+rwx,g+rwx,o+rx on Linux...So I dont think it was actually needed. I'll remove it.

bacongobbler:
@Blystad could you please rebase and address the comments when you have time? Thanks!

bacongobbler:
(as a note, I just tested this now and it seems to work for Go/ruby/python)

Blystad:
Planning to do so tomorrow (very busy week at my daytime job). Great that it worked during your manual tests.

Blystad:
There we go, rebased and fixed.

Seems we don't get away from running `chmod g+rwx on $TMP_DIR`.

`$TMP_DIR` is created using `mktemp` which sets [permissions to 700 / u+rwx](http://www.mktemp.org/manual.html). We got away with it before since before we changed the folders owner to `slug` where as we now just change the folders group to `slug`.

bacongobbler:
The test failure seems unrelated. I've kicked another job off for you.

builder/image/templates/builder
```diff
@@ -108,8 +102,13 @@ if [ ! -f Dockerfile ]; then
     BUILD_OPTS+=' -v '
     BUILD_OPTS+=$(echo $CACHE_DIR)
     BUILD_OPTS+=':/tmp/cache:rw'
-    # give non-root slugbuilder user R/W perms for docker volumes
-    chown -R 2000:2000 $TMP_DIR $CACHE_DIR
+    # give slug group ownership of TMP_DIR and CACHE_DIR.
+    chown -R :2000 $TMP_DIR
+    chown :2000 $CACHE_DIR
```
Blystad: Just thought I'd comment on this to avoid any potentional questions on why I had to do it this way:

`chown -R :2000` failed on `CACHE_DIR` since files in that folder is owned by `slug:slug` with permissions `rwxr-xr-x`. As such running recursive chown in this folder resulted in permission denied.

I doubt it will be any issue though, this folder seems only used by slugbuilder.

bacongobbler:
code LGTM

carmstrong:
LGTM. I hand-tested both example-ruby-buildpack and example-dockerfile-python, deployed them, scaled to 3 containers, then created a new commit and pushed that.

mboersma:
I re-ran the test suite locally and then tested this interactively with example-php and haven't found any issues. LGTM.

bacongobbler:
since $TMP_DIR and $CACHE_DIR are not persistent between builder restarts, the upgrade path looks fine as well. another LGTM.

Issue #279, opened 2013-11-01
Make a controller's docs internal


bacongobbler:
Currently, the controller's web UI's docs route to http://docs.deis.io/en/latest/. We should instead have an internal route to the current version of the docs that are residing on the server, so that the version is always for that current Deis controller version. Something like http://controller-fqdn.com/docs/. This would make it much easier for IT admins to reference users to the current version of the docs.

PR #2887, opened 2015-01-13
docs(image) image to test docs

lorieri:
Docker image to help commiters to test documentation changes

### DIFF
--- a/docs/image/Dockerfile
+++ b/docs/image/Dockerfile
@@ -1,0 +1,14 @@
+FROM ubuntu-debootstrap:14.04
+
+RUN apt-get update && apt-get install --yes python make python-dev libffi-dev git curl && apt-get clean
+
+RUN curl -sSL https://raw.githubusercontent.com/pypa/pip/1.5.6/contrib/get-pip.py | python -
+
+ADD ../docs_requirements.txt /
+ADD ./rundocs /
+
+RUN pip install -r /docs_requirements.txt
+
+WORKDIR /
+
+CMD ["/rundocs"]
--- a/docs/image/README.md
+++ b/docs/image/README.md
@@ -1,0 +1,33 @@
+# Docs Test
+this container helps to test the documentation before you push to github
+
+## How-to
+
+### Using local docker
+
+```
+DEIS_ROOT=/home/myuser/deis-git-repository
+
+$ docker build -t myuser/deis-docs .
+$ docker run --rm -t -i -v $DEIS_ROOT:/deis -p 8000:8000 local/deis-docs
+```
+
+Point your browser to http://localhost:8000
+
+### Using remote docker (e.g. boot2docker)
+
+When using a remote docker you must commit your changes and set GIT_REPO so the remote host can use your files.
+If you are using other branch than master, set the GIT_BRANCH too.
+
+```
+$ docker build -t myuser/deis-docs .
+$ docker run --rm -t -i -e "GIT_REPO=https://github.com/deis/deis" -e "GIT_BRANCH=release" -p 8000:8000 myuser/deis-docs
+```
+
+Point your browser to http://remotehost:8000
+
+### Using a HTTP Proxy
+
+```
+docker run -e "http_proxy=http://proxy:3128" -e "https_proxy=http://proxy:3128" --rm -t -i -v $DEIS_ROOT:/deis -p 8000:8000 myuser/deis-docs
+```
--- a/docs/image/rundocs
+++ b/docs/image/rundocs
@@ -1,0 +1,7 @@
+#!/bin/bash
+if [ $GIT_REPO ]   ; then git clone $GIT_REPO /deis || echo "#### PROBLEM CLONING $GIT_REPO" ; fi
+cd /deis
+if [ $GIT_BRANCH ] ; then git checkout $GIT_BRANCH || echo "#### ERROR TO CHECKOUT TO BRANCH $GIT_BRANCH" ; fi
+cd /deis/docs
+pip install -r docs_requirements.txt
+make server


docs/image/rundocs
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+pip install -r docs_requirements.txt
```
johanneswuerbach: Why is this executed here and in the [Dockerfile](https://github.com/deis/deis/pull/2887/files#diff-0f8ac35ad627718a859512ce1c59867cR10)? Removing this would also allow to remove this script and run `make server` directly.

docs/image/README.md
```diff
@@ -0,0 +1,16 @@
+# Docs Test
+this container helps to test the documentation before you push to github
+
+## How-to
+
+```
+$ docker build -t local/deis-docs .
+$ docker run --rm -t -i -v `pwd`/../../../deis/deis:/deis -p 8000:8000 local/deis-docs
+```
```
johanneswuerbach: It would be awesome to have this command available in `docs/Makefile`. Something like `make serve-preview`, which starts the server and prints that the doc preview is now available via `http://localhost:8000`.

mboersma: `make -C docs server` is what I have always used.

mboersma:
Here's how you do this currently:
```console
$ make -C docs test
$ . docs/venv/bin/activate  # so 'sphinx-build' is found
$ make -C docs server
```

docs/image/README.md
johanneswuerbach: Yes, but this requires to setup python, virtualenv, etc. on your system. Having everything already setup in a container a contributor could just download would be awesome.

lorieri:
I run the pip before to cache some dependencies in the image so if you not change it you don't wait for all downloads every time you run it

it would be best if we could have a image from registry

docs/image/README.md
lorieri: :+1:

docs/image/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+### Using a HTTP Proxy
```
bacongobbler: I don't think we need to call this out in the documentation. Users of docker will already know how to set this up on their own, and [docker will use your own proxy settings](https://coreos.com/docs/launching-containers/building/customizing-docker/#use-an-http-proxy) so this won't be necessary.

bacongobbler:
The issue I have with this PR has to do with the volume mount before we run `make server`. Volume mounts do not work on every machine (remote docker daemons e.g. boot2docker), which means this workflow will not work on all systems. Client/server autodocs won't be built without this volume mount, and because `docker build` only works in your current context, I fail to see a workaround unless we pulled the autodocs out (which won't be happening any time soon).

IMO best practices for volume mounts seem to be for saving container state on the host because of this limitation instead of mounting host resources into the container (exceptions to this would be mounting the docker socket, since that is guaranteed).

Other issues:

 - The `docker run -v` relative path arguments are very brittle. I'd rather reference an environment variable like $DEIS_ROOT to be more clear with the intention behind the volume.

If these two issues are addressed and everything else works fine, I'm happy with this as-is.

lorieri:
let me try to set an option to push it from git 

something like:

if [ $GITREPO ] git clone $GITREPO /deis

johanneswuerbach:
@lorieri I think instead of adding the docs using a volume you could just `ADD` them into the container and subsequent container builds are cached by docker to run faster. This would also handle changes in `docs_requirements` nicely.

lorieri:
@johanneswuerbach I would prefer to run from docker registry and never run build.

but it is up to you guys, I only changed docs once

lorieri:
added $GIT_REPO and $GIT_BRANCH for remote docker

changed relative path to $DEIS_ROOT

bacongobbler:
docker 1.5 is coming out with relative ADD commands, which would greatly simplify this PR. https://github.com/icecrime/docker/commit/7e803ba6b2ed52d9e67de8be1cb0eb459017f8cc?short_path=4ac32a7#diff-4ac32a78649ca5bdd8e0ba38b7006a1e

lorieri:
:+1:

docs/image/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM ubuntu-debootstrap:14.04
+
+RUN apt-get update && apt-get install --yes python python-pip postgresql-9.3 make postgresql-server-dev-9.3  python-dev libffi-dev git  && apt-get clean
```
mboersma: @lorieri we removed the need for PostgreSQL in #2946, so you can simplify these requirements.

docs/image/Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+
+RUN easy_install --upgrade pip
```
mboersma: Our standard one-liner for installing `pip` is this:
```Dockerfile
RUN curl -sSL https://raw.githubusercontent.com/pypa/pip/1.5.6/contrib/get-pip.py | python -
```
Could you change this line to match that?

lorieri:
@mboersma done

carmstrong:
@lorieri I think the consensus here is that we'd like to refactor this to use the new ADD commands in Docker 1.5.0. We can start to use that version of Docker for local development.

lorieri:
up to you guys, is there something I have to do ?

carmstrong:
> up to you guys, is there something I have to do ?

If you want to play around with Docker 1.5.0, you can update the PR to use the new ADD command to ADD relative paths to the build. The `docs/image/rundocs` shouldn't re-clone the code -- it should copy it into the container from the local clone of the Deis codebase when the container is built. Does that make sense?

This PR is a good change, but the implementation would be significantly improved by utilizing this feature in the new Docker. If you don't have the time, we can close this for now and resurrect it once one of us has the time.

bacongobbler:
A couple of changes:

- move the Dockerfile into `docs/` (doesn't make sense to put it in it's own rootfs unless we need to compile something and inject it into the rootfs)
- remove the `git clone` script
- switch to `ADD ../controller /controller` so we get the autodocs
- ADD the whole source in (`ADD . /docs`)
- run the sphinx generation tool with a RUN
- change the CMD to statically host the generated docs on some arbitrary port (8080?)

That should be a general gist to get things going

lorieri:
Hi, I'd prefer to close it. I only used docs once, I'm not the best person to make it since I barely use it. I just wanted to help people like me that are not developers and want to help with small fixes and get them merged faster. It is not meant to regular developers.

Some thoughts:

The rootfs and re-clone is meant so we can have a pre-built image, can't we have an image on docker hub with pre-installed dependencies to save time?

Git clone script is a requirement from you to use remote docker daemon, maybe I did'nt get it right, I don't have a clue how to do it in other way, will the relative path work remotely ?

bacongobbler:
It's okay, I'll eventually get to this. I've been wanting to dockerize the docs for a while: #279

lorieri:
I will wait for yours and then make a version were 'docker build' is unnecessary :) 

I wish you guys never work behind a proxy to understand how simple things can become a nightmare sometimes

bacongobbler:
You mean the `docker build` workflow is poor in proxied settings? The daemon should respect http_proxy and HTTP_PROXY flags, according to https://docs.docker.com/articles/systemd/.

> I wish you guys never work behind a proxy

I'd never want to wish that upon someone else :cry:

lorieri:
LOL

Lets say you are not authorized to access all URLs required by the dependencies and you have a internal docker-registry. If you are only doing small changes you can use a pre-built image.

Issue #2883, opened 2015-01-12
builder not displaying underlying error


bacongobbler:
When the builder fails, there is no error output. it just responds with `[remote rejected] master -> master (pre-receive hook declined)` with no reason for the error. This was introduced when we refactored the builder in #2335 to start refactoring the component over to Go.

This has been seen in previous issues such as #2876 and in #1748, so a fixup would help us better determine the underlying issue.

carmstrong:
:100:

bacongobbler:
This seems to have gone away -- if not, please open another issue on the specific error. :)

PR #2888, opened 2015-01-13
fix(builder): return the error message from the controller

aledbf:
Append the message after the generic error message

helps to #2883


builder/bin/publish-release-controller.go
```diff
@@ -77,24 +77,26 @@ func main() {
 
 	res, err := client.Do(req)
 
+	// Read json response from body
+	body, _ := ioutil.ReadAll(res.Body)
```
mboersma: We're ignoring a possible `error` return from `ioutil.ReadAll` here, then reporting any error that `client.Do` may have raised. I think we should handle `client.Do`'s error immediately afterward, and then have similar error handling for `ioutil.ReadAll`. I realize this isn't your fault, the error handling here was a bit disjointed already, and this is one of the uglier aspects of go coding, but I think explicit handling is better here.

aledbf:
The reason to print the error after the static message is because there is multiple reasons why the error occurred. In #2870 for instance the controller could not access the remote registry receiving a 401 when it was trying to push a release.

mboersma:
This didn't seem to work for me in testing, I was still getting the generic error message even through the controller had this:
```
Jan 15 18:36:43 deis-2 sh[3011]: File "/app/api/models.py", line 168, in scale
Jan 15 18:36:43 deis-2 sh[3011]: 'Container type {} does not exist in application'.format(container_type))
Jan 15 18:36:43 deis-2 sh[3011]: EnvironmentError: Container type web does not exist in application
Jan 15 18:36:43 deis-2 sh[3011]: 172.17.8.100 "POST /v1/hooks/build HTTP/1.1" 500 - "deis-builder"
```
I got the same generic error message when I tried to `git push deis master` without having added a key. Should this PR be covering those cases, or should I be testing other specific failures?

aledbf:
@mboersma I think that the title of this PR could be mis interpreted. The changes that I made only cover errors during the build phase and related to the go apps inside the builder script.

PR #2891, opened 2015-01-13
feat(*): support RFC 6598 for insecure registries

carmstrong:
An additional address range, 100.64.0.0/10, is reserved by RFC 6598.
This is also the private address space used by Azure, so this will
also allow us to support Azure without building a custom builder
component just for Azure, which was the previously-planned approach.

refs #2824


bacongobbler:
code LGTM

mboersma:
Code LGTM.

PR #2892, opened 2015-01-13
feat(client): add flag to disable SSL certificate verification

### LINKED ISSUE #1499: feat(client): add flag to disable SSL certificate verification
For users testing Deis with SSL, they may not be using a real cert right away. We should add a client flag which disables SSL verification for the Python requests.

mcano:
Resolves #1499. Addresses the issue when a user may not be using a proper 
certificate in their deis environment. This change allows the user to disable SSL
verification via a command line option during auth:login. The result is saved in the
user's client.json and is passed in subsequent commands made via the
client.

### DIFF
--- a/client/deis.py
+++ b/client/deis.py
@@ -390,6 +390,7 @@ class DeisClient(object):
         func = getattr(self._session, method.lower())
         controller = self._settings.get('controller')
         token = self._settings.get('token')
+        ssl_verify = self._settings.get('ssl_verify')
         if not token:
             raise EnvironmentError(
                 'Could not find token. Use `deis login` or `deis register` to get started.')
@@ -399,7 +400,7 @@ class DeisClient(object):
             'X-Deis-Version': __api_version__.rsplit('.', 1)[0],
             'Authorization': 'token {}'.format(token)
         }
-        response = func(url, data=body, headers=headers)
+        response = func(url, data=body, headers=headers, verify=ssl_verify)
         # check for version mismatch
         server_api_version = response.headers.get('X_DEIS_API_VERSION')
         if server_api_version is not None and server_api_version != __api_version__:
@@ -789,8 +790,11 @@ class DeisClient(object):
             provide a username for the account.
           --password=<password>
             provide a password for the account.
+          --ssl-verify=false
+            disables SSL certificate verification for API requests
         """
         controller = args['<controller>']
+        ssl_verify = True
         if not urlparse.urlparse(controller).scheme:
             controller = "http://{}".format(controller)
         username = args.get('--username')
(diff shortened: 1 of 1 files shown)


deis-admin:

client/deis.py
```diff
@@ -800,15 +804,20 @@ def auth_login(self, args):
         password = args.get('--password')
         if not password:
             password = getpass('password: ')
+        disable_ssl_verify = args.get('--disable_ssl_verify')
```
carmstrong: We seem to prefer using hyphens rather than underscores for option names, so this would become `--disable-ssl-verify`. I'm also wondering if `--no-ssl-verify` works just as well - it's certainly shorter!

bacongobbler: for consistency with other projects, `--ssl-verify=false` or `--tlsverify=false` seems to be the common theme.

https://github.com/berkshelf/berkshelf/blob/5d9fae311d4580e3d29027c86ad8e1ba83cb0362/lib/berkshelf/cli.rb#L188-L191
https://github.com/docker/docker/blob/8b95ad230e2ee76450ceb9a80aa8e942a56bb397/docker/docker.go#L93

mcano:
Updated to --ssl-verify=false.

client/deis.py
```diff
@@ -789,8 +790,11 @@ def auth_login(self, args):
             provide a username for the account.
           --password=<password>
             provide a password for the account.
+          --ssl-verify=false
+            disables ssl certificate verification
```
bacongobbler: Maybe something a bit more clear on the intent: `disables SSL certificate verification for API requests.`

client/deis.py
```diff
@@ -824,6 +833,7 @@ def auth_logout(self, args):
         self._settings['controller'] = None
         self._settings['username'] = None
         self._settings['token'] = None
+        self._settings['ssl_verify'] = None
```
bacongobbler: this is starting to get a bit unwieldly. Would you mind changing it to the following?

```python
for i in ['controller', 'username', 'token', 'ssl_verify']:
    self._settings[i] = None
```

client/deis.py
```diff
@@ -789,8 +790,11 @@ def auth_login(self, args):
             provide a username for the account.
           --password=<password>
             provide a password for the account.
+          --ssl-verify=false
+            disables ssl certificate verification for API requests
```
carmstrong: Could you change this to "SSL" (capitalized) here? Sorry for the nitpick! We just reviewed this and it's good to go otherwise.

mcano:
Updated to SSL.

bacongobbler:
Looks great! One more thing before we merge this: could you please squash your commits into the first? If not I'd be happy to do it for you :)

mcano:
Done.

carmstrong:
:ship:

mboersma:
This `--ssl-verify` option change LGTM.

bacongobbler:
thanks @mcano :)

PR #2896, opened 2015-01-14
chore(controller): update Django to 1.6.10 security release

mboersma:
See https://www.djangoproject.com/weblog/2015/jan/13/security/


gabrtv:
Code LGTM.  Given our use of the WSGI headers mentioned in the CVE we should consider a patch release.

carmstrong:
Code LGTM

PR #2898, opened 2015-01-14
chore(controller): bump django-rest-framework to 3.0.3

bacongobbler:
this is the beginning of #2854. Considering that bumping to DRF 3.0.3 broke a lot of old functionality, I wanted to mainly focus on bumping DRF before completely refactoring so there's less to review.

TODO:

 - [x] add unit tests to ensure the serialized response for all objects stays the same

### DIFF
--- a/controller/api/models.py
+++ b/controller/api/models.py
@@ -25,12 +25,12 @@ from django.dispatch import receiver
 from django.utils.encoding import python_2_unicode_compatible
 from django_fsm import FSMField, transition
 from django_fsm.signals import post_transition
-from docker.utils import utils
+from docker.utils import utils as dockerutils
 from json_field.fields import JSONField
 import requests
 from rest_framework.authtoken.models import Token
 
-from api import fields
+from api import fields, utils
 from registry import publish_release
 from utils import dict_diff, fingerprint
 
@@ -44,6 +44,15 @@ def log_event(app, msg, level=logging.INFO):
     app.log(msg)            # local filesystem
 
 
+def validate_id_is_docker_compatible(value):
+    """
+    Check that the ID follows docker's image name constraints
+    """
+    match = re.match(r'^[a-z0-9-]+$', value)
+    if not match:
+        raise ValidationError("App IDs can only contain [a-z0-9-].")
+
+
 def validate_app_structure(value):
     """Error if the dict values aren't ints >= 0."""
     try:
@@ -52,6 +61,12 @@ def validate_app_structure(value):
                 raise ValueError("Must be greater than or equal to zero")
     except ValueError, err:
         raise ValidationError(err)
+
+
+def validate_reserved_names(value):
+    """A value cannot use some reserved names."""
+    if value in ['deis']:
+        raise ValidationError('{} is a reserved name.'.format(value))
 
 
 def validate_comma_separated(value):
@@ -97,7 +112,9 @@ class App(UuidAuditedModel):
     """
 
     owner = models.ForeignKey(settings.AUTH_USER_MODEL)
-    id = models.SlugField(max_length=64, unique=True)
+    id = models.SlugField(max_length=64, unique=True, default=utils.generate_app_name,
+                          validators=[validate_id_is_docker_compatible,
+                                      validate_reserved_names])
     structure = JSONField(default={}, blank=True, validators=[validate_app_structure])
 
     class Meta:
@@ -334,7 +351,7 @@ class App(UuidAuditedModel):
 
         # check for backwards compatibility
         def _has_hostname(image):
-            repo, tag = utils.parse_repository_tag(image)
+            repo, tag = dockerutils.parse_repository_tag(image)
             return True if '/' in repo and '.' in repo.split('/')[0] else False
 
         if not _has_hostname(image):
@@ -587,6 +604,29 @@ class Config(UuidAuditedModel):
 
     def __str__(self):
         return "{}-{}".format(self.app.id, self.uuid[:7])
+
+    def save(self, **kwargs):
+        """merge the old config with the new"""
+        try:
+            previous_config = self.app.config_set.latest()
+            for attr in ['cpu', 'memory', 'tags', 'values']:
+                # Guard against migrations from older apps without fixes to
+                # JSONField encoding.
+                try:
+                    data = getattr(previous_config, attr).copy()
+                except AttributeError:
+                    data = {}
+                try:
+                    new_data = getattr(self, attr).copy()
+                except AttributeError:
+                    new_data = {}
+                data.update(new_data)
+                # remove config keys if we provided a null value
+                [data.pop(k) for k, v in new_data.items() if v is None]
+                setattr(self, attr, data)
+        except Config.DoesNotExist:
+            pass
+        return super(Config, self).save(**kwargs)
 
 
 @python_2_unicode_compatible
--- a/controller/api/serializers.py
+++ b/controller/api/serializers.py
@@ -9,10 +9,11 @@ import re
 
 from django.conf import settings
 from django.contrib.auth.models import User
+from django.utils import timezone
 from rest_framework import serializers
+from rest_framework.validators import UniqueTogetherValidator
 
 from api import models
-from api import utils
 
 
 PROCTYPE_MATCH = re.compile(r'^(?P<type>[a-z]+)')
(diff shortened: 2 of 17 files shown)


carmstrong:
Looks like the app is still in the database after a destroy:

```
=== appssample
BUILDPACK_URL: https://example.com
Git remote deis added
ok
apps:destroy --confirm=appssample
Destroying appssample... 
 ... done in 0s
Git remote deis removed
ok
apps:create appssample
Creating application...  ...  o.. done, created appssample
Git remote deis added
ok
apps:create appssample
Creating application...  ... 400 BAD REQUEST
{u'id': [u'This field must be unique.']}
error at command wait
--- FAIL: TestApps (12.01s)
	itutils.go:266: exit status 1
FAIL
exit status 1
FAIL	github.com/deis/deis/tests	12.010s
godep: go exit status 1
```

bacongobbler:
It's because the error code changed and the integration tests was expecting something else. I'll fix that up.

bacongobbler:
Seems like containers are not being deployed:

```
><> deis builds:create deis/example-go:latest --app=deispullsample --procfile="worker: while true; do echo hi; sleep 3; done"
Creating build... done, v2
><> fleetctl list-units
UNIT				MACHINE			ACTIVE	SUB
deis-builder.service		58adf8bc.../10.21.12.1	active	running
deis-cache.service		af60f361.../10.21.12.3	active	running
deis-controller.service		af60f361.../10.21.12.3	active	running
deis-database.service		00f7b8cc.../10.21.12.2	active	running
deis-logger.service		58adf8bc.../10.21.12.1	active	running
deis-logspout.service		00f7b8cc.../10.21.12.2	active	running
deis-logspout.service		58adf8bc.../10.21.12.1	active	running
deis-logspout.service		af60f361.../10.21.12.3	active	running
deis-publisher.service		00f7b8cc.../10.21.12.2	active	running
deis-publisher.service		58adf8bc.../10.21.12.1	active	running
deis-publisher.service		af60f361.../10.21.12.3	active	running
deis-registry@1.service		00f7b8cc.../10.21.12.2	active	running
deis-router@1.service		00f7b8cc.../10.21.12.2	active	running
deis-router@2.service		58adf8bc.../10.21.12.1	active	running
deis-router@3.service		af60f361.../10.21.12.3	active	running
deis-store-daemon.service	00f7b8cc.../10.21.12.2	active	running
deis-store-daemon.service	58adf8bc.../10.21.12.1	active	running
deis-store-daemon.service	af60f361.../10.21.12.3	active	running
deis-store-gateway.service	00f7b8cc.../10.21.12.2	active	running
deis-store-metadata.service	00f7b8cc.../10.21.12.2	active	running
deis-store-metadata.service	58adf8bc.../10.21.12.1	active	running
deis-store-metadata.service	af60f361.../10.21.12.3	active	running
deis-store-monitor.service	00f7b8cc.../10.21.12.2	active	running
deis-store-monitor.service	58adf8bc.../10.21.12.1	active	running
deis-store-monitor.service	af60f361.../10.21.12.3	active	running
deis-store-volume.service	00f7b8cc.../10.21.12.2	active	running
deis-store-volume.service	58adf8bc.../10.21.12.1	active	running
deis-store-volume.service	af60f361.../10.21.12.3	active	running
><> fleetctl status deis-controller
● deis-controller.service - deis-controller
   Loaded: loaded (/run/fleet/units/deis-controller.service; linked-runtime)
   Active: active (running) since Thu 2015-01-15 19:10:31 UTC; 3h 23min ago
 Main PID: 3699 (sh)
   CGroup: /system.slice/deis-controller.service
           ├─3699 /bin/sh -c IMAGE=`/run/deis/bin/get_image /deis/controller` && docker run --name deis-controller --rm -p 8000:8000 -e EXTERNAL_PORT=8000 -e HOST=$COREOS_PRIVATE_IPV4 -v /var/run/fleet.sock:/var/run/fleet.sock -v /var/lib/deis/store:/data $IMAGE
           └─3709 docker run --name deis-controller --rm -p 8000:8000 -e EXTERNAL_PORT=8000 -e HOST=10.21.12.3 -v /var/run/fleet.sock:/var/run/fleet.sock -v /var/lib/deis/store:/data leeroy.bacongobbler.com/deis/controller:git-2a78486

Jan 15 22:31:51 deis-3 sh[3699]: 10.21.12.1 "POST /v1/apps/test/builds HTTP/1.0" 201 - "python-requests/2.3.0 CPython/2.7.8 Darwin/14.0.0"
Jan 15 22:32:16 deis-3 sh[3699]: 2015-01-15T22:32:16Z 33febd7b79f0 confd[181]: ERROR Config check failed: exit status 1
Jan 15 22:33:12 deis-3 sh[3699]: 10.21.12.1 "DELETE /v1/apps/test HTTP/1.0" 204 - "python-requests/2.3.0 CPython/2.7.8 Darwin/14.0.0"
Jan 15 22:33:25 deis-3 sh[3699]: INFO deispullsample: config deispullsample-4e319b8 updated
Jan 15 22:33:25 deis-3 sh[3699]: INFO deispullsample: release deispullsample-v1 created
Jan 15 22:33:25 deis-3 sh[3699]: 10.21.12.1 "POST /v1/apps HTTP/1.0" 201 - "python-requests/2.3.0 CPython/2.7.8 Darwin/14.0.0"
Jan 15 22:33:35 deis-3 sh[3699]: INFO deispullsample: build deispullsample-6c95e92 created
Jan 15 22:33:35 deis-3 sh[3699]: INFO deispullsample: release deispullsample-v2 created
Jan 15 22:33:36 deis-3 sh[3699]: INFO deispullsample: bacongobbler scaled containers cmd=1
Jan 15 22:33:52 deis-3 sh[3699]: 10.21.12.1 "POST /v1/apps/deispullsample/builds HTTP/1.0" 201 - "python-requests/2.3.0 CPython/2.7.8 Darwin/14.0.0"
```

controller/api/models.py
```diff
@@ -519,7 +536,6 @@ class Push(UuidAuditedModel):
     class Meta:
         get_latest_by = 'created'
         ordering = ['-created']
-        unique_together = (('app', 'uuid'),)
```
mboersma: Why did we remove the `unique_together` Meta fields? This effectively changes the database schema by dropping a `UNIQUE` constraint: https://docs.djangoproject.com/en/1.6/ref/models/options/#unique-together

bacongobbler: The reasoning behind this is because Django Rest Framework **demands** the two keys to be present during a serializer's `.create()`, which in our case we do not create the `uuid` key until just before we [save it in the database](https://github.com/deis/deis/blob/c28e4ace5f22db14985a996aa00269aae96e7ea8/controller/api/fields.py#L31-L39).

From http://www.django-rest-framework.org/api-guide/serializers/:

> The ModelSerializer class is the same as a regular Serializer class, except that:

>- It will automatically generate a set of fields for you, based on the model.
>- It will automatically generate validators for the serializer, such as unique_together validators.
>- It includes simple default implementations of .create() and .update().

We could work around this by somehow disabling this specific UniqueTogether validation in the Serializers. Would that be reasonable?

controller/api/models.py
```diff
@@ -97,7 +112,9 @@ class App(UuidAuditedModel):
     """
 
     owner = models.ForeignKey(settings.AUTH_USER_MODEL)
-    id = models.SlugField(max_length=64, unique=True)
+    id = models.SlugField(max_length=64, unique=True, default=utils.generate_app_name,
+                          validators=[validate_id_is_docker_compatible,
```
gabrtv: :+1:

controller/api/serializers.py
```diff
@@ -69,71 +70,60 @@ class AdminUserSerializer(serializers.ModelSerializer):
...
+    owner = serializers.ReadOnlyField(source='owner.username')
+    structure = JSONFieldSerializer(required=False)
     created = serializers.DateTimeField(format=settings.DEIS_DATETIME_FORMAT, read_only=True)
     updated = serializers.DateTimeField(format=settings.DEIS_DATETIME_FORMAT, read_only=True)
 
     class Meta:
         """Metadata options for a :class:`AppSerializer`."""
         model = models.App
-
-    def validate_id(self, attrs, source):
```
gabrtv: I'm happy to move this validation logic into the model, however, I want to make sure we're still throwing 400's with proper error messages if the validations fail.

controller/api/tests/test_app.py
```diff
@@ -139,7 +139,7 @@ def test_app_errors(self):
         body = {'id': 'deis'}
         response = self.client.post(url, json.dumps(body), content_type='application/json',
                                     HTTP_AUTHORIZATION='token {}'.format(self.token))
-        self.assertContains(response, "App IDs cannot be 'deis'", status_code=400)
+        self.assertContains(response, 'deis is a reserved name.', status_code=400)
```
gabrtv: :+1:

controller/api/serializers.py
bacongobbler: see https://github.com/deis/deis/pull/2898/files#diff-05fcaef4cb94bfc712de7e3b13658284R82

controller/api/models.py
mboersma: It's just that this changes the schema definition:
```sql
CREATE TABLE "api_push" (
    "uuid" uuid NOT NULL PRIMARY KEY,
    "created" timestamp with time zone NOT NULL,
    "updated" timestamp with time zone NOT NULL,
    "owner_id" integer NOT NULL,
    "app_id" uuid NOT NULL REFERENCES "api_app" ("uuid") DEFERRABLE INITIALLY DEFERRED,
    "sha" varchar(40) NOT NULL,
    "fingerprint" varchar(255) NOT NULL,
    "receive_user" varchar(255) NOT NULL,
    "receive_repo" varchar(255) NOT NULL,
    "ssh_connection" varchar(255) NOT NULL,
    "ssh_original_command" varchar(255) NOT NULL,
    UNIQUE ("app_id", "uuid")
)
;
```
That last `UNIQUE` constraint / index is no longer in the table definition--ditto for two other tables. So now the database won't guarantee that there isn't an `api_push` table entry with duplicates for the `app` foreign key and the `uuid` field. I think that should remain in the schema definition. (And if not, we should provide a Django migration to alter those tables and drop the constraint.)

bacongobbler: I totally agree that this bump should not change our schema definitions. I'll look for a way around this. :)

controller/api/serializers.py
```diff
@@ -22,46 +23,64 @@
...
-    """Serialize a User model."""
+class ModelSerializer(serializers.ModelSerializer):
 
+    uuid = serializers.ReadOnlyField()
+
+    def get_validators(self):
+        """
+        Hack to remove DRF's UniqueTogetherValidator when it concerns the UUID.
+
+        See https://github.com/deis/deis/pull/2898#discussion_r23105147
```
mboersma: I think this is a better fix, even if it is a hack. Leaves the the schema intact and allows us to change the behavior in one place if DRF's validation behavior changes in this regard.

controller/api/serializers.py
```diff
@@ -22,46 +23,64 @@
...
-        for f in ('password',):
-            if f in d:
-                del d[f]
-        return d
+        fields = ['email', 'username', 'password', 'first_name', 'last_name', 'is_superuser',
+                  'is_staff', 'groups', 'user_permissions', 'last_login', 'date_joined',
+                  'is_active']
+        read_only_fields = ['is_superuser', 'is_staff', 'groups',
+                            'user_permissions', 'last_login', 'date_joined', 'is_active']
+        extra_kwargs = {'password': {'write_only': True}}
```
mboersma: +1 for `{'password': {'write_only': True}}`.

controller/api/tests/test_app.py
```diff
@@ -69,6 +67,22 @@ def test_app(self):
...
+                                    HTTP_AUTHORIZATION='token {}'.format(self.token))
+        for key in response.data.keys():
+            self.assertIn(key, ['uuid', 'created', 'updated', 'id', 'owner', 'url', 'structure'])
+        expected = {
+            'id': 'test',
+            'owner': self.user.username,
+            'url': 'test.deisapp.local',
+            'structure': {}
+        }
+        self.assertDictContainsSubset(expected, response.data)
```
mboersma: +1 for `assertDictContainsSubset`. :white_check_mark:

controller/api/urls.py
```diff
@@ -1,41 +1,36 @@
...
-from api import routers
 from api import views
 
 
-router = routers.ApiRouter()
-
 # Add the generated REST URLs and login/logout endpoint
 urlpatterns = patterns(
     '',
-    url(r'^', include(router.urls)),
```
mboersma: This was serving...the browseable REST API pages?

controller/api/views.py
```diff
@@ -2,176 +2,117 @@
...
 """
 
-from __future__ import absolute_import
-from __future__ import unicode_literals
-
 from django.conf import settings
-from django.contrib.auth.models import AnonymousUser, User
 from django.core.exceptions import ValidationError
-from django.utils import timezone
-from guardian.shortcuts import assign_perm
```
mboersma: I try to be [PEP8 compliant](https://www.python.org/dev/peps/pep-0008/#imports) by putting each import on a separate line, although `flake8` doesn't enforce this.

controller/api/urls.py
bacongobbler: so it may seem. I guess we'd still want them, even if we don't necessarily recommend them at the moment. :P

controller/api/views.py
bacongobbler: as you mentioned, if it's all from the same root package then the comma-separated `import` is PEP8 compliant. It's just the root imports that need separate line statements :)

controller/api/urls.py
mboersma: If possible, we should leave them in. I think the browseable API pages had their own login page so they were actually functional, since I didn't see how to use token auth with them. But I hadn't checked in a while.

mboersma:
These changes to our REST view code LGTM. (The views.py diff was really hard to read for some reason, it was easier to look at the source.)

controller/api/serializers.py
bacongobbler: totally agree :)

gabrtv:
Code LGTM.  We are leaning heavily on our test coverage here, which is good.

Thanks for leaving the refactoring out of this PR, I hope to see that one soon!  :smile:

PR #2901, opened 2015-01-15
docs(using_deis): add disabling registration docs to user creation

carmstrong:
@brynary suggested that it'd be helpful to note how to disable user registration after instructing users to create their superuser account.


brynary:
Nice. Thanks, Chris!

bacongobbler:
docs LGTM!

mboersma:
LGTM.

PR #2902, opened 2015-01-15
docs(system-requirements): Fix broken link

jgmize:


deis-admin:

bacongobbler:
I think [etcd disaster recovery](https://github.com/coreos/etcd/blob/master/Documentation/admin_guide.md#disaster-recovery) is more relevant here since we're specifically talking about the number of machines etcd requires to maintain quorum. lgtm otherwise!

jgmize:
@bacongobbler I agree, and I've amended the commit to link there instead.

carmstrong:
Thanks for finding this, @jgmize. It looks like we also use this old link on both:

docs/installing_deis/aws.rst
docs/installing_deis/rackspace.rst

Would you mind fixing all of those as well?

jgmize:
@carmstrong sure, I've amended the commit to fix those links, as well.

carmstrong:
LGTM. Thanks @jgmize!

PR #2903, opened 2015-01-15
docs(managing_deis): clarify reasons for disabling CoreOS upgrades

carmstrong:
etcd/fleet incompatibilities used to be a problem, but not so
much anymore - we should stop calling that out.

ping @gabrtv


gabrtv:
Language LGTM.  While I want to enable rolling updates, they're simply too disruptive today.

bacongobbler:
lgtm. Haven't seen any incompatibility issues in quite some time over here, either.

mboersma:
That is a better explanation, LGTM.

PR #2905, opened 2015-01-15
Refactor deisctl config logic and add docs on Router SSL

### LINKED ISSUE #2712: deisctl config router set sslCert="$(cat cert.crt)" not working
deisctl 1.0.2
deis 1.0.2
deis cluster running on 3 digitalocean instances (4GB) 

When attempting to set the router sslCert and sslKey, I get the following:

$ deisctl config router set sslCert="$(cat cert.crt)"
Error: invalid argument: sslCert=-----BEGIN CERTIFICATE-----
......
......
......
-----END CERTIFICATE-----

So I tried to get into deis-1 and set it directly via etcdctl:

$ etcdctl set /deis/router/sslCert $(cat cert.crt)
Incorrect Usage.

NAME:
   set - set the value of a key

USAGE:
   command set [command options] [arguments...]

DESCRIPTION:

OPTIONS:
   --ttl '0'                    key time-to-live
   --swap-with-value            previous value
   --swap-with-index '0'        previous index

I've made sure the cert.crt exists in the directory I am running the commands above from and echo $(cat cert.crt) returns the certificate.

gabrtv:
This PR fixes #2712 and cleans up the config logic in `deisctl`:

 * defines `fileKeys` which are read from local files
 * defines `b64Keys` which are base64-encoded before stored (needed for `sshPrivateKey`)
 * `valueForPath` is factored out with corresponding test coverage

I've also added documentation on how to configure the Deis routers to terminate SSL.

### DIFF
--- a/deisctl/config/config.go
+++ b/deisctl/config/config.go
@@ -9,6 +9,15 @@ import (
 	"github.com/deis/deis/deisctl/utils"
 )
 
+// fileKeys define config keys to be read from local files
+var fileKeys = []string{
+	"/deis/platform/sshPrivateKey",
+	"/deis/router/sslCert",
+	"/deis/router/sslKey"}
+
+// b64Keys define config keys to be base64 encoded before stored
+var b64Keys = []string{"/deis/platform/sshPrivateKey"}
+
 // Config runs the config subcommand
 func Config(args map[string]interface{}) error {
 	return doConfig(args)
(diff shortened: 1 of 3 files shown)


mboersma:
Code LGTM.

deisctl/config/config.go
```diff
@@ -106,13 +104,31 @@ func doConfigGet(client *etcdClient, root string, keys []string) ([]string, erro
...
+		if path == p {
+
+			// read value from filesystem
+			bytes, err := ioutil.ReadFile(utils.ResolvePath(v))
+			if err != nil {
+				return "", err
+			}
+
+			// see if we should return base64 encoded value
+			for _, pp := range b64Keys {
```
carmstrong: So b64keys can only be used if they are also fileKeys, and only makes sense in that context? I guess that makes sense.

carmstrong:
Code LGTM.

deisctl/config/config.go
gabrtv: Yep.  The base64 encoding is to sanitize multi-line input, which is typically coming from a file.

PR #2907, opened 2015-01-16
fix(controller): track create and start separately to catch early errors

### LINKED ISSUE #2655: `deis run` hangs on failed container
If you do a `deis run` and the container `fail`s, the command just hangs with no response, and never cleans up the scheduled admin container in fleet.

for example, before PR'ing https://github.com/deis/deis/pull/2642, I was testing with https://github.com/deis/example-dockerfile-http which does not have `bash` present in the image (we use that as the entrypoint for run commands):

```
><> deis run echo hi
running 'echo hi'...
^C
><> fleetctl list-units | grep dock_v2.run.1
dock_v2.run.1.service		2280c10f.../10.21.12.2	failed		failed
```

This is not a regression that was recently introduced, just a bug that's been there for a while.

gabrtv:
This PR fixes #2655 by monitoring for errors during container creation and container start.  While normal error cases (i.e. `exit 1`) were being caught and handled properly, in the case of #2655 the binary we were executing did not exist.  This caused the container to fail at create-time, which caused hangs during state polling.

### DIFF
--- a/controller/scheduler/fleet.py
+++ b/controller/scheduler/fleet.py
@@ -256,14 +256,27 @@ class FleetHTTPClient(object):
             rc, output = chan.recv_exit_status(), out.read()
             return rc, output
 
-        # wait for container to start
-        for _ in range(1200):
+        # wait for container to launch
+        for _ in range(60):
             rc, _ = _do_ssh('docker inspect {name}'.format(**locals()))
             if rc == 0:
                 break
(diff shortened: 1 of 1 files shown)


mboersma:
This is better logic and will fail faster in some important cases. Code LGTM.

carmstrong:
Code LGTM

Issue #1608, opened 2014-08-15
wildcard subdomain needs to be supported to host some heroku apps


dreamcodez:
Hi, there was one other issue which I saw related to this:

https://github.com/deis/deis/pull/563

I was wondering if this was still on the table, and if so when? I have a couple apps that I can't host on deis because of this currently.

Otherwise what is the workaround? Hack nginx config directly?

NOTE: If you can point me to work in progress branch, I can probably finish the work. Just sayin'

Thanks.

The error received after typing "deis domains:add '*.mydomain.com'" is :
{u'domain': [u'Adding a wildcard subdomain is currently not supported']}

jwaldrip:
+1 we need this as well!

christofferh:
+1

jnettome:
Any updates on that? +1

brunoalano:
+1

rukor:
Any idea if this is in the pipeline?

carmstrong:
> Any idea if this is in the pipeline?

This change won't make it into Deis v1, but @helgi is working on it for Deis v2 (see https://github.com/deis/workflow/pull/338). The opportunity to make breaking changes gave us the ability to reimplement our models to better represent things like this.

rukor:
Thanks

helgi:
Looks like #4762 will get into the LTS and the improvements be forced ported to v2 on top of all of my changes. It may be slightly different in v2 due to model changes @carmstrong mentioned but it should be mostly the same.

Hope it helps.

helgi:
Forward ported to v2*

bacongobbler:
this has been implemented in v2, [which is now generally available](https://deis.com/blog/2016/workflow-stable/). Closing!

PR #2911, opened 2015-01-16
Proposal: SSL support for custom domains

### LINKED ISSUE #18: Application SSL support

### LINKED ISSUE #1873: Add integration tests for Domains
Work on #1872 pointed out that we don't currently have integration tests that cover domains.  We need to cover:

1. Basic domain add/removal workflow
2. Test etcd keys are populated and purged properly

bacongobbler:
This is an initial COA for custom domain SSL support. Please read through the documentation and let me know what you think. Please feel free to scrutinize, nitpick and question all you like. Once it's been approved I will implement the feature, call for another round of review and then finally merge.

# TODO/THOUGHTS

- [x] figure out how permissions will work (who can attach certs to which domains?)
- [x] determine the final database schema
- [x] would the `Domain` model change to become a first-class `Route` instead?

# FAQ

## Why are the certs being attached to the domains?

We already have wildcard SSL support for the platform itself. [Heroku models it in this fashion as well](https://devcenter.heroku.com/articles/ssl-endpoint), so the default app routes would rely on old behaviour:

> SSL Endpoint is only useful for custom domains. All default appname.herokuapp.com domains are SSL-enabled already and can be accessed simply by using https, e.g., https://appname.herokuapp.com.

## Why does the client not specify the common name for the domain when adding a new certificate?

That information is already available [inside the certificate](http://unix.stackexchange.com/a/103464), so there shouldn't be a need to be explicit about it.

closes #18
closes #1873


carmstrong:
ping @nathansamson @jparkerCAA for feedback

nathansamson:
Hi

Looks okay to me...

Few questions
1) What happens if you add a cert with a CN that doesn't match any of the specified domains? Does it still accept it (with a warning), in the expectation you will add the custom domain  later?
2) What happens if you add a cert with a CN that does match the CN from a previously uploaded cert, eg with a greater expiry date? (I guess once a cert is created is is valid and doesn't have a valid since in the future you can safely overwrite the other one)

I guess thats it for me, it seems to follow the logic of heroku, which makes sense.... Now up to implementation :)

bacongobbler:
> Few questions
> 1) What happens if you add a cert with a CN that doesn't match any of the specified domains? Does it still accept it (with a warning), in the expectation you will add the custom domain later?

It will return HTTP_400_BAD_REQUEST. The expectation is that the custom domain [must be present](https://github.com/deis/deis/pull/2911/files#diff-98fe74250471c27aef86277741604ebdR41) before you add the domain.

> 2) What happens if you add a cert with a CN that does match the CN from a previously uploaded cert, eg with a greater expiry date? (I guess once a cert is created is is valid and doesn't have a valid since in the future you can safely overwrite the other one)

The certificate and key will be replaced, granted that you have the correct permissions to edit the domain's certificates.

carmstrong:
Could we use wildcard certificates with this proposal? If I have `coolapp.example.com` and `coolapp2.example.com` and I own a wildcard cert for `*.example.com`, I'd want to upload the cert once and then use it for both of these custom domains.

bacongobbler:
I've just started working on this. Please let me know if I missed anything in the tests. Otherwise, work is underway!

bacongobbler:
There's still a few things I need to fix up on the docs/controller end, but basic functionality is complete! :tada

```
$ deis domains:add www.bacongobbler.com
$ deis certs:add ~/.openssl/www.bacongobbler.com.cert ~/.openssl/private.key.nopass
$ dig www.bacongobbler.com
[...]
CNAME	go.fishworks.io.
$ curl https://www.bacongobbler.com
Powered by Deis
```

client/deis.py
```diff
@@ -20,6 +20,7 @@
   limits        manage resource limits for your application
   tags          manage tags for application containers
   releases      manage releases of an application
+  certs         manage ssl endpoints for an app
```
carmstrong: Let's capitalize all uses of "SSL" and "SSH" while we're in here.

carmstrong:
Some things I want to make sure we're testing off the bat:

* Support for certchains (i.e. intermediate cert appended to my domain cert should work as one "cert" uploaded to Deis)
* Attaching a wildcard cert to multiple domains
* Attaching a normal cert (i.e. subdomain.domain.com) to a domain which is already covered by a wildcard cert (i.e. a.domain.com, b.domain.com)

bacongobbler:
To answer some of your points:

> Support for certchains (i.e. intermediate cert appended to my domain cert should work as one "cert" uploaded to Deis)

I have not yet tested that, but I'd imagine that if nginx supports certchains in the ssl_certificate file, then that should work (the API may behave differently though, considering that it loads the certificate to retrieve the common name via https://github.com/pyca/pyopenssl)

> Attaching a wildcard cert to multiple domains

I tried to keep this PR simple. Wildcard certs are not supported yet -- it's just a 1-to-1 mapping between domains and certs in the API, though if you want that in for this release I'm game. :space_invader:

> Attaching a normal cert (i.e. subdomain.domain.com) to a domain which is already covered by a wildcard cert (i.e. a.domain.com, b.domain.com)

Again just trying to keep it simple. The API tries to determine the common name for the cert, and if it detects that a domain exists for that common name it will attach the cert, otherwise it'll get rejected. I just wanted to keep it as simple as possible and tackle the wildcard cert problem once we start tackling #1608. :)

Hope this helps!

controller/api/models.py
```diff
@@ -110,6 +112,16 @@ def validate_domain(value):
         raise ValidationError('"{}" contains unexpected characters'.format(value))
 
 
+def validate_domain_certificate(value):
+    try:
+        cert = crypto.load_certificate(crypto.FILETYPE_PEM, value)
+        Domain.objects.get(domain=cert.get_subject().CN)
+    except crypto.Error as e:
+        raise ValidationError('Could not load certificate: {}'.format(e))
+    except Domain.DoesNotExist:
```
gabrtv: Why do we require this?  Doesn't seem like this is a strict requirement, unless I'm missing something.

controller/api/models.py
```diff
@@ -820,12 +832,45 @@ class Domain(AuditedModel):
...
     app = models.ForeignKey('App')
     domain = models.TextField(blank=False, null=False, unique=True)
+    cert = models.ForeignKey('DomainCert', null=True)
 
     def __str__(self):
         return self.domain
 
 
 @python_2_unicode_compatible
+class DomainCert(AuditedModel):
```
gabrtv: Why `DomainCert` and not `Cert` or `Certificate`?  We refer to them as "certs" all over the place.. the CLI, the URL patterns, the etcd keyspace -- as well as as when we talk about them generally.

controller/api/models.py
bacongobbler: I thought that heroku's ssl addon **requires** that a domain needs to be added first. That doesn't seem to be the case upon closer inspection so I'll remove the check.

controller/api/models.py
bacongobbler: The idea behind `DomainCert` here was because certificates are intimately tied to a domain. It's cosmetic so I'd be happy to change if it would help clarify things across the board.

gabrtv: I prefer changing it, especially if we're removing the tight coupling to domains (which I prefer).

gabrtv:
@bacongobbler nice work so far!  The developer UX seems like what we want.  The cron implementation is clever.  My big concern at the moment is how we handle cert chains -- which are quite common.  To make sure we're covered, let's add a test case for them in the unit tests.

With regard to wildcard domains, I agree we can tackle that later if needed.  My guess is most folks will want to put a wildcard cert on the router to handle SSL for _all_ their apps, as you proposed.  This PR is really about custom SSL for specific apps, which will often have a custom domain.

I look forward to testing this!

bacongobbler:
@gabrtv check has been removed and DomainCert is no more :)

bacongobbler:
> My big concern at the moment is how we handle cert chains -- which are quite common. To make sure we're covered, let's add a test case for them in the unit tests.

Absolutely. Heroku handles them quite well according to their documentation... I'll have to look into that and hash that out. Asides from that this should be good to go.

One of my concerns with this PR is the "clever" cron job may actually kill nginx until the certificate and keys are populated from the cronjob. `nginx -t` expects the file to actually be available when validating the config, so we may have to check that out. Might also want to check with multiple certs and see if they all get populated.

gabrtv:
@bacongobbler it sounds like we should start putting together a detailed manual test plan that covers all these edge cases.  As usual, if any of the testing can be automated, let's write the tests now or at least open issues to do so.

smothiki:
Tested manually LGTM -- needs integration tests

bacongobbler:
Actually I realized that we don't have any domain integration tests (#1873). Looks like I'm killing another bird with this stone ;)

bacongobbler:
yusss :party:

bacongobbler:
things still left to cover:

 - [ ] cert chains

bacongobbler:
Tested it manually:

```
><> cat bacongobbler.com.cert bacongobbler.com.ca > bacongobbler.com.int
><> deis-init
Registered bacongobbler
Logged in as bacongobbler
Uploading me@bacongobbler.com to Deis...done
><> deis certs
No certs
><> deis certs:add bacongobbler.com.int private.key.nopass
Adding SSL endpoint... done
www.bacongobbler.com
```

I'll update the documentation on how users can add their own chain certs, if it's not in there already.

carmstrong:
@bacongobbler Are we just waiting on Jenkins?

bacongobbler:
yes... I'm just fixing up one last thing. ELB doesn't listen on port 443!

bacongobbler:
we finally got a green. @deis we ok to merge, or do you guys want another pass of reviews?

mboersma:
Code LGTM:exclamation:

PR #2915, opened 2015-01-18
feat(builder): replace cedarish with herokuish

aledbf:
This proposal:
- [x] removes docker in docker
- [x] eliminates the use of --privileged
- [x] removes the ENTRYPOINT
- [x] replaces slugbuilder with herokuish
- [x] removes the progrium/cedarish from the builder (and all the issues loading the tgz in docker)


aledbf:
Jenkins, test this please.

aledbf:
Jenkins, test this please.

aledbf:
ping @bacongobbler @gabrtv

gabrtv:
This looks very interesting.  Big +1's for moving to herokuish and removing the docker-in-docker requirement.  Will explore in more detail..

builder/image/bin/boot
```diff
@@ -47,29 +48,9 @@ done
...
-    gzip -dc /progrium_cedarish.tar.gz | docker load
-else
-    echo "Cedarish already loaded"
-fi
-
-# build required images
-docker build -t deis/slugbuilder /app/slugbuilder/
+# do not use cedarish from the registry
+# use the id from the opdemand cedarish tgz
+docker tag -f 4b458f40de0a progrium/cedarish:latest
```
bacongobbler: `docker load` from the tarball will give us `progrium/cedarish:latest` so this logic isn't necessary.

builder/image/bin/boot
```diff
@@ -15,6 +15,7 @@ export ETCD="$HOST:$ETCD_PORT"
 export ETCD_PATH=${ETCD_PATH:-/deis/builder}
 export ETCD_TTL=${ETCD_TTL:-10}
 export STORAGE_DRIVER=${STORAGE_DRIVER:-btrfs}
+export SLUGRUNNER="/deis/slugrunner"
```
bacongobbler: What is this for? This is unused in both the builder and in herokuish, AFAICT.

builder/image/bin/boot
aledbf: @bacongobbler for some reason it tries to use the image from the registry.

builder/image/bin/boot
aledbf: @bacongobbler you are right. I will remove that

builder/image/bin/boot
aledbf: I saw the "problem" after a couple of builds when I checked the parents of deis-builder and the apps using `docker images -tree`.

builder/image/templates/builder
```diff
@@ -75,50 +80,14 @@ if [ -f Dockerfile ]; then
...
-if [ $CODE -ne 0 ]; then
-    puts-warn $RESPONSE
-    exit 1
-fi
-
-BUILD_OPTS=()
-BUILD_OPTS+='/usr/bin/docker'
-BUILD_OPTS+=' run '
-# get application configuration
-BUILD_OPTS+=$(echo $RESPONSE | /app/bin/get-app-values)
```
bacongobbler: I think this is too important of a feature to lose. A build should have all of the relevant envvars present which the user set via `deis config:set`.

aledbf: @bacongobbler during the build?

bacongobbler: yes

johanneswuerbach: Shouldn't the `ENV_DIR` be used for that? https://devcenter.heroku.com/articles/buildpack-api#bin-compile-summary

aledbf: @johanneswuerbach you are right and herokuish has support for that  https://github.com/gliderlabs/herokuish/blob/db6541268f130e0700990d36d64a50cfdbda31ef/include/procfile.bash#L42

aledbf:
Jenkins, test this please.

aledbf:
Jenkins, test this please.

builder/image/Dockerfile
```diff
@@ -61,5 +52,7 @@ ADD etc /etc
 ADD . /app
 RUN chown -R root:root /app
 
+RUN /app/bin/install-buildpacks.sh
```
johanneswuerbach: I think it's is better to use the herokuish buildpack versions as those are tested. BP versions should be aligned soon:
https://github.com/gliderlabs/herokuish/pull/9
https://github.com/gliderlabs/herokuish/pull/10

builder/image/Dockerfile
```diff
@@ -13,6 +13,10 @@ RUN curl -sSL -o /usr/local/bin/etcdctl https://s3-us-west-2.amazonaws.com/opdem
 RUN curl -sSL -o /usr/local/bin/confd https://s3-us-west-2.amazonaws.com/opdemand/confd-v0.5.0-json \
     && chmod +x /usr/local/bin/confd
 
+# install herokuish
+RUN curl -sSL https://github.com/aledbf/herokuish/releases/download/v0.1.2/herokuish_0.1.2_linux_x86_64.tgz \
```
johanneswuerbach: I think it would be better to use the official release and upstream all changes to have full BP test coverage. @progrium would you mind creating a release when all changes required for this PR are in herokuish?

aledbf: @johanneswuerbach I want that too but in the current implementation he variable (slug_path)[https://github.com/gliderlabs/herokuish/issues/5] is a constant so that is a problem in case of concurrent builds

builder/image/Dockerfile
aledbf: @johanneswuerbach I just did that to keep control over the buildpack versions. I don't have any opinions about this.

builder/image/Dockerfile
johanneswuerbach: @aledbf could you send him a PR. I don't see any obvious issue with building multiple apps in the same container when we separate all dirs carefully. Somebody could kill the entire builder with a custom BP, but this is already a problem in our existing implementation.

aledbf: @johanneswuerbach the thing is that there is no docker in docker to build an app inside `deis-builder`. It executes `herokuish` directly.

johanneswuerbach: Sure, but herokuish is running all builds as different unprivileged users. Shouldn't this forbid accessing files of other users?

aledbf: @johanneswuerbach  yes but this is a problem [slug.bash](https://github.com/gliderlabs/herokuish/blob/master/include/slug.bash#L2) when you need to run `herokuish generate`

progrium:
Thank you for banging on Herokuish and working out the kinks. If you can write tests for any fix PRs to Herokuish, that would be rad.

johanneswuerbach:
@progrium sure, the only issue I encountered so far is fixed here: https://github.com/gliderlabs/herokuish/pull/10

bacongobbler:
@johanneswuerbach could you please elaborate more on this?

> Somebody could kill the entire builder with a custom BP, but this is already a problem in our existing implementation

AFAIK there's no way for someone to mess with someone else's build because each build is a new container (deis/slugbuilder). Even if you suply a custom BUILDPACK_URL, it will [only affect your builds](https://github.com/deis/deis/blob/master/builder/image/templates/builder#L115-L117).

Personally I'm a little concerned with running all the builds in the same container, as filesystem changes can modify another user's build.

johanneswuerbach:
@bacongobbler currently the slugbuilder has no resource constrains. This would allow me to kill the builder using a custom buildpack, which does something really cpu or memory intensive. Keeping builds inside containers would allow us to constrain builds itself in the future.

I talked with @aledbf earlier today and he is looking into building the slugs directly inside a new container (and mounting the app, cache, etc. as volumes), a bit like @progrium is doing it here https://github.com/progrium/buildstep/pull/129

carmstrong:
I think the concern here is moving the building of apps from an isolated environment to a process within the builder component itself, but @bacongobbler is our area expert and can add more to the conversation.

bacongobbler:
The best solution I've heard so far would be to move towards utilizing https://github.com/progrium/buildstep/pull/129 inside builder. That way we can drop both slugbuilder and slugrunner in favour of buildstep.

progrium:
And actually if I learn more about builder, buildstep is going to soon
become buildkit with various build strategies, including the default
dokku/buildstep/herokuish building. So I can be sure to include your
requirements in there.

On Thu, Feb 12, 2015 at 3:31 PM, Matthew Fisher <notifications@github.com>
wrote:

> The best solution I've heard so far would be to move towards utilizing
> progrium/buildstep#129 <https://github.com/progrium/buildstep/pull/129>
> inside builder. That way we can drop both slugbuilder and slugrunner in
> favour of buildstep.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/deis/deis/pull/2915#issuecomment-74156956>.
>



-- 
Jeff Lindsay
http://progrium.com

bacongobbler:
Sounds good. In that case let's close this out and collaborate on a solid API for buildkit, since that will significantly help simplify our build infrastructure. Once that's done we can revisit the builder again and run another round of refactoring centralized around buildkit's workflow.

PR #2918, opened 2015-01-18
fix(store): use hostnames for mon host

carmstrong:
Commit 690b380 had a typo which made the `mon host` config entry for
Ceph contain IP addresses, not hostnames. This doesn't seem to
be an issue for day-to-day operations, but could have unforeseen
consequences. It's best to fix this up.

### DIFF
--- a/store/base/templates/ceph.conf
+++ b/store/base/templates/ceph.conf
@@ -1,7 +1,7 @@
 [global]
 fsid = {{ .deis_store_fsid }}
 mon initial members = {{ .deis_store_monSetupLock }}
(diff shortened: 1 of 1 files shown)


gabrtv:
Code LGTM. Not sure what the implications are though.

carmstrong:
> Not sure what the implications are though.

Yeah, not sure there are any - it hasn't seemed to affect operations. For reference, here is the description of that config value: http://ceph.com/docs/master/rados/configuration/mon-config-ref/#minimum-configuration

mboersma:
Code LGTM.

bacongobbler:
code LGTM. Should be completely compatible with older releases, and seems to be a fixup from a previous PR.

Issue #2923, opened 2015-01-20
deis-controller too many clients


ngpestelos:
Please refer to this [gist](https://gist.github.com/karlo57/9307d81649020882f0eb).

Has anyone encountered the same problem?

carmstrong:
This looks like an error returned by the database. A few users have reported this, but not under normal operating load.

To confirm, you could try entering deis-database with `nse deis-database`, editing https://github.com/deis/deis/blob/1fbe1f8c9192de0d13d295c5c6590ef80efebe47/database/templates/postgresql.conf#L64 to a higher value, and then reloading postgres.

ngpestelos:
Thanks, @carmstrong.

ngpestelos:
Here's what we did:

1. `nse deis-database`
2. `sed -i 's/max_connections = 100/max_connections = 200/' /etc/postgresql/9.3/main/postgresql.conf
3. `kill -HUP <postgres-daemon-pid>`

carmstrong:
@ngpestelos You could also PR a change to that value to something like 250 - pretty sure it's either default or arbitrary.

carmstrong:
They could also be stale connections - we may want to lower the timeout value for clients.

PR #2924, opened 2015-01-20
fix(database): increases max_connections to 250

ngpestelos:
For context, please refer to https://github.com/deis/deis/issues/2923


carmstrong:
Code LGTM. A few users have reported seeing that error when stress-testing Deis.

mboersma:
Code LGTM. It's not that hard to use up > 100 connections temporarily, and 250 is still a conservative value for PostgreSQL.

Issue #2920, opened 2015-01-19
deis-controller generated fleet units should [X-Fleet] Conflicts with the others of the same app


ianblenke:
When there is an instance failure, units migrate to the remaining fleet machines. Over time, as fleet machines fail and are terminated by the AWS auto-scaling group and replaced with new instances,  this tends to leave the longest-running fleet machine with a good chance of having multiple if not all units for an app on it.

This just happened to me this evening. I had a 10 minute production outage where the deis-registry, deis-store-gateway, and one of my critical web apps had all 3 units somehow allocated on the the same fleet machine which suddenly stopped responding (thanks AWS!) The 3 app web units were trying to be relocated by fleet, but they were trying to docker pull from the deis-registry which was waiting on the deis-store-gateway to finish starting up which was waiting on ceph which was running in that normal degraded slow way ceph does when one of its monitors/OSDs isn't up. It did eventually repair itself, but that web app wasn't being serviced in the interim.

While it might be fair to say "well fleet shouldn't schedule all 3 on the same machine!" - how is Fleet to know?

At the moment, deis-controller generates fleet units without any kind of affinity for the Fleet scheduler to know better than to allocate all of the units of an app on the same fleet machine.

While using `deis tags` would afford us the ability to tag generated units with fleet machine affinity, it still doesn't prevent the occurrence of a fleet tagged machine from acquiring multiple or all units for an app. Meaning: fleet will merrily schedule multiple units to any available fleet machine with the app's tags.

What we need is a way of tagging deis-controller generated fleet units with an [X-Fleet] Conflicts.

For example, a myapp v3 web proctype unit would ideally have this appended to it:

    [X-Fleet]
    Conflicts=myapp_v3.web.*

Granted, on an N instance deis fleet cluster, this limits it to `deis ps:scale web=N`. Any more than N, and the units simply won't get scheduled by fleet.

As Conflicts is a simple glob, in order to allow overcommits of units, we would need to introduce a "oversubscription" ordinal component to the generated unit service name.

For example, if myapp were on a 3 machine fleet cluster, and we set `deis ps:scale web=6`, in order to allow for oversubscription of 2 (and only 2) units on any given fleet machine, we would need to generate service names like:

    myapp_v3.web.1.1
    myapp_v3.web.1.2
    myapp_v3.web.1.3
    myapp_v3.web.2.1
    myapp_v3.web.2.2
    myapp_v3.web.2.3

And have a unit [X-Fleet] Conflicts something like:

[...]

Issue #2725, opened 2014-12-11
Proposal: scale deis-store-gateway


aledbf:
Similar to `deis-router`, one per node and add the instances in the router

carmstrong:
closed by #2929

PR #2929, opened 2015-01-20
feat(store): Scalable store-gateway

johanneswuerbach:
Allow the store gateway to be scaled across multiple cluster nodes. This speeds up app rescheduling when the node hosting the gateway died

Helps with cases like https://github.com/deis/deis/issues/2920 
Basis for: https://github.com/deis/deis/issues/2725

An inactive gateway uses ~40mb memory, so it might make sense to schedule two gateways by default. https://www.dropbox.com/s/3uwz2brmn188ouu/Screenshot%202015-01-20%2015.36.00.png?dl=0

### DIFF
--- a/deisctl/README.md
+++ b/deisctl/README.md
@@ -39,7 +39,7 @@ $ deisctl install platform
 Logging subsystem...
 deis-logger.service: loaded
 Storage subsystem...
-deis-store-gateway.service: loaded
+deis-store-gateway@1.service: loaded
 Control plane...
 deis-cache.service: loaded
 deis-database.service: loaded
@@ -66,7 +66,7 @@ deis-logspout.service: running
 Storage subsystem...
 deis-store-daemon.service: running
 deis-store-monitor.service: running
-deis-store-gateway.service: running
+deis-store-gateway@1.service: running
 Control plane...
 deis-cache.service: running
 deis-database.service: running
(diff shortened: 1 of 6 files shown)


carmstrong:
Shouldn't we also change the nginx template to write all upstreams and round-robin between them?

johanneswuerbach:
Yes, but as the nginx access is only used by external tools and not deis itself I limited this PR to only deis internal changes for improved HA.

carmstrong:
> Yes, but as the nginx access is only used by external tools and not deis itself I limited this PR to only deis internal changes to improve HA.

:thumbsup:

carmstrong:
Code LGTM.

carmstrong:
We should account for #2995 with this, too.

carmstrong:
LGTM! I deployed a cluster, scaled up to 3 gateways, killed the first gateway, and then pushed a new version of my app. Things went off without a hitch.

I'm going to address #2995 in a separate PR.

bacongobbler:
yeah that's related but we can fix it later

code LGTM!

deisctl/README.md
```diff
@@ -84,7 +84,7 @@ Done.
 
 Note that the default start command activates 1 of each component.
 You can scale components with `deisctl scale router=3`, for example.
-The router and the registry are the only component that _currently_ scales beyond 1 unit.
+The router, the registry and the store gateway are the only component that _currently_ scales beyond 1 unit.
```
mboersma: Note for @carmstrong: subject-verb disagreement here should be cleaned up when you submit your other PR.

mboersma:
Code LGTM.

PR #2932, opened 2015-01-20
chore(contrib/rackspace): bump CoreOS to 522.5.0

carmstrong:


carmstrong:
Turns out this is 494.5.0 - Rackspace isn't there yet.

PR #2933, opened 2015-01-20
fix(controller): return 403 when a user does not have permission

### LINKED ISSUE #2858: Some API calls return 404 instead of 403 when a user doesn't have permission
The following CLI commands will return 404 instead of 403 (as they should):
* info
* scale
* logs
* run
* destroy

Related: #2854

bacongobbler:
closes #2858 

rebased off of #2898 since those are the main blockers for this change.

### DIFF
--- a/controller/api/tests/test_app.py
+++ b/controller/api/tests/test_app.py
@@ -10,7 +10,6 @@ import json
 import mock
 import os.path
 import requests
-import unittest
 
 from django.conf import settings
 from django.contrib.auth.models import User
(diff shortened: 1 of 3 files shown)


mboersma:
This needs to be rebased, but LGTM!

bacongobbler:
rebased

bacongobbler:
note that the code comes straight from https://github.com/tomchristie/django-rest-framework/blob/a7567efa8d6fd008ba0a48f0e8fa7028703af386/rest_framework/mixins.py#L39-L46

carmstrong:
:ship:

PR #2934, opened 2015-01-21
[WIP] chore(registry): update docker-registry to v0.9.1

mboersma:
See https://github.com/docker/docker-registry/compare/0.9.0...0.9.1


carmstrong:
Is the boto hack a known issue on their end?

mboersma:
I copied the boto `patch` from their Dockerfile. It was added here: docker/docker-registry#678

carmstrong:
> I copied the boto patch from their Dockerfile.

Gotcha. Code LGTM.

bacongobbler:
as noted offline, you will need to apply these patches to v0.9.1: https://github.com/deis/docker-registry/compare/new-repository-import

aledbf:
I'm trying to update the regsitry and I found a little detail:
```
Creating build... 503 SERVICE UNAVAILABLE
PUT Checksum Error (412: {"error": "X-Docker-Checksum-Payload not set.  If you are using the Docker daemon, you should upgrade to version 0.10 or later"})
```

So this requires to update the docker api in the controller

bacongobbler:
good catch. Our registry module in the controller is roughly based on docker 0.8's API calls as far as I can remember, so if there's a breaking change there then we'll have to patch that up as well.

mboersma:
I've done more work on this but won't be able to finish it for a few days, so I'm closing this PR until that time.

PR #2935, opened 2015-01-21
ref(Vagrantfile): remove redundant virtio NIC section

mboersma:
The paravirtualized "virtio" adapter is the default when creating a CoreOS 522.5.0 vagrant / VirtualBox instance, so this stanza is no longer needed.


carmstrong:
Code LGTM.

bacongobbler:
:+1:

PR #2941, opened 2015-01-21
fix(controller): return proper json messages on error

### LINKED ISSUE #2512: The controller must return json even for errors
Doing a push I received `aborting, failed to destroy some containers` in plain text (from `/v1/hooks/build`)

### LINKED ISSUE #2843: REST errors question

bacongobbler:
closes #2512
closes #2843

### DIFF
--- a/controller/api/__init__.py
+++ b/controller/api/__init__.py
@@ -2,4 +2,4 @@
 The **api** Django app presents a RESTful web API for interacting with the **deis** system.
 """
 
-__version__ = '1.1.1'
+__version__ = '1.1.2'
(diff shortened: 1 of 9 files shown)


carmstrong:
Code LGTM.

mboersma:
This code should make our responses consistently JSON formatted. I looked at `views.py` and it looks like this covers everything. Code LGTM.

PR #2945, opened 2015-01-22
ref(controller): API view refactor

### LINKED ISSUE #2854: ref(controller): API views need refactoring

bacongobbler:
The majority of code shifting was done earlier in #2898, so this optimizes the rest of the viewsets to use more generic calls, specifically the hooks and app perm viewsets which were basically re-implementing the permission calls.

closes #2854

### DIFF
--- a/controller/api/permissions.py
+++ b/controller/api/permissions.py
@@ -42,6 +42,20 @@ class IsOwner(permissions.BasePermission):
             return False
 
 
+class IsOwnerOrAdmin(permissions.BasePermission):
+    """
+    Object-level permission to allow only owners of an object or administrators to access it.
+    Assumes the model instance has an `owner` attribute.
+    """
+    def has_object_permission(self, request, view, obj):
+        if request.user.is_superuser:
+            return True
+        if hasattr(obj, 'owner'):
+            return obj.owner == request.user
+        else:
+            return False
+
+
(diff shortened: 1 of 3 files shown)


mboersma:
I had to re-read http://www.django-rest-framework.org/topics/3.0-announcement/ to grok this. Code LGTM.

controller/api/permissions.py
```diff
@@ -5,6 +5,18 @@
...
 
 
+def is_app_user(request, obj):
+    if request.user.is_superuser or \
+            isinstance(obj, models.App) and obj.owner == request.user or \
+            hasattr(obj, 'app') and obj.app.owner == request.user:
+        return True
+    elif request.user.has_perm('use_app', obj) or \
+            hasattr(obj, 'app') and request.user.has_perm('use_app', obj.app):
+        return request.method != 'DELETE'
```
gabrtv: We should have a comment explaining this logic.  Took me a while to work it out.

gabrtv:
Code LGTM.  The cleanup seems pretty straightforward.

PR #2946, opened 2015-01-23
ref(docs): remove postgresql from doc generation requirements

mboersma:
To automatically generate python code documentation, Sphinx needs to import the code, which requires Django, which requires a database driver. This change sets the driver to python's built-in `sqlite3` at runtime to simplify requirements for building the docs.

Refs #2887.

### DIFF
--- a/docs/docs_requirements.txt
+++ b/docs/docs_requirements.txt
@@ -17,7 +17,6 @@ djangorestframework==3.0.3
 docker-py==0.6.0
 gunicorn==19.1.1
 paramiko==1.15.2
-psycopg2==2.5.4
 python-etcd==0.3.2
 PyYAML==3.11
 South==1.0.2
--- a/.gitignore
+++ b/.gitignore
@@ -40,6 +40,7 @@ deisctl/package/
 deisctl/makeself/
 docs/_build/
 docs/docs.zip
+docs/dummy.sqlite3
 logspout/build/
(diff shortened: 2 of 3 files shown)


carmstrong:
Code LGTM.

docs/docs_requirements.txt
```diff
@@ -17,7 +17,6 @@ djangorestframework==3.0.3
 docker-py==0.6.0
 gunicorn==19.1.1
 paramiko==1.15.2
-psycopg2==2.5.4
```
bacongobbler: are there other requirements here which we could also remove, or will it need every other dependency to generate the controller autodocs?

bacongobbler:
my comment is more of a question if there's ways we can further remove controller dependencies from docs_requirements.txt, but this PR LGTM!

docs/docs_requirements.txt
mboersma: I think everything else in the first block is imported directly by our code, so the other packages need to remain. (`psycopg2` was imported indirectly, as a string value at runtime.) Same for the client packages and other items in `doc_requirements.txt`.

We could try to see if there are others we can remove, but it's quite tedious to try:
```console
$ # first remove a package from docs_requirements.txt, then
$ rm -rf docs/venv
$ make -C docs test
```

I'm fairly confident we can't remove any others, but we should test that assumption at some point.

PR #2947, opened 2015-01-23
ref(docs): remove local_settings.py hack for secret key

mboersma:
Doc generation creates a local_settings.py file, because at one time having `SECRET_KEY` unset caused an error when importing Django code. That was essentially a temp file that was never cleaned up, and it no longer seems to be necessary.


carmstrong:
Code LGTM.

bacongobbler:
code LGTM!

Issue #1601, opened 2014-08-15
AWS doesn't support subnets in us-east-1a


clayzermk1:
CloudFormation doesn't support subnets in us-east-1a. The defaults tried to make one there.

I cloned master@1aa32511 and followed the instructions in `contrib/ec2/README.md`. (I ran into the EC2-Classic/ELB/SG issue in 0.10.0 so I decided to try master.)

cloudformation.json
```
[
    {
        "ParameterKey":     "InstanceType",
        "ParameterValue":   "t1.micro"
    },
    {
        "ParameterKey":     "KeyPair",
        "ParameterValue":   "deis"
    }
]
```

CloudFormation log:

> Value (us-east-1a) for parameter availabilityZone is invalid. Subnets can currently only be created in the following availability zones: us-east-1b, us-east-1c, us-east-1d.

![screen shot 2014-08-14 at 5 56 21 pm](https://cloud.githubusercontent.com/assets/1075015/3929060/b7bda794-2417-11e4-86bb-e3b7dda42c4e.png)

deis looks amazing, I'm really looking forward to getting it going. I'm going to look around for a way to specify the availability zone. Thanks for all your hard work everyone! Cheers :beers:

carmstrong:
Closing, as there's nothing actionable. Folks who run into this can find the issue.

PR #2949, opened 2015-01-23
Change default Availability Zones' index

beechnut:
I received the following error when running `/provision-ec2-cluster.sh` in the us-east-1, which resulted in the deletion of my CloudFormation stack every time:

`Value "us-east-1a" for parameter availabilityZone is invalid. Subnets can currently only be created in the following availability zones: us-east-1b, us-east-1e, us-east-1c, us-east-1d.`

AZs are selected by choosing the 0th and 1st returned AZs, but the 0th one in us-east-1 is us-east-1a, which is an invalid zone. I changed it to select the 1st and 2nd (1b and 1c, probably), and it builds the stack successfully.


deis-admin:

mboersma:
I think **currently** is the salient word in their error message, however. If you tried tomorrow, you might find `us-east-1a` is back and `us-east-1b` is now unavailable. So shifting the index by one isn't really a safe or long-term fix.

carmstrong:
Yep. And, unfortunately, your availability zones are not mine - see #1601. There's nothing we can account for here unless we rewrite the provision script to probe for AZs and try them all, one-by-one.

mboersma:
And what's really going to bake your noodle:

> your availability zones are not mine

That makes it doubly hard.

> unless we rewrite the provision script to probe for AZs and try them all, one-by-one.

That seems like the right way to do it. Not simple though.

carmstrong:
> And what's really going to bake your noodle:

woah

beechnut:
Thanks, y'all -- that was a great crash course in Availability Zones.

PR #2950, opened 2015-01-23
ref(tests): remove unnecessary time.Sleep()

mboersma:
The artificial delay in `RunCommandWithStdoutStderr` has always been there, without explanation. I couldn't discern why, so I removed it and ran test-integration.sh three times without any errors. Hopefully Jenkins agrees, as this shaves a couple minutes off the overall test time.

### DIFF
--- a/tests/utils/utils.go
+++ b/tests/utils/utils.go
@@ -14,7 +14,6 @@ import (
 	"path/filepath"
 	"strings"
 	"syscall"
-	"time"
 )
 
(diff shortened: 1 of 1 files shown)


carmstrong:
:ship:

bacongobbler:
code LGTM if it goes green :+1:

mboersma:
This failed once with what appeared to be an unrelated error during `git push`. I'm going to run it a few more times to see if that error recurs, and let this PR gestate for a while to make sure it isn't problematic.

PR #2951, opened 2015-01-23
feat(store): add store-admin container

### LINKED ISSUE #2925: store: add deis-store-admin container for management tasks
While troubleshooting #2914 with @mgartner, I realized that we have no easy way to get a shell in a store component that **isn't running**, which is necessary when forcefully recovering a Ceph cluster that has lost quorum. I also realized we don't publish `deis-store-base` to Docker Hub, so I couldn't run that interactively. I figured out a reasonable workaround with `docker run -i --volumes-from=deis-store-monitor-data --entrypoint=/bin/bash deis/store-monitor:v1.2.0`, but being able to run `deis/store-admin` would be nice :) This container would also run confd periodically to re-template `/etc/ceph/ceph.conf`, as most of the store containers only use the config file to connect to the monitors on startup, so don't care if they're up-to-date (whereas a user certainly would).

carmstrong:
Adds a new optional component, `deis-store-admin`, to aid in diagnosing and recovering from issues with store.

Most of the store containers only write the Ceph config file once before the /bin/boot script execs the Ceph daemon, so when a user comes along and tries to query cluster state, it's likely that he's using a very old config file and will see a bunch of monitor faults. This admin container uses confd to rewrite the config file constantly, ensuring that the latest cluster state is reflected in the config.

This change also moves the Troubleshooting Store docs into its own page, since they were overwhelming the Troubleshooting Deis docs. I had to include a toctree in the main Troubleshooting docs to satisfy Sphinx.

closes #2925

### DIFF
--- a/store/Makefile
+++ b/store/Makefile
@@ -1,8 +1,10 @@
 include ../includes.mk
 
-TEMPLATE_IMAGES=daemon gateway metadata monitor
+TEMPLATE_IMAGES=admin daemon gateway metadata monitor
 BUILT_IMAGES=base $(TEMPLATE_IMAGES)
 
+ADMIN_IMAGE = $(IMAGE_PREFIX)store-admin:$(BUILD_TAG)
+ADMIN_DEV_IMAGE = $(DEV_REGISTRY)/$(ADMIN_IMAGE)
 DAEMON_IMAGE = $(IMAGE_PREFIX)store-daemon:$(BUILD_TAG)
 DAEMON_DEV_IMAGE = $(DEV_REGISTRY)/$(DAEMON_IMAGE)
 GATEWAY_IMAGE = $(IMAGE_PREFIX)store-gateway:$(BUILD_TAG)
@@ -68,6 +70,8 @@ run: install start
 dev-release: push set-image
 
 push: check-registry
+	docker tag $(ADMIN_IMAGE) $(ADMIN_DEV_IMAGE)
+	docker push $(ADMIN_DEV_IMAGE)
 	docker tag $(DAEMON_IMAGE) $(DAEMON_DEV_IMAGE)
 	docker push $(DAEMON_DEV_IMAGE)
 	docker tag $(GATEWAY_IMAGE) $(GATEWAY_DEV_IMAGE)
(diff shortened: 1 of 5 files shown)


bacongobbler:
do we want to add this unit to `refresh-units` at the very least?

docs/troubleshooting_deis/troubleshooting-store.rst
```diff
@@ -0,0 +1,124 @@
...
+
+``deis-store-admin`` is an optional component that is helpful when diagnosing problems with ``deis-store``.
+It contains the ``ceph`` client and writes the necessary Ceph configuration files so it always has the
+most up-to-date configuration for the cluster.
+
+To use ``deis-store-admin``, install and start it with ``deisctl``:
+
+.. code-block:: console
+
+    $ deisctl install store-admin
```
bacongobbler: following up from my last comment about refresh-units, this will result in an error unless they already have downloaded the unit to their filesystem themselves, so we either need to add documentation around that or toss it into refresh-units.

carmstrong:
> do we want to add this unit to refresh-units at the very least?

Yeah, added.

mboersma:
Code LGTM for this Ceph troubleshooting container.

store/admin/bin/boot
```diff
@@ -0,0 +1,14 @@
...
+#
+
+ETCD_PORT=${ETCD_PORT:-4001}
+ETCD="$HOST:$ETCD_PORT"
+
+confd -node $ETCD -config-file /app/confd.toml &
+
+# loop forever until the container is stopped
+while true; do
+  sleep 5
```
bacongobbler: Would this mean that at the longest, it'd take 5 seconds for the container to die? Would there be an issue if we reduced it to `sleep 1`?

carmstrong: It really doesn't matter - I just changed it to 1s.

store/Makefile
```diff
@@ -78,12 +82,14 @@ push: check-registry
...
 
 set-image: check-deisctl
+	deisctl config store-admin set image=$(ADMIN_DEV_IMAGE)
 	deisctl config store-daemon set image=$(DAEMON_DEV_IMAGE)
 	deisctl config store-gateway set image=$(GATEWAY_DEV_IMAGE)
 	deisctl config store-metadata set image=$(METADATA_DEV_IMAGE)
 	deisctl config store-monitor set image=$(MONITOR_DEV_IMAGE)
 
 release:
+	docker push $(ADMIN_IMAGE)
```
bacongobbler: I don't think this should be published to DockerHub if it's an optional component

carmstrong: Why not? How else would people use it? Asking them to build it themselves is a bit unreasonable.

mboersma: I'm ok with it being published to Docker Hub--we also have deis/test-etcd and deis/test-postgresql out there for convenience, and they're kind of optional containers.

bacongobbler:
code lgtm

PR #2953, opened 2015-01-23
docs(*): add Ceph quorum documentation

### LINKED ISSUE #2926: docs: Add documentation for recovering Ceph from quorum failure
The resolution to #2914 was a manual recovery of the cluster because it had lost enough monitors due to the addition of new hosts and could no longer achieve quorum. We successfully recovered the cluster with just one host that was part of the original, healthy cluster.

The line of responsibility for this documentation is unclear, as the relevant documentation exists in several places within the Ceph documentation, but we have no expectation that Deis users are familiar with Ceph (and arguably shouldn't be), so I think it'd be helpful to provide some general steps and examples for cluster recovery.

ping @mgartner - he wanted to see this, since I was silently performing these tasks without reporting in for several nerve-wracking minutes ;)

The general steps are:

1. Trying `ceph -s` on all hosts and noticing that none ever connect to the cluster - instead, they show timeouts and/or monitor faults
2. In a `store-monitor` container, using the [admin socket](http://ceph.com/docs/master/rados/troubleshooting/troubleshooting-mon/#using-the-monitor-s-admin-socket) to query [mon_status](http://ceph.com/docs/master/rados/troubleshooting/troubleshooting-mon/#understanding-mon-status), identifying that there are enough stale entries to prevent Ceph from gaining quorum 
3. Stopping the platform with `deisctl stop platform` so components stop trying to write data to store (note that leaving the routers up will allow application containers to remain up, unaffected)
4. Cleaning up stale entries in `/deis/store/hosts` so that dead monitors are not written out to clients
5. Updating `/deis/store/monSetupLock` to point to the healthy monitor -- note that this isn't strictly necessary, as this value is only used if wiping clean and starting a fresh cluster from scratch with no data, but it's good cleanup
6. Start the healthy monitor and use the admin socket to get the current state of the cluster. 
  * Given the cluster state as the monitor sees it, use [monmaptool](http://ceph.com/docs/master/man/8/monmaptool/) to manually remove stale monitor entries from the monmap (i.e. `monmaptool --rm mon.<hostname> --clobber /etc/ceph/monmap`)
7. On the same host as the healthy monitor, get a shell in a store-monitor container without a monitor running (see #2925), and inject the prepared monmap into the monitor with `ceph-mon -i <hostname> --inject-monmap /etc/ceph/monmap`)
8. Start the monitor and ensure it achieves quorum by itself (use `ceph -s` and/or query mon_status on the admin socket)
9. Start the other monitors and ensure they connect
10. Start the OSDs with `deisctl start store-daemon`.
11. Observe the OSD map with `ceph osd dump`
  * For each OSD that is no longer with us, follow [Removing an OSD](http://docs.deis.io/en/latest/managing_deis/add_remove_host/#removing-an-osd) -- take care to ensure that the data is relocated before marking another OSD as `out`.
12. Once the OSD map reflects the now-healthy OSDs, start the remaining store services in order:
  * `deisctl start store-metadata`
  * `deisctl start store-gateway`
13. Confirm that the cluster is healthy with the metadata servers added, and then start `store-volume` with `deisctl start store-volume`.
14. Start the remaining services with `deisctl start platform`

carmstrong:
Note that PR also includes the doc change from #2951.

closes #2926

### DIFF
--- a/docs/managing_deis/add_remove_host.rst
+++ b/docs/managing_deis/add_remove_host.rst
@@ -111,6 +111,8 @@ that the store services on this host will be leaving the cluster.
 In this example we're going to remove the first node in our cluster, deis-1.
 That machine has an IP address of ``172.17.8.100``.
 
+.. _removing_an_osd:
+
 Removing an OSD
 ~~~~~~~~~~~~~~~
 
--- a/docs/managing_deis/add_remove_host.rst
+++ b/docs/managing_deis/add_remove_host.rst
@@ -25,11 +25,11 @@ Inspecting health
 -----------------
 
 Before we begin, we should check the state of the Ceph cluster to be sure it's healthy.
-We can do this by logging into any machine in the cluster, entering a store container, and then querying Ceph:
+To do this, we use ``deis-store-admin`` - see :ref:`using-store-admin`.
 
 .. code-block:: console
 
-    core@deis-1 ~ $ nse deis-store-monitor
+    core@deis-1 ~ $ nse deis-store-admin
     root@deis-1:/# ceph -s
         cluster 20038e38-4108-4e79-95d4-291d0eef2949
          health HEALTH_OK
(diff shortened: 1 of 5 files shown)


mboersma:
This Ceph troubleshooting documentation LGTM.

bacongobbler:
docs LGTM.

Issue #291, opened 2013-11-04
Failed to register user after recreating controller


bengrunfeld:
```
(deis)ben$ deis > deis register http://<controller>
username: <usrName>
password: 
password (confirm): 
email: <usrEmail>
Registered <usrName>
Login failed
Login failed
```

ianmiell:
deis register http://<domain>:8000

worked for me. Could this be made clearer in the docs? Quite a lot of time was wasted for me on this :)

ie a note similar to the Vagrant one here:

http://docs.deis.io/en/latest/using_deis/register-user/

PR #2961, opened 2015-01-26
docs(docs/using_deis/register-user.rst): note re controller on 8000 for

ianmiell:
dockers

### DIFF
--- a/docs/using_deis/register-user.rst
+++ b/docs/using_deis/register-user.rst
@@ -28,6 +28,10 @@ Note that you always use ``deis.<domain>`` to communicate with the controller.
 
     For Vagrant clusters: ``deis register http://deis.local3.deisapp.com``
 
+.. note::
+
+    For Docker-ized clusters: ``deis register http://deisapp.com:8000``
+
 .. important::
 
     The first user to register with Deis receives "superuser" privileges. Additional users who


deis-admin:

bacongobbler:
If you set up the DNS records correctly as per http://docs.deis.io/en/latest/managing_deis/configure-dns/#configure-dns, `http://deis.deisapp.com` will forward traffic to your controller. Could you please explain your issue more clearly as well as the instructions on how you got there?

For example, I have my DNS records set up for Digital Ocean like so (no load balancing available, so I'm using round robin DNS instead):

```
><> dig +noall +answer deis.fishworks.io
deis.fishworks.io.	1796	IN	CNAME	deis-1.fishworks.io.
deis-1.fishworks.io.	1796	IN	CNAME	deis-2.fishworks.io.
deis-2.fishworks.io.	1796	IN	CNAME	deis-3.fishworks.io.
deis-3.fishworks.io.	1796	IN	A	104.130.198.72
```

Note to others, this refs https://github.com/deis/deis/issues/291#issuecomment-71442655

ianmiell:
That seems to contradict what's on here:

http://docs.deis.io/en/latest/installing_deis/digitalocean/

or am I missing something?

bacongobbler:
From that same section:

> One way to configure this on DigitalOcean is to setup round-robin DNS via the DNS control panel. To do this, add the following records to your domain:

>- A wildcard CNAME record at your top-level domain, i.e. a CNAME record with * as the name, and @ as the canonical hostname
>- For each CoreOS machine created, an A-record that points to the TLD, i.e. an A-record named @, with the droplet’s public IP address

After you're done that, `deis.domain.com` will route to the controller, provided that there's a router on the host you connected to.

mboersma:
You shouldn't access the controller at port 8000, but at port 80 prefixed by `deis`.
```console
$ deis register http://deis.mydomain.com/
```
Otherwise, your DNS configuration isn't working correctly.

ianmiell:
Thanks (and thanks for the quick response). I'll try that now.

BTW, this is for this blog post: https://zwischenzugs.wordpress.com/2015/01/26/set-up-a-deis-docker-friendly-paas-on-digital-ocean-for-0-18-per-hour-in-six-easy-steps-using-shutit/

(so you can even repro the steps to set up yourself if you're so inclined; it's how I'm testing it now).

The code is here: https://github.com/ianmiell/shutit-coreos-do/blob/master/deis/deis.py

ianmiell:
That worked a treat, thanks.

"Note that you always use deis.<domain> to communicate with the controller."

I failed to read the docs properly. Closing.

ianmiell:
(Actually, I recall reading that and thinking it was referring to the deis binary and that the dot was a typo. Duh.)

Issue #2960, opened 2015-01-26
Push Failure after Restoration of from Backup


babarinde:
Hi,
I just did a restored my apps from a backup as described [here](http://docs.deis.io/en/latest/managing_deis/backing_up_data/) without any major issues except [this which was not backing up](http://docs.deis.io/en/latest/managing_deis/backing_up_data/#database-data).

When i try to push new changes  to my app , the git push fails (only via the buildpacks workflow , there are no problems with the Dockerfile workflow), but from my little observation, the push seems to be pushing the entire code base and not the latest changes, I thought this might be a good starting point.
Please see [gist](https://gist.github.com/babarinde/b9def784d4cbc576fcc0#file-gistfile1-txt)
Also I just need to be sure that when doing a recovery, does the new  deis platform version have to be the same version as the backed up one or is not necessary?

Thanks

bacongobbler:
could you please post some logs from your push failure? We need to see something that resembles your issue so we can try to identify the problem. :)

babarinde:
@bacongobbler 
https://gist.github.com/babarinde/b9def784d4cbc576fcc0#file-gistfile1-txt

bacongobbler:
that looks like a buildpack issue. Could you please try `deis config:set BUILDPACK_URL=https://github.com/heroku/heroku-buildpack-php` before pushing again and see if that fixes the issue? If so we may need to update buildpack-php.

babarinde:
@bacongobbler setting the buildpack url make the push progress but its now stuck at launching...
pls see ```deisctl journal deis-builder``` [logs](https://gist.github.com/babarinde/b9def784d4cbc576fcc0#comment-1380545)

bacongobbler:
Unfortunately the builder logs doesn't show anything useful. If you want to figure out why the application isn't launching, you'll have to take a look at the controller logs and `fleetctl list-units` to start digging into the root issue.

babarinde:
@bacongobbler 
After a while at the launching stage it shows this:
-----> Launching... 
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x10 pc=0x40179b]

goroutine 1 [running]:
main.main()
remote: /home/jenkins/workspace/test-acceptance/src/github.com/deis/deis/builder/bin/publish-release-controller.go:80 +0xb5b
To ssh://git@deis.abacus.com.ng:2222/ischool.git
 ! [remote rejected] master -> master (pre-receive hook declined)
[...]

PR #2964, opened 2015-01-26
fix(builder): correctly handle error on request timeout

bacongobbler:
refs https://github.com/deis/deis/issues/2960#issuecomment-71518301

### DIFF
--- a/builder/bin/publish-release-controller.go
+++ b/builder/bin/publish-release-controller.go
@@ -5,6 +5,7 @@ import (
 	"flag"
 	"fmt"
 	"io/ioutil"
+	"log"
 	"net/http"
 	"os"
 	"regexp"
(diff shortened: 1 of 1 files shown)


builder/bin/publish-release-controller.go
```diff
@@ -77,15 +78,12 @@ func main() {
...
+	if err != nil {
+		log.Fatalln(err)
 	}
 
-	if err != nil || res.StatusCode != 200 {
-		fmt.Println("failed retrieving config from controller")
-		fmt.Println(res.Body)
-		os.Exit(1)
+	if res.StatusCode != 200 {
+		log.Fatalf("failed retrieving config from controller: %s\n", res.Body)
```
carmstrong: Shouldn't we still exit here?

bacongobbler: `log.Fatalf` [does exactly that](http://golang.org/pkg/log/#Fatalf):

> Fatalf is equivalent to Printf() followed by a call to os.Exit(1).

carmstrong: Cool, that looks fine to me then. I still kind of like the explicit check for 404, as that's a clue that the controller is dead.

bacongobbler: The thing that confuses me with the check for a 404 response is that the router [returns a 503](https://github.com/deis/deis/blob/master/router/image/templates/nginx.conf#L83-L85) if the controller is down. The only time we'd ever respond with a 404 for a push hook is if [the user or application is not found](https://github.com/deis/deis/blob/master/controller/api/views.py#L270-L271), which means that the error message was misleading. Should I change that to a 503 check instead?

ping @aledbf in case he knows more about this specific check

aledbf: @bacongobbler the 404 check is pre 503 status code in the router

bacongobbler: gotcha. One 503 check coming right up!

carmstrong:
Code LGTM.

mboersma:
Code LGTM.

PR #2966, opened 2015-01-26
test(*): enforce go fmt

### LINKED ISSUE #2412: "make test-style" should use "gofmt -l" to verify formatting
In our coding standards we ask for all go code to be run through `gofmt`. We could probably enforce this using `gofmt -l ./...` and checking the output or return code in the appropriate Makefiles.

### LINKED ISSUE #2827: Additional checks for go code
I found this script https://gist.github.com/hailiang/0f22736320abe6be71ce that could complement the existing checks in go:

1. [gofmt](http://golang.org/cmd/gofmt/)
2. [goimports](https://github.com/bradfitz/goimports)
3. [golint](https://github.com/golang/lint)
4. [go vet](http://golang.org/cmd/vet)
5. [race detector](http://blog.golang.org/race-detector)
6. [test coverage] (http://blog.golang.org/cover)

bacongobbler:
This PR enforces and suggests that the codebase should follow a specific
style using a combination of `gofmt`, `golint` and `go vet`. `gofmt` is
enforced; the latter are suggested and displayed.

deis/publisher also now uses `godep go build` to build the binary,
similar to our other components.

closes #2827
closes #2412


bacongobbler:
ha, case and point ;)

carmstrong:
Code LGTM, but change should probably be separated into 3 commits as discussed:
* Changing publisher over to Godep
* Removing the client.yaml code
* Enforcing gofmt

bacongobbler:
We missed that there was a 4th change in here, which was to remove the logspout Dockerfile which we no longer use.

mboersma:
The publisher fix here is the important one. Ideally these would be separate PRs, but I'm fine with all the changes. Code LGTM.

carmstrong:
:ship:

PR #2968, opened 2015-01-27
feat(coreos): daily cleanup of docker images

aledbf:
This script runs every day at midnight. It removes stopped containers and unused images.
Also removes older application images leaving only the last 4 versions.

### DIFF
--- a/contrib/coreos/user-data.example
+++ b/contrib/coreos/user-data.example
@@ -74,6 +74,29 @@ coreos:
       [Service]
       Type=oneshot
       ExecStart=/bin/sh -c "sysctl -w net.netfilter.nf_conntrack_max=262144"
+  - name: daily-docker-cleanup.timer
+    command: start
+    content: |
+      [Unit]
+      Description=Daily docker cleanup timer
+
+      [Timer]
+      OnCalendar=daily
+      Persistent=true
+      Unit=daily-docker-cleanup.service
+
+      [Install]
+      WantedBy=timers.target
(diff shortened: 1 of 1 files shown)


carmstrong:
I'm hesitant to automatically run any sort of destructive operations. I think I'd personally prefer that this be installed but not started - we can provide some documentation in "Managing Deis" for folks who want to run this.

mboersma:
I'm first going to try using `docker-cleanup.sh` on our test nodes, to avoid the downtime we sometimes have for exactly this purpose.

carmstrong:
> I'm first going to try using docker-cleanup.sh on our test nodes

:thumbsup:

aledbf:
@carmstrong well, the point of this PR is to not run the cleanup manually :)
How do you suggest that I modify this PR?
Removing the timer and indicate the use of the script in "Managing Deis" is enough?

carmstrong:
> How do you suggest that I modify this PR?
> Removing the timer and indicate the use of the script in "Managing Deis" is enough?

Ah, didn't realize it was a `.timer` service. With a normal unit, simply removing `command: start` would be enough (as this just installs the service to systemd), but I imagine installing a timer service essentially activates ("starts") it.

carmstrong:
> Removing the timer and indicate the use of the script in "Managing Deis" is enough?

Let's remove this from Deis itself and instead just put the unit and some documentation in the docs. We now have a page that would be great for this - http://docs.deis.io/en/latest/managing_deis/disk_usage/

PR #2971, opened 2015-01-27
feat(Vagrant): Manage specific instances

johanneswuerbach:
Allows to start or stop specific instances in a Vagrant cluster.
Example:
`INSTANCE_IDS=1,2 vagrant halt` will shut down deis-1 and deis-2

This simplifies manual HA testing with a local Vagrant cluster.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -15,6 +15,7 @@ CONFIG = File.join(File.dirname(__FILE__), "config.rb")
 # Defaults for config options defined in CONFIG
 $num_instances = 1
 $update_channel = ENV["COREOS_CHANNEL"] || "stable"
+$instance_ids = ENV["INSTANCE_IDS"]
 $enable_serial_logging = false
 $vb_gui = false
 $vb_memory = 1024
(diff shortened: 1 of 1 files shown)


carmstrong:
Code LGTM, although we'll want to manually test this.

johanneswuerbach:
Sure. I currently testing ceph issues (losing a single instance), but wanted to PR this already.

Now it's easier to test what happens when an instance is replaced:
```bash
make dev-cluster
INSTANCE_IDS=3 vagrant destroy --force # hard kill an instance or vagrant halt to gracefully stop
INSTANCE_IDS=4 vagrant up # add a new one to the cluster
```

bacongobbler:
vagrant has this built into the client:

`vagrant halt deis-3`

I don't think this is necessary unless this resolves a different problem.

johanneswuerbach:
Oh, I didn't know that you could specify an instance. If there is also a way to start `deis-4` and add it to the cluster, this PR can be closed.

carmstrong:
> If there is also a way to start deis-4 and add it to a running cluster, this PR can be closed.

That's the main benefit as I see it.

mboersma:
> If there is also a way to start deis-4 and add it to a running cluster...

You can `export NUM_INSTANCES=4` instead of 3 and do `vagrant up` in that case. Only the new node will be provisioned. But that doesn't give you the specific control over the instance IDs that your PR does. I like this but I'm not convinced it's that generally useful.

carmstrong:
> I like this but I'm not convinced it's that generally useful.

I'm in agreement on this one - specifying the instance IDs is pretty straightforward.

PR #2973, opened 2015-01-27
feat(ec2/cloudformation): Select SSD EBS volumes by default

ianblenke:
This enables a CloudFormation parameter named EC2EBSVolumeType that may
be one of: 'gp2' for SSD, 'io1' for guaranteed IOPs, or 'standard' for magnetic

### DIFF
--- a/contrib/ec2/deis.template.json
+++ b/contrib/ec2/deis.template.json
@@ -78,6 +78,13 @@
       "AllowedValues" : [ "PV", "HVM" ],
       "ConstraintDescription" : "must be either PV or HVM"
     },
+    "EC2EBSVolumeType" : {
+      "Description" : "EC2 EBS VolumeType",
+      "Type": "String",
+      "Default": "gp2",
+      "AllowedValues" : [ "gp2", "io1", "standard" ],
+      "ConstraintDescription" : "must be either 'gp2' for SSD, 'io1' for guaranteed IOPs, or 'standard' for magnetic"
+    },
     "AssociatePublicIP": {
       "Description": "Whether to associate a public IP address to the instances",
       "Type": "String",
@@ -251,11 +258,11 @@
         "BlockDeviceMappings" : [
           {
             "DeviceName" : { "Fn::FindInMap": [ "RootDevices", { "Ref": "EC2VirtualizationType" }, "Name" ] },
-            "Ebs" : { "VolumeSize" : "50" }
+            "Ebs" : { "VolumeSize" : "50", "VolumeType": { "Ref": "EC2EBSVolumeType" } }
           },
           {
             "DeviceName" : "/dev/xvdf",
-            "Ebs" : { "VolumeSize" : "100" }
+            "Ebs" : { "VolumeSize" : "100", "VolumeType": { "Ref": "EC2EBSVolumeType" } }
           }
         ]
       }


carmstrong:
Holy crap - the default has been `standard` this whole time!! http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateVolume.html
http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/EBSVolumeTypes.html

That means we were using *magnetic* storage this whole time. Also, general purpose isn't as costly as provisioned iops: http://aws.amazon.com/ebs/pricing/

Good find, @ianblenke! Code LGTM.

johanneswuerbach:
Wouldn't it make sense to use the instance store from everything but `/var/lib/docker` and ceph instead of an EBS? m3.large have a 32GB SSD, which is free and removes any kind of failures because of bad network.

carmstrong:
> Wouldn't it make sense to use the instance store from everything but /var/lib/docker and ceph instead of an EBS? m3.large have a 32GB SSD, which is free and removes any kind of failures because of bad network.

The SSD is 32GB, but due to how the drive is partitioned, the root partition is only 8GB. You can confirm to see if that's still the case, but I suspect it is. We moved to an EBS-backed root volume in ff24d62c43f0724c7ba47e13bc3cbb48809bec04.

mboersma:
```console
core@ip-10-21-2-33 ~ $ mount | grep ssd
/dev/xvda9 on / type btrfs (rw,relatime,ssd,space_cache)
/dev/xvdf on /var/lib/docker type btrfs (rw,relatime,ssd,space_cache)
```
I also deployed v1.2.2 and tested some apps, everything looks good using SSD volumes as expected. LGTM!

contrib/ec2/deis.template.json
```diff
@@ -78,6 +78,13 @@
       "AllowedValues" : [ "PV", "HVM" ],
       "ConstraintDescription" : "must be either PV or HVM"
     },
+    "EC2EBSVolumeType" : {
+      "Description" : "EC2 EBS VolumeType",
+      "Type": "String",
+      "Default": "gp2",
+      "AllowedValues" : [ "gp2", "io1", "standard" ],
```
carmstrong: I don't think we can support `io1` here, as another parameter `Iops` would be required in that case: http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-blockdev-template.html

ianblenke: Yep, you are absolutely correct. We could wrap that in an Fn::If. Will need to TDD that. Have some rspec ready (really need to blog about it soon), will give it a try.

ianblenke: Also note: if we allow io1, it is also important to modify the autoscaling group's launch configuration to enable provisioned IOPs, or you're not getting the dedicated gig-e iSCSI interface to the provisioned IOPs EBS volumes which results in pretty much the same performance on the EBS volume had you not enabled IOPs on the volume at all

gabrtv:
LGTM, see no reason we shouldn't allow for the other volume types.

contrib/ec2/deis.template.json
carmstrong: Maybe we simplify this by just hard-coding `gp2`, and if folks want to use provisioned iops they can hack the template with the necessary values?

carmstrong:
I added a commit to remove iops as an option - see #3065. If a user really wants to support that, they can mess with the template :)

PR #2976, opened 2015-01-28
Add enforceHttps support

### LINKED ISSUE #2838: feat(router): Add HTTP to HTTPs redirect
This was attempted in #2635 and again in #2782, but the implementation in those PRs doesn't seem to work. This is a desired feature, so I wanted to track it separately so it wasn't lost.

I've attempted this myself, but it appears the `server` blocks keep conflicting in the nginx template. After several failed attempts at this, I resorted to [asking StackOverflow](http://stackoverflow.com/a/27711250/644992) for help. An answer seems worthy of attempting, but I haven't had the time to get back to this, and likely will be working on other things for the next Deis release.

I wanted to document this so hopefully someone else can pick it up.

ianblenke:
feat(router/enforceHttps): Allow https enforcement

Setting /deis/router/enforceHttps to anything other than "false" will result in any requests to http being redirected to https

closes #2838

### DIFF
--- a/router/image/templates/nginx.conf
+++ b/router/image/templates/nginx.conf
@@ -127,6 +127,7 @@ http {
 
     ## start service definitions for each application
     {{ $useSSL := or .deis_router_sslCert "false" }}
+    {{ $enforceHttps := or .deis_router_enforceHttps "false" }}
     {{ $domains := .deis_domains }}{{ range $service := .deis_services }}{{ if $service.Nodes }}
     upstream {{ Base $service.Key }} {
         {{ range $upstream := $service.Nodes }}server {{ $upstream.Value }};
@@ -156,6 +157,12 @@ http {
             proxy_set_header            Connection        $connection_upgrade;
 
             proxy_next_upstream         error timeout http_502 http_503 http_504;
+
+            {{ if ne $enforceHttps "false" }}
+            if ($http_x_forwarded_proto != "https") {
+              rewrite ^(.*)$ https://$host$1 permanent;
+            }
+            {{ end }}
 
             add_header                  X-Deis-Upstream   $upstream_addr;
 


router/image/templates/nginx.conf
```diff
@@ -157,6 +158,12 @@ http {
 
             proxy_next_upstream         error timeout http_502 http_503 http_504;
 
+            {{ if ne $enforceHttps "false" }}
```
bacongobbler: This is a double negative. `{{ if $enforceHttps }}` should work

ianblenke: Completely agree. Fixing now.

ianblenke: As a followup, just above this section, there is:

    {{ if ne $useSSL "false" }}

This is the reason I had used this method of comparison, the other similar references in this template used this same "trick". Is it idiomatic? Is it necessary? If the bare case works, I'm fine with the clearer expression.

bacongobbler: IMO it's more clear to be explicit about true rather than false, so either checking if the key exists (`{{ if $enforceHttps }}`) or if the key is set explicitly to "true" (`{{ if eq $enforceHttps "true" }}`) is preferred.

We should've used that for the other key values, but that's a behavioural change so we can't change it now.

aledbf: @bacongobbler just in case the $enforceHttps returns a string that's why the comparison is required. 
Unless the actual value is not important and only the existence of the key indicates the use of a "feature".

bacongobbler: that's what I said :)

aledbf: @bacongobbler sorry I didn't saw your last comment when I submitted :)

carmstrong:
@ianblenke Do you think this variable should redirect traffic for the controller too? It's probably just a case of adding the code to that upstream, too, right?

ianblenke:
Yes. I think that it should redirect traffic for the controller too. That's how we have it setup for our use, it enforces HTTPS for all HTTP traffic from the ELB. Didn't I work that into this PR? Hrm.

carmstrong:
> Didn't I work that into this PR?

Doesn't appear so from the code (I would think we'd have to also configure it in the controller upstream), but I can manually verify.

carmstrong:
I'll take this over in #3075. Thanks @ianblenke!

Issue #2394, opened 2014-11-05
Proposal: Use in-house mirroring for slow internet connections


johanneswuerbach:
Docker 1.3 made it fairly simple to use an in-house registry to pull public docker images: https://github.com/docker/docker/blob/master/docs/sources/articles/registry_mirror.md

Providing a `deisctl` command to use an in-house registry when your outbound internet connection is slow, would allow pre-seeding, faster new node boot times and decrease the overall deis download size as layer downloads like ubuntu-debootstrap are shared between nodes.

//cc @aledbf , @jperville

This would allow to remove all download timeout hacks like https://github.com/deis/deis/pull/2305 and the local building of slugrunner and -builder.

johanneswuerbach:
A fixed version of docker is now available in CoreOS stable and I created a PR #2981.

PR #2981, opened 2015-01-28
docs(*): Explain how to use a local Docker Hub mirror

johanneswuerbach:
Instructions how to use a local Docker Hub mirror

I would prefer to document adding another `docker.service.d` drop in instead of modifying a line, but couldn't find a way to append and not overwrite `DOCKER_OPTS` from another service file. Ideas?

Fixes https://github.com/deis/deis/issues/2394

//cc @aledbf are you already using on-premises mirroring?


aledbf:
@johanneswuerbach it makes no sense to me the use of mirroring. 
I have a dedicated private registry for production. The registry contains the only dependency of deis (ubuntu-debootstrap) and all the deis releases (I don't use the official versions from the docker) and all the "dev" application images.

johanneswuerbach:
@aledbf thanks for the reply. Mirroring is just removing the manual management part of deis releases and dependencies, but requires another registry as mirroring and standalone doesn't seem to work together.

aledbf:
@johanneswuerbach maybe the problem is https://github.com/docker/docker-registry/pull/908

johanneswuerbach:
Not really, because I want to mirror the public hub and the issue is about mirroring a private registry. It's more that mirroring requires to disable standalone mode, which you generally want for an internal registry.

For development I'm currently using two registries. One handles custom builds and one mirrors the Docker Hub to allow caching of containers between destroying and rebuilding my local Vagrant cluster. It would simplify things to merge those two, but I don't think this is possible.

bacongobbler:
Thanks for the article, but sadly we do not want to put this in the documentation. This is a very advanced article with a very specific use case, and the guide you linked to has most of the information on how to set it up, so I'd imagine that this would not have much benefit on its own.

That being said, if you want to write up a blog post on how you set up a mirror as well as how it interacts with your build process, we would happily post it on our blog! Our blog can be found at [deis/deis.io](/deis/deis.io). PRs welcome :)

Issue #980, opened 2014-05-15
Deis platform log aggregation


carmstrong:
We need to have a solution for draining application and relevant system logs to an external source.

PR #2982, opened 2015-01-28
Proposal: Platform-wide log drain

rjocoleman:
Currently [platform logging](http://docs.deis.io/en/latest/managing_deis/platform_logging/) is done via `journalctl -o short -f | ncat --ssl logs2.papertrailapp.com 23654`

There are a couple of disadvantages to this:
1. A unit needs to be created (and customised) by end users.
2. Container logs aren't prefixed with application name.

I propose adding an external drain to `logger` that would send to external syslog servers in addition to the current write to disk process.

Interface would be:

```
$ deisctl config logs set drain/host logs2.papertrailapp.com
$ deisctl config logs set drain/port 23654
```

Logger pulls this configuration out of etcd and sends the messages on.

related #980

### DIFF
--- a/logger/drain/drain.go
+++ b/logger/drain/drain.go
@@ -1,0 +1,60 @@
+package drain
+
+import (
+	"fmt"
+	"log"
+	"net"
+	"net/url"
+	"os"
+
+	"github.com/coreos/go-etcd/etcd"
+)
+
+func GetDrain() string {
+	host := getopt("HOST", "127.0.0.1")
+
+	etcdPort := getopt("ETCD_PORT", "4001")
+	etcdPath := getopt("ETCD_PATH", "/deis/logs")
+
+	client := etcd.NewClient([]string{"http://" + host + ":" + etcdPort})
+
+	s, err := client.Get(etcdPath+"/drain", true, false)
+	if err != nil {
+		return ""
+	}
+
+	return s.Node.Value
+}
+
+func SendToDrain(m string, drain string) error {
+	u, err := url.Parse(drain)
+	if err != nil {
+		log.Fatal(err)
+	}
+	uri := u.Host + u.Path
+	switch u.Scheme {
+	case "syslog":
+		sendToSyslogDrain(m, uri)
+	default:
+		log.Println(u.Scheme + " drain type is not implemented.")
+	}
+	return nil
+}
+
+func sendToSyslogDrain(m string, drain string) error {
+	conn, err := net.Dial("udp", drain)
+	if err != nil {
+		log.Fatal(err)
+	}
+	defer conn.Close()
+	fmt.Fprintf(conn, m)
+	return nil
+}
+
+func getopt(name, dfault string) string {
+	value := os.Getenv(name)
+	if value == "" {
+		value = dfault
+	}
+	return value
+}
--- a/logger/syslogd/syslogd.go
+++ b/logger/syslogd/syslogd.go
@@ -9,6 +9,8 @@ import (
 	"regexp"
 
 	"github.com/deis/deis/logger/syslog"
+
+	"github.com/deis/deis/logger/drain"
 )
 
 const logRoot = "/data/logs"
@@ -81,6 +83,10 @@ func (h *handler) mainLoop() {
 		m := h.Get()
 		if m == nil {
 			break
+		}
+		d := drain.GetDrain()
+		if d != "" {
+			drain.SendToDrain(m.String(), d)
 		}
 		err := writeToDisk(m)
 		if err != nil {
--- a/docs/managing_deis/platform_logging.rst
+++ b/docs/managing_deis/platform_logging.rst
@@ -16,8 +16,19 @@ and sends their logs to ``/deis/logs/host`` and ``/deis/logs/port``.
 when a client runs ``deis logs``. This component publishes its host and port to ``/deis/logs/host``
 and ``/deis/logs/port``, and is typically the service which consumes logs from ``deis-logspout``.
 
-Routing logs to a custom location
----------------------------------
+Application log drain
+---------------------
+
+Application logs can be drained to an external syslog server (or compatible service such as Logstash, Papertrail, Splunk etc).
+
+.. code-block:: console
+
+    $ deisctl config logs set drain=syslog://logs2.papertrailapp.com:23654
+
+This will send all application logs - there is currently no way to drain logs per application.
+
+Routing host logs to a custom location
+--------------------------------------
 
 Logging to an external location can be achieved without modifying the log flow within Deis -
 we can simply send the master journal on a CoreOS host using ``ncat``. For example, if I'm using the


deis-admin:

mboersma:
Jenkins, add to whitelist.

rjocoleman:
I've added the docs - if everyone's happy with this as the interface I'll get started on logger.

gabrtv:
@rjocoleman to make it atomic and remove the slash in the key name, let's change the interface to:

```
$ deisctl config logs set drain=logs2.papertrailapp.com:23654
```

Otherwise, looks awesome. Let us know if we can help!

theaboutbox:
Perhaps I'm over-thinking it, but would you also want to add a protocol to the drain setting? Example:

    $ deisctl config logs set drain=syslog://logs2.papertrailapp.com:23654

I can think of a few advantages:

* Explicit: The setting doesn't make it obvious whether it's sending TCP, UDP, Syslog, HTTPS, etc. Also if a user tries something that's not implemented, it's easier to give an informative error message.
* Future-proof: Right after syslog working, I'd want to dig in and add tcp and tcp+tls
* Can use URI parsing libraries to get protocol, host, port, etc.

gabrtv:
@theaboutbox I see little downside to this. +1.

rjocoleman:
I've just pushed a first pass at this, feedback please (I'm pretty new to Go)

logger/drain/drain.go
```diff
@@ -0,0 +1,59 @@
...
+	host := getopt("HOST", "127.0.0.1")
+
+	etcdPort := getopt("ETCD_PORT", "4001")
+	etcdPath := getopt("ETCD_PATH", "/deis/logs")
+
+	client := etcd.NewClient([]string{"http://" + host + ":" + etcdPort})
+
+	s, err := client.Get(etcdPath+"/drain", true, false)
+	if err != nil {
+		return ""
```
bacongobbler: It seems odd that we're silently passing an error here

logger/syslogd/syslogd.go
```diff
@@ -82,6 +84,9 @@ func (h *handler) mainLoop() {
 		if m == nil {
 			break
 		}
+		if drain.GetDrain() != "" {
+			drain.SendToDrain(m.String(), drain.GetDrain())
```
bacongobbler: With each call to `.GetDrain()`, a new etcd client is created. We should be re-using the client instead of spawning a new connection.

logger/drain/drain.go
rjocoleman: How would you prefer this done?
etcd gives out an error `100: Key not found (/deis/logs/drain) [2674]`

logger/syslogd/syslogd.go
rjocoleman: Good point, will adjust to:
```go
		d := drain.GetDrain()
		if d != "" {
			drain.SendToDrain(m.String(), d)
		}
```

logger/drain/drain.go
rjocoleman: @bacongobbler any more feedback on this?

bacongobbler: sorry! This was a recent issue: #3007

rjocoleman: np, I've rebased this.
I'm still not sure what to do about the silent error - it's probably preferable to every logged line firing a "drain not configured". I can't see a better way to pull from etcd.

bacongobbler: neither can I. I think it's fine as-is unless we can think of a better solution. :)

bacongobbler:
Hmm, the Jenkins issue seems unrelated. I punted another job for you :)

logger/drain/drain.go
```diff
@@ -0,0 +1,59 @@
...
+	case "syslog":
+		sendToSyslogDrain(m, uri)
+	default:
+		log.Println(u.Scheme + " drain type is not implemented.")
+	}
+	return nil
+}
+
+func sendToSyslogDrain(m string, drain string) error {
+	conn, err := net.Dial("udp", drain)
```
bacongobbler: we'll need to make a call to close the connection after we're done with it via `defer conn.Close()`

bacongobbler:
asides from deferring to close the connection, this LGTM! We'll still need to manually test, but this looks awesome :fireworks:

rjocoleman:
@bacongobbler I've added in the `defer conn.Close()` now, thanks!

logger/drain/drain.go
```diff
@@ -0,0 +1,60 @@
...
+	defer conn.Close()
```
bacongobbler: we need to call `defer` *after* we check if it's nil. Can't call `.Close()` on a `nil` type!

rjocoleman: Oops! force push coming to fix.

bacongobbler:
apologies, there was a bug in the tests which is making this fail. I've personally seen this go green but we have a strict green policy. Would you mind rebasing, please? :heart:

bacongobbler:
also a squash would be great, too

rjocoleman:
> apologies, there was a bug in the tests which is making this fail. I've personally seen this go green but we have a strict green policy before merging PRs. Would you mind rebasing, please?

I've rebased and squashed, I hope Jenkins likes this now!

carmstrong:
This looks great. Code LGTM. Thanks @rjocoleman!

mboersma:
Ideally this would have tests and comments on the public funcs, but we don't yet have a test framework in the logger/syslog package and I will backfill the comments in #3118. Code LGTM.

carmstrong:
This works exactly as advertised! I set:
```console
$ deisctl config logs set drain=syslog://logs2.papertrailapp.com:23654
```
(which is the host/port for my test Papertrail account), and moments later log entries started appearing in my dashboard. `deis logs` also still showed the logs there, which is what we want.

Thanks @rjocoleman!

carmstrong:
@rjocoleman Could you please rebase this PR on master one last time? It failed to work for me locally until I had rebased on master, due to some fixes unrelated to this PR. Thanks!

rjocoleman:
@carmstrong sure, this is done now

wenzowski:
Really excited to see app logs prefixed with the app's name instead of just `sh`. grep ftw. love it.

carmstrong:
This is ready to ship once Jenkins passes it a last time.

rjocoleman:
@carmstrong not sure what happened here? Is this a Jenkins issue? `Error response from daemon: No such container: deis-cache-jenkins-test-integration-2180`

mboersma:
@rjocoleman yes, we're having some CI issues (some filesystem corruption on a test node). I think it's better now and I'll run this one again.

parndt:
Looks like Jenkins passed this a last time so this seems good to merge and ship, to me. :+1:

bacongobbler:
There's no tests on this one so it's a false positive, hence the "requires-manual-testing" label :)

mboersma:
Code LGTM.

carmstrong:
> There's no tests on this one so it's a false positive, hence the "requires-manual-testing" label :)

I manually tested myself. :shipit:

PR #2983, opened 2015-01-28
fix(controller): relax domain regex to allow dash in last label

### LINKED ISSUE #2979: deis blocks domain creation under improper circumstances
`$ deis domains:add django.paas-sandbox`
`Adding django.paas-sandbox to rubber-undertow... 400 BAD REQUEST`
`{u'domain': [u'Hostname does not look like a valid hostname. Only lowercase characters are allowed.']}`

deis will not let me create a domain where the top of the domain is "paas-sandbox", as displayed above, however this is not a top level domain nor is it intended as one. Our corporation uses an internal domain that is set as the dns search on all corporate computers. For example, if someone types `http://www/` in their browser, it will come up because their dns searches `corp.example.com` for non qualifying domains. They are really going to the site `www.corp.example.com` but all they type and all the host header on the server sees is "www". deis requires the domain to exist how the user will display it in the browser in order to render the page. Our DNS system is almost identical to Active Directory so anyone using Active Directory can relate but this is not a MS / Active Directory specific scenario. This can be done using only dhcp and most ISP's will do this from their dhcp as well.

Anyways, please fix. Thank you.

mboersma:
Closes #2979.


carmstrong:
Code LGTM.

PR #2988, opened 2015-01-29
feat(builder): Try shallow cloning buildpack repos

johanneswuerbach:
Try to shallow clone buildpack repositories before falling back to deep cloning. This improves clone times significantly.

Shallow cloning only works for tags and branches, but not for a specific commit.

Benchmark:
```
localhost:deis johannes$ time git clone https://github.com/heroku/heroku-buildpack-ruby.git
Cloning into 'heroku-buildpack-ruby'...
...

real	0m12.764s
user	0m0.610s
sys	0m0.565s
localhost:deis johannes$ rm -rf heroku-buildpack-ruby/
localhost:deis johannes$ time git clone --branch v129 --depth 1 https://github.com/heroku/heroku-buildpack-ruby.git
Cloning into 'heroku-buildpack-ruby'...
...

real	0m1.657s
user	0m0.078s
sys	0m0.092s
```

This saves around 10 seconds per buildpack and we are cloning 7 buildpacks using a tag currently.

PR #2989, opened 2015-01-29
chore(builder): Updated gradle and play buildpack

johanneswuerbach:
Both buildpacks are now also tagged when released.

Changes:
https://github.com/heroku/heroku-buildpack-gradle/compare/24a8ebe...v12
https://github.com/heroku/heroku-buildpack-play/compare/9c137b4...v22

This improves https://github.com/deis/deis/pull/2988 even further. @bacongobbler do you have example apps to tests those buildpacks?  Otherwise I would create them.


bacongobbler:
We don't have a gradle app, but there's https://github.com/deis/example-play

johanneswuerbach:
The example-play app was not actually testing the play BP. I submitted https://github.com/deis/example-play/pull/1 to fix this and tested this PR successfully.

mboersma:
LGTM. I tested with deis/example-play#1.

bacongobbler:
I haven't manually tested the buildpacks, but it looks good to me! Nice to have these buildpacks following a proper tag :)

PR #2991, opened 2015-01-29
fix(controller): improve domain name validation per RFC 1123

### LINKED ISSUE #2979: deis blocks domain creation under improper circumstances

mboersma:
Closes #2979, replaces #2983.

### DIFF
--- a/controller/api/serializers.py
+++ b/controller/api/serializers.py
@@ -239,23 +239,22 @@ class DomainSerializer(ModelSerializer):
         """
         Check that the hostname is valid
         """
-        match = re.match(
-            r'^(\*\.)?(' + settings.APP_URL_REGEX + r'\.)*([a-z0-9-]+)\.([a-z0-9]{2,})$',
-            value)
-        if not match:
+        if len(value) > 255:
+            raise serializers.ValidationError('Hostname must be 255 characters or less.')
+        if value[-1:] == ".":
+            value = value[:-1]  # strip exactly one dot from the right, if present
+        labels = value.split('.')
+        if labels[0] == '*':
             raise serializers.ValidationError(
-                "Hostname does not look like a valid hostname. "
-                "Only lowercase characters are allowed.")
-
+                'Adding a wildcard subdomain is currently not supported.')
+        allowed = re.compile("^(?!-)[a-z0-9-]{1,63}(?<!-)$", re.IGNORECASE)
+        for label in labels:
+            match = allowed.match(label)
+            if not match or '--' in label or label[-1].isdigit() or label.isdigit():
+                raise serializers.ValidationError('Hostname does not look valid.')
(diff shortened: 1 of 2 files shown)


controller/api/serializers.py
```diff
@@ -239,23 +239,22 @@ def validate_domain(self, value):
         """
         Check that the hostname is valid
         """
-        match = re.match(
-            r'^(\*\.)?(' + settings.APP_URL_REGEX + r'\.)*([a-z0-9-]+)\.([a-z0-9]{2,})$',
-            value)
-        if not match:
+        if len(value) > 255:
```
bacongobbler: I believe a Validator already exists for this: https://github.com/django/django/blob/master/django/core/validators.py#L294

There's some pretty useful validators at `django.core.validators`. We should use them more often :)

bacongobbler: note: we can use these validators if we shift this over to the model, which ay be a better idea (I did this in https://github.com/deis/deis/pull/2898)

controller/api/tests/test_domain.py
```diff
@@ -48,24 +48,34 @@ def test_response_data(self):
...
-        self.assertEqual(response.status_code, 204)
-        url = '/v1/apps/{app_id}/domains'.format(app_id=self.app_id)
-        response = self.client.get(url, content_type='application/json',
-                                   HTTP_AUTHORIZATION='token {}'.format(self.token))
-        self.assertEqual(0, response.data['count'])
+        test_domains = [
+            'test-domain.example.com',
+            'django.paas-sandbox',
+            'domain',
+            'not.too.loooooooooooooooooooooooooooooooooooooooooooooooooooooooooooong',
```
bacongobbler: :laughing:

bacongobbler:
we can probably clean this up and move the validation to the model, but this works for now. Code LGTM!

gabrtv:
Code LGTM too.  Nice work with the test coverage!

PR #2993, opened 2015-01-29
ref(controller): query fleet state

### LINKED ISSUE #2766: Don't store app state in the database - query it fresh on a `deis ps`
When I released a bad version of an app that was crashing, I noticed that `deis ps` was reporting the containers were up when `fleetctl list-units` was showing that they were dead. Ideally `deis ps` would show me that something was wrong.

```
$ deis ps
=== lob-images Processes

--- cmd: 
cmd.1 up (v21)
cmd.2 up (v21)
```

```
$ fleetctl list-units
UNIT					MACHINE			ACTIVE		SUB
...
lob-images_v21.cmd.1.service		7af5ddc0.../10.21.1.111	inactive	dead
lob-images_v21.cmd.2.service		e00d7927.../10.21.2.134	inactive	dead
```

### LINKED ISSUE #2987: Deis hangs in launching state though launced successfully
-----> Building Docker image
remote: Sending build context to Docker daemon 83.54 MB
remote: build context to Docker daemon 
Step 0 : FROM deis/slugrunner
 ---> f4cf8ef17328
Step 1 : RUN mkdir -p /app
 ---> Using cache
 ---> 18ef30ab747b
Step 2 : WORKDIR /app
 ---> Using cache
 ---> 52b7562312b0
Step 3 : ENTRYPOINT /runner/init
 ---> Using cache
 ---> a84e69fba887
Step 4 : ADD slug.tgz /app
 ---> 8a08ae8d9de5
Removing intermediate container b040657d0d9d
Step 5 : ENV GIT_SHA c268e1b65d9d20862b6b1e416ffd410f4a227923
 ---> Running in bccd0c98ace8
 ---> d1d2381d2a41
Removing intermediate container bccd0c98ace8
Successfully built d1d2381d2a41
-----> Pushing image to private registry

-----> Launching... 

And yeah ... just stuck there ... ctl-c exits, application is at the appropriate endpoint, running just fine.

bacongobbler:
Before, we relied on django-fsm to give us a general idea of what state
the container object in the model is at. This did not give users an
accurate idea of the state of their containers. Introducing a new
.state() method to the scheduler as well as changing Container.state to
call it's own scheduler's .state() method allows users to directly
understand what state their containers are in.

closes #2766 
closes #2987

This was thoroughly tested by hand since the controller does not wrap unit tests around fleet.py, but this should cover it :)

Points: PTAL at the systemdActiveStateMap in fleet.py and let me know if the mappings between the [dbus API's ActiveState](http://www.freedesktop.org/wiki/Software/systemd/dbus/) and what used to be DjangoFSM's JobState is correct.

### DIFF
--- a/controller/scheduler/fleet.py
+++ b/controller/scheduler/fleet.py
@@ -1,13 +1,16 @@
 import cStringIO
 import base64
 import copy
-import json
 import httplib
+import json
 import paramiko
 import socket
 import re
 import time
+
 from django.conf import settings
+
+from .states import JobState
 
 
 MATCH = re.compile(
@@ -42,16 +45,21 @@ class FleetHTTPClient(object):
     # connection helpers
 
     def _put_unit(self, name, body):
-        headers = {'Content-Type': 'application/json'}
-        self.conn.request('PUT', '/v1-alpha/units/{name}.service'.format(**locals()),
-                          headers=headers, body=json.dumps(body))
-        resp = self.conn.getresponse()
-        data = resp.read()
-        if not 200 <= resp.status <= 299:
-            errmsg = "Failed to create unit: {} {} - {}".format(
-                resp.status, resp.reason, data)
-            raise RuntimeError(errmsg)
-        return data
+        for attempt in range(RETRIES):
+            try:
+                headers = {'Content-Type': 'application/json'}
+                self.conn.request('PUT', '/v1-alpha/units/{name}.service'.format(**locals()),
+                                  headers=headers, body=json.dumps(body))
+                resp = self.conn.getresponse()
+                data = resp.read()
+                if not 200 <= resp.status <= 299:
+                    errmsg = "Failed to create unit: {} {} - {}".format(
+                        resp.status, resp.reason, data)
+                    raise RuntimeError(errmsg)
+                return data
+            except:
+                if attempt >= (RETRIES - 1):
+                    raise
 
     def _delete_unit(self, name):
         headers = {'Content-Type': 'application/json'}
@@ -131,14 +139,8 @@ class FleetHTTPClient(object):
             tagset = ' '.join(['"{}={}"'.format(k, v) for k, v in tags.items()])
             unit.append({"section": "X-Fleet", "name": "MachineMetadata",
                          "value": tagset})
-        # post unit to fleet and retry
-        for attempt in range(RETRIES):
-            try:
-                self._put_unit(name, {"desiredState": "launched", "options": unit})
-                break
-            except:
-                if attempt == (RETRIES - 1):  # account for 0 indexing
-                    raise
+        # post unit to fleet
+        self._put_unit(name, {"desiredState": "loaded", "options": unit})
 
     def _get_hostname(self, application_name):
         hostname = settings.UNIT_HOSTNAME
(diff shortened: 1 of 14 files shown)


bacongobbler:
Something that @gabrtv brought up was that we should have a benchmark test for the old `deis ps` and the new `deis ps`. This one will definitely take longer since it's a direct http call... Might be better to cache the request somewhere, somehow. Open to ideas!

bacongobbler:
I had to add a new `close_db_connections` decorator to make up for the post_transition decorator that was replaced. This is necessary in multi-threaded function calls which modify a model's state. Source code courtesy of https://code.djangoproject.com/ticket/22420#comment:17

aledbf:
> Might be better to cache the request somewhere, somehow. Open to ideas!

@bacongobbler why the cache? I know that it's going to be "slower" than a sql query but I need the real state not something from 60 seconds ago

gabrtv:
@aledbf when i discussed with @bacongobbler I was referring to an in-memory cache per request that avoided N Fleet UnitState queries for N containers (as currently implemented).  The result would be "real state", just a more efficient query.

bacongobbler:
We could optimize by querying the global unit state instead of querying for each individual job state, then filtering for the specific job. That would allow us to retrieve the state of all containers in one request rather than N requests, as @gabrtv mentions :)

bacongobbler:
After looking through fleet's API with @gabrtv, it seems like the global unit state request is paginated. That may be more trouble than it's worth; it could potentially hurt our common use case (between 1-5 containers). I'll have to do some benchmarks, but as long as it's not terribly slow when dealing with 20+ containers then we should be okay.

bacongobbler:
while fixing this up, I'm finding that `deis destroy` takes forever, but that's because we destroy each container in series rather than parallel. Shall I fix that up, similar to how we scale containers?

bacongobbler:
As an example:

```
><> deis destroy

 !    WARNING: Potentially Destructive Action
 !    This command will destroy the application: foo
 !    To proceed, type "foo" or re-run this command with --confirm=foo

> foo
Destroying foo...
done in 99s
Git remote deis removed
```

bacongobbler:
@gabrtv I was testing `deis ps` with 20 containers scaled up. The response time was < 1 second. I've also made App.destroy() delete its containers in parallel so we should be good now for manual testing :)

controller/scheduler/fleet.py
```diff
@@ -314,6 +327,36 @@ def _do_ssh(cmd):
...
+            "failed": "crashed",
+            "activating": "down",
+            "deactivating": "down",
+        }
+        try:
+            # NOTE (bacongobbler): this call to ._get_unit() also acts as a pre-emptive check to
+            # determine if the job no longer exists (will raise a RuntimeError on 404)
+            unit = self._get_unit(name)
+            state = self._wait_for_container_state(name)
+            activeState = state['systemdActiveState']
```
mboersma: Heh, ActiveState.

mboersma:
This should fix container state disagreement that we have seen from time to time. Code LGTM. Nice work @bacongobbler.

carmstrong:
Code LGTM. We can also remove the last note from http://docs.deis.io/en/latest/managing_deis/backing_up_data/ but in the sake of getting this PR merged, that can be done in a follow-up PR.

PR #2994, opened 2015-01-29
chore(controller): update djangorestframework to 3.0.4

mboersma:
See https://github.com/tomchristie/django-rest-framework/compare/3.0.3...3.0.4


bacongobbler:
If only we could use [this](https://github.com/tomchristie/django-rest-framework/commit/5bb348605e5dad3b58495b1fc56ea393194b89fb) with our current implementation of UUID fields. Sigh.

code lgtm

carmstrong:
Code LGTM.

PR #2998, opened 2015-01-30
ref(*): refactor dev- targets in Makefiles

carmstrong:
When running `make build`, sometimes components fail to build,
aborting the Makefile run. After fixing the issue, building again,
and running `make dev-release`, some components
have already been tagged and emit an error like "Tag asdf already
references image bcde".

This commit makes the following changes:
* `docker tag` will now force the tag, preventing a semi-successful build from preventing releases
* Prefixes all of the developer targets with `dev-` to clearly denote they are only for local development
* Splits the tag target out from the push target

These changes make the Makefile both more resilient and easier to
read/follow.


mboersma:
> `docker tag` will now force the tag

That's great--I run into that issue often. Code LGTM.

bacongobbler:
I think the dev-* targets look silly asides for dev-release (I'd change them back to the way they were so it's `make build tag push set-image` rather than `make build dev-tag dev-push dev-set-image`) but otherwise LGTM.

carmstrong:
> (I'd change them back to the way they were so it's make build tag push set-image rather than make build dev-tag dev-push dev-set-image)

Why wouldn't you just do `make build dev-release` in that case (and in most cases)? The change here is mostly for readability's sake in the Makefile, not that the targets would often be chained together like that, although they can be targeted separately if a tag succeeds but the push doesn't, for example. It doesn't really make sense that `tag` and `push` don't have `dev-`, whereas `release` does - they're all only used in local development.

Note that the only "fix" is to force the tags.

bacongobbler:
Sometimes when hacking along, I already have the images built, tagged and pushed to my local registry, but I need to spin up a new cluster before testing. I have a script that boots up a cluster, runs `make set-image`, installs the units and runs off with the cluster to do typical workflow testing. In that sense, I like to keep things the same unless we're simplifying the makefiles.

I guess I'm fine with the changes. I just don't like changing target names unless absolutely necessary as  others may have written tooling around them.

FYI you'll also need to update the root Makefile for these updated target names.

PR #3000, opened 2015-01-30
chore(release): bump user-data.example to v1.3.0-dev

ngpestelos:


mboersma:
There is no v1.3.0-dev `deisctl` installer or Docker Hub images, and we don't plan to create them, only released versions. So leaving these two strings at the previous release version ensures you can still provision from the `master` branch.

bacongobbler:
I agree with @mboersma on this one. When you deploy from master, the latest public release is used by default. This is because a public release is more explicit about the commit, whereas someone reporting an issue about 1.3.0-dev is ambiguous (any commit from 1.2.2 to 1.3.0 could be 1.3.0-dev).

If you want to run off master, you'll have to follow the guide to [Hacking on Deis](http://docs.deis.io/en/latest/contributing/hacking/) which includes instructions on setting up your own dev images for testing.

ngpestelos:
That makes sense, thank you.

PR #3006, opened 2015-01-30
Revert "fix(contrib): use 1GB per VM"

carmstrong:
Many tears were shed because the testing infrastructure was choking due to having too little memory. Trying to run Deis with too little memory invites a slew of issues. 

Reverts deis/deis#2746


mboersma:
The tradeoff here seems to be that if we use > 1024MB per node, you can't even get Deis started on a Mac with 8GB RAM. We lowered it to 1024MB to enable Deis clusters to be provisioned on this common configuration, but clearly you may run into issues if you stress it.

Currently the CI jobs work around this with a one-liner:
```sh
echo '$vb_memory = 2048' > config.rb
```
I'm fine with continuing to do that; I'm worried that we'll get complaints again from users just trying to test drive Deis.

carmstrong:
> I'm worried that we'll get complaints again from users just trying to test drive Deis.

If Deis doesn't work well on 1GB, we shouldn't easily allow people to run it that way. Our system requirements documentation even calls out that instances should have at least 2GB minimum. Folks can always manually edit the file if they want to try 1GB, but in that case it's clear that it's not recommended.

Deis is a distributed system - 8GB of RAM really isn't a ridiculous requirement. There's always Digital Ocean test machines, or Dokku for 1-node production workloads.

mboersma:
That sounds reasonable. Should we revert the original commit, or just change https://github.com/deis/deis/blob/master/Vagrantfile#L20 to 2048? We don't support `$num_instances < 3`.

gabrtv:
> Deis is a distributed system - 8GB of RAM really isn't a ridiculous requirement. There's always Digital Ocean test machines, or Dokku for 1-node production workloads.

Amen. LGTM.

carmstrong:
> Should we revert the original commit, or just change https://github.com/deis/deis/blob/master/Vagrantfile#L20 to 2048?

Yeah, I'll make that change.

carmstrong:
> Yeah, I'll make that change.

In another PR.

PR #3017, opened 2015-02-03
chore(builder, logger, logspout): remove unused references to BUILD_IMAGE

ngpestelos:
...from the Makefile

### DIFF
--- a/logspout/Makefile
+++ b/logspout/Makefile
@@ -2,7 +2,6 @@ include ../includes.mk
 
 COMPONENT := logspout
 DOCKER_IMAGE := deis/$(COMPONENT)
-BUILD_IMAGE := $(DOCKER_IMAGE)-build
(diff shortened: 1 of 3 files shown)


mboersma:
You're right, we don't actually use `$BUILD_IMAGE` in logger or builder. Code LGTM.

carmstrong:
Code LGTM. Thanks @ngpestelos!

bacongobbler:
We also don't use it in logspout, either. Could you please remove those references there as well? :) Thanks!

logspout/Makefile
```diff
@@ -12,12 +11,11 @@ build: check-docker
 
 clean: check-docker check-registry
```
bacongobbler: we no longer need these pre-trip checks for `make clean`

mboersma:
Thanks for being so flexible @ngpestelos. Code LGTM.

PR #3030, opened 2015-02-04
ref(store): rewrite store-volume to use ceph-fuse

### LINKED ISSUE #2970: store: Consider using ceph-fuse rather than the kernel client for Ceph FS (deis-store-volume)
When Ceph fixes bugs in the Linux kernel driver for Ceph FS, we have to wait for the patch to get accepted, a kernel to be released, and then a CoreOS release with that kernel version to make it to the stable channel.

Instead of using the kernel client, we could instead use [ceph-fuse](http://ceph.com/docs/master/man/8/ceph-fuse/), which allows us to upgrade to the latest any time the container is rebuilt. If we go this route, we would certainly have to test upgrades from the kernel client filesystem to a FUSE filesystem.

I'm hoping this will prevent Ceph bugs from taking down instances (which @ianblenke has seen), but it definitely allows us to quickly adopt upstream fixes (specifically [this one](http://lists.ceph.com/pipermail/ceph-users-ceph.com/2015-January/045838.html) reported by @lorieri, which has a patch pending).

carmstrong:
Several users have reported host crashes, presumably due to Ceph
kernel driver issues. This commit refactors the store-volume component
into a Docker container that uses ceph-fuse to mount the CephFS volume
instead of the kernel driver, so bugs will hopefully at most crash the
volume container and not the host.

**TODO:**
- [ ] Benchmark performance compared to the kernel driver 
- [ ] Test upgrades / switching from kernel-driven `deis-store-volume`

closes #2970

### DIFF
--- a/deisctl/units/deis-store-volume.service
+++ b/deisctl/units/deis-store-volume.service
@@ -5,10 +5,10 @@ Description=deis-store-volume
 EnvironmentFile=/etc/environment
 ExecStartPre=/usr/bin/mkdir -p /var/lib/deis/store
 ExecStartPre=/bin/sh -c "echo waiting for store-monitor... && until etcdctl get /deis/store/monSetupComplete >/dev/null 2>&1; do sleep 2; done"
-ExecStartPre=/bin/bash -c "HOSTS=`etcdctl ls /deis/store/hosts | cut -d/ -f5 | awk '{if(NR == 1) {printf $0} else {printf \",\"$0}}'` && cat /proc/mounts |grep '/var/lib/deis/store' || mount -t ceph $HOSTS:/ /var/lib/deis/store -o name=admin,secret=`etcdctl get /deis/store/adminKeyring | grep 'key =' | cut -d' ' -f3`"
-ExecStart=/usr/bin/tail -f /dev/null
-ExecStartPost=/bin/sh -c "test -d /var/lib/deis/store/logs || mkdir -p /var/lib/deis/store/logs"
-ExecStopPost=-/usr/bin/umount /var/lib/deis/store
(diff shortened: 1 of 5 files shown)


carmstrong:
Note that there's no guarantee we're merging this - rather, we want to see if it's more stable than the kernel driver.

ping @mgartner @ianblenke since I know they've seen crashes firsthand.

carmstrong:
An update: I think there's a timing issue with the mount. On the host where logger lives, we see the log files in `/var/lib/deis/store` on the host, but we don't in the `deis-store-volume` container. Not sure what's going on here, as the `store-volume` containers are successfully mounting the filesystem with ceph-fuse.

carmstrong:
Closing until this is fixed.

mgartner:
Sorry @carmstrong I didn't have time to test this yet. It feels like the issues @icblenke and I have been seeing should be fairly reproducible given the number of times we've seen crashes, but it remains elusive. I feel like it is somehow related to traffic or load on my servers though because my staging cluster lasted weeks without a single crash.

carmstrong:
@mgartner One observation we had in IRC the other day - are you using PV or HVM instances on EC2? @ianblenke is seeing crashes with HVM, but I haven't seen one and I'm running PV.

Any chance that is a difference between your clusters?

mgartner:
@carmstrong my staging cluster is PV, but I'm not sure what my production cluster was. I'm fairly certain it was also PV though, because I didn't adjust the default: https://github.com/deis/deis/blob/master/contrib/ec2/deis.template.json#L77

carmstrong:
> because I didn't adjust the default:

Yeah, then it's PV. Ok :(

PR #3036, opened 2015-02-05
fix(logger): create logRoot on startup

### LINKED ISSUE #2398: Potential file descriptor/memory leak in deis-logger
When deis-logger can't create or write to its log files, it seems to make runaway attempts to open the file numerous times per second. We have anecdotal evidence that this leads to high load and memory usage on the CoreOS host, leading to components crashing/restarting and having other oddities.

See https://github.com/deis/deis/issues/2372#issuecomment-61867924 for details.

carmstrong:
Currently, logger expects logRoot to exist. If it doesn't,
logger will throw errors hundreds of times per second as it tries
to write a logfile.

This commit makes logger create logRoot on startup.

Note that this doesn't necessarily close #2398, as it's conceivable that permissions could change (or directories removed) while logger is already running, so a single log file write could still fail.

refs #2398

### DIFF
--- a/logger/syslogd/syslogd.go
+++ b/logger/syslogd/syslogd.go
@@ -93,6 +93,13 @@ func (h *handler) mainLoop() {
 // Listen starts a new syslog server which runs until it receives a signal.
 func Listen(signalChan chan os.Signal, cleanupDone chan bool) {
 	fmt.Println("Starting syslog...")
+	// If logRoot doesn't exist, create it
+	// equivalent to Python's `if not os.path.exists(filename)`
+	if _, err := os.Stat(logRoot); os.IsNotExist(err) {
+		if err = os.MkdirAll(logRoot, 0777); err != nil {
+			log.Fatalf("unable to create logRoot at %s", logRoot)
+		}
+	}
 	// Create a server with one handler and run one listen gorutine
 	s := syslog.NewServer()
 	s.AddHandler(newHandler())


logger/syslogd/syslogd.go
```diff
@@ -93,6 +93,15 @@ func (h *handler) mainLoop() {
 // Listen starts a new syslog server which runs until it receives a signal.
 func Listen(signalChan chan os.Signal, cleanupDone chan bool) {
 	fmt.Println("Starting syslog...")
+	// If logRoot doesn't exist, create it
+	// equivalent to Python's `if not os.path.exists(filename)`
+	if _, err := os.Stat(logRoot); os.IsNotExist(err) {
+		err = os.Mkdir(logRoot, 0777)
+		if err != nil {
+			fmt.Printf("unable to create logRoot at %s", logRoot)
+			return
```
bacongobbler: let's change this with a `log.Fatalf` so it exits correctly, otherwise it'll return exit(0)

carmstrong: Done.

mboersma:
Code LGTM.

bacongobbler:
code LGTM! This should prevent 99% of the cases where this issue pops up. It's that danged `for loop` sitting right above which is the main issue, though we want to handle all messages quickly so I'm not sure how to clean that up atm.

PR #3037, opened 2015-02-05
feat(logspout): use custom datetime

### LINKED ISSUE #3032: Consider datetime format to ISO8601
Currently deis is using:
```
DEIS_DATETIME_FORMAT = '%Y-%m-%dT%H:%M:%S%Z'

const DEIS_DATETIME_FORMAT string = "2006-01-02T15:04:05MST"
```

ISO8601:
```
DEIS_DATETIME_FORMAT = '%Y-%m-%dT%H:%M:%S%z'

const DEIS_DATETIME_FORMAT string = "2006-01-02T15:04:05Z0700"
```

ISO8601 expects UTC be represented by 'Z', Go supports this, however Pythons strftime does not which makes the example above not 100% correct. I think instead we'd need to call .isoformat() - which kinda breaks the pattern of defining time format in one spot.

This would make it a little simpler for log ingestion etc.

Would this adversely affect existing clusters?

carmstrong:
This commit enables users to specify their own datetime format
by supplying the `DATETIME_FORMAT` environment variable when
using logspout. If none is supplied, logspout will continue to use
`DEIS_DATETIME_FORMAT`.

closes #3032

### DIFF
--- a/logspout/logspout.go
+++ b/logspout/logspout.go
@@ -73,7 +73,7 @@ func syslogStreamer(target Target, types []string, logstream chan *Log) {
 		// HACK: Go's syslog package hardcodes the log format, so let's send our own message
 		_, err = fmt.Fprintf(conn,
 			"%s %s[%s]: %s",
-			time.Now().Format(dtime.DEIS_DATETIME_FORMAT),
+			time.Now().Format(getopt("DATETIME_FORMAT", dtime.DEIS_DATETIME_FORMAT)),
 			tag,
 			pid,
 			logline.Data)
--- a/logspout/README.md
+++ b/logspout/README.md
@@ -48,6 +48,10 @@ By default, routes are ephemeral. But if you mount a volume to `/mnt/routes`, th
 
 See [Routes Resource](#routes-resource) for all options.
 
+#### Using a custom timestamp format
+
+By default, logspout will use the timestamp format `2006-01-02T15:04:05MST`. A custom format can be specified by setting the `DATETIME_FORMAT` environment variable.
+
 ## HTTP API
 
 ### Streaming Endpoints


logspout/logspout.go
```diff
@@ -73,7 +77,7 @@ func syslogStreamer(target Target, types []string, logstream chan *Log) {
 		// HACK: Go's syslog package hardcodes the log format, so let's send our own message
 		_, err = fmt.Fprintf(conn,
 			"%s %s[%s]: %s",
-			time.Now().Format(dtime.DEIS_DATETIME_FORMAT),
+			time.Now().Format(format),
```
bacongobbler: There's a function defined above which will save you a few lines of trouble:

```
getopt("DATETIME_FORMAT", dtime.DEIS_DATETIME_FORMAT)
```

bacongobbler:
lgtm, though how do we expose this to users? By asking them to modify the logspout unit file?

carmstrong:
> By asking them to modify the logspout unit file?

Yeah, I was thinking about that. Using an etcd key here would be ideal, but adding confd to logspout just for this one change doesn't make sense. I think that it's a rare enough situation when a user would want to modify this that it's not the worst thing to have to maintain a separate unit file for logspout, but I'm open to suggestions.

Another thought is making this customizable upstream in progrium/logspout.

logspout/logspout.go
carmstrong: Kewl!

jasoncox:
Could we set a platform value in etcd like /deis/logspout/datetime_format and query etcd directly in logspout.go as is done for logger host?

bacongobbler:
> Could we set a platform value in etcd like /deis/logspout/datetime_format and query etcd directly in logspout.go as is done for logger host?

Personally I'd rather steer away from further etcd integrations if at all possible. etcd is a distributed key/value store for service discovery, not necessarily a centralized configuration management solution... Though we seem to have already tied ourselves down that route, so I guess we've already set that precedent.

bacongobbler:
that and we also utilize go-etcd in logspout to figure out where logger's hiding so I'm partially okay with it

bacongobbler:
I'd be quite happy with documenting how to customize the date by sending in an envvar in logspout's README. Does that seem reasonable? I don't imagine that this would be a common use case which we'd need to expose in etcd.

carmstrong:
> I don't imagine that this would be a common use case which we'd need to expose in etcd.

I agree. I don't think this is a common enough scenario to warrant another etcd dependency. I'll update the commit to add the environment variable to the readme.

carmstrong:
I updated the README to specify the environment variable.

bacongobbler:
:+1:

mboersma:
Code LGTM. An env var seems like a good enough solution for the power users who will want to change the date/time format.

PR #3039, opened 2015-02-06
Let docker-in-docker choose its storage driver

### LINKED ISSUE #2975: logspout: /var/lib/docker/init/dockerinit-1.4.1: cannot allocate memory
Builder ends up restarting over and over, and looking at the instances, logspout was restarting on all:
```console
core@deis-2 ~ $ journalctl -fu deis-logspout
-- Logs begin at Tue 2015-01-27 22:56:15 UTC. --
Jan 27 23:36:46 deis-2 systemd[1]: deis-logspout.service holdoff time over, scheduling restart.
Jan 27 23:36:46 deis-2 systemd[1]: Stopping deis-logspout...
Jan 27 23:36:46 deis-2 systemd[1]: Starting deis-logspout...
Jan 27 23:36:46 deis-2 sh[5108]: Error: No such image or container: deis-logspout
Jan 27 23:36:46 deis-2 systemd[1]: Started deis-logspout.
Jan 27 23:36:46 deis-2 systemd[1]: deis-logspout.service: main process exited, code=exited, status=1/FAILURE
Jan 27 23:36:46 deis-2 sh[5119]: time="2015-01-27T23:36:46Z" level="fatal" msg="Error response from daemon: Cannot start container 070c2884756f7a5b93b7a43b1c5a01f91f5276aa9d136a87eb39ac9a0336473f: fork/exec /var/lib/docker/init/dockerinit-1.4.1: cannot allocate memory"
Jan 27 23:36:46 deis-2 systemd[1]: Unit deis-logspout.service entered failed state.
Jan 27 23:36:46 deis-2 systemd[1]: deis-logspout.service failed.
Jan 27 23:36:46 deis-2 docker[5137]: deis-logspout
Jan 27 23:36:51 deis-2 systemd[1]: deis-logspout.service holdoff time over, scheduling restart.
Jan 27 23:36:51 deis-2 systemd[1]: Stopping deis-logspout...
Jan 27 23:36:51 deis-2 systemd[1]: Starting deis-logspout...
Jan 27 23:36:51 deis-2 sh[5187]: Error: No such image or container: deis-logspout
Jan 27 23:36:51 deis-2 systemd[1]: Started deis-logspout.
Jan 27 23:36:52 deis-2 sh[5198]: 2015/01/27 23:36:52 attach: 0c82c510868b deis-logspout success
Jan 27 23:36:52 deis-2 sh[5198]: 2015/01/27 23:36:52 attach: 1ae37b8d7083 deis-router success
Jan 27 23:36:52 deis-2 sh[5198]: 2015/01/27 23:36:52 attach: 66d720dfdcda deis-database success
Jan 27 23:36:52 deis-2 sh[5198]: 2015/01/27 23:36:52 attach: a12d3b495c70 deis-cache success
Jan 27 23:36:52 deis-2 sh[5198]: 2015/01/27 23:36:52 attach: 390fff8c3f4f deis-publisher success
Jan 27 23:36:52 deis-2 sh[5198]: 2015/01/27 23:36:52 attach: 0aa2a15097ac deis-store-metadata success
Jan 27 23:36:52 deis-2 sh[5198]: 2015/01/27 23:36:52 attach: 84ef4afc802b deis-store-daemon success
Jan 27 23:36:52 deis-2 sh[5198]: 2015/01/27 23:36:52 attach: d1b335cdaf35 deis-store-monitor success
Jan 27 23:36:52 deis-2 sh[5198]: 2015/01/27 23:36:52 etcd: http://172.17.8.101:4001
Jan 27 23:36:53 deis-2 sh[5198]: 2015/01/27 23:36:53 url: 100: Key not found (/deis/logs) [21297]
Jan 27 23:36:53 deis-2 systemd[1]: deis-logspout.service: main process exited, code=exited, status=1/FAILURE
Jan 27 23:36:53 deis-2 docker[5240]: Error response from daemon: No such container: deis-logspout
Jan 27 23:36:53 deis-2 docker[5240]: time="2015-01-27T23:36:53Z" level="fatal" msg="Error: failed to remove one or more containers"
```
CoreOS beta channel and stable seem to be fine. See https://ci.deis.io/job/test-coreos-alpha/5/console for an example.

mboersma:
Docker at one point had to be coerced into using `btrfs` on CoreOS--which had to be overridden with `aufs` for functional testing on non-`btrfs` hosts. In testing with CoreOS alpha, our choice of `btrfs` fails, since it won't work on a host filesystem formatted with `ext4`.

This change removes the `--storage-driver` override code and allows Docker(-in-Docker) to choose the best driver. This is still `btrfs` when the host filesystem matches--I verified several times inside the container. On CoreOS alpha, Docker chooses `devicemapper` (both `aufs` and `overlay` fail with "not supported"). This would be fine except there are no `/dev/loopN` devices yet, only the master `/dev/loop-control` device, and Docker will fail with "no loopback devices available." Calling `losetup -f` once before starting deis-builder ensures that the standard set of loopback devices is present and that the `devicemapper` backend will work.

(For reasons I couldn't discern, `losetup -f` from within the builder boot script fails, but works fine as a separate `ExecStartPre` task.)

I've tested this multiple times with `test-integration.sh` against all CoreOS channels.

Closes #2975.

### DIFF
--- a/deisctl/units/deis-builder.service
+++ b/deisctl/units/deis-builder.service
@@ -7,6 +7,7 @@ TimeoutStartSec=30m
 ExecStartPre=/bin/sh -c "docker inspect deis-builder-data >/dev/null 2>&1 || docker run --name deis-builder-data -v /var/lib/docker ubuntu-debootstrap:14.04 /bin/true"
 ExecStartPre=/bin/sh -c "IMAGE=`/run/deis/bin/get_image /deis/builder` && docker history $IMAGE >/dev/null || docker pull $IMAGE"
 ExecStartPre=/bin/sh -c "docker inspect deis-builder >/dev/null && docker rm -f deis-builder || true"
+ExecStartPre=-/bin/sh -c "/sbin/losetup -f"
 ExecStart=/bin/sh -c "IMAGE=`/run/deis/bin/get_image /deis/builder` && docker run --name deis-builder --rm -p 2223:22 --volumes-from=deis-builder-data -c 800 -e EXTERNAL_PORT=2223 -e HOST=$COREOS_PRIVATE_IPV4 --privileged $IMAGE"
 ExecStartPost=/bin/sh -c "echo 'Waiting for builder on 2223/tcp...' && until echo 'dummy-value' | ncat $COREOS_PRIVATE_IPV4 2223 >/dev/null 2>&1; do sleep 1; done"
 ExecStartPost=/usr/bin/docker exec deis-builder /usr/local/bin/push-images
--- a/builder/README.md
+++ b/builder/README.md
@@ -22,7 +22,6 @@ install, and start **deis/builder**.
   its configuration (default: */deis/builder*)
 * **ETCD_TTL** sets the time-to-live before etcd purges a configuration
   value, in seconds (default: *10*)
-* **STORAGE_DRIVER** sets the Docker-in-Docker storage driver (default: btrfs)
 * **PORT** sets the TCP port on which the builder listens (default: *2222*)
 
 ## License
--- a/builder/image/bin/boot
+++ b/builder/image/bin/boot
@@ -14,7 +14,6 @@ export ETCD_PORT=${ETCD_PORT:-4001}
 export ETCD="$HOST:$ETCD_PORT"
 export ETCD_PATH=${ETCD_PATH:-/deis/builder}
 export ETCD_TTL=${ETCD_TTL:-10}
-export STORAGE_DRIVER=${STORAGE_DRIVER:-btrfs}
 
 # wait for etcd to be available
 until etcdctl --no-sync -C $ETCD ls >/dev/null 2>&1; do
@@ -51,7 +50,7 @@ CONFD_PID=$!
 test -e /var/run/docker.sock && rm -f /var/run/docker.sock
 
 # spawn a docker daemon to run builds
-docker -d --storage-driver=$STORAGE_DRIVER --bip=172.19.42.1/16 --insecure-registry 10.0.0.0/8 --insecure-registry 172.16.0.0/12 --insecure-registry 192.168.0.0/16 --insecure-registry 100.64.0.0/10 &
+docker -d --bip=172.19.42.1/16 --insecure-registry 10.0.0.0/8 --insecure-registry 172.16.0.0/12 --insecure-registry 192.168.0.0/16 --insecure-registry 100.64.0.0/10 &
 DOCKER_PID=$!
 
 # wait for docker to start
--- a/builder/tests/builder_test.go
+++ b/builder/tests/builder_test.go
@@ -49,7 +49,6 @@ func TestBuilder(t *testing.T) {
 			"--rm",
 			"-p", port+":22",
 			"-e", "PORT=22",
-			"-e", "STORAGE_DRIVER=aufs",
 			"-e", "HOST="+host,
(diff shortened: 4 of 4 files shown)


carmstrong:
Interesting that devicemapper is the default on alpha - I thought with the move to ext4 they were using overlayfs...

mboersma:
They are using `overlay` on CoreOS host, but I couldn't get it to work inside the container: "not supported." I'm not sure why not. It's still experimental from Docker's point of view, so you have to specify it since the chooser logic won't ever choose `overlay`. I'm also wondering if they took `aufs` support out of the kernel, because that ought to have worked (and it's the first one Docker tries).

carmstrong:
> They are using overlay on CoreOS host, but I couldn't get it to work inside the container

Interesting. So on the host we're using overlayfs, but in the builder we're using devicemapper? Shouldn't make a difference that I can think of, but it's interesting...

mboersma:
@gabrtv had a more radical idea I want to try: bind-mount in `/var/lib/docker` and the `docker` binary directly, and forget about Docker-in-Docker. I would like to try that approach, but it will break currently on CoreOS alpha and beta since they have Docker 1.4.1, which if used in our builder will have auth errors against deis-registry. That's a separate problem that needs to be fixed before his approach can be viable.

deisctl/units/deis-builder.service
```diff
@@ -7,6 +7,7 @@ TimeoutStartSec=30m
 ExecStartPre=/bin/sh -c "docker inspect deis-builder-data >/dev/null 2>&1 || docker run --name deis-builder-data -v /var/lib/docker ubuntu-debootstrap:14.04 /bin/true"
 ExecStartPre=/bin/sh -c "IMAGE=`/run/deis/bin/get_image /deis/builder` && docker history $IMAGE >/dev/null || docker pull $IMAGE"
 ExecStartPre=/bin/sh -c "docker inspect deis-builder >/dev/null && docker rm -f deis-builder || true"
+ExecStartPre=-/bin/sh -c "/sbin/losetup -f"
```
bacongobbler: Is it intended to be run on the host before we run the builder? In your OP, you say

> For reasons I couldn't discern, losetup -f from within the builder boot script fails, but works fine as a separate ExecStartPre task.

Which means that you intended to run this inside the container rather than on the host. Is that correct?

carmstrong: I read that as he first tried doing that inside the container, which didn't work, so he instead moved it to the host.

mboersma: Yes, what @carmstrong said, sorry I didn't explain it well. Inside the container, `losetup -f` returns an error rather than just-in-time creating the `/dev/loopN` devices. On the host, `losetup` works fine, and then those devices are available inside the deis-builder container.

bacongobbler: okay. thanks for clearing that up!

bacongobbler:
I'm all for letting the daemon choose the proper device driver, granted that it's smart enough to do so. LGTM :)

mboersma:
> granted that it's smart enough to do so

Docker tries `aufs`, `btrfs`, `devicemapper`, `vfs` in order and uses the first one that passes the filesystem and device tests specific to each driver. (I think when we hard-coded `btrfs` it was labelled experimental, as `overlay` is currently.)

carmstrong:
> as overlay is currently.

That makes sense. Code LGTM.

carmstrong:
Code LGTM.

PR #3046, opened 2015-02-10
chore(deisctl): update vendored fleet code to 0.9.0

mboersma:
This changes little from `deisctl`'s point of view, mostly updating the location of the `golang.org/x/crypto/ssh` package and its implementation.


Godeps/Godeps.json
```diff
@@ -40,53 +35,53 @@
 		},
 		{
 			"ImportPath": "github.com/coreos/fleet/client",
-			"Comment": "v0.3.2-976-g96d7f98",
-			"Rev": "96d7f986de72c3fca04b627ddd6578f7206cb7e4"
+			"Comment": "v0.9.0",
```
bacongobbler: is there a way we could remove all of this duplicate import paths with github.com/coreos/fleet instead, or is this simply autogenerated Godeps?

bacongobbler:
:+1: for bumping versions

carmstrong:
Code LGTM.

carmstrong:
We're using fleet 0.9.0 as of #3046.

Godeps/Godeps.json
mboersma: It's autogenerated; that's how `godep` likes to track individual packages. (And in this case there is no actual code in `github.com/coreos/fleet`, so it's not a Go package strictly speaking, neither is `github.com/deis/deis`.)

PR #3056, opened 2015-02-10
fix(controller): remove timeout on container launch

### LINKED ISSUE #3054: Remove scheduler timeouts / Taking too long to pull a run container fails with 'failed to create container'
Debugging Deis on Azure led us to initially suspect an Azure-specific issue, but we have seen this on other platforms. Any time a `deis run` takes longer than 60 seconds to complete the `docker pull`, the client errs with `failed to create container`.

This leads to a deeper issue with our timeouts -- the router uses a [20 minute timeout](https://github.com/deis/deis/blob/f03078180bd830cec7a23bd67acfc19e5f966f42/router/image/templates/nginx.conf#L152-L153). In the fleet scheduling logic, we have various loops which perform actions:

* [_wait_for_container](https://github.com/deis/deis/blob/f03078180bd830cec7a23bd67acfc19e5f966f42/controller/scheduler/fleet.py#L163)
* [_wait_for_destroy](https://github.com/deis/deis/blob/f03078180bd830cec7a23bd67acfc19e5f966f42/controller/scheduler/fleet.py#L180)
* [run](https://github.com/deis/deis/blob/f03078180bd830cec7a23bd67acfc19e5f966f42/controller/scheduler/fleet.py#L216)
* [wait for launch](https://github.com/deis/deis/blob/f03078180bd830cec7a23bd67acfc19e5f966f42/controller/scheduler/fleet.py#L260)
* [wait for start](https://github.com/deis/deis/blob/f03078180bd830cec7a23bd67acfc19e5f966f42/controller/scheduler/fleet.py#L269)
* [wait for completion](https://github.com/deis/deis/blob/f03078180bd830cec7a23bd67acfc19e5f966f42/controller/scheduler/fleet.py#L282)

We really shouldn't be "budgeting" the 20 minutes across these timeouts, as we have no idea how long any one will take at any one time. For example, on Azure, a `docker pull` for my run container took 5 minutes, but our pollwait loop only waits for 60s. Letting each of these loops run indefinitely and instead letting the worker/router handle the total request timeout is probably preferred here.

carmstrong:
We can't reliably guess how long this will take.

closes #3054

### DIFF
--- a/controller/scheduler/fleet.py
+++ b/controller/scheduler/fleet.py
@@ -257,7 +257,8 @@ class FleetHTTPClient(object):
             return rc, output
 
         # wait for container to launch
-        for _ in range(60):
+        # we loop indefinitely here, as we have no idea how long the docker pull will take
+        while True:
             rc, _ = _do_ssh('docker inspect {name}'.format(**locals()))
             if rc == 0:
                 break


bacongobbler:
Code LGTM. Thanks for catching this!

mboersma:
Code LGTM. I've run into this several times.

PR #3061, opened 2015-02-11
chore(controller): update djangorestframework to 3.0.5

mboersma:
See http://www.django-rest-framework.org/topics/release-notes/


carmstrong:
Code LGTM.

bacongobbler:
Seems like there may be some more improvements which we can make in the views. We still reference `self.model` in a few of the views, which seems like it's more or less deprecated: https://github.com/tomchristie/django-rest-framework/pull/2486

Code LGTM, just thought it'd be a good topic to bring up

PR #3064, opened 2015-02-11
fix(client): display default answer for auth:cancel prompt

bacongobbler:
Just making it clear that if an answer is not supplied, the default answer is no.


mboersma:
Code LGTM.

carmstrong:
:ship:

PR #3065, opened 2015-02-11
feat(ec2/cloudformation): use SSD EBS volumes by default

carmstrong:
Added a commit to remove provisioned iops as an option based on the discussion starting at https://github.com/deis/deis/pull/2973#discussion_r23650395

replaces #2973


mboersma:
I tested these changes in #2973, and this removes the problematic `io1` option. Code LGTM.

bacongobbler:
haven't tested this manually but it looks like it captures the use case. Code LGTM!

PR #3068, opened 2015-02-12
fix(controller): kill processes removed from procfile

### LINKED ISSUE #2930: processes removed from Profile cannot be stopped
When an item is removed from the source code procfile, deis does not stop the containers. After the item is no longer in the Procfile it becomes impossible to stop the containers:
```
Brendans-MacBook-Air:hercules bfosberry$ deis ps | grep transacted
--- transacted_tender_importer:
transacted_tender_importer.1 up (v109)
--- transacted_item_importer:
transacted_item_importer.1 up (v109)
Brendans-MacBook-Air:hercules bfosberry$ deis scale transacted_tender_importer=0 
400 BAD REQUEST
Container type transacted_tender_importer does not exist in application
```

bacongobbler:
When a process type is removed in a new build, the old containers are
not killed, and there's no way for the user to stop them. This change
will destroy the containers that no longer exist in the procfile.

closes #2930

### DIFF
--- a/controller/api/models.py
+++ b/controller/api/models.py
@@ -255,6 +255,9 @@ class App(UuidAuditedModel):
     def _destroy_containers(self, to_destroy):
         """Destroys containers via the scheduler"""
         destroy_threads = []
+        if not to_destroy:
+            # do nothing if we didn't request any containers
+            return
         for c in to_destroy:
             destroy_threads.append(threading.Thread(target=c.destroy))
         [t.start() for t in destroy_threads]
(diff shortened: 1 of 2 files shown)


carmstrong:
I like the clean implementation. Code LGTM!

mboersma:
Great to have this fixed. Code LGTM.

aledbf:
@bacongobbler with this PR I see this error in the controller:
```ERROR example: aborting, failed to destroy some containers```
the build is ok but and version is not incremented

bacongobbler:
The unit tests are failing so there's bound to be issues :)

bacongobbler:
fixed and tests passing, PTAL when you have time :)

carmstrong:
Code LGTM.

mboersma:
Code LGTM still. :smile:

Issue #3022, opened 2015-02-04
Need a way to stop one-off stale commands


glogiotatidis:
With `deis info` I get 

```
=== xx-prod Application
{
  "updated": "2015-02-04T10:08:37UTC", 
  "uuid": "b1812xxx1", 
  "created": "2015-01-20T12:46:39UTC", 
  "url": "xx-prod.masterfirefoxos.com", 
  "owner": "xx", 
  "id": "xx-prod", 
  "structure": {
    "cmd": 3, 
    "run": 32
  }
}
```
 1. Deis believes that there are 32 "run" processes but "docker ps" in all hosts returns 0 "run" processes
 2. Even if there were 32 running processes, we need a way to stop them

`deis scale run=0` returns 
```
400 BAD REQUEST
Container type run does not exist in application
```

bacongobbler:
I think this is just false info that shouldn't be shown in the client. run processes are the process type when you perform `deis run`, which mean you've run 32 `deis apps:run` commands against it. :)

glogiotatidis:
I didn't explain myself correctly:
It should totally be shown in the client, when they do exist. These don't exist (no docker containers related to the "run" command" and I've no way to reset Deis.

bacongobbler:
These containers are [killed after they're executed](https://github.com/deis/deis/blob/master/controller/scheduler/fleet.py#L306-L308), so they do not exist in the scheduler. It's just an implementation detail of the platform so we can run multiple one-off commands in parallel. We keep the database entry around such that each one-off job has a unique number, so there's nothing that you need to do on your end to "reset Deis".

bacongobbler:
what version of deis are you you running, by the way?

glogiotatidis:
OK I get that but in this case why other apps on which I've used `deis run` don't display this line?

```
=== xx-dev Application
{
  "updated": "2015-01-20T14:59:25UTC", 
  "uuid": "93d069d8-xx-4bc8-9ef3-163c1203c502", 
  "created": "2015-01-20T12:45:17UTC", 
  "url": "xx-dev.masterfirefoxos.com", 
  "owner": "xx", 
  "id": "xx-dev", 
  "structure": {
    "cmd": 3
  }
}
```

(I'm running 1.1.1)

glogiotatidis:
More info:
Running other commands in the same app doesn't increase the displayed "run" number, i.e. the output of `deis info` is exactly the same. I guess that Deis believes that there are 32 currently running "run" commands.

[...]

PR #3084, opened 2015-02-13
fix(controller): ignore run procs when updating structure field

mboersma:
Refs #3022, which still needs a cleanup fix for the fleet unit files that can be orphaned in that situation if a worker actually times out. ~~It wasn't reasonable to write a unit or integration test for this case unfortunately, IMHO, but I await the judgment of other maintainers.~~ Implemented a controller unit test that fails before this fix, passes afterward. :green_heart:

### DIFF
--- a/controller/api/models.py
+++ b/controller/api/models.py
@@ -228,7 +228,8 @@ class App(UuidAuditedModel):
             if to_remove:
                 self._destroy_containers(to_remove)
         # save new structure to the database
-        vals = self.container_set.values('type').annotate(Count('pk')).order_by()
+        vals = self.container_set.exclude(type='run').values(
+            'type').annotate(Count('pk')).order_by()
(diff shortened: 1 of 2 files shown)


bacongobbler:
I think this can easily be captured in the unit tests. Create an app, add a build, run a command, scale up web/cmd, then check that the structure contains only the web/cmd proctype. :)

mboersma:
> I think this can easily be captured in the unit tests.

Oh, in test_container I missed we had passing tests for `deis run`. Your wish is my command @bacongobbler.

mboersma:
I updated this with a unit test. Ready for review again.

carmstrong:
Code LGTM. Nice tests. :thumbsup:

gabrtv:
Code to tests ratio is pretty funny :smiley_cat: LGTM.

PR #3088, opened 2015-02-14
feat(router): add optional query string affinity support

wenzowski:
To be enabled with extreme prejudice. Twelve-factor processes should be
stateless and share-nothing. Sticky sessions should not be used to cache
user data in memory. Use a backing store to persist session state.

Valid uses, however, do exist and are widespread: eg. see primus/primus#147

> HTTP implementations are expected to engage in connection management,
> which includes maintaining the state of current connections,
> establishing a new connection or reusing an existing connection,
> processing messages received on a connection, detecting connection
> failures, and closing each connection.
> <cite>[RFC 7230]</cite>

Many real time systems rely on detection of connection failures at the
application level. Without session affinity, detection of connection
failures stops at the load balancer and causes undesirable side-effects
for many implementations. For instance, the Google BrowserChannel implementation
that may have led @zag2art to #2060 relies on detection of connection
failures, as does socket.io

Care should be taken when using this setting in conjunction with a 3rd party
Layer-7 load balancer. At this time, Amazon's ELB and Rackspace's Cloud Load
Balancer, for instance, only support cookie-based affinity which must be
manually enabled. Many OSI Layer-4 load balancers, such as the Azure Load
Balancer and GCE Network Load Balancing, currently perform [RFC 2474] 5-tuple
"microflow" hashing by default.

[RFC 2474]: https://tools.ietf.org/html/rfc2474
[RFC 7230]: https://tools.ietf.org/html/rfc7230

### DIFF
--- a/docs/customizing_deis/router_settings.rst
+++ b/docs/customizing_deis/router_settings.rst
@@ -37,6 +37,7 @@ setting                                      description
 /deis/controller/host                        host of the controller component (set by controller)
 /deis/controller/port                        port of the controller component (set by controller)
 /deis/domains/*                              domain configuration for applications (set by controller)
+/deis/router/affinityArg                     for requests with the indicated query string variable, hash its contents to perform session affinity (default: undefined)
 /deis/router/bodySize                        nginx body size setting (default: 1m)
 /deis/router/builder/timeout/connect         proxy_connect_timeout for deis-builder (default: 10000). Unit in miliseconds
 /deis/router/builder/timeout/read            proxy_read_timeout for deis-builder (default: 1200000). Unit in miliseconds
--- a/router/build.sh
+++ b/router/build.sh
@@ -10,8 +10,8 @@ if [[ -z $DOCKER_BUILD ]]; then
   exit 1
 fi
 
-export VERSION_NGINX=nginx-1.6.2
-export VERSION_TCP_PROXY=0.4.5
+export VERSION_NGINX=nginx-1.7.10
+export VERSION_TCP_PROXY=f2156eff9f6621aaf601eaa8dee40c6820dea0b0
 export VERSION_NAXSI=0d53a64ed856e694fcb4038748c8cf6d5551a603
 
 export BUILD_PATH=/tmp/build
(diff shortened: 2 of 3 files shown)


deis-admin:

wenzowski:
I have no idea how to write tests for this

wenzowski:
ugh, will fail with `VERSION_NGINX=nginx-1.6.2` as the hash directive appeared in 1.7.2

wenzowski:
rebased on #3090 which this now depends on

carmstrong:
@wenzowski So we discussed this in our code review today, and have three comments:

* We should bump nginx to 1.7.10 so we're at the latest point release in that branch
* How does the [hash](http://nginx.org/en/docs/http/ngx_http_upstream_module.html#hash) directive support multiple nginx instances? That is, does it hash consistently across routers?
* I'm not sure why we're using ` $arg_{{ $affinityArg }}` -- it seems like we want just `{{ $affinityArg }}` there, but we should document the options here (i.e. `$request_uri`, etc.)

wenzowski:
I'll resubmit with 1.7.10 First time I built it, it didn't work correctly with 1.7.10, but I'm guessing that will be solved by bumping tcp proxy (as was done here).

I'm running this on a cluster as-is: `$arg_{{ $affinityArg }}` is definitely intended.

$arg_*whatever* an [embedded variable](http://nginx.org/en/docs/http/ngx_http_core_module.html#variables) announced in nginx 0.7.8 that allows us to extract a specific var from the query string.

In my case, the query string `a` represents random id generated by a javascript client which persists until refresh, window close, or any other javascript-stopper.

The nginx team stole the `hash $var consistent;` directive from [libketama](https://github.com/RJ/ketama). As long as all upstreams are reachable from all routers, it won't matter which router the request passes through. Once you start thinking about failure modes though, it all goes haywire. I'm running microflow to nginx as a precaution since my entire stack is wildly experimental.

But libketama is designed for minimal bucket reassignment in case of upstream failure: only those buckets served by the now missing upstream will be reassigned.

wenzowski:
broken or needs rerun?

carmstrong:
> broken or needs rerun?

Waiting on Jenkins to come back from the dead. We should also add some documentation around this, showing examples of the affinity variable. At the very least, we should link to the nginx documentation.

wenzowski:
Should docs for this just update [router_settings.rst](https://github.com/deis/deis/blob/master/docs/customizing_deis/router_settings.rst)?
I'll start with that but please let me know if more is needed.

carmstrong:
> Should docs for this just update router_settings.rst?
> I'll start with that but please let me know if more is needed.

I think that's probably sufficient. Something more common would probably warrant a note in "Managing Deis", but pointing folks to the router settings is probably sufficient.

wenzowski:
cool, thanks @carmstrong. have this + #3264 deployed in wenzowski/deis-router:git-ea6e3af and it's working nicely.

wenzowski:
cherry-picked to master & docs squashed into ce038ed

let me know if anything else is needed to land this

carmstrong:
> let me know if anything else is needed to land this

Just Jenkins passing, at this point I think.

wenzowski:
`Error response from daemon: No such container: deis-store-volume` is the same failure from #3298 which is docs-only so either master is red or there are still some lingering ci woes. Either way, let me know when I should rebase & rerun.

wenzowski:
cffi failure pip installing wal-e deps isn't due to this commit

carmstrong:
> cffi failure pip installing wal-e deps isn't due to this pr

That should be fixed now - can you rebase this on master, @wenzowski?

wenzowski:
rebased

carmstrong:
Code LGTM.

deis-admin:

mboersma:
Code LGTM. Great stuff, thanks for sticking with this one @wenzowski!

wenzowski:
glad to see it landed, thanks @carmstrong & @mboersma

PR #3089, opened 2015-02-14
feat(router): add optional ip_hash affinity support

wenzowski:
When enabled, this makes all requests to all apps sticky. I'd much
prefer #3088 but it looks like that means an upgrade to mainline
since the hash directive appeared in nginx-1.7.2

### DIFF
--- a/docs/customizing_deis/router_settings.rst
+++ b/docs/customizing_deis/router_settings.rst
@@ -57,6 +57,7 @@ setting                                      description
 /deis/router/gzipVary                        nginx gzipVary setting (default: on)
 /deis/router/gzipDisable                     nginx gzipDisable setting (default: "msie6")
 /deis/router/gzipTypes                       nginx gzipTypes setting (default: "application/x-javascript application/xhtml+xml application/xml application/xml+rss application/json text/css text/javascript text/plain text/xml")
+/deis/router/ipHash                          nginx ip_hash setting (default: false)
 /deis/router/serverNameHashMaxSize           nginx server_names_hash_max_size setting (default: 512)
 /deis/router/serverNameHashBucketSize        nginx server_names_hash_bucket_size (default: 64)
 /deis/router/sslCert                         cluster-wide SSL certificate
--- a/router/image/templates/nginx.conf
+++ b/router/image/templates/nginx.conf
@@ -129,6 +129,8 @@ http {
     {{ $useSSL := or .deis_router_sslCert "false" }}
     {{ $domains := .deis_domains }}{{ range $service := .deis_services }}{{ if $service.Nodes }}
     upstream {{ Base $service.Key }} {
+        {{ $ipHash := or .deis_router_ipHash "false" }}{{ if eq $ipHash "true" }}ip_hash;
+        {{ end }}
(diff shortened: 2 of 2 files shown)


deis-admin:

wenzowski:
another alternative to #3088 might be [evanmiller/nginx_upstream_hash](https://github.com/evanmiller/nginx_upstream_hash)

bacongobbler:
I think we specifically closed out some proposals for sticky sessions because it breaks 12-factor best practices: #2060 

A general concern I have with the router now is that we cannot solve everyone's use case... Maybe it's time to take a better look at the router and see if there's a way we can make it more programmatic to customize, like with https://github.com/progrium/nginx-appliance.

All those concerns aside, I do recall our BDFL talking about potentially adding in something like sticky sessions for some legacy app support, even though the "true solution" in 12-factor apps is to store session data server-side. ping @gabrtv

wenzowski:
Sticky sessions don't inherently break 12-factor ideals...but they have historically had a tendency to do so by enabling application state to be stored in the running instance. To be absolutely clear here, my goal is not to store application state but to ensure app instances can be made aware of http connection state as appropriate.

The WebSockets upgrade handshake, for example, has a [standard mechanism] to signal initiation of this connection-level selective-stickyness such that the connection to the application instance is kept alive though a proxy that would [normally terminate] it. Should the addition of WebSockets preclude the possibility of a 12-factor app?

I contest the statement that
> Sticky sessions are a violation of twelve-factor and should never be used or relied upon. Session state data is a good candidate for a datastore that offers time-expiration, such as Memcached or Redis. <cite>[12factor.net]</cite>

...if the application implementation is stateless from a _user_  or _auth_ session perspective. Storing protocol state in memcached/redis results in a game of race condition whackamole (at least for me). What I'm looking for here are supporting pragmatic equivalents to the [keepalive] directive in an HTTP/1.1 world.

> Clients are also encouraged to consider the use of Connection:
> keep-alive in requests carefully; while they can enable persistent
> connections with HTTP/1.0 servers, clients using them will need to
> monitor the connection for "hung" requests (which indicate that the
> client ought stop sending the header field), and this mechanism ought
> not be used by clients at all when a proxy is being used.
> <cite>[RFC 7230 A.1.2.]</cite>

For a web application needing a full duplex stream to a remote server, be that via WebSockets, socket.io, engine.io, sockjs, browserchannel, webchannel, or something hand-rolled, knowing when that HTTP/1.1 keepalive-by-default connection has "hung" is critical.

This is an area where many existing PaaS providers are lagging. Would be really cool to see Deis competitve with Modulus, Nodejitsu, and maybe even MDG's Galaxy!

[standard mechanism]: https://tools.ietf.org/html/rfc6455#section-1.8
[normally terminate]: http://nginx.com/blog/http-keepalives-and-web-performance/
[12factor.net]: http://12factor.net/processes#vi_processes
[keepalive]: http://nginx.org/en/docs/http/ngx_http_upstream_module.html#keepalive
[RFC 7230 A.1.2.]: https://tools.ietf.org/html/rfc7230#appendix-A.1.2
[RFC 7230 p.53]: https://tools.ietf.org/html/rfc7230#page-53

carmstrong:
Closing in favor of #3088.

PR #3091, opened 2015-02-14
feat(scheduler): graceful shutdown with SIGTERM

wenzowski:
Aiming for Heroku parity, closes #3086

PR #3093, opened 2015-02-14
feat(*): add pre-commit hook to run flake8

wenzowski:
I angered flake8 in #3091. Why not use standardized pre-commit hooks to run linters?

This enables the hook for `flake8`.

Considering I wrote my first line of golang yesterday, perhaps someone with a few more hours in can patch [hooks.yaml] to run `go vet` or `golint` or `whatever-it-actually-is`?

[hooks.yaml]: https://github.com/pre-commit/pre-commit-hooks/blob/master/hooks.yaml

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,4 @@
+- repo: git://github.com/pre-commit/pre-commit-hooks
+  sha: c4c3c735fd99be1f70b89c816739c0cb0f1e9b53
+  hooks:
+    - id: flake8


wenzowski:
hmm, would `gofmt` be what I'm looking for here?

wenzowski:
oh. `go fmt` `go vet` and `golint` all called by `make test-style`

carmstrong:
I'd like @mboersma's input on this - I think that while generally we want PRs to be checked with `make test-style`, I'm not sure we want to strictly enforce it for **all** commits.

mboersma:
I tried this out but it didn't catch a python error I introduced:
```console
$ make -C controller/ test-style
venv/bin/pip install -q -r requirements.txt -r dev_requirements.txt
  Could not find a tag or branch '7413317', assuming commit.
venv/bin/flake8
./api/urls.py:8:1: F401 'multiprocessing' imported but unused
make: *** [test-style] Error 1
$ git add controller
$ git commit -m "syntax error"
[dialoghq-feature/pre-commit a4e986e] syntax error
 1 file changed, 1 insertion(+)
$
```

Maybe the default `flake8` hook doesn't know to recurse into the client and controller directories? (And does it respect the different `flake8` settings we have in `setup.cfg` for each?) Deis doesn't make it easy since we are a hybrid Go/Python project and need to run different code checkers in different directories.

wenzowski:
the yelp pre-commit hooks are designed for polyglot apps, which is one of the reasons I rely on them so heavily myself

if you voluntarily choose to grab the hookrunner with [`pip install pre-commit`](http://pre-commit.com/#install) then you still have to `pre-commit install` to enable them for the repo you're working on

conversely turning them off for a single commit is a simple `pre-commit uninstall` which works wonders with suckier linters than flake8

it should be picking up the config from the local repo...will have to check on that

if you want this in your workflow as well, then the relevant stuff probably deserves a place in the Makefile, otherwise at worst case this will help me avoid sending you shitty prs by adding a single config file to your repo

the nifty thing about this tool is it only runs the linters on the staged diff–if you have a bunch of horrible code that hasn't been staged, no biggie

mboersma:
If this were a simple matter of adding a file and getting increased code checking, I'd be all for it. But it seems you have to run additional commands to enable it. I think we'd rather keep it simple with `make test-style`.

Thanks though @wenzowski!

PR #3095, opened 2015-02-14
feat(contrib/ec2): Add optional extra WebSocket/TCP ELB

Blystad:
This is my follow up to #2717.

AWS ELB does not currently support WebSockets out of the box with load balancers set in HTTP mode. However, if the ELB is configured in TCP mode WebSockets work, but `deis-router` will be unable to inform the application about the origin request IP and scheme - (`X-Forwarded-For`, `X-Forwarded-Scheme`).

By setting `ExtraTCPELB` to `true` in `cloudformation.json`, two load balancers will be configured: The default one in HTTP mode a extra one in TCP mode allowing WebSocket access over a seperate endpoint.

This is a non-ideal (hopefully temporarily) workaround until AWS proper WebSocket support to ELB.

### Example usage:
Application at `example.com` with WebSocket endpoint at `example.com/ws`

    Configure www.example.com to access DeisWebELB (CNAME)
    Configure  ws.example.com to access DeisTCPELB (CNAME)
    
Add both `www.example.com` and `ws.example.com` as domains to the application.
    
Access websocket endpoint at `ws.example.com/ws`.

--

This PR will require **manual testing** .

Steps to test:
1. In `cloudformation.json` set `ExtraTCPELB` to `true`.
2. Provision
3. You should see a extra ELB being provisioned, and it being added to the AutoScaling group (which is the key to the PR, as the autoscaling group cannot add more load balancers once created). 
4. In `cloudformation.json` set `ExtraTCPELB` to `false` (or nothing)
5. Everything should be as before. No extra load balancer should be created nor added to autoscaling group.

### DIFF
--- a/contrib/ec2/deis.template.json
+++ b/contrib/ec2/deis.template.json
@@ -405,11 +405,15 @@
   },
   "Outputs" : {
     "DNSName" : {
-      "Description" : "DNS Name of the HTTP ELB",
+      "Description" : "(Deprecated) DNS Name of Deis Web ELB",
       "Value" :  { "Fn::GetAtt" : ["DeisWebELB", "DNSName"] }
     },
-    "TCPELBName" : {
-      "Description" : "DNS Name of the TCP ELB",
+    "WebELBDNSName" : {
+        "Description" : "DNS Name of Deis Web ELB",
+        "Value" :  { "Fn::GetAtt" : ["DeisWebELB", "DNSName"] }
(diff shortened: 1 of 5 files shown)


bacongobbler:
Sorry for the late delay on this one. This looks much cleaner. Thanks @Blystad!

carmstrong:
Closing in favor of #3096

PR #3103, opened 2015-02-17
fix(tests): source /etc/environment only if it exists

mboersma:
The proxy settings added in #2825 included a [workaround](https://github.com/deis/deis/blob/master/builder/tests/builder_test.go#L69) for the builder functional tests that does not work for `boot2docker`. The mount either fails or appears as a directory inside the container, which hangs up `/app/bin/entry/` or `/app/bin/boot`.

This changes each `source /etc/environment_proxy` line in builder to be conditional on whether that file exists. This fixes running the functional tests on a Mac, and seems to me more backward-compatible.

### DIFF
--- a/builder/image/bin/boot
+++ b/builder/image/bin/boot
@@ -6,7 +6,9 @@
 # fail hard and fast even on pipelines
 set -eo pipefail
 
-source /etc/environment_proxy
+if [[ -f /etc/environment_proxy ]]; then
+	source /etc/environment_proxy
+fi
 
 # set debug based on envvar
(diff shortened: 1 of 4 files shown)


carmstrong:
Great change. Code LGTM.

gabrtv:
Code LGTM. Wish we had a good way to test the proxy functionality in our end-to-end suite...

bacongobbler:
> Code LGTM. Wish we had a good way to test the proxy functionality in our end-to-end suite...

step 1: acquire a proxy server ;)

PR #3104, opened 2015-02-18
fix(contrib/azure): fix virtual network name

withinboredom:
@carmstrong - probably need more try-except blocks...

Added in virtual-network-name to deployment config as well.

### DIFF
--- a/contrib/azure/azure-coreos-cluster
+++ b/contrib/azure/azure-coreos-cluster
@@ -220,11 +220,14 @@ for i in range(args.num_nodes):
 
     try:
       if i == 0:
-          result = sms.create_virtual_machine_deployment(
+          try:
+            result = sms.create_virtual_machine_deployment(
                       args.cloud_service_name, deployment_name=args.cloud_service_name,
                       deployment_slot='production', label=vm_name,
-                      role_name=vm_name, system_config=system, os_virtual_hard_disk=os_hd,
+                      role_name=vm_name, system_config=system, os_virtual_hard_disk=os_hd, virtual_network_name=args.virtual_network_name,
                       role_size=args.vm_size, network_config=network, data_virtual_hard_disks=data_disks)
+          except Exception, e:
+            print "Caught %s" % str(e)
       else:
           result = sms.add_role(
                       args.cloud_service_name, deployment_name=args.cloud_service_name,
--- a/contrib/azure/azure-coreos-cluster
+++ b/contrib/azure/azure-coreos-cluster
@@ -220,14 +220,11 @@ for i in range(args.num_nodes):
 
     try:
       if i == 0:
-          try:
-            result = sms.create_virtual_machine_deployment(
+          result = sms.create_virtual_machine_deployment(
                       args.cloud_service_name, deployment_name=args.cloud_service_name,
                       deployment_slot='production', label=vm_name,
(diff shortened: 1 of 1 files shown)


deis-admin:

carmstrong:
Thanks @withinboredom. What sort of errors were you seeing?

withinboredom:
Was missing the vnet name parameter, so it was throwing a "virtual network name cannot be null error" Having it output the error made it much easier for me to identify what was going wrong with the deployment instead of it saying "started vm in 0 seconds" with no error shown.

carmstrong:
> Was missing the vnet name parameter, so it was throwing a "virtual network name cannot be null error"

Interesting. Were you adding additional instances to an existing deployment in that case? I've never seen that error before myself.

contrib/azure/azure-coreos-cluster
```diff
@@ -220,11 +220,14 @@ for i in range(args.num_nodes):
 
     try:
       if i == 0:
-          result = sms.create_virtual_machine_deployment(
+          try:
+            result = sms.create_virtual_machine_deployment(
                       args.cloud_service_name, deployment_name=args.cloud_service_name,
                       deployment_slot='production', label=vm_name,
-                      role_name=vm_name, system_config=system, os_virtual_hard_disk=os_hd,
+                      role_name=vm_name, system_config=system, os_virtual_hard_disk=os_hd, virtual_network_name=args.virtual_network_name,
```
withinboredom: Only got the error when subnet-name was set. Had to review the python client code to find this parameter, which interestingly sms.add_role doesn't even have the option. It appears to attach to the same virtual network that the virtual machine deployment was attached to.

withinboredom:
It was initial deployment.

mboersma:
Nitpick: this is really just logging the error rather than error checking of any kind, so the commit subject should probably change.

Also, looking at the enclosing try/except code block, I think all this will actually change is that a `WindowsAzureError` that does not contain the string "Forbidden" will now be logged instead of [disappear silently](https://github.com/withinboredom/deis/blob/oops/contrib/azure/azure-coreos-cluster#L244). Any other errors will cause the Python interpreter to exit the script, where it will print the error type along with a stack trace. 

It seems to me that swallowing some errors with a `pass` statement is problematic, but maybe you just want this extra logging.

carmstrong:
@withinboredom Sorry for the delay in this. Let's remove the error handling code and just fix up the virtual network name. You should also amend your commit to something like: `fix(contrib/azure): fix virtual network name`. We should be able to merge shortly after the change. Thanks!

deis-admin:

withinboredom:
@carmstrong No worries, all done.

carmstrong:
@withinboredom Could you squash and amend the commits? We can't merge until it follows our style guide for commits, like: `fix(contrib/azure): fix virtual network name`

withinboredom:
@carmstrong I can do that, but doesn't that go against gitish culture of not changing history once its been made public?

carmstrong:
> I can do that, but doesn't that go against gitish culture of not changing history once its been made public?

Once it's been merged to master, most definitely. We will frequently ask people to squash and rebase in their own branches until the commits are ready to be merged. 

If it makes you feel better, you could close this PR and open a fresh one from a new branch with a single commit with these changes.

carmstrong:
@withinboredom I did this for you in #3340.

Issue #2995, opened 2015-01-29
Uninstall Platform leaves deis-registry@2.service hanging around


withinboredom:
True story ...

Don't have any logs, just tore down the whole cluster. :/

bacongobbler:
Unfortunately, `deisctl uninstall platform` is not smart enough to figure out how many registries/routers you have available: https://github.com/deis/deis/blob/master/deisctl/cmd/cmd.go#L519

Maybe it'd make sense to scale back down to the default platform instances and then uninstall?

carmstrong:
This be a bug...

https://github.com/deis/deis/blob/0459ccc19e76d3c3bf28891c248776322a544057/deisctl/cmd/cmd.go#L275
and
https://github.com/deis/deis/blob/0459ccc19e76d3c3bf28891c248776322a544057/deisctl/cmd/cmd.go#L519

We should also ensure #2929 accounts for this.

PR #3106, opened 2015-02-18
feat(deisctl): start / stop all installed units

johanneswuerbach:
Start all installed units using `deisctl start router@*`. This works for start, stop and restart.

Fixes https://github.com/deis/deis/issues/2995

### DIFF
--- a/deisctl/Makefile
+++ b/deisctl/Makefile
@@ -32,14 +32,16 @@ install:
 	godep go install -v .
 
 setup-root-gotools:
-	sudo GOPATH=/tmp/tmpGOPATH go get -u -v code.google.com/p/go.tools/cmd/cover
-	sudo GOPATH=/tmp/tmpGOPATH go get -u -v code.google.com/p/go.tools/cmd/vet
+	sudo GOPATH=/tmp/tmpGOPATH go get -u -v golang.org/x/tools/cmd/cover
+	sudo GOPATH=/tmp/tmpGOPATH go get -u -v golang.org/x/tools/cmd/vet
 	sudo rm -rf /tmp/tmpGOPATH
 
 setup-gotools:
-	go get -v github.com/golang/lint/golint
+	go get -u github.com/golang/lint/golint
+	go get -u golang.org/x/tools/cmd/cover
+	go get -u golang.org/x/tools/cmd/vet
 
-test: test-style test-unit test-functional
+test: test-style test-unit
 
 test-style:
 # display output, then check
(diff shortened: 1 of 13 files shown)


johanneswuerbach:
Oh this is about `deisctl stop platform`, I'll update this PR accordingly.

deisctl/Makefile
```diff
@@ -32,14 +32,16 @@ install:
...
-	sudo GOPATH=/tmp/tmpGOPATH go get -u -v code.google.com/p/go.tools/cmd/cover
-	sudo GOPATH=/tmp/tmpGOPATH go get -u -v code.google.com/p/go.tools/cmd/vet
+	sudo GOPATH=/tmp/tmpGOPATH go get -u -v golang.org/x/tools/cmd/cover
+	sudo GOPATH=/tmp/tmpGOPATH go get -u -v golang.org/x/tools/cmd/vet
 	sudo rm -rf /tmp/tmpGOPATH
 
 setup-gotools:
-	go get -v github.com/golang/lint/golint
+	go get -u github.com/golang/lint/golint
+	go get -u golang.org/x/tools/cmd/cover
```
mboersma: It doesn't work for me to install `go cover` and `go vet` this way:
```console
$ go get -u golang.org/x/tools/cmd/cover
go install golang.org/x/tools/cmd/cover: open /usr/local/go/pkg/tool/darwin_amd64/cover: permission denied
```
That's why there is the temp GOPATH workaround in the setup-root-gotools target. Those tools are special and need to go in the root (default) GOPATH where go itself lives: https://groups.google.com/d/msg/golang-nuts/lz0nPiUwfUk/E92u9uZhMHYJ

mboersma:
Could you `git commit --amend` this? It refers to `deis-cli` but should be `deisctl`. :nail_care: 

I want to do manual testing on this, but on my first read it looks good. This is a great improvement, thanks @johanneswuerbach!

deisctl/Makefile
johanneswuerbach: `setup-gotools` works fine for me on OS X 10.9 with go 1.4.2 installed using brew. I guess it's because the brew folder is user-writeable by default.
```
ls -la /usr/local/opt/go
lrwxr-xr-x  1 johannes  admin  18 Feb 18 17:55 /usr/local/opt/go -> ../Cellar/go/1.4.2
```

carmstrong:
You beat me to the punch, @johanneswuerbach! This looks like a better implementation than I was trying :thumbsup:

mboersma:
I tested with many `deisctl` variations on the scalable components: router, registry, and store-gateway. This works well. LGTM.

deisctl/Makefile
carmstrong: @mboersma Are you okay with this? Code LGTM otherwise.

mboersma: It doesn't break any test scripts, so I'm fine with having `go vet` and `go cover` be in both targets.

carmstrong:
Thanks @johanneswuerbach!

Issue #1078, opened 2014-05-31
Add header to indicate the node that processed the request


aledbf:
Right now it's not possible to know which node processed the request (I think that is important to know the node in case of an error or to know how is the load is balanced).

Just add `add_header    X-Deis-Upstream $upstream_addr;` inside `location /` section in the deis-router nginx templates

TrollWarlord:
I would like to point out the HUGE security flaw this also opens. This is revealing a lot more about your internal cluster than a client should ever know. If there is an error, look through physical logs.

TrollWarlord:
https://www.owasp.org/index.php/Fingerprint_Web_Server_%28OTG-INFO-002%29

Allowing server fingerprinting is a HORRIBLE security flaw. STOP IT!

PR #3108, opened 2015-02-18
feat(router): Removed X-Deis-Upstream

johanneswuerbach:
I'm with @TrollWarlord and his comment in https://github.com/deis/deis/issues/1078#issuecomment-74879266, that leaking internal IPs is a bad thing.

The information, which instance handled which request, should be also available via `deis logs`.

### DIFF
--- a/router/image/templates/nginx.conf
+++ b/router/image/templates/nginx.conf
@@ -171,8 +171,6 @@ http {
             }
             {{ end }}
 
-            add_header                  X-Deis-Upstream   $upstream_addr;
-
             proxy_pass                  http://{{ Base $service.Key }};
         }
         {{ else }}


gabrtv:
Agreed.  We should not break the platform abstraction by leaking host information to application consumers.  Instead of adding upstream info, using globally unique request identifiers is a better way to map requests to the host that serviced them -- even though it requires some log analytics infrastructure. 

/cc @aledbf @bacongobbler

aledbf:
First, I understand the reason for this change. (In my case deis routers are not exposed directly to internet)

One detail of this header is that returns the number of upstream servers that processed the request, i.e. if the first node that received the request dies, or takes too long it will jump to the next server in the upstream list and the header will return 2 ip addresses as value.

> using globally unique request

To keep this simple why not use the machine id that fleet returns?

gabrtv:
> To keep this simple why not use the machine id that fleet returns?

That still leaks information about the internal cluster.  

I was thinking something along the lines of: https://github.com/newobj/nginx-x-rid-header which uses [v4 UUID](http://en.wikipedia.org/wiki/Universally_unique_identifier#Version_4_.28random.29).

TrollWarlord:
@aledbf You never want to leak ANYTHING. if you want logs, store them on ceph and dig through them that way.

carmstrong:
I think removing this for now is best - we can discuss a UUID approach if folks find it useful. Code LGTM.

Thanks @johanneswuerbach @TrollWarlord

gabrtv:
Code LGTM as is.  We can add a request UUID in a follow-on PR.

TrollWarlord:
Keep the logs internal on CEPH, use them with the fleet id, I don't really care. Just don't expose it externally.

bacongobbler:
Not to beat a dead horse, but I agree that internal IP addresses should not have been exposed externally. In a public cloud setting, exposing any information about the cluster to the client/application is a bad idea. We did this in https://github.com/deis/deis/commit/a770a2c61cc2119ce6f2896e300e78e21cd0eef1 for the same security concerns that @TrollWarlord brought up.

To follow up, we should also inspect the rest of the application's headers to make sure we're not giving away any other platform information (such as nginx version).

TrollWarlord:
@bacongobbler I think in nginx unless you recompile the source .. it still displays a banner even without the name.

Issue #2299, opened 2014-10-28
Log which keys are missing


johanneswuerbach:
When a component is stuck in a confd reboot loop, it's really difficult to find the missing keys. It would be awesome to have a list (or at least one) of missing keys printed into the journal.

carmstrong:
We bumped confd so this happens now.

Issue #2160, opened 2014-10-17
Use confd watch


johanneswuerbach:
Currently it takes up to 5 seconds until a new app is added to nginx (5 seconds polling). Watch support was recently added to confd https://github.com/kelseyhightower/confd/pull/150 so it's time to add it to deis.

Might fix: https://github.com/deis/deis/issues/1927

mboersma:
We are (almost) completely moved over to [`confd` 0.8](https://github.com/kelseyhightower/confd/releases/tag/v0.8.0) now thanks to the hard work of @bacongobbler. I am retitling this to reference `confd watch` only and assigning it.

bacongobbler:
Thanks! let's get this in for 1.5 :D

PR #3113, opened 2015-02-19
WIP: ref(*): Use confd 0.7

johanneswuerbach:
Upgrade confd from 0.5 to 0.7 and converted all templates to the new syntax. 

This is **not completed yet**, but I opened this already for early feedback and to start a small discussion how to convert some template patterns.

Requirement for https://github.com/deis/deis/issues/2160


controller/templates/confd_settings.py
```diff
@@ -1,6 +1,6 @@
 # security keys and auth tokens
-SECRET_KEY = '{{ .deis_controller_secretKey }}'
-BUILDER_KEY = '{{ .deis_controller_builderKey }}'
+SECRET_KEY = '{{ getv "/deis/controller/secretKey" }}'
+BUILDER_KEY = '{{ getv "/deis/controller/builderKey" }}'
 
 # scheduler settings
 SCHEDULER_MODULE = '{{ or (.deis_controller_schedulerModule) "fleet" }}'
```
johanneswuerbach: I couldn't find a nice way to express default values in confd 0.7

0.5
`{{ or (.deis_controller_schedulerModule) "fleet" }}`
0.7
`{{ if exists "/deis/controller/schedulerModule" }}{{ getv "/deis/controller/schedulerModule" }}{{ else }}fleet{{ end }}`

Deis uses default values a lot, somebody has a better idea?

aledbf:
@johanneswuerbach  :+1:

mboersma:
+1 for updating confd. I don't know the new template idioms yet but I'll try to help out here if I can.

mboersma:
It would be outstanding if this could also resolve #2299 (Log which keys are missing).

johanneswuerbach:
Yep, but it's currently blocked on not having a default value pattern in confd 0.7 https://github.com/kelseyhightower/memkv/pull/4

bacongobbler:
@johanneswuerbach text/template has a native `or` function. I'm one of the maintainers of confd now so I'll see if I can get some good examples going.

http://golang.org/pkg/text/template/#hdr-Functions

bacongobbler:
sadly I couldn't find a better alternative than `if exists/else` in the new confd since the memkv functions all return errors now instead of `<no value>`.

johanneswuerbach:
Yes, I tried to add another one, but it isn't merged yet. https://github.com/kelseyhightower/memkv/pull/4 definitely open for other solutions.

bacongobbler:
We're gonna have to address the router later because confd 0.8 is *practically* impossible to work with for our use case, so we're gonna have to stick with a very old version of confd for that component. As for all the others, I've filed #3360 #3360 #3364 #3365 and #3366 in separate PRs so each component can get the love it deserves.

Thanks for spear-heading this, @johanneswuerbach :)

johanneswuerbach:
Thanks, those PRs are looking great. The router template was also the biggest barrier for me :-D

PR #3114, opened 2015-02-20
docs(contrib): link to community projects

### LINKED ISSUE #2952: docs: Link to community members' custom components, tools, and Deis extensions
We should have a page somewhere in the documentation which point to useful tools folks from our community have developed. Off the top of my head:

* @aledbf's Ceph store-dashboard container: https://github.com/aledbf/deis/tree/optional_store_dashboard
* @ianblenke's various junk drawer tools: https://github.com/ianblenke/coreos-vagrant-kitchen-sink/tree/master/cloud-init

I think @lorieri has some good tools, too.

carmstrong:
closes #2952

### DIFF
--- a/contrib/README.md
+++ b/contrib/README.md
@@ -9,3 +9,20 @@ This does not preclude their usefulness.
 
 Please add any issues you find with this software to the
 [Deis project](https://github.com/deis/deis/issues).
+
+## Community contributions
+
+Various community members have modified Deis components, created new components, or provided tools
+that may be useful to Deis users. While these are not supported by the Deis team, they can be helpful
+in certain scenarios.
+
+Some of these projects include:
+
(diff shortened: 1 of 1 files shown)


mboersma:
Outstanding--it makes me very happy to see these Deis ecosystem projects listed, and I think the contrib README is the right place. LGTM.

gabrtv:
LGTM too.  Over time we might also want to link from the main project README to give this stuff some more exposure.  We don't need to do that now, though.

carmstrong:
> Over time we might also want to link from the main project README to give this stuff some more exposure.

I thought about that, but the main README is so clean with "here's how to get up and running" that I was worried this would be out of place. We'll see if people don't stumble on this enough.

Issue #2158, opened 2014-10-17
fix logspout golint warnings


mboersma:
There is some minor go code cleanup and documentation that should be done in `logspout/`, as pointed out by `golint` [in the tests](http://ci.deis.io/job/test-integration/542/console):
```console
$ golint .
attacher.go:13:6: exported type AttachManager should have comment or be unexported
attacher.go:20:1: exported function NewAttachManager should have comment or be unexported
attacher.go:115:1: exported method AttachManager.Get should have comment or be unexported
attacher.go:121:1: exported method AttachManager.Listen should have comment or be unexported
attacher.go:152:6: exported type LogPump should have comment or be unexported
attacher.go:159:1: exported function NewLogPump should have comment or be unexported
attacher.go:197:1: exported method LogPump.AddListener should have comment or be unexported
attacher.go:203:1: exported method LogPump.RemoveListener should have comment or be unexported
attacher.go:92:10: should omit 2nd value from range; this loop is equivalent to `for ch := range ...`
attacher.go:191:10: should omit 2nd value from range; this loop is equivalent to `for ch := range ...`
logspout.go:43:6: exported type Colorizer should have comment or be unexported
logspout.go:45:1: comment on exported method Colorizer.Get should be of the form "Get ..."
logspout.go:88:9: if block ends with a return statement, so drop this else and outdent its block
routes.go:16:6: exported type RouteStore should have comment or be unexported
routes.go:23:6: exported type RouteManager should have comment or be unexported
routes.go:30:1: exported function NewRouteManager should have comment or be unexported
routes.go:34:1: exported method RouteManager.Load should have comment or be unexported
routes.go:46:1: exported method RouteManager.Get should have comment or be unexported
routes.go:56:1: exported method RouteManager.GetAll should have comment or be unexported
routes.go:66:1: exported method RouteManager.Add should have comment or be unexported
routes.go:94:1: exported method RouteManager.Remove should have comment or be unexported
routes.go:108:6: exported type RouteFileStore should have comment or be unexported
routes.go:110:1: exported method RouteFileStore.Filename should have comment or be unexported
routes.go:114:1: exported method RouteFileStore.Get should have comment or be unexported
routes.go:126:1: exported method RouteFileStore.GetAll should have comment or be unexported
routes.go:144:1: exported method RouteFileStore.Add should have comment or be unexported
routes.go:148:1: exported method RouteFileStore.Remove should have comment or be unexported
routes.go:59:2: can probably use "var routes []*Route" instead
[...]

PR #3118, opened 2015-02-20
ref(*): clean up most code linter warnings

mboersma:
This change cleans up some go language constructs and adds comments. It also makes the `go vet` command in `make test-style` mandatory. (`golint` always returns 0, so it can't easily be made mandatory, and that is not recommended anyway.)

I left a few warnings alone--all those in logspout, plus:
```
backend/fleet/fleet.go:6:6: type name will be used as fleet.FleetClient by other packages, and that stutters; consider calling this Client
syslog/message.go:9:6: type name will be used as syslog.SyslogMessage by other packages, and that stutters; consider calling this Message
```
That code originated elsewhere and I wasn't willing to change the package names without more thought. And logspout has no code comments at all, even upstream. :disappointed: 

Refs #2158, which we can still fix separately if we want to, although we might want to also look at changes in logspout upstream at the same time.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -92,11 +92,8 @@ test-style:
 # display output, then check
 	$(GOFMT) $(GO_PACKAGES)
 	@$(GOFMT) $(GO_PACKAGES) | read; if [ $$? == 0 ]; then echo "gofmt check failed."; exit 1; fi
-# FIXME: make this mandatory
-	-$(GOVET) $(GO_PACKAGES_REPO_PATH)
-# FIXME: make this mandatory
+	$(GOVET) $(GO_PACKAGES_REPO_PATH)
 	@for i in $(addsuffix /...,$(GO_PACKAGES)); do \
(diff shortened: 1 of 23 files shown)


carmstrong:
Code LGTM.

bacongobbler:
I recall some issues with adding `.` and `tests` to the package list when I initially PR'd this. Doesn't seem to be showing up now, though. lgtm

mboersma:
> I recall some issues with adding . and tests to the package list

For logger at least, it was a problem, so I separated a [`GO_TESTABLE_PACKAGES_REPO_PATH`](https://github.com/deis/deis/pull/3118/files#diff-b0fdc981a48d729a02916b0fb9ace04fR9). The other components were ok with adding `tests`--I wanted the test code to be style checked as well.

bacongobbler:
good idea. :+1:

Issue #2917, opened 2015-01-18
deisctl: Print proper error message when getting status of global units


carmstrong:
Some of the Deis components, notably: `store-monitor`, `store-daemon`, `store-metadata`, `store-volume`, `logspout`, and `publisher`are fleet global units. This means that we schedule them once to fleet, and fleet will ensure they run on all hosts in the cluster. 

The output of `deisctl status` on a global unit is confusing: `Unit deis-store-volume.service does not appear to be running.`. This is because the status of global units cannot be queried in this way. We should detect this in deisctl and provide appropriate output so as not to confuse users.

PR #3119, opened 2015-02-20
fix(deisctl) : deisctl Journal and status custom error message for global units

smothiki:
Fixes https://github.com/deis/deis/issues/2917

### DIFF
--- a/deisctl/backend/fleet/journal.go
+++ b/deisctl/backend/fleet/journal.go
@@ -25,6 +25,10 @@ func runJournal(name string) (exit int) {
 		fmt.Fprintf(os.Stderr, "Error retrieving Unit %s: %v", name, err)
 		return 1
 	}
+	if suToGlobal(*u) {
+		fmt.Fprintf(os.Stderr, "Unable to get journal for global unit %s. Check the logs on the host using journalctl.\n", name)
+		return 1
+	}
 	if u == nil {
 		fmt.Fprintf(os.Stderr, "Unit %s does not exist.\n", name)
 		return 1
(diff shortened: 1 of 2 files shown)


mboersma:
@smothiki could you add a newline at the end of each message? `fmt.Printf` doesn't:
```console
$ deisctl journal store-daemon
Unable to get status for global unit deis-store-daemon.service. Check the logs on the host using journalctl.gradishar:deis matt$ 
```
Also could you `git commit --amend` [the subject](http://docs.deis.io/en/latest/contributing/standards/#commit-style) to be `fix(deisctl): add custom error message for global units`?

smothiki:
Sorry , Will change the commit subject and add new line .

mboersma:
No worries, I was just testing and reviewing. Thanks for fixing this!

deisctl/backend/fleet/journal.go
```diff
@@ -25,6 +25,10 @@ func runJournal(name string) (exit int) {
 		fmt.Fprintf(os.Stderr, "Error retrieving Unit %s: %v", name, err)
 		return 1
 	}
+	if suToGlobal(*u) {
+		fmt.Fprintf(os.Stderr, "Unable to get status for global unit %s. Check the logs on the host using journalctl.", name)
```
carmstrong: This should be something like "Unable to get journal for global unit" -- we replaced status in the second sentence, but not the first.

mboersma:
These are much better error messages when dealing with global units through `deisctl`. Code LGTM.

carmstrong:
Thanks for fixing this up, @smothiki. Code LGTM.

Issue #2758, opened 2014-12-16
How to restart an app in deis


wiselyman:
how can I restart an app in deis cluster?

Issue #1929, opened 2014-09-27
Feature: deis restart


johanneswuerbach:
After encountering issues like #1925 or internal app errors an app restart is required.

Today there a multiple workarounds (change the config / push a new release), but having a `deis restart` command would be much nicer.

Issue #467, opened 2014-01-15
Should be able to restart container processes


mboersma:
Heroku has `ps:restart` and Deis needs something similar to bounce databases or other processes that may become stuck.

rjocoleman:
I'm going to take a look at this, how does this interface work for you guys?

```
$ heroku help restart
Alias: restart redirects to ps:restart
Usage: heroku ps:restart [DYNO]

 restart an app dyno

 if DYNO is not specified, restarts all dynos on the app

Examples:

 $ heroku ps:restart web.1
 Restarting web.1 dyno... done

 $ heroku ps:restart web
 Restarting web dyno... done

 $ heroku ps:restart
 Restarting dynos... done
```

carmstrong:
> I'm going to take a look at this, how does this interface work for you guys?

Could you open a proposal like #2911? I think that's the vehicle we're using for planning features with the community moving forward.

rjocoleman:
sure @carmstrong  proposal is here #3110

PR #3110, opened 2015-02-20
Proposal: deis ps:restart

rjocoleman:
This is a proposal for a `deis` cli command, `ps:restart` with the alias `restart`

`deis ps:restart` should restart an application.

I'll investigate the possibility of only restarting specific container types like the similar Heroku command:

The use case for this command is restarting an app in times of instability without using the existing workaround of the `ps:scale` command.

```
$ heroku help restart
Alias: restart redirects to ps:restart
Usage: heroku ps:restart [DYNO]

 restart an app dyno

 if DYNO is not specified, restarts all dynos on the app

Examples:

 $ heroku ps:restart web.1
 Restarting web.1 dyno... done

 $ heroku ps:restart web
 Restarting web dyno... done

 $ heroku ps:restart
 Restarting dynos... done
```
I'll convert this to a PR and attach docs once I've looked at the relevant code and have feedback that this is a desired function + any opinions on the interface.

Refs #467, #1929, #2758

### DIFF
--- a/docs/reference/api-v1.1.rst
+++ b/docs/reference/api-v1.1.rst
@@ -407,6 +407,69 @@ Example Response:
     X_DEIS_PLATFORM_VERSION: 1.3.1
 
 
+Restart Application
+```````````````````
+
+Example Request:
+
+.. code-block:: console
+
+    POST /v1/apps/example-go/restart/ HTTP/1.1
+    Host: deis.example.com
+    Content-Type: application/json
+    Authorization: token abc123
+
+Example Response:
+
+.. code-block:: console
+
+    HTTP/1.1 204 NO CONTENT
+    X_DEIS_API_VERSION: 1.1
+    X_DEIS_PLATFORM_VERSION: 1.3.1
+
+
+Restart Application Containers by Type
+``````````````````````````````````````
+
+Example Request:
+
+.. code-block:: console
+
+    POST /v1/apps/example-go/restart/web HTTP/1.1
+    Host: deis.example.com
+    Content-Type: application/json
+    Authorization: token abc123
+
+Example Response:
+
+.. code-block:: console
+
+    HTTP/1.1 204 NO CONTENT
+    X_DEIS_API_VERSION: 1.1
+    X_DEIS_PLATFORM_VERSION: 1.3.1
+
+
+Restart Specific Container
+``````````````````````````
+
+Example Request:
+
+.. code-block:: console
+
+    POST /v1/apps/example-go/restart/web/1 HTTP/1.1
+    Host: deis.example.com
+    Content-Type: application/json
+    Authorization: token abc123
+
+Example Response:
+
+.. code-block:: console
+
+    HTTP/1.1 204 NO CONTENT
+    X_DEIS_API_VERSION: 1.1
+    X_DEIS_PLATFORM_VERSION: 1.3.1
+
+
 Configuration
 -------------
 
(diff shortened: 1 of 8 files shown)


rjocoleman:
@carmstrong I've added the controller API docs, I'll need to do the client docs last as they're automagic

docs/reference/api-v1.1.rst
```diff
@@ -407,6 +407,69 @@ Example Response:
...
+    POST /v1/apps/example-go/restart/ HTTP/1.1
+    Host: deis.example.com
+    Content-Type: application/json
+    Authorization: token abc123
+
+Example Response:
+
+.. code-block:: console
+
+    HTTP/1.1 204 NO CONTENT
```
bacongobbler: This should be a 200 OK

docs/reference/api-v1.1.rst
```diff
@@ -407,6 +407,69 @@ Example Response:
...
+    POST /v1/apps/example-go/restart/web HTTP/1.1
+    Host: deis.example.com
+    Content-Type: application/json
+    Authorization: token abc123
+
+Example Response:
+
+.. code-block:: console
+
+    HTTP/1.1 204 NO CONTENT
```
bacongobbler: again, 200 OK :)

docs/reference/api-v1.1.rst
```diff
@@ -407,6 +407,69 @@ Example Response:
...
+
+
+Restart Application Containers by Type
+``````````````````````````````````````
+
+Example Request:
+
+.. code-block:: console
+
+    POST /v1/apps/example-go/restart/web HTTP/1.1
```
bacongobbler: I think `web` should be a query type, so it'd look like `/v1/apps/example-go/restart?procType=web`

docs/reference/api-v1.1.rst
```diff
@@ -407,6 +407,69 @@ Example Response:
...
+
+
+Restart Specific Container
+``````````````````````````
+
+Example Request:
+
+.. code-block:: console
+
+    POST /v1/apps/example-go/restart/web/1 HTTP/1.1
```
bacongobbler: Similar to the former: `/v1/apps/example-go/restart?procType=web&num=1`

bacongobbler:
Hmm, Heroku seems to have two endpoints:

```
DELETE /apps/{app_id_or_name}/dynos/{dyno_id_or_name}
DELETE /apps/{app_id_or_name}/dynos
```

Not sure why they decided on DELETE, but I like the idea of the endpoint being `POST /v1/apps/{app_name}/containers/restart` with query parameters to allow more control over the restart strategy (all, by process type or some other process).

johanneswuerbach:
Maybe because you delete one / all dynos and heroku auto-scales (creates dynos) to the previously specified number.

bacongobbler:
right, that's my thinking too. Just looks awkward when looking at it from a pure REST standpoint (DELETE should mean "delete this dyno permanently", not "delete this dyno, which will be re-created").

paralin:
+1 I like this command :)

bacongobbler:
@rjocoleman are you able to get to this for v1.5? If not that's okay; I can get started on it :)

rjocoleman:
@bacongobbler I've pushed my work in progress, it's by no means complete. I got stuck in the API urls section and de-railed.

I'm flat out with a big project launching at the start of next month so I don't have the capacity to look at this for ~1 month sorry

carmstrong:
> I'm flat out with a big project launching at the start of next month so I don't have the capacity to look at this for ~1 month sorry

No worries. We're going to close this for now, as there have been a lot of changes and this will have to be rebased. If you have time to get back to this and have any questions about its implementation, please ping us and we'll be happy to reopen the PR.

Thanks!

PR #3123, opened 2015-02-21
feat(docs): use local 375GB ssds on gce

wenzowski:
This is roughly how I'm deploying to GCE with [SCSI-attached 375GB local SSDs](https://cloud.google.com/compute/docs/disks/#local_ssds)

I'm finding this a minimum usable default for testing, but happy to split this if `gcloud compute` is useful but SSDs are not.

Still a few `gcutil` calls that need to be ported.


wenzowski:
Running an OSD and a monitor or a metadata server on a single disk–irrespective of partitions–is NOT a good idea either.
    <cite>[ceph hardware recommendations]</cite>

This explains the massive performance boost I've seen moving from standard Persistent Disks to locally attached ssd. How would we go about setting things up so journal and data are on separate drives?

[ceph hardware recommendations]: http://ceph.com/docs/master/start/hardware-recommendations/

carmstrong:
> This explains the massive performance boost I've seen moving from standard Persistent Disks to locally attached ssd.

But even in that case, both the OSD and monitor are writing to the same location, which happens to be a Docker volume. So it's faster, but it sounds like we should still split them out to separate disks.

> How would we go about setting things up so journal and data are on separate drives?

That requires a not-insignificant refactor of the store components to write to the host CoreOS filesystem rather than writing to Docker volumes. It's worth doing if we'd see a significant performance improvement, but upgrading clusters would be tricky - we'd have to provide a script to migrate the data from the Docker volume to the host filesystem.

wenzowski:
Indeed, it looks like writing OSD and monitor to the same hard disk results in some level of seek thrashing. Since seek time with the SSD is significantly reduced the performance gain is nontrivial.

I have minimal experience with tuning ceph. Fantastic articles by @leseb on Ceph IO patterns: [the good], [the bad], and [the ugly] indicate that in addition to performance penalty incurred by running [OSD && (monitor || metadata server)] off a single disk (to be addressed in #3124) it seems placement of the OSD journal matters as well.

It seems to that the ideal would be two physical disks per OSD. But it also seems that moving the OSD's journal to the beginning of the hard disk using a raw partition does help as well. This is exceedingly difficult to do on many cloud providers.

Another option here might just be to require a GCE persistent SSD pool instead of an HDD pool. A 375GB persistent SSD pool on GCE is documented to deliver 10^4 IOPS while the ephemeral local ssd documented in this PR is documented to deliver 10^5 IOPS. Following coreos/bugs#238 two local SSDs should be able to provide 10^6 IOPS over NVMe if work is equally balanced.

By comparison, the default 32GB Persistent Disk this PR replaces was delivering 10^1 write and 10^0 read IOPS at $1.29 per provisioned disk. A 64GB Persistent SSD would at least deliver 10^3 IOPS at $10.88 per provisioned disk.

[the good]: http://www.sebastien-han.fr/blog/2014/02/10/ceph-io-patterns-the-good/
[the bad]: http://www.sebastien-han.fr/blog/2014/02/17/ceph-io-patterns-the-bad/
[the ugly]: http://www.sebastien-han.fr/blog/2014/02/24/ceph-io-patterns-the-ugly/

carmstrong:
> Another option here might just be to require a GCE persistent SSD pool instead of an HDD pool.

I think that makes a lot of sense, and would probably alleviate much of the performance wall. We've already switched the EC2 provision script to use SSDs for all volumes, so a default here makes sense as well.

wenzowski:
Ok will shoot for a resubmit of this on Sat with 64GB persistent ssd pool & the rest of the `gcutil` calls.

carmstrong:
> Ok will shoot for a resubmit of this on Sat with 64GB persistent ssd pool & the rest of the gcutil calls.

Sounds great. Thank you!

PR #3125, opened 2015-02-23
docs(contributing): remove reference to `deisctl ssh`

carmstrong:
@mhahn pointed out in IRC that this command doesn't actually exist.

### DIFF
--- a/docs/contributing/hacking.rst
+++ b/docs/contributing/hacking.rst
@@ -202,7 +202,8 @@ Django Shell
 
 .. code-block:: console
 
-    $ deisctl ssh controller   # SSH into the controller
+    $ deisctl list             # determine which host runs the controller
+    $ ssh core@<host>          # SSH into the controller host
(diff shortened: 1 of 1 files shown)


carmstrong:
I killed Jenkins after it passed the docs: https://ci.deis.io/job/test-integration/2162/console

bacongobbler:
Doesn't `deisctl list` show the internal IPs which means that you cannot directly SSH into the nodes unless you're in the cluster?

carmstrong:
> Doesn't deisctl list show the internal IPs which means that you cannot directly SSH into the nodes unless you're in the cluster?

Yes, but I'm assuming they're either on the cluster or can translate the internal IP to map to the proper machine.

bacongobbler:
Okay, sounds good. lgtm :)

mboersma:
Oops, thanks for catching the doc-bug. LGTM.

PR #3126, opened 2015-02-23
fix(docs+controller): update botbot.me and stackoverflow links

mboersma:
The footer links at http://docs.deis.io/en/latest/ and content in two controller web templates didn't match the footer links at http://deis.io/.  cc: @JoshuaSchnell.


bacongobbler:
nice. Just verified on deis.io and these are indeed the correct links. :+1:

carmstrong:
:ship:

PR #3127, opened 2015-02-23
fix(controller): ensure cleanup of "deis run" fleet units

### LINKED ISSUE #3022: Need a way to stop one-off stale commands

mboersma:
If a controller task times out during `deis run`, the cleanup code to destroy its fleet unit isn't called. This wraps the important code in a try/finally to ensure the unit is destroyed even when the code doesn't follow the happy path.

I tested by first setting the [worker timeout](https://github.com/deis/deis/blob/master/controller/deis/gconf.py#L4) to 20, then running `deis run sleep 30`, waiting until it fails and then looking at the output of `fleetctl list-units`. Before this change, a deis run unit was orphaned each time; after the change, the unit is removed.

In conjunction with #3084 (merged), closes #3022.

### DIFF
--- a/controller/scheduler/fleet.py
+++ b/controller/scheduler/fleet.py
@@ -221,92 +221,95 @@ class FleetHTTPClient(object):
             time.sleep(1)
         else:
             raise RuntimeError('container did not report state')
-        machineID = state.get('machineID')
 
-        # find the machine
-        machines = self._get_machines()
-        if not machines:
-            raise RuntimeError('no available hosts to run command')
+        try:
+            machineID = state.get('machineID')
 
-        # find the machine's primaryIP
-        primaryIP = None
-        for m in machines.get('machines', []):
-            if m['id'] == machineID:
-                primaryIP = m['primaryIP']
-        if not primaryIP:
-            raise RuntimeError('could not find host')
+            # find the machine
+            machines = self._get_machines()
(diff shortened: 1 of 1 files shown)


mboersma:
I'm not sure why this diff is so hard to read. All this does is add the `try:` and `finally:` statements and indent their blocks by four spaces.

carmstrong:
> I'm not sure why this diff is so hard to read. All this does is add the try: and finally: statements and indent their blocks by four spaces.

Thanks - I was trying to figure out what changed. Code LGTM.

bacongobbler:
:shipit:

PR #3130, opened 2015-02-23
fix(user-data): ensure nf_conntrack kernel module is loaded

### LINKED ISSUE #3105: Questions on increase-nf_conntrack-connections.service
There's service `increase-nf_conntrack-connections.service` installed by Deis. On new CoreOS installation it fails because there's no conntrack module loaded by default in CoreOS 557. I have some questions/proposals regarding it. Here's what (as it seems to me) should be changed, please let me know if I misunderstand the purpose of this service.

1. Conntrack connections limit should persist after reboot. This can be done by either installing `increase-nf_conntrack-connections.service` or by adding following to `write_files:` section:

```
# /etc/sysctld.d/conntrack.conf
net.netfilter.nf_conntrack_max=262144
```

2. Same for module `nf_conntrack`. It should be either loaded in the service before calling `sysctl` or added as a file:

```
# /etc/modules-load.d/conntrack.conf
nf_conntrack
```

mboersma:
At some point CoreOS stopped loading the `nf_conntrack` kernel module by default, which broke the increase-nf_conntrack-connections.service we added in #2942. I tried to fix it in a more elegant way as @ineu suggested in #3105, but for some reason a /etc/modules-load.d/conntrack.conf file did not cause the module to be loaded on boot, so instead I added a `modprobe` command to the existing one-shot unit.

Before:
```console
core@deis-01 ~ $ sysctl net.netfilter.nf_conntrack_max
sysctl: cannot stat /proc/sys/net/netfilter/nf_conntrack_max: No such file or directory
core@deis-01 ~ $ lsmod | grep conntrack
core@deis-01 ~ $ 
```
After:
```console
core@deis-01 ~ $ sysctl net.netfilter.nf_conntrack_max
net.netfilter.nf_conntrack_max = 262144
```

Closes #3105. ping @aledbf.

### DIFF
--- a/contrib/coreos/user-data.example
+++ b/contrib/coreos/user-data.example
@@ -73,6 +73,7 @@ coreos:
 
       [Service]
       Type=oneshot
+      ExecStartPre=/usr/sbin/modprobe nf_conntrack
       ExecStart=/bin/sh -c "sysctl -w net.netfilter.nf_conntrack_max=262144"
 write_files:
   - path: /etc/deis-release


carmstrong:
Code LGTM.

aledbf:
@mboersma good catch. I just found the same issue with another module.

bacongobbler:
:+1:

ineu:
Nice one, thanks, @mboersma. What do you think about `enable` for this unit?

bacongobbler:
`enable` is not a valid command for units in coreos-cloudinit. It's bitten us a few times: https://github.com/deis/deis/pull/2844

PR #3133, opened 2015-02-23
fix(go): go 1.4 static binaries

### LINKED ISSUE #3131: bug(go): some binaries are not static
```
docker run -it deis/cache:v1.3.1 /bin/bash
root@ce60d047494f:/app# ldd /app/bin/boot
	linux-vdso.so.1 =>  (0x00007fffc7f07000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f792e126000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f792dd60000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f792e348000)
```

workaround: `CGO_ENABLED=0 go build -a -installsuffix cgo`

https://github.com/golang/go/issues/9344
https://github.com/golang/go/issues/9369

aledbf:
closes #3131

### DIFF
--- a/builder/Makefile
+++ b/builder/Makefile
@@ -15,7 +15,7 @@ BINARY_DEST_DIR := image/bin
 
 build: check-docker
 	for i in $(BINARIES); do \
-		GOOS=linux GOARCH=amd64 CGO_ENABLED=0 godep go build -a -v -ldflags '-s' -o $(BINARY_DEST_DIR)/$$i bin/$$i.go || exit 1; \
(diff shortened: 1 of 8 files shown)


carmstrong:
Code LGTM. Thanks for fixing, @aledbf!

aledbf:
Jenkins, test this please.

mboersma:
This does cause some spew when building--all the go standard packages get listed each time. But based on the discussion in golang/go#9344, this is now how we should do static compilation. Code LGTM.

PR #3128, opened 2015-02-23
fix(database): Commence recovery if database is started with stale data container

romansergey:
I popped up with this on IRC on Friday, but to recap, the fail scenario is as follows:

1. start db on node A
2. force reschedule to node B
3. perform some updates
4. force reschedule to node A

Result: db has started successfully on node A with an existing, outdated `deis-database-data` container and is serving old data.

The PR provides a fix for that with a test script.

Note on PR:

* The test takes twice as long to complete compared to `database_test.go`, but it is just as isolated, so I've placed it in `database/tests`.
* I had to patch Mock-S3 in mock-store for wal-e not to frustrate over inability to delete old wal segments. I'm not sure what is the best way to keep it until upstream is updated.

// edit by @carmstrong to add referenced tickets
fixes #3276
fixes #3234

PR #3140, opened 2015-02-24
ref(tests): mock store lazy build with git ref tracking

romansergey:
Emerged from the need in https://github.com/deis/deis/pull/3128 to pick up changes in mock-store and recompile it as well. The way it is currently, it wouldn't recompile as long as at least some `deis/mock-store` image exists.

When `LAZY=1` is specified, it would rebuild `mock-store` only if docker tag corresponding to the current git ref for `tests/fixtures/mock-store` is missing.

### DIFF
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -9,7 +9,17 @@ GO_FILES = apps_test.go auth_test.go builds_test.go config_test.go integration_t
 GO_PACKAGES = dockercli etcdutils mock utils
 GO_PACKAGES_REPO_PATH = $(addprefix $(repo_path)/,$(GO_PACKAGES))
 
-MOCK_STORE_IMAGE = $(IMAGE_PREFIX)mock-store:latest
+MOCK_STORE_IMAGE = $(IMAGE_PREFIX)mock-store
+MOCK_STORE_IMAGE_PATH = fixtures/mock-store
+
+MOCK_STORE_TAG = $(shell git log -n 1 --format="%h" -- $(MOCK_STORE_IMAGE_PATH))
+ifeq (${LAZY},1)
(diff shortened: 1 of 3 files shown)


tests/Makefile
```diff
@@ -35,8 +45,13 @@ setup-root-gotools:
 setup-gotools:
 	go get -v github.com/golang/lint/golint
 
+
```
carmstrong: Are these newlines intentional?

carmstrong:
Code LGTM, but would like @mboersma's review. May need to manually test this.

tests/Makefile
romansergey: @carmstrong, apologies - not intentional. Should I remove them?

romansergey:
@carmstrong, I'm actually having thoughts of withdrawing this one. I think it is better to rely on docker cache with injection of that same GIT_SHA as the first step to void the cache when changes kick in.

carmstrong:
> I think it is better to rely on docker cache with injection of that same GIT_SHA as the first step to void the cache when changes kick in.

:thumbsup: We actually [do this](https://github.com/deis/deis/blob/master/builder/image/Dockerfile#L66) for all of our release images so they are rebuilt when we publish a Deis release. Perhaps we should do this for our ancillary components as well? Anywhere we have the `DEIS_RELEASE

romansergey:
@carmstrong, yes, exactly, I remember I saw that here.
But as far as I understand, for it to take effect in cache voiding it has to precede all instructions that we want to affect. (https://docs.docker.com/articles/dockerfile_best-practices/#build-cache)

romansergey:
replaced with https://github.com/deis/deis/pull/3151

PR #3142, opened 2015-02-25
ref(controller): remove duplicate _start_containers() logic

bacongobbler:


mboersma:
That does (did) look like duplicated code to me. `Showing 1 changed file with 4 additions and 26 deletions` is always nice. Code LGTM.

carmstrong:
Interesting. Definitely looks like duplication. Not sure how that got in there.

Code LGTM.

bacongobbler:
old code cruft, my best guess :)

PR #3143, opened 2015-02-25
feat(cloud-init): check if deisctl does not exists before install it

aledbf:
It makes no sense to download `deisctl` in every boot. It only delays the end of `systemd` boot


bacongobbler:
good change. Code LGTM!

mboersma:
> It makes no sense to download deisctl in every boot

Couldn't agree more. Code LGTM.

PR #3151, opened 2015-02-25
ref(tests): rely on docker build cache in mock-store build

romansergey:
Replaces https://github.com/deis/deis/pull/3140

Also adds *dockerfeed* util (https://github.com/itsafire/dockerfeed) to pack context on the fly.

NOTE: Requires gettext-base (or gettext) installed on building machine.

### DIFF
--- a/database/Makefile
+++ b/database/Makefile
@@ -52,6 +52,6 @@ test-style:
 	@echo no style tests
 
 test-functional:
-	@docker history deis/mock-store >/dev/null 2>&1 || $(MAKE) -C ../tests/ mock-store
+	$(MAKE) -C ../tests/ mock-store
 	@docker history deis/test-etcd >/dev/null 2>&1 || docker pull deis/test-etcd:latest
 	GOPATH=`cd ../tests/ && godep path`:$(GOPATH) go test -v ./tests/...
--- a/registry/Makefile
+++ b/registry/Makefile
@@ -46,7 +46,7 @@ deploy: build dev-release restart
 test: test-style test-unit test-functional
 
 test-functional:
-	@docker history deis/mock-store >/dev/null 2>&1 || $(MAKE) -C ../tests/ mock-store
+	$(MAKE) -C ../tests/ mock-store
 	@docker history deis/test-etcd >/dev/null 2>&1 || docker pull deis/test-etcd:latest
 	GOPATH=`cd ../tests/ && godep path`:$(GOPATH) go test -v ./tests/...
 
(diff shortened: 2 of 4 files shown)


romansergey:
@mboersma, would it be possible to get `gettext-core` package installed on jenkins machines?

mboersma:
@romansergey I will install the `gettext` package on the Ubuntu machines and `brew install gettext` on the Mac, but I'm hesitant to introduce requirements like that and `dockerfeed` just to use the Makefile. I think it should stay simple and portable.

`mock-store` is supposed to be (relatively) unchanging, like `deis/test-etcd` and `deis/test-postgresql`. We want them to be cached to speed things up, but obviously that requires out-of-band actions when the images are actually updated, and makes development on mock-store itself tedious. But so far it's been manageable, and we will have to continue doing so for test-etcd and test-postgresql.

mboersma:
It looks like the `gettext` package for Mac OS X Homebrew doesn't include `envsubst`.

romansergey:
@mboersma I agree that that's an overkill for a thing as small as mock store. But I see dockerfeed+envsubst being very useful to DRY many other things - including release packing.

And you find this PoC fine, I can amend similar changes for test-etcd and test-postgresql.

Instead of envsubst we could use some other template engine that wouldn't put a strain on controlling which env vars get substituted and which vars are left for docker build itself.

Option `#2` to guarantee consistency between build/test executions is to add practice of adding and substituting some cache binding instruction/comment in Dockerfile, for time being, until docker has an option to bind cache to context checksum. Along the lines of:

```
ENV DOCKER_CACHE_BIND_KEY 63f0e04e-52e8-4bd7-a193-4e1e74e1de1b
```

mboersma:
This PR points out a valid issue--when changes are made to the mock-store component, the functional tests won't notice that. It requires an out-of-band action to run `make -C tests/ mock-store` before running the tests in this case, which is easy to forget.

But I'm not comfortable adding any external requirements to the `Makefile`, frankly. The test containers--deis/test-etcd, deis/test-postgresql, and deis/mock-store--are considered "external" test artifacts so that we can test components outside of their normal context in Deis. They change rarely, and when they do it's not that painful to update them on my box and the several test nodes.

So this is a lingering issue for those of us who run the tests, but it almost never happens, and personally I would rather leave it as-is rather than complicate the Makefiles at all. Maybe some note added to the testing docs would be sufficient?

deis-admin:

PR #3156, opened 2015-02-27
feat(publisher): Check if app port is open before publishing it. Fixes 3115

### LINKED ISSUE #3115: Deis should check if container's port is accepting connections before adding it to the pool
Sometimes you need to run commands to setup your application before you run the web server. E.g. you may want to run `./manage.py syncdb` in a django app to apply database migrations before you run  `gunicorn`.

Deis should check if the exposed port of the container accepts connections, before adding the container to the pool.  Otherwise the container gets into the pool, router sends traffic to it and nginx returns 502 (bad gateway) because no server listens to the exposed port of a second or two.

Is this something that can go into [IsPublishableApp](https://github.com/deis/deis/blob/master/publisher/server/publisher.go#L104)?

(Apologies if this has being already reported. I've this memory but I cannot find the issue)

glogiotatidis:
fixes #3115

### DIFF
--- a/publisher/server/publisher.go
+++ b/publisher/server/publisher.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"net"
 	"os"
 	"regexp"
 	"strconv"
@@ -91,8 +92,9 @@ func (s *Server) publishContainer(container *docker.APIContainers, ttl time.Dura
 		keyPath := fmt.Sprintf("/deis/services/%s/%s", appName, containerName)
 		for _, p := range container.Ports {
 			port := strconv.Itoa(int(p.PublicPort))
-			if s.IsPublishableApp(containerName) {
-				s.setEtcd(keyPath, host+":"+port, uint64(ttl.Seconds()))
+			hostAndPort := host + ":" + port
+			if s.IsPublishableApp(containerName) && s.IsPortOpen(hostAndPort) {
+				s.setEtcd(keyPath, hostAndPort, uint64(ttl.Seconds()))
 			}
 			// TODO: support multiple exposed ports
 			break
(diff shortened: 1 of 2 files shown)


aledbf:
@glogiotatidis why use ExecStartPost like in deis-builder?. 
With that it will appear as start-post in `fleetctl list-units` and not as started.
https://github.com/deis/deis/blob/master/deisctl/units/deis-builder.service#L12

publisher/server/publisher.go
```diff
@@ -113,12 +115,29 @@ func (s *Server) IsPublishableApp(name string) bool {
 		log.Println(err)
 		return false
 	}
-	if version >= latestRunningVersion(s.EtcdClient, appName) {
+
+	if version >= latestRunningVersion(s.EtcdClient, appName) && isPortOpen(hostAndPort) {
```
bacongobbler: let's separate this call out. IsPublishableApp determines if the container name follows a regex, where IsPortOpen would be useful on its own.

publisher/server/publisher.go
```diff
@@ -113,12 +115,29 @@ func (s *Server) IsPublishableApp(name string) bool {
...
 
+func isPortOpen(hostAndPort string) bool {
+	portOpen := false
+	if hostAndPort == "localhost:-1" {
+		portOpen = true
+	} else if hostAndPort == "localhost:-2" {
+		portOpen = false
+	} else {
+		conn, err := net.Dial("tcp", hostAndPort)
+		if err != nil {
```
carmstrong: Shouldn't this be reversed? If we *don't* get an error (i.e. `err == nil`), that means the port is open.

glogiotatidis: Correct. I re-wrote the tests following the paradigm of https://golang.org/src/net/dial_test.go  to avoid such errors.

publisher/server/publisher.go
glogiotatidis: Updated.

glogiotatidis:
@aledbf I'm not aware of this fleet functionality. I'd prefer thought that this goes to publisher so that Deis gets to have the final call and it's fleetd independent.

mboersma:
https://get.docker.io is currently down, so this may fail until we can merge #3162 and rebase it off that.

glogiotatidis:
Rebased

mboersma:
Having publisher check that the app port is open before publishing it seems like a good idea. Code LGTM.

bacongobbler:
ooh, tests and everything! LGTM!

jgmize:
nice work, @glogiotatidis

lorieri:
Great!

PR #3158, opened 2015-02-27
Docs: Missing note for Vagrant users

### LINKED ISSUE #2528: docs: SSH keys when using vagrant
Working through the vagrant install, I ran into an issue understanding which SSH keys needed to be added.

The Vagrant instructions ask you to generate an SSH key, but then have you using the vagrant insecure key. When you move into "Installing the Deis Platform" it has you set the private key for the servers to talk to one another and references "~/.ssh/deis" which is the key you create in the Vagrant instructions.

This is confusing since you should be using the vagrant insecure key at this point.

I'd suggest modifying the Vagrant install instructions to not have the user generate a deis key and explain that they will be using the Vagrant insecure key locally.

phspagiari:
Vagrant Users by default have to set the sshPrivateKey to .vagrant.d/insecure_private_key instead of the deis key. Even the make dev-cluster does this:

```shell
dev-cluster: discovery-url
        vagrant up
        ssh-add ~/.vagrant.d/insecure_private_key
        deisctl config platform set sshPrivateKey=$(HOME)/.vagrant.d/insecure_private_key
        deisctl config platform set domain=local3.deisapp.com
        deisctl install platform
```
In all my installations and tests I always use this key instead the deis key, since the key is not used anywhere when you're in vagrant mode.

closes #2528 -- *edited by @bacongobbler*

### DIFF
--- a/docs/installing_deis/vagrant.rst
+++ b/docs/installing_deis/vagrant.rst
@@ -31,16 +31,18 @@ The ``Vagrantfile`` requires the plugin `vagrant-triggers`_. To install the plug
     website.
 
 
-Generate SSH Key
-----------------
-
-.. include:: ../_includes/_generate-ssh-key.rst
-
-
 Generate a New Discovery URL
 ----------------------------
 
 .. include:: ../_includes/_generate-discovery-url.rst
+
+
+Generate SSH Key
+----------------
+
+.. note::
+
+    For Vagrant clusters you don't need to create a key pair, instead use the insecure_private_key located in ``~/.vagrant.d/insecure_private_key``.
 
 
 Boot CoreOS
--- a/docs/installing_deis/install-platform.rst
+++ b/docs/installing_deis/install-platform.rst
@@ -58,6 +58,10 @@ hosts during ``deis run``:
 
     $ deisctl config platform set sshPrivateKey=~/.ssh/deis
 
+.. note::
+
+    For Vagrant clusters: ``deisctl config platform set sshPrivateKey=$(HOME)/.vagrant.d/insecure_private_key``
+
 We'll also need to tell the controller which domain name we are deploying applications under:
 
 .. code-block:: console


deis-admin:

bacongobbler:
I thought we mentioned this in the vagrant installation docs but I don't see one. LTGM!

carmstrong:
Perfect. Thanks @phspagiari! LGTM.

carmstrong:
I think this actually closes #2528

carmstrong:
> I think this actually closes #2528

Or maybe not. We should probably remove "Generate SSH Key" from the Vagrant docs, and instead show the user the insecure_vagrant_key. @phspagiari You up for that change?

phspagiari:
Done @carmstrong! Instead of only removing I left a note for vagrant users.

phspagiari:
And I imagine that this closes #2528

bacongobbler:
Indeed. I edited your OP so that it'll automagically close that issue once we merge :)

docs/installing_deis/vagrant.rst
```diff
@@ -31,18 +31,20 @@ The ``Vagrantfile`` requires the plugin `vagrant-triggers`_. To install the plug
...
-.. include:: ../_includes/_generate-ssh-key.rst
-
-
 Generate a New Discovery URL
 ----------------------------
 
 .. include:: ../_includes/_generate-discovery-url.rst
 
 
+SSH Key
```
bacongobbler: I still think this should be "Generate SSH Key" to be consistent with the other installation docs. Either that or add some actionable verb before "SSH Key" so people know what the paragraph is going to talk about -- "SSH Key" is too ambiguous.

phspagiari: On second thought "Generate" with the note will be better than only "SSH Key" or with another word. It will be equal than any other doc and the ``note`` alerts that it's not a "Key generation".

phspagiari:
Done, I did a squash of the last 3 commits.

mboersma:
Just to note, the `insecure_private_key` is essentially deprecated in vagrant now. (Because...wait for it...it's insecure.) We continue to use it thanks to [this code](https://github.com/deis/deis/blob/master/Vagrantfile#L53-L54), because generating a key just-in-time and using it for the rest of our workflow is a big pain, but it's worth noting this isn't the default now for vagrant.

Code LGTM.

carmstrong:
Thanks @phspagiari!

PR #3162, opened 2015-02-28
fix(builder): install docker from get.docker.com

mboersma:
The endpoint https://get.docker.io is serving up a hearty breakfast of 504s today. The Docker [install docs](https://docs.docker.com/installation/ubuntulinux/) reference https://get.docker.com instead. This change points builder (and our CI node setup script) to what is apparently the preferred endpoint.


carmstrong:
:ship:

bacongobbler:
:shipit:

PR #3164, opened 2015-03-01
ref(Makefile): break build stage up into build and compile

Joshua-Anderson:
This can be really useful if you need to build the go components using nonstandard methods.

For example:

I use gvm, a go version switcher. It manages go and `$GOPATH` for me locally. This make it impossible to run `make build` because if I run with root I don't have go in my `$PATH` and for security reasons you can't pass it, and if I run without root than I can't create docker images.

Now, I can run `make compile` to build the go components and then `sudo make build` to create the docker images.

This can also make cross compiling go easier as well. You can compile go however you want and then run `make build` to create the docker image.

Since build go binaries and building a docker image are two fairly separate steps anyway, I think it makes a lot of sense to separate them.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,7 @@ repo_path = github.com/deis/deis
 GO_PACKAGES = pkg/time version
 GO_PACKAGES_REPO_PATH = $(addprefix $(repo_path)/,$(GO_PACKAGES))
 
+COMPILED_COMPONENTS=builder logger logspout publisher
 COMPONENTS=builder cache controller database logger logspout publisher registry router store
 START_ORDER=publisher store logger logspout database cache registry controller builder router
 CLIENTS=client deisctl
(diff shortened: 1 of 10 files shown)


carmstrong:
Code LGTM.

mboersma:
This changes the base workflow so that `make build` isn't self-sufficient now. That means that test-integration.sh, which only calls `make build`, won't work.

Joshua-Anderson:
Thanks @mboersma:

I updated test acceptance, test integration, and test smoke.

If you would prefer, I could also separate building docker images into its own stage and then update build to run both of these stages, keeping the behavior of build the same.

Joshua-Anderson:
Any feedback to get this or a similar fix merged? It would make it ***much*** easier for me to build deis. :wink:

bacongobbler:
We're trying to resurrect Jenkins from the dead. Once he's back to life then we can start running it against this PR.

bacongobbler:
Also you can add yourself to the dockers group via `user mod -aG docker user`, nullifying this PR. I'd personally prefer docs for that as its a device environment.

Joshua-Anderson:
Cool tip about the docker group, that works well. I still think it makes sense to separate the two stages, but have no super need for it.

mboersma:
> you can add yourself to the docker group via `usermod -aG docker user`

That's the solution I use on the CI nodes and for my own development. I really prefer that solution rather than separating `make build` into `make compile` and `make build`. I'm going to close this PR @Joshua-Anderson since it sounds like being in the docker group is sufficient to get things built. Sorry to let this dangle for so long @Joshua-Anderson--let me know if you're not ok with `usermod` as a workaround.

Joshua-Anderson:
It is OK with me.

PR #3167, opened 2015-03-02
Fix URL to official docs for azure installation

gvilarino:
Tiny link fix, might be useful for newcomers


carmstrong:
I wonder when all of our links got switched to requiring .html... this seems to be a recent change. @mboersma are you aware of any changes there?

mboersma:
I think it must be an issue at our ReadTheDocs.org host. All the existing links work fine when you build the docs locally.

mboersma:
This looks like it was a (hopefully transient) issue with our docs host. I rebuilt the docs in their admin panel (which isn't usually necessary--doc builds are triggered by changes at GitHub), and the link http://docs.deis.io/en/latest/installing_deis/azure/ seems to work again.

bacongobbler:
yeah same issue reported here deis/deis.io#76

PR #3171, opened 2015-03-02
Add .git to builds

ianblenke:
feat(scripts/test-cluster): Add .git to builds

The current deis git push behavior strips .git from the build. This happens during the "git archive" step in the deis-builder template/builder script. In order to use submodules, the "git archive" method of sending the tarball to "docker build" needs to be replaced. As the repo is bare, it needs to be cloned to the TMP_DIR and checked out to the $GIT_SHA of the deployment.


ianblenke:
My bad, wrong bsae.

Issue #1094, opened 2014-06-02
git submodules hook/support ala Heroku


azurewraith:
I attempted to reproduce some Heroku'esque workflow involving a vendored Redmine plugin and found this:

https://devcenter.heroku.com/articles/git-submodules

90% sure it probably wouldn't have worked for me because git submodules are a PITA but it seems like Heroku handles this in their post-receive hook before delegating to the buildpack...

ianblenke:
This is duplicated in deis#3160 - closing that in favor of this on @bacongobbler's suggestion.

We have a couple of projects that use submodules. The precise SHA1 is important to us.

deis-builder uses this in `./builder/image/templates/builder' to generate the tarball that is sent to the docker slugbuilder:

    git archive $GIT_SHA | tar -xmC $TMP_DIR

Unfortunately, git archive strips the `.git` directory, and we need that for our submodules.

@bacongobbler mentioned [gh1094](https://github.com/deis/deis/issues/1094) as related to this, and rightfully explained that submodules are also an issue on Heroku.

To add .git, I've had to modify our deis/builder image to do this instead:

    git ls-files --cached --full-name --no-empty-directory -z $GIT_SHA | tar cf - -T - --null |  tar -xmC $TMP_DIR

It would be nice to somehow flag an app to selectively include .git, as no doubt others will run into this limitation.

Thanks!

paralin:
+1 just ran into this issue.

bacongobbler:
closed but relevant: https://github.com/deis/deis/pull/3375

bacongobbler:
Just an FYI by bburky on IRC:

```
16:05:00 <bburky> Hey, can I ask you someting about Deis?
16:09:05 <bacongobbler> sure
16:09:14 <bacongobbler> what's up? :)
16:10:08 <bburky> Something that I'm fairly certian isn't actually a security issue, but some stuff I wanted to make sure about
16:10:26 <bburky> For starters, deis is currently NOT multi tenant at all right?
16:10:52 <bacongobbler> correct. We explicitly say that in the docs that it's not intended to be deployed in a multi-tenant environment
16:11:12 <bburky> Cool
16:11:35 <bburky> And also how isolated/unprivledged is slugbuilder?
16:12:07 <bburky> I'm asking because I discovered CVE-2015-7545 awhile ago. Cloning git submodules could run arbitrary code
[...]

PR #3172, opened 2015-03-02
Add .git to builds

ianblenke:
feat(deis-builder/add_dot_git): Add .git to builds

The current deis git push behavior strips .git from the build. This happens during the "git archive" step in the deis-builder template/builder script. In order to use submodules, the "git archive" method of sending the tarball to "docker build" needs to be replaced. As the repo is bare, it needs to be cloned to the TMP_DIR and checked out to the $GIT_SHA of the deployment. This also allows for the use of submodules in deployments.

### DIFF
--- a/builder/image/templates/builder
+++ b/builder/image/templates/builder
@@ -56,7 +56,14 @@ mkdir -p $BUILD_DIR $CACHE_DIR
 TMP_DIR=$(mktemp -d --tmpdir=$BUILD_DIR)
 
 cd $REPO_DIR
-git archive $GIT_SHA | tar -xmC $TMP_DIR
+git clone $REPO_DIR/ "${TMP_DIR}.raw/"
+pushd "${TMP_DIR}.raw"
+export GIT_DIR="${TMP_DIR}.raw/.git" GIT_WORK_TREE="${TMP_DIR}.raw"
+git config core.bare false
+git checkout $(git branch -a | tail -1)
+( echo .git; git ls-tree --full-name -z -r --name-only  $GIT_SHA ) | tar cf - -T - --null | tar -xmC $TMP_DIR
+rm -fr "${TMP_DIR}.raw"
+popd
 
 # switch to app context
 cd $TMP_DIR
--- a/builder/image/templates/builder
+++ b/builder/image/templates/builder
@@ -61,7 +61,6 @@ pushd "$TMP_DIR"
 export GIT_DIR="$TMP_DIR/.git" GIT_WORK_TREE="$TMP_DIR"
 git config core.bare false
 git reset --hard $GIT_SHA
-( echo .git; git ls-tree --full-name -z -r --name-only $GIT_SHA ) | tar cf - -T - --null | tar -xmC "$TMP_DIR"
 popd
 
 # switch to app context
--- a/builder/image/templates/builder
+++ b/builder/image/templates/builder
@@ -57,14 +57,12 @@ TMP_DIR=$(mktemp -d --tmpdir=$BUILD_DIR)
 
 cd $REPO_DIR
 git clone $REPO_DIR/ "$TMP_DIR/"
-pushd "$TMP_DIR"
-export GIT_DIR="$TMP_DIR/.git" GIT_WORK_TREE="$TMP_DIR"
-git config core.bare false
-git reset --hard $GIT_SHA
-popd
 
 # switch to app context
 cd $TMP_DIR
+export GIT_DIR="$TMP_DIR/.git" GIT_WORK_TREE="$TMP_DIR"
+git config core.bare false
+git reset --hard $GIT_SHA
 
 USING_DOCKERFILE=false
 
--- a/builder/image/templates/builder
+++ b/builder/image/templates/builder
@@ -56,13 +56,12 @@ mkdir -p $BUILD_DIR $CACHE_DIR
 TMP_DIR=$(mktemp -d --tmpdir=$BUILD_DIR)
 
 cd $REPO_DIR
-git clone $REPO_DIR/ "${TMP_DIR}.raw/"
-pushd "${TMP_DIR}.raw"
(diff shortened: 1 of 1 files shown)


builder/image/templates/builder
```diff
@@ -56,7 +56,12 @@ mkdir -p $BUILD_DIR $CACHE_DIR
 TMP_DIR=$(mktemp -d --tmpdir=$BUILD_DIR)
 
 cd $REPO_DIR
-git archive $GIT_SHA | tar -xmC $TMP_DIR
+git clone $REPO_DIR/ "$TMP_DIR/"
+pushd "$TMP_DIR"
```
bacongobbler: on line 67 we `cd` to $TMP_DIR. It'd be cleaner if we moved these operations below that line and removed the call to `cd $REPO_DIR` on line 58.

builder/image/templates/builder
```diff
@@ -56,7 +56,12 @@ mkdir -p $BUILD_DIR $CACHE_DIR
...
+export GIT_DIR="$TMP_DIR/.git" GIT_WORK_TREE="$TMP_DIR"
+git config core.bare false
+git reset --hard $GIT_SHA
```
bacongobbler: why not a `git checkout` instead? That way we're not destroying the tree's history

ianblenke:
Thankfully, in this context, we aren't destroying anything.

The TMP_DIR is actually cloned from a bare git repo, it is empty except for the .git directory that was copied over.

That TMP_DIR is indeed temporary (see the mktemp a few lines up). As such, there is no history there to be lost.

The big problem here is that you can't simply do a "git checkout $GIT_SHA" as it will blow up telling you that the .git/HEAD mentions "refs/origin/master" and that doesn't exist if you didn't git push a master branch yet. So, what you end up doing, is doing a: "git checkout $(git branch -a)" followed by a "git checkout $GIT_SHA", all to end up with effectively the same thing you would have had by doing a "git reset --hard $GIT_SHA" in the first place. One way you would be in a local branch named after whatever branch they pushed checked out to the specific SHA1, the other way you end up in a disassociated branch checked out to the specific SHA1. The net effect is ambiguous after build time anyway, it's really a question of what the slug runtime sees when it tries to use that .git tree.

bacongobbler:
After a bit of reading between the two, essentially they do the same thing for this operation so I guess it doesn't really matter either way.

carmstrong:
@ianblenke would this close #1094?

carmstrong:
Code LGTM.

mboersma:
Code LGTM. We should test this against a project with submodules.

ianblenke:
Aye, this would potentially close #1094, with the caveat that you still must manage your ssh keys with your own buildpack (which we might build an example of) or in a Dockerfile as a RUN step. This fix merely makes .git available for use by a buildpack/Dockerfile.

In some cases, it might actually impair customers who expect a smaller resultant layer due to the fact that their .git has been ignored until now.

It might not be a bad idea to make this behavior conditional somehow to avoid that.

Principle of least surprise though: which is more surprising, finding that .git isn't included when doing a deploy, or finding that your smallish images are suddenly larger thanks to .git being included in a newer version of Deis?

bacongobbler:
IIRC Heroku strips this directory from the final slug. It might be better to clone the repository, initialize submodules then strip .git so the behaviour stays the same.

ianblenke:
The question then becomes when to initialize the submodules. If we move that functionality into the deis build script, we have to take into account private credentials for private git repos. If we leave it up to the buildpack/Dockerfile, do we leave it up to those mechanisms to make that decision or whether or not to strip the .git folder? While we could conceivably strip the .git folder after building a slug using a buildpack, we can't really remove the space overhead in a layer generated by an app's Dockerfile - unless we just want to flatten all such app Dockerfile generates images into a slug-like form before deploying to further help _possibly_ minimizing deploys (but not sharing anything between deploys as a result)...

bacongobbler:
hmm, I'm wondering if we should close this in favour of #3259 since that seems to cover all bases proposed here (private buildpacks, SSH keys, **and** heroku submodules).

ianblenke:
@bacongobbler, I ran into problems attempting to use a shallow git clone with submodules, so I don't believe we can use the `--depth 1` option as in #3259.

bacongobbler:
> I ran into problems attempting to use a shallow git clone with submodules, so I don't believe we can use the --depth 1 option as in #3259

I don't see a `--depth=1` being used in #3259..?

ianblenke:
Unfortunately #3305 has to do with ssh keys. So does #3259. While that does help access to private repos, and is an important thing to consider as it is tangentially related to this issue, neither really addresses the submodules _versioning_ problem.

According to this StackOverflow post, the commit sha1 of the submodule "[is stored in Git's object database directly. The tree object for the directory where the submodule lives will have an entry for the submodule's commit (this is the so-called "gitlink")](http://stackoverflow.com/questions/5033441/where-does-git-store-the-sha1-of-the-commit-for-a-submodule)".

I just tried to create a pair of local git repos and have one file:// reference the other as a submodule, and I could shallow clone that with `--depth=1` and it would indeed retain the submodule sha1, even after having added a few commits. The submodule sha1 metadata is being retained even in a shallow clone.

The point of this PR is really to make sure we clone the entire tree, including the .git folder. Without .git, there is no way to access this sha1 version metadata, and that is the larger problem that the ssh related issues don't address.

withinboredom:
@ianblenke If you are looking for a git repo with lots of submodules and commits, see https://github.com/FuturesJS/FuturesJS - it isn't an app, but you may find it useful to test with what you're working on.

deis-admin:

bacongobbler:
@ianblenke after our PR review this morning, we evaluated this and felt like this is a very special use case. This is a perfect use case for customizing the builder, but considering that the SSH_KEY PR should cover most of problems towards initializing submodules we're going to close this as a wontfix.

Issue #3135, opened 2015-02-24
Proposal: LDAP/AD auth on controller


phspagiari:
As we discussed at #1578, in order to use deis at large companies we need a way to use LDAP authentications on deis. 

I would be happy if I start to contribute with deis using this idea, indeed I already tested some ideas, but it will be very good if some of the core developers help me with the design of the idea making easy to use and administer.

So far I learned about how integrate ldap with django and already tested with a simple authentication app in django, so far so good.
Now I want to implement this on the controller and extend the configuration to etcd + deis client since the ldap configs are located in settings.py of the controller.

My idea is something like this:

 - django-auth-ldap, ldap and python-etcd as requirements of deis-controller.
 - Add LDAP settings into settings.py of controller and getting all values of attibutes from etcd.
 - Activating the LDAP with something like this:

```shell
deisctl config ldap set host=ldap.domain.com
deisctl config ldap set dn="dc=ldap,dc=domain"
deisctl config ldap set user-search="ou=Users,dc=my"
deisctl config ldap set group-search="ou=Groups"
```
or

```shell
deisctl config platform set ldap="ldap://ldap.domain.com:ou=Groups,ou=Users,dc=my,dc=ldap,dc=domain"
```

- The user will make a register with the user/password of LDAP and the register will make a verification if ldap is enabled and instead of creating a new user he will make a copy (avoiding that every time when the user makes login the controller go to de LDAP to authenticate the user).
- The user will be replicated in deis users database.
- The user will make a normal login, receive the token (now with a expiration time)

I really dont know if this is a good implementation, so lets discuss about this before we start to implement.

bacongobbler:
IMO I think that external auth libraries are great, but we should probably be laying out the groundwork first. In order to allow community members to plug in their auth solutions like LDAP, we should think of how we can break out the authentication system and make it extensible. Right now the auth system is anything *but* extensible, which means that LDAP support would be largely untested.

phspagiari:
I understand and agree with you @bacongobbler. So we need to "decouple" the actual auth model in a new module and make it extensible for any new auth solutions like Oauth, LDAP and more.

Actually the auth is under api.views using the api.serializers.UserSerializer. The idea is modifying all to a new module like api.auth and make this extensible for new auths ( class LdapAuth(api.auth) for ex. ) right? We cant forget that we are using django and he have your own user model and all new auth models have to use his User model.

mvanholsteijn:
[...]

PR #3174, opened 2015-03-03
feat(controller): Adding LDAP/AD auth support

phspagiari:
This is the Pull Request for the Proposal that we discussed at #3135 using the [PoC in the gist](https://gist.github.com/phspagiari/76165b35bae1b8e5a891).

I didn't write any test because I imagine that it's not good have specific tests for auth with ldap, so my opinion is that if the actual auth tests pass its good enough.

If sounds good I will make a doc with the README content and create another PR.

### DIFF
--- a/controller/deis/settings.py
+++ b/controller/deis/settings.py
@@ -8,6 +8,10 @@ import random
 import string
 import sys
 import tempfile
+import ldap
+
+from django_auth_ldap.config import LDAPSearch, GroupOfNamesType
+
 
 PROJECT_ROOT = os.path.normpath(os.path.join(os.path.dirname(__file__), '..'))
 
@@ -138,6 +142,7 @@ INSTALLED_APPS = (
     'django.contrib.sites',
     'django.contrib.staticfiles',
     # Third-party apps
+    'django_auth_ldap',
     'guardian',
     'json_field',
     'gunicorn',
@@ -151,6 +156,7 @@ INSTALLED_APPS = (
 )
 
 AUTHENTICATION_BACKENDS = (
+    "django_auth_ldap.backend.LDAPBackend",
     "django.contrib.auth.backends.ModelBackend",
     "guardian.backends.ObjectPermissionBackend",
 )
@@ -324,6 +330,16 @@ SECURE_PROXY_SSL_HEADER = ('HTTP_X_FORWARDED_PROTO', 'https')
 #  server       - Hostname based on CoreOS server hostname
 UNIT_HOSTNAME = 'default'
 
+# LDAP DEFAULT SETTINGS (Overrided by confd later)
+LDAP_ENDPOINT = ""
+BIND_DN = ""
+BIND_PASSWORD = ""
+USER_BASEDN = ""
+USER_FILTER = ""
+GROUP_BASEDN = ""
+GROUP_FILTER = ""
+GROUP_TYPE = ""
+
 # Create a file named "local_settings.py" to contain sensitive settings data
 # such as database configuration, admin email, or passwords and keys. It
 # should also be used for any settings which differ between development
@@ -334,9 +350,41 @@ try:
 except ImportError:
     pass
 
-
 # have confd_settings within container execution override all others
 # including local_settings (which may end up in the container)
 if os.path.exists('/templates/confd_settings.py'):
     sys.path.append('/templates')
     from confd_settings import *  # noqa
+
+# LDAP Backend Configuration
+# Should be always after the confd_settings import.
+LDAP_USER_SEARCH = LDAPSearch(
+    base_dn=USER_BASEDN,
+    scope=ldap.SCOPE_SUBTREE,
+    filterstr="(%s=%%(user)s)" % USER_FILTER
+)
+LDAP_GROUP_SEARCH = LDAPSearch(
+    base_dn=GROUP_BASEDN,
+    scope=ldap.SCOPE_SUBTREE,
+    filterstr="(%s=%s)" % (GROUP_FILTER, GROUP_TYPE)
+)
+AUTH_LDAP_SERVER_URI = LDAP_ENDPOINT
+AUTH_LDAP_BIND_DN = BIND_DN
+AUTH_LDAP_BIND_PASSWORD = BIND_PASSWORD
+AUTH_LDAP_USER_SEARCH = LDAP_USER_SEARCH
+AUTH_LDAP_GROUP_SEARCH = LDAP_GROUP_SEARCH
+AUTH_LDAP_GROUP_TYPE = GroupOfNamesType()
+AUTH_LDAP_USER_ATTR_MAP = {
+    "first_name": "givenName",
+    "last_name": "sn",
+    "email": "mail",
+    "username": USER_FILTER,
+}
+AUTH_LDAP_GLOBAL_OPTIONS = {
+    ldap.OPT_X_TLS_REQUIRE_CERT: False,
+    ldap.OPT_REFERRALS: False
+}
+AUTH_LDAP_ALWAYS_UPDATE_USER = True
+AUTH_LDAP_MIRROR_GROUPS = True
+AUTH_LDAP_FIND_GROUP_PERMS = True
+AUTH_LDAP_CACHE_GROUPS = False
--- a/controller/templates/confd_settings.py
+++ b/controller/templates/confd_settings.py
@@ -45,3 +45,16 @@ REGISTRATION_ENABLED = bool({{ getv "/deis/controller/registrationEnabled" }})
 WEB_ENABLED = bool({{ getv "/deis/controller/webEnabled" }})
 {{ end }}
 UNIT_HOSTNAME = '{{ if exists "/deis/controller/unitHostname" }}{{ getv "/deis/controller/unitHostname" }}{{ else }}default{{ end }}'
+
+# AUTH
+# LDAP
+{{ if exists "/deis/controller/auth/ldap/endpoint" }}
+LDAP_ENDPOINT = '{{ if exists "/deis/controller/auth/ldap/endpoint" }}{{ getv "/deis/controller/auth/ldap/endpoint"}}{{ else }} {{ end }}'
+BIND_DN = '{{ if exists "/deis/controller/auth/ldap/bind/dn" }}{{ getv "/deis/controller/auth/ldap/bind/dn"}}{{ else }} {{ end }}'
+BIND_PASSWORD = '{{ if exists "/deis/controller/auth/ldap/bind/password" }}{{ getv "/deis/controller/auth/ldap/bind/password"}}{{ else }} {{ end }}'
+USER_BASEDN = '{{ if exists "/deis/controller/auth/ldap/user/basedn" }}{{ getv "/deis/controller/auth/ldap/user/basedn"}}{{ else }} {{ end }}'
+USER_FILTER = '{{ if exists "/deis/controller/auth/ldap/user/filter" }}{{ getv "/deis/controller/auth/ldap/user/filter"}}{{ else }} {{ end }}'
+GROUP_BASEDN = '{{ if exists "/deis/controller/auth/ldap/group/basedn" }}{{ getv "/deis/controller/auth/ldap/group/basedn"}}{{ else }} {{ end }}'
+GROUP_FILTER = '{{ if exists "/deis/controller/auth/ldap/group/filter" }}{{ getv "/deis/controller/auth/ldap/group/filter"}}{{ else }} {{ end }}'
+GROUP_TYPE = '{{ if exists "/deis/controller/auth/ldap/group/type" }}{{ getv "/deis/controller/auth/ldap/group/type"}}{{ else }} {{ end }}'
+{{ end }}
--- a/controller/build.sh
+++ b/controller/build.sh
@@ -16,7 +16,7 @@ DEBIAN_FRONTEND=noninteractive
 # HACK: install git so we can install bacongobbler's fork of django-fsm
 # install openssh-client for temporary fleetctl wrapper
 apt-get update && \
-    apt-get install -yq python-dev libffi-dev libpq-dev libyaml-dev git
+    apt-get install -yq python-dev libffi-dev libpq-dev libyaml-dev git libldap2-dev libsasl2-dev
 
 # install pip
 curl -sSL https://raw.githubusercontent.com/pypa/pip/6.0.8/contrib/get-pip.py | python -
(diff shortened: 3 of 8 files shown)


dserodio:
:+1:

lorieri:
:+1:
after this, production \o/

controller/deis/settings.py
```diff
@@ -136,6 +140,8 @@
     'django.contrib.sites',
     'django.contrib.staticfiles',
     # Third-party apps
+    'django_auth_ldap',
+    'django_fsm',
```
bacongobbler: this shouldn't be here, right? :)

bacongobbler:
looks good, but how do you replace the token auth with LDAP? I don't see any changes in /api/urls.py or in the views. As far as I can tell this doesn't actually replace the `deis login` endpoints.

gabrtv:
@bacongobbler looks like the PR is modifying the authentication backend, which means token auth and API endpoints should work without any changes -- which is terrific.

@phspagiari nice work, I look forward to seeing this merged!  A few questions:

1. It's not clear to me how this functionality is enabled/disabled. What happens for the majority of users who won't have this configured?
2. Without automated tests I'm concerned we could break this functionality over time with, for example, a library update.  Thoughts?

phspagiari:
@bacongobbler  as @gabrtv said, token auth and API endpoints works with ldap or with default auth without any changes. If you configure LDAP, the /auth/login will try to login via LDAP, if you dont configure, the django will pass the ldap backend and try to login with other backend (default).

@gabrtv If all the values of LDAP is left blank (not configured) django will pass  to the next auth backend and try to login on it. At first I thought that I would need a ugly if to add or remove the ldap auth backend, but latter I saw that django does this automatically.

controller/deis/settings.py
phspagiari: I'm not a django expert, some people said that would be better if we put the module here, some people don't. For safe precautions I prefer to have it explicit. I can remove if you think that its better. :+1:

bacongobbler:
Sounds good to me! :+1:

controller/deis/settings.py
bacongobbler: We don't use django_fsm any more, so it shouldn't be necessary. The only reason we keep it around any more are for legacy South migrations. Let's remove it so the module's not taking up precious resources when the server boots. django_auth_ldap should be kept though, of course.

bacongobbler:
@gabrtv I guess we'll just have to hope that django_auth_ldap doesn't make any critically breaking changes... Either that or we're gonna have to set up an LDAP server in the future to test updates ;)

phspagiari:
@bacongobbler @gabrtv A question: I want to make a documentation of how to use LDAP with Deis. What is the best way to do it? Put together with this PR or creating another one after this PR is accepted?

controller/deis/settings.py
phspagiari: Haha, I'm talking about the django_auth_ldap into this list, probably came in some merge I did, had not noticed, I obviously do not use this module, sorry I will remove. :smile:

bacongobbler:
I'd throw it into this PR

phspagiari:
I added the documentation :smile:. I use the "Managing Deis" tree because I saw the "Using a Proxy Server" here, I can move if you guys prefer in another place.

docs/managing_deis/using-a-ldap-auth.rst
```diff
@@ -0,0 +1,48 @@
...
+Using a LDAP Auth
+=================
+
+Deis Controller supports Single Sign On access control, for now Deis is able to authenticate using LDAP or Active Directory.
+
+Settings used by LDAP
+---------------------
+
+=========================================           =================================================================================
+setting                                             description
```
bacongobbler: We can keep the header, but this should be moved over to http://docs.deis.io/en/latest/customizing_deis/controller_settings/

docs/managing_deis/using-a-ldap-auth.rst
```diff
@@ -0,0 +1,48 @@
...
+/deis/controller/auth/ldap/group/filter             The field that we will locate your groups with LDAPSearch. (In most cases is objectClass)
+/deis/controller/auth/ldap/group/type               The Groups type of LDAP. (Use groupOfNames if you don't know)
+=========================================           =================================================================================
+
+Configuring LDAP on Controller
+------------------------------
+
+.. important::
+
+    It's important that you have a local admin user ( see :ref:`Register a User <register-user>` ) and have disabled the registration ( see :ref:`disable_user_registration` ) in order to only LDAP users and the local admin can login into controller.
```
bacongobbler: Could you please elaborate more on why they must have registered a local admin user and have disabled registration? I understand the reasoning behind this but it'd be nice to help users understand why it's important.

phspagiari:
Another thing: I said some considerations at docs about the admin user and disabling the default registration, It would be great if you guys give your opinions about it. 
First I was imagining that I would have to implement the "disable registration" if the LDAP is enabled, but we can use the default method to this.
I tested and so far so good :)

docs/managing_deis/using-a-ldap-auth.rst
phspagiari: Perfect, I will try to elaborate more about the problems if the users don't do this.

docs/managing_deis/using-a-ldap-auth.rst
phspagiari: I wrote this doc at controller_settings and moved to a separated doc before. I think that since this is optional, should be equal to the proxy settings. But if its better on controller I can move again.

azurewraith:
:+1: This would be useful...

deis-admin:

lorieri:
just tested it, works like a charm

lorieri:
good to disable registration

# doesn't work
core@coreos1 ~ $ deisctl config controller set registrationEnabled='False'
false

# works
core@coreos1 ~ $ etcdctl set /deis/controller/registrationEnabled
False

phspagiari:
@lorieri  In LDAP Docs is referenced the [disable_user_registration](http://docs.deis.io/en/latest/managing_deis/operational_tasks/?highlight=disable%20user%20registration#disabling-user-registration). The key registrationEnabled is boolean, not string. This changes the value in [confd_settings.py L41](https://github.com/deis/deis/blob/master/controller/templates/confd_settings.py#L41).

phspagiari:
Guys, I tested again against a new cluster of Deis running in 3 production servers. Works so far so good :)

carmstrong:
Code LGTM!

bacongobbler:
We don't have an Active Directory server here to manually test this PR. Considering that multiple community members have tested this so far with success I think it's suffice to say that it works.

I'm wondering if we should add documentation around that this is maintained by the community. I don't have any experience with LDAP so I'm the last guy you'd want to approach for issues around LDAP auth. @carmstrong? :)

carmstrong:
> We don't have an Active Directory server here to manually test this PR. Considering that multiple community members have tested this so far with success I think it's suffice to say that it works.

I would agree. Let's add a note to the documentation added as part of this PR to clarify that LDAP support is maintained by the community.

phspagiari:
It seems fair enough to me and if you guys wish you can put me as the main member of the community to support LDAP. :) @carmstrong @bacongobbler

carmstrong:
> It seems fair enough to me and if you guys wish you can put me as the main member of the community to support LDAP. :)

Sounds good to me. Below "Using LDAP Auth" let's add something like:

```
.. note::

    LDAP support was contributed by community member Pedro Spagiari (@phspagiari) and is unsupported by the Deis core team. 
```

mboersma:
For me, `make -C controller/ test-unit` fails, because it couldn't install `python-ldap` without having the Ubuntu packages `libldap2-dev` and `libsasl2-dev` installed. Could you add those to the `apt-get install` line in [`setup-node.sh`](https://github.com/deis/deis/blob/master/tests/bin/setup-node.sh#L36)?

Also `make -C docs/ test` fails with `No module named ldap`. I think you need to add the two required python libraries to [docs_requirements.txt](https://github.com/deis/deis/blob/master/docs/docs_requirements.txt) as well.

deis-admin:

phspagiari:
@mboersma I saw this when I install the local docs server but I forgot to fix. But its not a good way using a "duplicated" requirements from controller into docs_requirements, right? We can do something like "-r ../controller/requirements.txt" into docs_requirements or its too ugly :smile: ?

About the libldap2-dev, I put at the [build.sh script](https://github.com/deis/deis/pull/3174/files#diff-fcfcde2541cd2d833fa057cc28420e9cR19). If we change the setup-node.sh to run the build.sh instead duplicate the dependencies its not a better option?

bacongobbler:
Actually that may be a good middle ground. I didn't realize you could send option flags inside requirements.txt

phspagiari:
@bacongobbler yes you can, people normally put a ``-r requirements.txt`` inside a dev_requirements.txt for example. I will do all the changes (including the docs). Waiting about the build.sh script, but for now I will put the libldap2-dev into setup-node.sh

mboersma:
The reason we have `docs_requirements.txt` is because we host our documentation at ReadTheDocs.org, where the docs build is configured by a web form that allows only one `pip` requirements file to be specified. So yes: we need to maintain both `requirements.txt` and `docs_requirements.txt` in paralell.

bacongobbler:
Dang it... knew there was a reason. =(

phspagiari:
@mboersma @bacongobbler Sad :( Ok, I will make the changes. :+1:

phspagiari:
Done and squashed :+1:

bacongobbler:
code LGTM still stands with latest changes

bacongobbler:
https://ci.deis.io/job/test-ec2/178/ if you want to follow the job

bacongobbler:
@phspagiari could you please rebase? `test-ec2.sh` is on master :)

phspagiari:
@bacongobbler Done! :)

controller/templates/confd_settings.py
```diff
@@ -46,3 +46,14 @@
 {{ end }}
 
 UNIT_HOSTNAME = '{{ or (.deis_controller_unitHostname) "default" }}'
+
+# AUTH
+# LDAP
+LDAP_ENDPOINT = '{{ .deis_controller_auth_ldap_endpoint }}'
```
bacongobbler: You'll have to make these conditional or else confd will bail forever when these keys are unset: https://ci.deis.io/job/test-ec2/213/

phspagiari: weird, I created a deis cluster without the keys and everything works fine, but indeed the confd will wait for the keys forever.

@bacongobbler You have a suggestion for these conditional? 
I imagine something like "controller/auth/ldap/enabled" with a false by default. Its enough ?

bacongobbler:
I think it's safe to assume if /deis/controller/auth/ldap/endpoint is set, then that's the auth system we'd want to use.

phspagiari:
@bacongobbler there is a reason for confd_settings.py uses a "syntax" different from the default from confd?
Example: confd_settings.py uses {{ .key_for_something }} instead  of {{ getv /key/for/something }} and when I try to test this template with confd into my mac I get error of "bad syntax".

bacongobbler:
It's based on an older version of confd. 0.4.6, I believe. We're in the process of upgrading to 0.8 in #3361

phspagiari:
I see.. and #3361 answered my doubt about why the ``or default`` works using the older version and not using the new version with getv.
I prefer the ``or default``` instead of keep the variables with None value in the ``settings.py`` and a ``if`` into endpoint key. You agree?

bacongobbler:
I was thinking more along the lines of it being enabled equivalent to [this](https://github.com/deis/deis/blob/master/registry/templates/config.yml#L44-L53). In that example, if the user sets /deis/registry/smtpHost then the registry will send emails based on that config.

phspagiari:
I went with the ``if key`` option... but my idea was to remove the default attributes at settings.py like [this](https://github.com/deis/deis/pull/3174/files#diff-688613f62c7e8e2421cff9128bab2822R334). With my idea django will break if the keys doesn't exists. After thinking a little I supose that will be better if the attributes have a default value in settings file for "understanding purposes".

I made a rebase too. :) Lets test it again :+1:

bacongobbler:
Sorry @phspagiari. This'll need another rebase as we just merged #3361 so you'll need to update the confd_settings.py template.

phspagiari:
@bacongobbler holy cow... this merge is just the confd update =( We are talking about this earlier... Ok I will change all to the new confd template version.

phspagiari:
Updated syntax to confd 0.8.0 :)

phspagiari:
I saw the error, the problem is dont know another way to do this 
```python
 filterstr="(" + USER_FILTER + "=%(user)s)"
```
The usually pythonic way to do this ```"%s=%(user)s" % (USER_FILTER)``` or even ``"%s=%s" % (USER_FILTER, user) doesnt work because of the ``user`` is a placeholder of django-ldap.
 String concatenation its ugly but I cant imagine another way to do this.

bacongobbler:
Set up `USER_FILTER` as an empty string instead. The format wants a unicode string, not a NoneType.

bacongobbler:
you can also run `make -C docs` to check if it works

phspagiari:
@bacongobbler Yes, I imagine that was a lint error first, I am fixing the two problems right now.

phspagiari:
Sorry for my mistake, all fixed.

mboersma:
Code LGTM. Thanks again @phspagiari!

controller/templates/confd_settings.py
```diff
@@ -45,3 +45,16 @@
 WEB_ENABLED = bool({{ getv "/deis/controller/webEnabled" }})
 {{ end }}
 UNIT_HOSTNAME = '{{ if exists "/deis/controller/unitHostname" }}{{ getv "/deis/controller/unitHostname" }}{{ else }}default{{ end }}'
+
+# AUTH
+# LDAP
+{{ if exists "/deis/controller/auth/ldap/endpoint" }}
+LDAP_ENDPOINT = '{{ if exists "/deis/controller/auth/ldap/endpoint" }}{{ getv "/deis/controller/auth/ldap/endpoint"}}{{ else }} {{ end }}'
```
bacongobbler: I'd remove the whitespace between `{{else}}` and `{{end}}` so the end result is an empty string if the key's unset.

bacongobbler:
small nit, but LGTM :)

controller/templates/confd_settings.py
mboersma: Oops, sorry I missed this comment before merging. Should we typo-fix the space change?

phspagiari:
Typo :( Thanks guys, I'm very help to be contributing with Deis :smile:

controller/templates/confd_settings.py
bacongobbler: nah, it's not a huge issue. Everything will sort itself out once all ldap keys are set :)

bacongobbler:
Thanks for pushing this one through, @phspagiari :balloon:

PR #3175, opened 2015-03-03
Proposal ref(*): explore ubuntu alternatives

aledbf:
This is a proposal to change (where is possible) the base image from `ubuntu-debootsrap` to `gliderlabs/alpine:3.1`
 
After two weeks with alpine linux this is the result:
- the migration was transparent (I upgraded an existent cluster)
- the reduction of the size is considerable
- the images contains only the essential things
- is not possible to use alpine linux and ceph. The deis store components are unchanged
    http://bugs.alpinelinux.org/issues/3241
    if they add support the store images can be a lot smaller than now.

- the changes in Dockerfile and scripts are related to busybox (for instance sed doesn't support the -u flag)
- small change to wal-e flags to be busybox compatible https://github.com/wal-e/wal-e/issues/165
- the binary https://s3-us-west-2.amazonaws.com/opdemand/confd-v0.5.0-json is not a static one
 
**This is current size of the images that I'm running in the development cluster**
 ```
deis/builder                        git-7e51baf             38b6fdb64eeb        4 days ago          515.2 MB
deis/cache                          git-7e51baf             f9b12ca66623        4 days ago          10.61 MB
deis/controller                     git-7e51baf             92f83775cd45        4 days ago          112.5 MB
deis/database                       git-7e51baf             0c3659e8fe52        4 days ago          166.4 MB
deis/logger                         git-7e51baf             68fde222937c        4 days ago          9.802 MB
deis/logspout                       git-7e51baf             f38940508009        4 days ago          10.47 MB
deis/publisher                      git-7e51baf             defaf8e17a02        4 days ago          10.04 MB
deis/registry                       git-7e51baf             3151c1bdc38d        4 days ago          148.5 MB
deis/router                         git-7e51baf             8a664f0f5add        4 days ago          31.68 MB
 ```
 ```
deis/builder                        v1.3.1                  f6757e43b95c        2 weeks ago         751.1 MB
deis/cache                          v1.3.1                  b8ae20fb03c3        2 weeks ago         95.87 MB
deis/controller                     v1.3.1                  90be93e1858f        2 weeks ago         272.3 MB
deis/database                       v1.3.1                  7deb3e8fcb63        2 weeks ago         349.9 MB
deis/logspout                       v1.3.1                  4732d3250eae        2 weeks ago         10.06 MB
deis/publisher                      v1.3.1                  ac0ad90e7aca        2 weeks ago         5.036 MB
deis/registry                       v1.3.1                  54e709265d9f        2 weeks ago         316.6 MB
deis/router                         v1.3.1                  10aa6b6d4ac8        2 weeks ago         151.1 MB
```

I know that this is a "radical" change but the difference in the size is more than considerable

### CODE REVIEW

**Comment by bacongobbler** [comment]
been considering this one as well, just didn't have the time to start thinking about it. :+1: for me on moving to a more minimal distro, just wondering if we should wait until ceph-rdb support is in alpine linux first.

**Comment by mboersma** [comment]
It's too bad Ceph won't work in alpine linux currently, but I like this change very much. Separately, I reworked a couple components to play around with alpine and found it very approachable and sensible.

I guess if we're not able to use gliderlabs/alpine for all components, we should do the tests to see if this actually speeds things up overall. (It may be that optimizing some containers actually costs us more download time if the overall set of images ends up sharing fewer layers.) I'm optimistic about this change though. Alpine is really designed for this purpose, and ubuntu-debootstrap isn't.

**Comment by carmstrong** [comment]
> is not possible to use alpine linux and ceph.

That bug is specific to the RBD kernel module, which I don't believe we use. What about the Ceph FS kernel driver?

**Comment by mboersma** [comment]
> the binary https://s3-us-west-2.amazonaws.com/opdemand/confd-v0.5.0-json is not a static one

I'm glad you noticed that. We should fix that on its own.

**Comment by mboersma** [comment]
@aledbf something's going wrong in builder in the tests:
```
Step 5 : USER slug
 ---> Running in c56736e88642
Driver vfs failed to create image rootfs 92274cb5b6fc2e1f3a98d1d368d168096f9e2ebc6522ef811293e59672056238: Untar exit status 1 write /usr/share/lintian/overrides/xml-core: no space left on device
```

**Comment by aledbf** [comment]
@carmstrong  ceph is not ready to run in "any" linux distro not just because of the silly [res_nquery](https://github.com/aledbf/ceph/commit/fb4a30f9acbc4e08787b9938b38839766ec8f88e) error.
This are the steps to fix that but even then the number of libraries to install is huge.

```
docker run -it gliderlabs/alpine:3.1 /bin/sh

echo "@testing http://nl.alpinelinux.org/alpine/edge/testing" >>/etc/apk/repositories

apk-install yasm \
  libtool \
  util-linux-dev \
  udev-dev \
  automake \
  autoconf \
  build-base \
  git \
  perl \
  keyutils \
  keyutils-dev \
  crypto++@testing

git clone --recursive https://github.com/aledbf/ceph.git

./autogen.sh

./configure \
  --prefix=/usr \
  --sysconfdir=/etc \
  --mandir=/usr/share/man \
  --infodir=/usr/share/info \
  --without-profiler \
  --without-debug \
  --without-fuse \
  --without-tcmalloc \
  --without-jdk-dir \
  --without-libatomic-ops \
  --without-ocf \
  --without-kinetic \
  --without-librocksdb \
  --without-librocksdb-static \
  --without-rest-bench \
  --without-libaio \
  --without-libxfs \
  --without-librocksdb \
  --without-librocksdb-static
```

and given how important ceph is in deis I think that is not an option to compile it from source (for now).

carmstrong:
> and given how important ceph is in deis I think that is not an option to compile it from source (for now).

Agreed 1000%. I wasn't thinking about built Ceph packages - of course we'd need those.

aledbf:
Jenkins, test this please.

mboersma:
If we can't also base the deis/store-* components off of `alpine:3.1`, I'm curious if this saves us time and bandwidth overall for the entire set of Docker images, since we still have to bring in `ubuntu-debootstrap` for Ceph. (I think it still might.)

I would like to do more testing with this at some point, but it needs a rebase.

aledbf:
> I'm curious if this saves us time and bandwidth overall for the entire set of Docker images,

Good point.
Running this I "see" the difference in 3 images: registry, controller and database when I kill the node where is running and is scheduled in another (if the image is not in the node, obviously).

carmstrong:
> I'm curious if this saves us time and bandwidth overall for the entire set of Docker images,

I personally think we should close this and revisit if and when we can move all the containers over to a new base image. I suspect the maintenance burden of tracking multiple upstream bases outweighs the potential benefit from size reduction. 

Of course, I've been wrong before :smile:

aledbf:
> bases outweighs the potential benefit from size reduction.

@carmstrong ok, didn't mention this in the response to @mboersma but specifically this change reduces by half the size (even if I add the 90MB of `ubuntu-debootstrap`).

I know that I must prove this statement before any possible change :)

mboersma:
>  I suspect the maintenance burden of tracking multiple upstream bases outweighs the potential benefit from size reduction.

Very possibly--we don't know until we've done some testing & benchmarking. 

I don't think it's a big maintenance headache to leave deis-store-* as-is, but move other containers to `alpine`. It means we have to speak two Linux dialects in the Dockerfiles, but specifically it means we need to know `apk-install` which isn't much to learn IMHO.

It's also worth noting that another big benefit of `alpine:3.1` is that the containers also *build* much faster than on Ubuntu, so final image size isn't the only benefit here.

carmstrong:
> It's also worth noting that another big benefit of alpine:3.1 is that the containers also build much faster than on Ubuntu, so final image size isn't the only benefit here.

That's certainly not insignificant...

I'm fine with whatever you guys decide on this, @aledbf and @mboersma. I just suggest that we close the PR and work in a branch if it's not something we can necessarily merge soon, but it sounds like maybe we can.

aledbf:
> I just suggest that we close the PR and work in a branch if it's not something we can necessarily merge soon, but it sounds like maybe we can.

ok. I will open a new PR after the slowdown of changes in the repo (keeping this up to date is killing me)

aledbf:
The `wal-e` change is not required upstream after this https://github.com/wal-e/wal-e/pull/171

Issue #1532, opened 2014-08-07
Push configuration from existing environment file


andyshinn:
it would be nice to have a command to push configuration from an existing `.env` file. There already appears to be a `config:pull` command to download the `.env` file. This is a little bit related to #1531 in that it would help configure existing Procfile based applications in one go.

gvilarino:
+1000

I think the `CLI` api should mimic [`ddolar/heroku-config`](https://github.com/ddollar/heroku-config) (a plugin for `heroku toolbelt`)

`deis config:pull` should keep it's functionality
`deis config:push` should push contents of a `.env` (if present) as if it were `deis config:set [..]`

bacongobbler:
`deis config:pull` is already implemented. Personally I'm interested why a `config:push` command is necessary when `deis config:set $(cat .env)` does the same thing... Any reasons why the latter won't work?

gvilarino:
tl;dr: I think most users would expect `config:push` to work if `config:pull` works.

I wouldn't call it 'necessary' as much as I would call it 'expected'. It's pretty safe to assume that over 80% of Deis implementors have been or still are Heroku users. With that in mind, the most sensible approach would be to mimic Heroku's API as much as possible, in order to provide what's expected by Deis' userbase. That being said, Heroku dosen't provide `config:pull` out of the box; that feature is instead provided by the `[heroku-config](https://github.com/ddollar/heroku-config)` plugin. But that plugin provides `config:push` as well, all based on the `.env` contents.

bacongobbler:
I understand the motivation behind convenience functions, but after giving this more thought on this I believe that they should be a plugin, much like Heroku's client. If there's a way we can make the client more pluggable, this functionality could easily be added into the client and convenience functions like `config:pull` and `config:push` could be moved out into a plugin. I'm all for it; I'd just prefer that we focus on the "core" functionality and allow users to extend the client the way they prefer.

Does that seem reasonable to you?

gvilarino:
:+1: 
I agree on `config:push/pull` being plugins as a right approach. I just thought that as `pull` was part of the core, then `push` might be just as well :)

carmstrong:
> I just thought that as pull was part of the core, then push might be just as well

That's a valid point. @bacongobbler we probably need to have a larger discussion around plugin architecture.

PR #3177, opened 2015-03-03
Proposal: deis client plugins

bacongobbler:
I've talked about this before and I think this would be really beneficial for commands like `deis config:push` etc. as discussed in #1532.

This article describes the changes involved in the client in order to manage client plugins. It essentially behaves the same as `heroku plugins`, following the theme of winning big with the Heroku -> Deis migration path. I didn't feel like it fit in any of the other document paths, and it follows the same URL path as [heroku's articles](https://devcenter.heroku.com/articles). Docker does the same thing as well: http://docs.docker.com/articles/

If this is approved, I'll get to work on implementing a PoC plugin called `deis-accounts`, similar to https://github.com/ddollar/heroku-accounts.

### DIFF
--- a/client/deis.py
+++ b/client/deis.py
@@ -2208,7 +2208,17 @@ def main():
     if hasattr(cli, cmd):
         method = getattr(cli, cmd)
     else:
-        raise DocoptExit('Found no matching command, try `deis help`')
+        # split by : to execute the proper command
+        split_cmd = args['<command>'].split(':')
+        dash_separated_command = 'deis-{}'.format(split_cmd[0])
+        arglist = args['<args>']
+        if len(split_cmd) > 1:
+            # safety precaution in case users want to use more than one colon in their command
+            arglist = split_cmd[1:] + arglist
+        try:
+            sys.exit(subprocess.call([dash_separated_command] + arglist))
+        except OSError:
+            raise DocoptExit('Found no matching command, try `deis help`')
     # re-parse docopt with the relevant docstring
     docstring = trim(getattr(cli, cmd).__doc__)
     if 'Usage: ' in docstring:
--- a/docs/customizing_deis/cli-plugins.rst
+++ b/docs/customizing_deis/cli-plugins.rst
@@ -1,0 +1,36 @@
+:title: CLI Plugins
+:description: How to manage plugins for the Deis CLI.
+
+.. _cli_plugins:
+
+CLI Plugins
+===========
+
+Plugins allow developers to extend the functionality of the :ref:`Deis Client <install-client>`,
+adding new commands or features.
+
+If an unknown command is specified, the Client will attempt to execute the command as a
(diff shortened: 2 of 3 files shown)


bacongobbler:
Alternatively, here's a different implementation which could also be implemented in deis. Less moving parts, though I'm not sure how the whole `resource:command` syntax would work in this context: https://github.com/fishworks/dis#extending-dis

bacongobbler:
I'm leaning more towards the latter, actually. It means that plugins can be implemented in any language (which means we have an upgrade story for moving to a Go client) and is overall much simpler to implement :)

rjocoleman:
I like this feature, and the implementation you're suggesting is like the git cli? If so [`packer`](https://github.com/mitchellh/packer) works like this too.

This also opens up the possibility of users using any language for their plugins which is nice.

One caveat, updates are then left up to the individual plugin?

bacongobbler:
yes, precisely. Everything is left up to the plugin.

bacongobbler:
> One caveat, updates are then left up to the individual plugin?

Note that the plugin manager could also be implemented as a plugin, too ;)

bacongobbler:
Work here on the client side is done. Now just making a plugin...

bacongobbler:
done!

https://github.com/fishworks/deis-accounts

phspagiari:
:+1: This will be very useful here since we want to make a different flow for our pushes and with plugins will be very easy :)

bacongobbler:
unfortunately you cannot override existing commands' behaviour, but you can create a new command which would utilize your custom workflow. I think `git` behaves the same way

carmstrong:
Code LGTM.

bacongobbler:
Note that this PR brings a new "Articles" section to the docs as I couldn't see a relevant section for "handling third-party plugins". In #2911 I'm throwing the documentation into the "reference" docs but IMO it might be hard to find if it's stuck in the "reference" section. "reference" is typically reserved for reference guides, such as API/client documentation.

Does this seem fair, or is there a better place to put this documentation?

mboersma:
> is there a better place to put this documentation?

I think it would fit in as a top-level item under [Customizing Deis](http://docs.deis.io/en/latest/customizing_deis/).

mboersma:
I had problems building the [`deis-accounts`](https://github.com/fishworks/deis-accounts) plugin binary which I will add as issues to that project, but a trivial plugin worked fine:
```console
$ sudo vi /usr/bin/deis-hello
$ sudo chmod +x /usr/bin/deis-hello 
$ deis-hello
Hello!
$ deis hello
Hello!
```

I still think the docs for this would go better under [Customizing Deis](http://docs.deis.io/en/latest/customizing_deis/), but otherwise this looks good.

bacongobbler:
@mboersma done

mboersma:
LGTM.

bacongobbler:
jenkins is acting funky but this PR passed the docs test and has gone green before. Good to merge?

PR #3179, opened 2015-03-04
ref(security): REJECT target uses port-unreachable

ineu:
This allows for more clear error messages. Instead of
coreos-3.example.com [x.x.x.x] 554 (rtsp) : No route to host
it gives
coreos-3.example.com [x.x.x.x] 554 (rtsp) : Connection refused

### DIFF
--- a/contrib/util/custom-firewall.sh
+++ b/contrib/util/custom-firewall.sh
@@ -42,7 +42,7 @@ template=$(cat <<EOF
 
 # Log and drop everything else
 -A Firewall-INPUT -j LOG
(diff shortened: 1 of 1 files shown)


carmstrong:
Code LGTM. I think this is clearer.

carmstrong:
Ping @aledbf - he wrote the firewall implementation

mboersma:
Code LGTM, better error message for sure.

aledbf:
@carmstrong the intention of that flag is to make the host a black hole (it doesn't return information of ports not opened). But I'm not against this change if this confuses a user (to see the difference is required to scan the host with the 2 alternatives)

PR #3181, opened 2015-03-04
fix(client): Older dates are incorrectly parsed as today and yesterday.

glogiotatidis:
readable_datetime function believes that datestring '2015-02-04T09:14:34UTC' is today (when on 2015-03-04)


client/deis.py
```diff
@@ -299,7 +299,7 @@ def readable_datetime(datetime_str):
     now = datetime.now(timezone)
     delta = relativedelta.relativedelta(now, dt)
     # if it happened today, say "2 hours and 1 minute ago"
-    if delta.days <= 1 and dt.day == now.day:
+    if dt > (now - relativedelta.relativedelta(days=1)):
```
bacongobbler: we should probably move line 321 (the line where we initiate `yesterday`) above this line such that there's no duplicate code here

client/deis.py
```diff
@@ -316,13 +316,14 @@ def readable_datetime(datetime_str):
             return 'Just now'
         else:
             return "{}{}ago".format(hour_str, min_str)
+
     # if it happened yesterday, say "yesterday at 3:23 pm"
-    yesterday = now + relativedelta.relativedelta(days=-1)
-    if delta.days <= 2 and dt.day == yesterday.day:
```
bacongobbler: I think this logig's a bit cleaner and accomplishes the same thing as line 322. Am I missing something?

glogiotatidis: relativedelta obj is not acting the way this code expects:
```python
In [1]: from dateutil import tz

In [2]: from dateutil import parser

In [3]: from datetime import datetime

In [4]: from dateutil import relativedelta

In [5]: 

In [5]: now = datetime.now(tz.tzlocal())

In [6]: datetime_str = '2015-02-04T09:14:34UTC'

In [7]: dt = parser.parse(datetime_str).astimezone(tz.tzlocal())

In [8]: r = relativedelta.relativedelta(now, dt)

In [9]: r
Out[9]: relativedelta(months=+1, days=+1, hours=+2, minutes=+55, seconds=+55, microseconds=+527323)

In [10]: r.days
Out[10]: 1
```

`r.days` is `1` and not the actual number of days between the two dates.

bacongobbler: Ah, I see your point. I thought that r.days would have been 31 in that case, but it's only showing as 1, causing it to show "Yesterday" every month interval. :+1:

bacongobbler:
code LGTM!

bacongobbler:
@glogiotatidis would you mind rebasing, please?

mboersma:
This should fix the date calculation for that case. Code LGTM.

glogiotatidis:
unintentionally i closed this, so i created #3283

PR #3183, opened 2015-03-04
Proposal ref(store): remove apache2 from ceph gateway

aledbf:
This PR removes `apache2` as web server and uses the the web server that `radosgw` includes (inside the same binary) [civetweb](https://github.com/bel2125/civetweb)


bacongobbler:
ping @carmstrong as I know he explicitly wanted apache2 as per the ceph docs, but if they embed some other router in there then that may be the "new" recommended way.

carmstrong:
I'm ok with this in principle, but the documentation still suggests Apache: http://ceph.com/docs/master/radosgw/config/

I wonder why they don't recommend the built-in server?

aledbf:
@carmstrong  I just saw this https://github.com/ceph/ceph-docker/pull/33

carmstrong:
> I just saw this

Yeah, but I don't think "we" in that sentence refers to the Ceph team :) I'd rather wait on this until the official documentation suggests it.

lorieri:
Hi,
the ceph modified apache package was created to handle manipulation of objects up to 5G

Ulexus:
My 2 satoshi:

[We](http://github.com/ceph/ceph-docker) just merged the change to use civetweb by default.  We are using the official debian packages from Ceph.com, so this is not an outside patch.  It is, however, less well-tested than using Apache.  From my testing, though, it is much faster to start and (in the framework of Docker, due to the single-binary execution) is a more reliable, simpler deployment.

carmstrong:
> From my testing, though, it is much faster to start and (in the framework of Docker, due to the single-binary execution) is a more reliable, simpler deployment.

Fair enough. Thanks, @Ulexus!

@aledbf I'm okay with this PR once Jenkins resurrects and confirms the change.

aledbf:
Jenkins, test this please.

aledbf:
Jenkins, test this please.

aledbf:
Jenkins, test this please.

aledbf:
Jenkins, test this please.

aledbf:
Jenkins, test this please.

aledbf:
Jenkins, test this please.

aledbf:
ping @carmstrong

carmstrong:
Code LGTM.

bacongobbler:
I'm personally concerned with the "less well-tested as Apache" statement here. I'm all for it, but as long as it proves to be just as reliable and can handle the current workload in production. Considering that ceph-docker is relying on it gives me a bit more comfort, so it (tenatively) LGTM.

carmstrong:
> Considering that ceph-docker is relying on it gives me a bit more comfort

That's the boat I'm in, as well. We're probably going to look at using ceph-docker for the store components (#3198), so the fact that ceph-docker uses it is enough for me. I'd love to be able to run a test suite and load test this change, though...

Ulexus:
Also keep in mind that the webserver is really just a patchwork for the
fastcgi handler of radosgw.  The new ceph/radosgw is even pluggable, such
that you can just run radosgw and hook up your own webserver.  That said,
I'm pretty comfortable with the in-binary civetweb.

Ulexus:
Patchwork->passthrough

carmstrong:
Merging, based on the LGTMs from myself and @bacongobbler, and the community support from @Ulexus and @lorieri.

Issue #3185, opened 2015-03-04
Registration through API on 1.4.0 return internal server error


daemonza:
Running the following

        curl -i -X POST \
         -H "Content-Type: application/json" \
          -d '{"username":"demouser","password":"123"}' \
         http://DEIS/v1/auth/register/

as per the documentation here : 
http://docs.deis.io/en/latest/reference/api-v1.1/

I get the following : 

    HTTP/1.1 500 INTERNAL SERVER ERROR
    Content-Type: text/html
    Date: Wed, 04 Mar 2015 15:33:44 GMT
    Server: nginx/1.6.2
    X_DEIS_API_VERSION: 1.1
    X_DEIS_PLATFORM_VERSION: 1.4.0
    transfer-encoding: chunked
    Connection: keep-alive
    Server Error (500)

Running fleetctl journal on the controller this is the error that comes back : 

```
Mar 04 15:33:44 ip-10-21-1-205.eu-central-1.compute.internal sh[2355]: ERROR Internal Server Error: /v1/auth/register/
Mar 04 15:33:44 ip-10-21-1-205.eu-central-1.compute.internal sh[2355]: Traceback (most recent call last):
Mar 04 15:33:44 ip-10-21-1-205.eu-central-1.compute.internal sh[2355]: File "/usr/local/lib/python2.7/dist-packages/django/core/handlers/base.py", line 112, in get_response
Mar 04 15:33:44 ip-10-21-1-205.eu-central-1.compute.internal sh[2355]: response = wrapped_callback(request, *callback_args, **callback_kwargs)
Mar 04 15:33:44 ip-10-21-1-205.eu-central-1.compute.internal sh[2355]: File "/usr/local/lib/python2.7/dist-packages/django/views/decorators/csrf.py", line 57, in wrapped_view
Mar 04 15:33:44 ip-10-21-1-205.eu-central-1.compute.internal sh[2355]: return view_func(*args, **kwargs)
Mar 04 15:33:44 ip-10-21-1-205.eu-central-1.compute.internal sh[2355]: File "/usr/local/lib/python2.7/dist-packages/rest_framework/viewsets.py", line 85, in view
Mar 04 15:33:44 ip-10-21-1-205.eu-central-1.compute.internal sh[2355]: return self.dispatch(request, *args, **kwargs)
[...]

PR #3186, opened 2015-03-04
docs(reference): email is a mandatory parameter

bacongobbler:
closes #3185

PR #3187, opened 2015-03-04
moved email from optional to required json parameter

daemonza:
#3185


bacongobbler:
dupe of #3186 which I also had to fix up in another document. Sorry!

PR #3194, opened 2015-03-04
docs(managing_deis): remove note on upgrading

carmstrong:
The note for Deis v1.0.2 upgrades confuses folks, and considering
we're several versions beyond now, it's probably safe to remove


mboersma:
Agreed, I hope we have very few people upgrading from 1.0.0 or 1.0.1 at this point anyway. Code LGTM.

bacongobbler:
Yup, lgtm!

PR #3196, opened 2015-03-05
fix(deisctl/units): eat error output for history and inspect

carmstrong:
Currently, whenever we run `docker history` and `docker inspect`
before pulling components for the first time, we emit an
"Image not found" error, which is confusing.

This commit suppresses this output.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@ CLIENTS=client deisctl
 all: build run
 
 dev-registry: check-docker
-	@docker inspect registry >/dev/null && docker start registry || docker run --restart="always" -d -p 5000:5000 --name registry registry:0.9.1
+	@docker inspect registry >/dev/null 2>&1 && docker start registry || docker run --restart="always" -d -p 5000:5000 --name registry registry:0.9.1
 	@echo
 	@echo "To use local boot2docker registry for Deis development:"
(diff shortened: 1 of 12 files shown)


carmstrong:
Note that the store units already do this.

mboersma:
`docker history` is just a test, we really don't want to read its output. Great cleanup, code LGTM.

aledbf:
@carmstrong why not include the same thing for `docker inspect`?

```
Welcome to Deis			Powered by CoreOS
core@nodo-3 ~ $ docker inspect deis-registry >/dev/null
Error: No such image or container: deis-registry
core@nodo-3 ~ $ docker inspect deis-registry >/dev/null 2>&1
core@nodo-3 ~ $
```

bacongobbler:
There's [a few spots](https://github.com/deis/deis/pull/3196/files#diff-39efd56dbeb43d46e3a98f4fed6e8ecaR7) where we did the `2>&1` though it does seem like we missed most of the calls to `docker inspect`. good call

carmstrong:
Yep, updated again to get all inspect and history.

aledbf:
Code LGTM.

mboersma:
Code still LGTM after changes.

PR #3202, opened 2015-03-05
feat(docker): load overlay module

### LINKED ISSUE #3192: Enable OverlayFS kernel module for Docker
See https://github.com/deis/deis/issues/1678#issuecomment-77048129

aledbf:
closes #3192

### DIFF
--- a/contrib/coreos/user-data.example
+++ b/contrib/coreos/user-data.example
@@ -88,6 +88,14 @@ coreos:
       Type=oneshot
       ExecStartPre=/usr/sbin/modprobe nf_conntrack
       ExecStart=/bin/sh -c "sysctl -w net.netfilter.nf_conntrack_max=262144"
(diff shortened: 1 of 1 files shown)


carmstrong:
It sounds like we may want to wait for [CoreOS 591](https://coreos.com/releases/#591.0.0) before enabling overlay.

mboersma:
Code LGTM, but I agree we should wait for CoreOS 591 or later to land in stable before merging.

mboersma:
[CoreOS 607.0.0](https://coreos.com/releases/) with Docker 1.5.0 landed in the stable channel just now. We are prepping a PR and then we can revisit this.

carmstrong:
Closing in favor of #3319 - if we intend to enable overlayfs, we'll need to do so in concert with some filesystem changes.

PR #3204, opened 2015-03-05
feat(publisher): remove application when is stopped

### LINKED ISSUE #3176: stop publishing an app to etcd when container is stopped
see discussion in #2044

aledbf:
The reason of `/deis/publisher/containers/<id container> -> <app name>/<app type instance>` is required because in the docker stop event only the id of the stopped container is available.

closes #3176

### DIFF
--- a/publisher/server/publisher.go
+++ b/publisher/server/publisher.go
@@ -1,10 +1,8 @@
 package server
 
 import (
-	"errors"
 	"fmt"
 	"log"
-	"net"
 	"os"
 	"regexp"
 	"strconv"
@@ -44,6 +42,8 @@ func (s *Server) Listen(ttl time.Duration) {
 					continue
 				}
 				s.publishContainer(container, ttl)
+			} else if event.Status == "stop" {
+				s.removeContainer(event.ID)
 			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


bacongobbler:
Personally I'd rather keep the IDs in an in-memory map as it's a lot faster and more reliable to access than etcd. We fall back with TTLs so it's not a huge concern if the process is restarted, losing the map's data.

aledbf:
> Personally I'd rather keep the IDs in an in-memory map

that was my first approach but I thought use etcd like in the rest of the services :)
I will change this

bacongobbler:
code LGTM!

mboersma:
That's great cleanup code, thanks @aledbf. Code LGTM, but this needs a manual test after CI.

aledbf:
Jenkins, test this please.

aledbf:
Jenkins, test this please.

aledbf:
Jenkins, test this please.

aledbf:
Jenkins, test this please.

mboersma:
Note that this passed testing, although the way that GitHub test statuses stack now doesn't reflect that on the summary page with a green :white_check_mark:

mboersma:
I tested this interactively on vagrant. LGTM.

PR #3214, opened 2015-03-06
feat(controller): disable swap usage if there is a memory limit

### LINKED ISSUE #3205: Proposal ref(controller): disable swap usage if there is a memory limit
If the app uses `deis limits:set -m type=XXX -a XXX` add the flag `--memory-swap=-1`
```
docker help run
....
  --memory-swap=""           Total memory usage (memory + swap), set '-1' to disable swap (format: <number><optional unit>, where unit = b, k, m or g)
```

without that if the node uses swap it will not use the specified limit

aledbf:
This change requires docker 1.5.0

closes #3205

### DIFF
--- a/controller/deis/settings.py
+++ b/controller/deis/settings.py
@@ -6,6 +6,7 @@ from __future__ import unicode_literals
 import os.path
 import random
 import string
+import semantic_version
 import sys
 import tempfile
 
@@ -340,3 +341,12 @@ except ImportError:
 if os.path.exists('/templates/confd_settings.py'):
     sys.path.append('/templates')
     from confd_settings import *  # noqa
+
+
+DOCKER_VERSION = os.environ.get('DOCKER_VERSION', '1.4.1')
+
+DISABLE_SWAP = ""
+
+if (semantic_version.validate(DOCKER_VERSION) and
+        semantic_version.Version(DOCKER_VERSION) >= semantic_version.Version('1.5.0')):
+    DISABLE_SWAP = "--memory-swap=-1"
--- a/controller/requirements.txt
+++ b/controller/requirements.txt
@@ -16,6 +16,7 @@ paramiko==1.15.2
 psycopg2==2.6
 python-etcd==0.3.2
 PyYAML==3.11
+semantic_version==2.3.1
 setproctitle==1.1.8
 static==1.1.1
 South==1.0.2
--- a/controller/scheduler/fleet.py
+++ b/controller/scheduler/fleet.py
@@ -129,7 +129,7 @@ class FleetHTTPClient(object):
         # prepare memory limit for the container type
         mem = kwargs.get('memory', {}).get(l['c_type'], None)
         if mem:
-            l.update({'memory': '-m {}'.format(mem.lower())})
+            l.update({'memory': '-m {} --memory-swap=-1'.format(mem.lower())})
         else:
             l.update({'memory': ''})
         # prepare memory limit for the container type
--- a/controller/scheduler/fleet.py
+++ b/controller/scheduler/fleet.py
@@ -129,7 +129,7 @@ class FleetHTTPClient(object):
         # prepare memory limit for the container type
         mem = kwargs.get('memory', {}).get(l['c_type'], None)
         if mem:
-            l.update({'memory': '-m {} --memory-swap=-1'.format(mem.lower())})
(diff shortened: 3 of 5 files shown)


mboersma:
[CoreOS 607.0.0](https://coreos.com/releases/) with Docker 1.5.0 landed in the stable channel just now. We are prepping a PR and then we can revisit this.

mboersma:
#3316 was merged, so we're on Docker 1.5.0 now.

bacongobbler:
Code LGTM!

@aledbf would you mind rebasing this, please? :)

bacongobbler:
My only concern is with older CoreOS clusters running deis v1.5. Is that supported or are we suggesting users to upgrade to CoreOS 607 with the latest release>

carmstrong:
> My only concern is with older CoreOS clusters running deis v1.5 which won't have docker 1.5 installed. Is that supported or are we suggesting users to upgrade to CoreOS 607 with the latest deis release?

We'll need to support upgrades. This PR will have to be upgrade-proof, and fail silently on Docker < 1.5.

carmstrong:
#3319 is similarly going to support older CoreOS versions.

aledbf:
@carmstrong ok. 
Check the CoreOS version in the node using the values in `/etc/os-release` and if that is `< 591.0.0` then `docker < 1.5.x`?

carmstrong:
> Check the CoreOS version in the node using the values in /etc/os-release and if that is < 591.0.0 then docker < 1.5.x?

I'm not sure the clean way to do this in the controller yet, unfortunately. I'd really rather not bind-mount that file just for this one use.

aledbf:
>  I'd really rather not bind-mount that file just for this one use.

Me neither, but I don't see an easier way (and this never happened before)

bacongobbler:
IMO I'd prefer the "Heroku way" of configuration through environment variables. How about we set $COREOS_VERSION in deis-controller.service and disable mem swap if the envvar's unset or higher than `607.0.0`? e.g. if I were running the controller locally on boot2docker I won't have COREOS_VERSION set, therefore memswap should be disabled

bacongobbler:
though we shouldn't be reading the envvar directly... Instead we should be reading from `django.settings.COREOS_VERSION` or w/e seems relevant

carmstrong:
> Instead we should be reading from django.settings.COREOS_VERSION or w/e seems relevant

I suppose that's probably the cleanest approach, although I'd prefer getting the Docker version directly rather than inferring it from CoreOS version.

Having the controller know about what the CoreOS version is seems like a leaky abstraction, but I can't see another way.

bacongobbler:
Maybe if fleet actually leaked information about the host through the api, we could figure out the docker version... swarm would have that information, I'd think, but fleet != swarm :sob:

mboersma:
> I'd prefer getting the Docker version directly

I agree--COREOS_VERSION seems indirect. I can't think of a cleaner way than setting a DOCKER_VERSION env var and/or django.setting though.

carmstrong:
Code LGTM pending requirements change to get it passing Jenkins.

mboersma:
This needs the `semantic_version==2.3.1` line added to [docs_requirements.txt](https://github.com/deis/deis/blob/master/docs/docs_requirements.txt) as well so `make -C docs/ test` will pass.

mboersma:
Code LGTM.

bacongobbler:
I think this is as clean as we can get for detecting the cluster's docker version (with Fleet, at least).

Code LGTM!

aledbf:
I just found this https://github.com/docker/docker/pull/10633

aledbf:
This is the error in the tests right now:
```
Mar 22 03:28:56 deis-01 systemd[1]: Started configsample_v9.cmd.1.
Mar 22 03:28:56 deis-01 sh[10200]: docker: invalid size: '-1'. See 'docker run --help'.
```

mboersma:
> The strange thing is that the test suite pass in many places where memory limits are applied

That is confusing. The `invalid size:` message isn't a warning but an error, I would assume. We need to revert this for now and follow up on the test inconsistency.

mboersma:
We can revisit this change now that CoreOS stable has Docker 1.7.1. Maybe after #4581 we can use the `docker-py` library that's already in controller to infer the Docker version.

@aledbf I'd be glad to do that unless you're already on it--let me know.

PR #3215, opened 2015-03-06
feat(makefile): check that generated go binaries are statically linked

aledbf:
Check to avoid problems like https://github.com/deis/deis/pull/3133

### DIFF
--- a/builder/Makefile
+++ b/builder/Makefile
@@ -16,6 +16,7 @@ BINARY_DEST_DIR := image/bin
 build: check-docker
 	for i in $(BINARIES); do \
 		GOOS=linux GOARCH=amd64 CGO_ENABLED=0 godep go build -a -installsuffix cgo -v -ldflags '-s' -o $(BINARY_DEST_DIR)/$$i bin/$$i.go || exit 1; \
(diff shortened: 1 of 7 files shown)


bacongobbler:
I think this should be part of some form of validation check rather than `make build`

bacongobbler:
otherwise LGTM

carmstrong:
> I think this should be part of some form of validation check rather than make build

:+1: That would allow people to bypass it as well, by skipping the makefile target for `check-static`.

aledbf:
Jenkins, test this please.

aledbf:
Jenkins, test this please.

mboersma:
Is the suggestion that there be separate, optional `make check-static` targets? I think it's ok to enforce static go binary checking as part of `make build`.

aledbf:
What I need to do: keep this or add a `check-static` target?

mboersma:
This LGTM as-is. ping @bacongobbler @carmstrong--what changes do you want to see in this PR?

bacongobbler:
I think it's fine. lgtm

Issue #2978, opened 2015-01-28
user-data error in service debug-etcd.service


aledbf:
```
Jan 28 14:39:46 nodo-3 coreos-cloudinit[607]: 2015/01/28 14:39:46 Calling unit command "enable" on "debug-etcd.service"'
Jan 28 14:39:46 nodo-3 coreos-cloudinit[607]: Failed to apply cloud-config: Unsupported systemd command "enable"
```

bacongobbler:
looks like it's missing `command: start`: https://github.com/deis/deis/blob/master/contrib/coreos/user-data.example#L59-L60

aledbf:
Not an issue. Older copy of use-data fault :)

PR #3222, opened 2015-03-07
fix(store): check and sync the clock every hour

aledbf:
Helping @gvilarino in to debug an error running an in-place upgrade I saw a huge difference of time between the nodes.

5 nodes cluster running in Azure:
```
node 1 - step time server xxxxxx offset 11.356867 sec
node 2 - step time server xxxxxx offset 12.610266 sec
node 3 - step time server xxxxxx offset 6.312251 sec
node 4 - step time server xxxxxx offset 14.479774 sec
node 5 - step time server xxxxxx offset 8.707206 sec
````

this timer will run every hour and keep in sync the clocks

### CODE REVIEW

**Comment by gvilarino** [comment]
:+1:

File: contrib/coreos/user-data.example [hunk 1]
```diff
@@ -76,6 +76,28 @@ coreos:
...
+  - name: sync-clock-hourly.service
+    command: enable
+    content: |
+
+      [Unit]
+      Description=sync clock service
+
+      [Service]
+      Type=simple
+      ExecStart=/bin/bash -c "/usr/sbin/ntpdate -b -u 0.pool.ntp.org"
```
johanneswuerbach: I think we should use the coreos ntp server group here https://github.com/coreos/coreos-overlay/blob/471cd901c4aa8987d24db4e392d582fb212c1f52/net-misc/ntp/files/ntp.conf#L2-L5 Is there a way to not specify a server directly and use the ntp config shipped with coreos instead?

aledbf:
@johanneswuerbach you need to specify the server/s

aledbf:
@johanneswuerbach done. Now it reads the servers defined in `/etc/ntp.conf`

gvilarino:
This module/feature could be called Parker Lewis™ :joy:

carmstrong:
I'm really bummed this is necessary - maybe I don't understand why https://github.com/deis/deis/blob/8826c83b28a573334083edc83571488238857924/contrib/coreos/user-data.example#L45-L59 isn't sufficient? I would think that running ntpdate would keep the clock in sync, but maybe it only occurs once at boot.

If we go with the approach in this PR, can we clean up that other logic now?

gvilarino:
@carmstrong I guess it's platform-related. I'm running on Azure and I get >1s offsets after a 1 or 2 hours of not running ntp update

aledbf:
> isn't sufficient?

No because the issue occurs after the boot. At least in azure is not enough.

aledbf:
@carmstrong and this issue with the clock can provoke another issues with fleet for instance of at app level if is using jwt tokens for instance.

aledbf:
> If we go with the approach in this PR, can we clean up that other logic now?

It can be removed but the node could boot with the clock out of synchronization

carmstrong:
OK. Then I'm okay with this as-is. We'll run Jenkins on all the existing PRs once it's happy.

mboersma:
It's odd that there is so much clock drift in the wild, but running `ntpdate` hourly seems like an acceptable fix. Code LGTM.

contrib/coreos/user-data.example
```diff
@@ -76,6 +76,28 @@ coreos:
...
+      Description=sync clock timer
+
+      [Timer]
+      OnCalendar=hourly
+      Unit=sync-clock-hourly.service
+
+      [Install]
+      WantedBy=multi-user.target
+  - name: sync-clock-hourly.service
+    command: enable
```
bacongobbler: I think this has bitten us before: https://github.com/deis/deis/issues/2978

aledbf: @bacongobbler you are right

aledbf:
This issue could be related to https://github.com/coreos/bugs/issues/304

aledbf:
Jenkins, test this please.

PR #3237, opened 2015-03-09
docs(customizing_deis): remove unused registry secret key

bacongobbler:
This key was removed in a recent release of the docker registry so it
is no longer in use.


carmstrong:
If it isn't used, we should probably clean up https://github.com/deis/deis/blob/8826c83b28a573334083edc83571488238857924/registry/bin/boot#L45 as well.

bacongobbler:
good catch, done

carmstrong:
Code LGTM.

mboersma:
Cleaning up the unused registry secret key is good, I always like to see negative LOC. Code LGTM.

Issue #3231, opened 2015-03-07
Docs on using Deis to locally develop dockerfiles?


haf:
I was wondering what the recommended way of using the registry that comes with deis is?

 - What are the security blocks on the registry? Can I push to it from any node without going through the builder?
 - If I can't, then from where can I push to it?
 - In what way is it configured in comparison to [docker-registry](https://github.com/docker/docker-registry) or [this guide to configuring it](https://www.digitalocean.com/community/tutorials/how-to-set-up-a-private-docker-registry-on-ubuntu-14-04)?

ianblenke:
@haf, this might help:

http://docs.deis.io/en/latest/customizing_deis/registry_settings/

The deis-registry's port 5000 is not publicly exposed by default. It stores everything via the deis-store-gateway (ceph radosgw). There is a secretKey, though I'm not sure how this is really used. It is painless to push to it. You do not need to go through the registry. I often push to it directly from the underlying coreos hosts using docker tag followed by a docker push. It is actually running docker-registry:

https://github.com/deis/deis/blob/master/registry/build.sh#L29-L30

The source tree there is very small, mostly shell scripts and template config files, very easy to read.

haf:
@ianblenke thank you! I will study those! Great to know that someone else has a similar requirement.

bacongobbler:
The registry is simply docker-registry with a few patches on top of v0.9.1 for some of the features utlized in `deis pull`. As @ianblenke said, it's not publicly exposed by default.

> There is a secretKey, though I'm not sure how this is really used

That's a key which was used in older versions of the registry: https://github.com/docker/docker-registry/blob/0.6.0/config_sample.yml#L7-L8

It's probably safe to remove from the documentation now, considering that it's no longer in use.

carmstrong:
> It's probably safe to remove from the documentation now, considering that it's no longer in use.

We should probably file this so we don't forget about it.

bacongobbler:
it's ok I filed #3237 :)

haf:
Here's another thing, that I think should go in the docs:

```
sudo coreos-cloudinit --from-file share/deis-vagrant/contrib/coreos/user-data
```

If you want to run cloud-init in your machine, to access docker-tcp.

Also docker-tcp doesn't work out of the box.

Also if you try `systemctl start docker-tcp` you get errors saying that the socket is already bound.

There should be a mention of whether I need to mount a NFS share or if the 'context' of the docker invocation is automatically sent.

[...]

PR #3242, opened 2015-03-09
Closes #3231, docs(using_deis): docker images

haf:
Work in progress

PR #3243, opened 2015-03-09
support local dev ref #3231

haf:


haf:
Goes with PR  https://github.com/deis/deis/pull/3242

PR #3246, opened 2015-03-09
chore(client+controller): update python requests lib to 2.6.0

mboersma:
See http://docs.python-requests.org/en/latest/community/updates/#id1


carmstrong:
Code LGTM.

mboersma:
Apparently there is a conflict between PyInstaller and newer versions of requests: http://stackoverflow.com/questions/28775276/pyinstaller-error-importerror-no-module-named-requests-packages-chardet-sys
We may have to stick with requests 2.5.1 for now.

carmstrong:
@mboersma This may be a valid supporting argument to us sticking with working releases of libraries until new features/fixes require us to upgrade.

mboersma:
CVE-2015-2296 would have been nice to have.

PR #3259, opened 2015-03-10
feat(builder): add ssh key variable to support private repositories

rvaralda:
Include SSH_KEY to builder.sh in slugbuilder to able to build private buildpacks and dependency repositories

To set the key use:

Linux:
$ deis config:set SSH_KEY="`base64 -w 0 <private-key>`"

MacOS:
$ deis config:set SSH_KEY="`base64 <private-key>`"

BREAKING CHANGE: Set SSH_KEY var to builder

### DIFF
--- a/docs/installing_deis/gce.rst
+++ b/docs/installing_deis/gce.rst
@@ -57,7 +57,7 @@ Initialize Compute Engine
 Google Computer Engine won't be available via the command line tools until it is initialized in the
 web console. Navigate to *COMPUTE* -> *COMPUTE ENGINE* -> *VM Instances* in the project console.
 The Compute Engine will take a moment to initialize and then be ready to create resources via
-``gcutil``.
+``gcloud compute``.
 
 
 Cloud Init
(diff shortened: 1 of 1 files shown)


RTodorov:
+1

marcelo-nunes:
+1

klaussilveira:
Desperately needing this: http://stackoverflow.com/questions/28881419/private-repositories-on-deis

bacongobbler:
@rvaralda could you please add documentation around this such that users will know how to use this feature? :+1: from me!

gutomaia:
+1

ericof:
+1

rvaralda:
@bacongobbler There is the doc. :)

builder/image/slugbuilder/builder/build.sh
```diff
@@ -87,6 +87,20 @@ if [[ -n "$BUILDPACK_URL" ]]; then
...
+        mkdir -p ~/.ssh/
+        chmod 700 ~/.ssh/
+
+        echo $SSH_KEY | base64 -d > ~/.ssh/id_rsa
+        chmod 400 ~/.ssh/id_rsa
+
+        echo 'StrictHostKeyChecking=no' > ~/.ssh/config
+        chmod 600 ~/.ssh/config
+
+        chmod 600 ~/.ssh/config
```
mboersma: This `chmod` command got included twice.

rvaralda:
removed, thanks @mboersma

carmstrong:
Awesome. An improvement on this would be to have `deisctl` automatically read from the file and base64 encode it. It should be pretty easy: https://github.com/deis/deis/blob/46661f13da0dd467cb53a91dd41d87b08bbfe7c7/deisctl/config/config.go#L12-L19

docs/using_deis/using-buildpacks.rst
```diff
@@ -27,6 +27,23 @@ Use ``deis create`` to create an application on the :ref:`controller`.
     Creating application... done, created unisex-huntress
     Git remote deis added
 
+Optional: Add SSH KEY to use private repositories
```
carmstrong: Could we add a sentence or two here detailing *why* someone would need this? "To pull code from private repositories, you'll need to provide Deis with a private SSH key which has access. You can do so with deisctl:"

rvaralda: Sure, no problem

rvaralda:
@carmstrong 
It should be put in `deis-client` side, once this is an application config.

I can include a hard coded check for SSH_KEY in `deis-client` config method, b64 it, and let it pass to the controller already b64 coded.

The final command must be this:
`$ deis config:set SSH_KEY=~/.ssh/<private-key>`

What do you think?

deis-admin:

bacongobbler:
I wonder if we should allow both types of behaviour, both filepath and raw keys. For example, if I were in a big organization I could imagine that company policy would be to disallow saving company private keys on your workstation.

To elaborate:

1. We should try to see if SSH_KEY path exists
  a. If it does, load er' up
  b. If it doesn't, assume it's a raw key value
2. base64 encode the key
3. Upload to deis

Does that seem fair?

bacongobbler:
@rvaralda is there a reason for closing this PR?

rvaralda:
No, I did mess with merges here :/

rvaralda:
There is a way to reopen this or I need to open a new PR?

bacongobbler:
Yeah if you re-push to the same branch on your repository (in this case, master) you can re-open using the button below.

rvaralda:
yeah... I did it, but didn't work

bacongobbler:
oaky, just open a new PR :)

rvaralda:
yep, I will open a new and reference this one.

PR #3260, opened 2015-03-10
Run integration tests against EC2 clusters

### LINKED ISSUE #1680: End-to-end acceptance test on EC2
While there is a certain quiet dignity in the ragtag, fugitive fleet of computers that currently powers http://ci.deis.io/, we should move these resources to a hosted facility outside Boulder with predictable bandwidth. I'm going to pilot Rackspace-on-Metal for this.

mboersma:
This forks `test-integration.sh` into `test-integration-ec2.sh` so we can target both providers. EC2 testing ~~seems to be more reliable currently than vagrant and~~ will allow us to scale CI more easily. This has run several times as [test-ec2](https://ci.deis.io/job/test-ec2/).

Closes #1680.

### DIFF
--- a/includes.mk
+++ b/includes.mk
@@ -10,7 +10,10 @@ SELF_DIR := $(dir $(lastword $(MAKEFILE_LIST)))
 DOCKER_HOST = $(shell echo $$DOCKER_HOST)
 REGISTRY = $(shell echo $$DEV_REGISTRY)
 GIT_SHA = $(shell git rev-parse --short HEAD)
-IMAGE_PREFIX := deis/
+
+ifndef IMAGE_PREFIX
+  IMAGE_PREFIX = deis/
+endif
 
 ifndef BUILD_TAG
   BUILD_TAG = git-$(GIT_SHA)
--- a/tests/bin/route53-wildcard.py
+++ b/tests/bin/route53-wildcard.py
@@ -1,0 +1,63 @@
+#!/usr/bin/env python
+"""
+The route53-wildcard utility is used to add and remove wildcard
+DNS entries from an AWS Route53 zonefile.
+
+Usage: route53-wildcard.py <action> <name> <value> [options]
+
+Options:
+
+  --zone=<zone>      name of zone to use [defaults to parsing name]
+  --region=<region>  AWS region to use [default: us-west-2].
+  --ttl=<ttl>        record TTL to use [default: 60].
+
+Examples:
+
+./route53-wildcard.py create mycluster.gabrtv.io deis-elb.blah.amazonaws.com
+./route53-wildcard.py delete mycluster.gabrtv.io deis-elb.blah.amazonaws.com
+"""
+import boto.route53
+import docopt
+import json
+import subprocess
+import time
+
+
+def parse_args():
+    return docopt.docopt(__doc__)
+
+def create_cname(args):
+    conn = boto.route53.connect_to_region(args['--region'])
+    zone = conn.get_zone(args['<zone>'])
+    status = zone.add_cname(args['<name>'], args['<value>'], ttl=args['--ttl'])
+    print("waiting for record to sync: {}".format(status))
+    while status.update() != "INSYNC":
+        time.sleep(2)
+    print(status)
+
+def delete_cname(args, block=False):
+    conn = boto.route53.connect_to_region(args['--region'])
+    zone = conn.get_zone(args['<zone>'])
+    status = zone.delete_cname(args['<name>'])
+    if block:
+        print("waiting for record to sync: {}".format(status))
+        while status.update() != "INSYNC":
+            time.sleep(2)
+        print(status)
+
+if __name__ == '__main__':
+    args = parse_args()
+
+    # calculate zone from provided name
+    zone = '.'.join(args['<name>'].split('.')[-2:])
+    args['<zone>'] = zone
+
+    # add a * to the provided name
+    args['<name>'] = u'\\052.'+unicode(args['<name>'])
+
+    if args['<action>'] == 'create':
+        create_cname(args)
+    elif args['<action>'] == 'delete':
+        delete_cname(args)
+    else:
+        raise NotImplementedError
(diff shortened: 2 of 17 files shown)


carmstrong:
Code LGTM!

carmstrong:
@mboersma Can we ship this? I think it's about time we run PRs on EC2.

mboersma:
@carmstrong it just needs another review and I'll merge it and start testing PRs against EC2. Although preliminary results suggest we see the same fleet-related errors there as we do on vagrant.

includes.mk
```diff
@@ -10,7 +10,10 @@ SELF_DIR := $(dir $(lastword $(MAKEFILE_LIST)))
 DOCKER_HOST = $(shell echo $$DOCKER_HOST)
 REGISTRY = $(shell echo $$DEV_REGISTRY)
 GIT_SHA = $(shell git rev-parse --short HEAD)
-IMAGE_PREFIX := deis/
```
bacongobbler: IIRC you could overwrite this at runtime: `make IMAGE_PREFIX="bacongobbler/" build`

Is there a particular reason we need to wrap it in `ifndef`'s? Just curious

tests/bin/route53-wildcard.py
```diff
@@ -0,0 +1,80 @@
...
+    name = args['<name>']
+    status = zone.add_cname(name, args['<value>'], ttl=args['--ttl'])
+    print("waiting for record to sync: {}".format(status))
+    while status.update() != "INSYNC":
+        time.sleep(2)
+    print(status)
+    print('waiting for wildcard domain to become available...', end='')
+    # AWS docs say it can take up to 30 minutes for route53 changes to happen, although
+    # it seems to be almost immediate.
+    for i in xrange(120):
```
bacongobbler: This should be okay most of the time, but there are some cases where it has run over 2 minutes for myself on namecheap... I'm wondering if we should display a warning if we don't retrieve a connection by the end of it.

tests/bin/route53-wildcard.py
```diff
@@ -0,0 +1,80 @@
...
+        try:
+            random_hostname = str(uuid.uuid4())[:8]
+            if socket.gethostbyname("{}.{}".format(random_hostname, name)):
```
bacongobbler: what happens if the old records are still up and responding such as when we are migrating to another cluster?

There's also a very slim chance of it hitting an existing record with the same name, though that's highly unlikely... ~1 in 40,000.

tests/bin/test-integration-ec2.sh
```diff
@@ -0,0 +1,158 @@
+#!/bin/bash
+#
+# Preps a test environment and runs `make test-integration`
+# against artifacts produced from the current source tree
+#
+
+# fail on any command exiting non-zero
+# set -eo pipefail
+set -eo
```
bacongobbler: forgot a `pipefail here`

tests/bin/test-setup.sh
```diff
@@ -67,8 +67,10 @@ go get -v github.com/tools/godep
 # cleanup any stale example applications
 rm -rf $DEIS_ROOT/tests/example-*
 
-# generate ssh key if it doesn't already exist
+# generate ssh keys if they don't already exist
 test -e ~/.ssh/$DEIS_TEST_AUTH_KEY || ssh-keygen -t rsa -f ~/.ssh/$DEIS_TEST_AUTH_KEY -N ''
+# TODO: parameterize this key required for keys_test.go?
+test -e ~/.ssh/deiskey || ssh-keygen -q -t rsa -f ~/.ssh/deiskey -N '' -C deiskey
```
bacongobbler: Maybe we could generate a key and dump it into `test/fixtures`? We may need to do the same for SSL keys in the future.

bacongobbler: also what makes this one different from $DEIS_TEST_AUTH_KEY?

tests/ps_test.go
```diff
@@ -72,18 +71,18 @@ func psScaleTest(t *testing.T, params *utils.DeisTestConfig, cmd string) {
...
-		"-o", "PasswordAuthentication=no",
-		"core@deis."+params.Domain, "ls")
-	out, err := sshCmd.Output()
-	if err != nil {
-		t.Fatal(err)
-	}
-	if strings.Contains(string(out), ".service") {
-		t.Fatalf("systemd files left on filesystem: \n%s", out)
-	}
+	// // Regression test for https://github.com/deis/deis/pull/1347
```
bacongobbler: Let's just remove this entirely. That's what a VCS is for :)

tests/bin/test-integration-ec2.sh
mboersma: No, `pipefail` won't work because of [this line](https://github.com/mboersma/deis/commit/57ca0c0f5a12fdf629a806ce87414d57956711f3#diff-5b34da9043b2bea3c75b7e5dd653b036R73). I'm open to suggestions.

tests/bin/route53-wildcard.py
mboersma: This will wait a long time as AWS docs suggest, but in practice returns on the first loop since the domain is available after the zone's status goes to `INSYNC`. If the route53 domain somehow isn't actually available, the script that calls this code will print some type of socket error, so I'm not sure that's needed.

tests/bin/test-integration-ec2.sh
bacongobbler: drat. We should remove the `-o` flag then since that's used to specify option flags.

includes.mk
mboersma: > Is there a particular reason we need to wrap it in ifndef's?

I'm not sure, you'd have to ask BDFL.

mboersma: > IIRC you could overwrite this at runtime

Yes, and for our CI setup it's crucial that we set `IMAGE_PREFIX=deisci/` so the go in the correct org.

tests/bin/test-setup.sh
mboersma: Because the keys_test.go has it hard-coded for a specific test evidently. I didn't intend to fix everything about the tests in this PR, this just moved a hidden requirement noticed by @smothiki out into the open.

tests/bin/route53-wildcard.py
mboersma: We would have an odd test failure in that case, but I'm ok with those odds.

bacongobbler:
all comments LGTM.

PR #3264, opened 2015-03-11
fix(router) enforce HTTPS correctly when not behind an elb

### LINKED ISSUE #3245: redirect loop when enforceHTTPS is enabled on 1.4.1
first registration with enforceHTTPS enabled results in redirect loop

wenzowski:
$http_x_forwarded_proto is only set by a load balancer. The rest of us just want to check if `$https = "on"`

closes #3245

### DIFF
--- a/router/image/templates/nginx.conf
+++ b/router/image/templates/nginx.conf
@@ -55,6 +55,12 @@ http {
         ''      close;
     }
 
+    # trust http_x_forwarded_proto headers correctly indicate ssl offloading
+    map $http_x_forwarded_proto $access_scheme {
+      default $http_x_forwarded_proto;
+      ''      $scheme;
+    }
+
     {{ $enforceHTTPS := or .deis_router_enforceHTTPS "false" }}
 
     ## start deis-controller
@@ -92,7 +98,7 @@ http {
         }{{ end }}
 
         {{ if eq $enforceHTTPS "true" }}
-        if ($http_x_forwarded_proto != "https") {
+        if ($access_scheme != "https") {
           rewrite ^(.*)$ https://$host$1 permanent;
         }
         {{ end }}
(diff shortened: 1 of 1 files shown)


wenzowski:
...and yes, I know, [If is Evil](http://wiki.nginx.org/IfIsEvil) but I don't see any other way to do this that doesn't break elb compat

wenzowski:
can change `$redirect` to something else if folks object

wenzowski:
after chatting with nginx folks, `$scheme` seems to be idiom. changing

wenzowski:
this probably the most idiomatic I can make it

carmstrong:
This code LGTM. ping @aledbf too

bacongobbler:
I'm not quite sure what this change entails. Is there some documentation/explanation somewhere that would help better explain these changes?

wenzowski:
the `map` is equivalent to
```
if $http_x_forwarded_proto then
  $access_scheme = $http_x_forwarded_proto
else
  $access_scheme = $scheme
```

where the `HTTP_X_FORWARDED_PROTO` header exists if ssl is offloaded to an elb (your current check) and the `$scheme` builtin otherwise indicates the [connection protocol](http://nginx.org/en/docs/http/ngx_http_core_module.html#var_scheme)...which will be wrong when ssl is offloaded

wenzowski:
From a security perspective, it would be better if `HTTP_X_FORWARDED_PROTO` were ignored unless the appropriate etcd key eg. `sslOffloadingEnabled` were set. You don't really want to trust that header unless you're deployed into an environment where you have to, but since the current version does a fix for that would be a separate issue.

wenzowski:
> Error response from daemon: No such container: deis-store-volume
https://ci.deis.io/job/test-vagrant/2392/console

router/image/templates/nginx.conf
```diff
@@ -92,7 +98,7 @@ http {
         }{{ end }}
 
         {{ if eq $enforceHTTPS "true" }}
-        if ($http_x_forwarded_proto != "https") {
+        if ($access_scheme != "https") {
           rewrite ^(.*)$ https://$host$1 permanent;
```
aledbf: Why not just `return 301 https://$host$request_uri;` instead of the rewrite?

wenzowski: my question as well, which is why I opened #3266

wenzowski:
rebased

mboersma:
Code LGTM.

PR #3269, opened 2015-03-11
fix(cache): expire all keys when out of memory

### LINKED ISSUE #3218: Meta issue: "Tag not found" error when using `deis pull`
This issue is meant to coordinate the discussion of the "Tag not found" error some users see when using `deis pull`. This issue seemed to have two different root causes which manifested in the same way, making it difficult to resolve:

* The deis-registry component sometimes timed out when pulling the layers from the remote registry. This was resolved in https://github.com/deis/docker-registry/pull/10 which shipped in Deis v1.4.0.
* The deis-cache component, which was configured to be used as an LRU cache for registry in #1858, seems to be configured to only use [50MB](https://github.com/deis/deis/blob/8826c83b28a573334083edc83571488238857924/cache/main.go#L23) of memory, which is suspected to be too little to effectively serve as an LRU cache.

Users who are still seeing this issue after upgrading to Deis v1.4.0 are encourgaed to try disabling registry's LRU cache. This is achieved by logging into the host which runs `deis-registry` and:
```console
$ nse deis-registry
$ vi /app/templates/config.yml
$ # comment the lines for cache_lru
$ /app/bin/reload
```

If `deis pull` works reliably after this for some time, it points to the cache as the culprit. We have not been able to confirm this in our testing, so we can't resolve this until we have some confidence in the fix. If cache is indeed the culprit, we should consider bumping its memory (although we'd be choosing an arbitrary value that may also be too small), or removing the registry's LRU cache entirely. We should ideally do some testing to see what effect (if any) the registry LRU cache has on performance.

/cc @johanneswuerbach 

refs #3116
refs #2916

carmstrong:
I haven't had a chance to test this yet, as I am on hotel WiFi. If this doesn't solve the `deis pull` issue in #3218, we'll have to revert the LRU cache entirely, as removing that configuration seems to fix the issue.

closes #3218

### DIFF
--- a/cache/image/redis.conf
+++ b/cache/image/redis.conf
@@ -146,9 +146,9 @@ dbfilename dump.rdb
 #
 # The DB will be written inside this directory, with the filename specified
 # above using the 'dbfilename' configuration directive.
-# 
+#
 # The Append Only File will also be created inside this directory.
-# 
+#
 # Note that you must specify a directory here, not a file name.
 dir ./
 
@@ -250,7 +250,7 @@ slave-priority 100
 #
 # This should stay commented out for backward compatibility and because most
 # people do not need auth (e.g. they run their own servers).
-# 
+#
 # Warning: since Redis is pretty fast an outside user can try up to
 # 150k passwords per second against a good box. This means that you should
 # use a very strong password otherwise it will be very easy to break.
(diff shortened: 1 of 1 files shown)


johanneswuerbach:
Can't the redis container only run out of memory when the instance has no memory left? I think this is definitely not solving the issue and a fix in the registry is required.

bacongobbler:
this did not work for me. I had to completely disable the cache in order for `deis pull` to start working for me. I imagine this whole thing would've been circumvented if we didn't hook into the registry. OTOH, it was necessary or else builds would've been taking < 10 minutes for even the simplest of images :(

bacongobbler:
I'll propose a counter-PR

PR #3272, opened 2015-03-11
fix(controller): allow labels with trailing numbers

### LINKED ISSUE #3270: Domain name validation: w3.domain.com
I tried to add a domain to my app: `deis domains:add w3.domain.com`
and got the following error :
`{u'domain': [u'Hostname does not look valid.']}`
w3.domain.com is a valid domain.

bacongobbler:
fixes #3270 

pinging @mboersma -- do you know if I should be checking that `domain1` is still considered invalid, or is this now a valid domain?

### DIFF
--- a/controller/api/serializers.py
+++ b/controller/api/serializers.py
@@ -251,7 +251,8 @@ class DomainSerializer(ModelSerializer):
         allowed = re.compile("^(?!-)[a-z0-9-]{1,63}(?<!-)$", re.IGNORECASE)
         for label in labels:
             match = allowed.match(label)
-            if not match or '--' in label or label[-1].isdigit() or label.isdigit():
+            if not match or '--' in label or label.isdigit() or \
+               len(labels) == 1 and any(char.isdigit() for char in label):
                 raise serializers.ValidationError('Hostname does not look valid.')
         if models.Domain.objects.filter(domain=value).exists():
             raise serializers.ValidationError(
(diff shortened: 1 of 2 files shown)


controller/api/serializers.py
```diff
@@ -249,9 +249,15 @@ def validate_domain(self, value):
...
+        all_digits = True
         for label in labels:
+            if label.isalpha():
+                all_digits = False
             match = allowed.match(label)
-            if not match or '--' in label or label[-1].isdigit() or label.isdigit():
+            if not match or \
+              '--' in label or \
+              len(label) > 3 and any(char.isdigit() for char in label[1:-1]) or \
+              label.isdigit() and not all_digits:
```
mboersma: I don't think `all_digits` is necessary. [`str.isdigit()`](https://docs.python.org/2/library/stdtypes.html#str.isdigit) does that test already.

mboersma:
>  that domain1 is still considered invalid

I think it's still illegal to have a TLD ending with a digit, but for our purposes it's probably ok to remove that test case.

controller/api/serializers.py
bacongobbler: Ah, it's not going to work anyways. logic like `1.example.com` would work which is an invalid case. I'll just throw away this commit since it's not part of the original issue anyways.

bacongobbler:
I'm in a very not-so-breaking-change mood today. I've amended and fixed up such that it still fails on TLDs like `domain1`.

mboersma:
Code LGTM.

carmstrong:
Code LGTM.

PR #3273, opened 2015-03-11
fix(registry): remove cache from install/start platform

### LINKED ISSUE #3218: Meta issue: "Tag not found" error when using `deis pull`

bacongobbler:
For existing clusters, they'll have to run the following to disable
the cache:

        $ deisctl stop cache
        $ deisctl uninstall cache
        $ deisctl restart registry@1

I opted away from removing the LRU cache -> registry links as we will still want those in the future once we figure out the root cause of this issue. This is just a stopgap until the root issue arises.

closes #3269

### DIFF
--- a/deisctl/cmd/cmd.go
+++ b/deisctl/cmd/cmd.go
@@ -183,17 +183,13 @@ func startDefaultServices(b backend.Backend, wg *sync.WaitGroup, outchan chan st
 	b.Start([]string{"logspout"}, wg, outchan, errchan)
 	wg.Wait()
 
-	// optimization: start all remaining services in the background
-	// cache is a special case because it must start before registry
-	b.Start([]string{"cache"}, &_wg, _outchan, _errchan)
-	wg.Wait()
 	b.Start([]string{
 		"database", "registry@*", "controller", "builder",
 		"publisher", "router@*"},
 		&_wg, _outchan, _errchan)
 
 	outchan <- fmt.Sprintf("Control plane...")
-	b.Start([]string{"cache", "database", "registry@*", "controller"}, wg, outchan, errchan)
+	b.Start([]string{"database", "registry@*", "controller"}, wg, outchan, errchan)
 	wg.Wait()
 	b.Start([]string{"builder"}, wg, outchan, errchan)
 	wg.Wait()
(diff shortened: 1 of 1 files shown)


carmstrong:
@aledbf Says the cache is an immediate remedy, but not a long-term fix. It sounds like registry is doing something wonky.

bacongobbler:
> All fingers point to @bacongobbler's hack

=3

bacongobbler:
> could not find unit: deis-cache.service

missed a spot. Thanks for running a job, @mboersma

mboersma:
It does appear that removing the registry cache improves stability, even if it may not be a definitive fix to the "tag not found" error. This change leaves cache intact but makes installing it optional. Code LGTM.

carmstrong:
Code LGTM (although I'm sad this is necessary), but it shouldn't close #3218 based on @aledbf's comment that it doesn't definitively fix the "Tag not found" issue: https://github.com/deis/deis/issues/3218#issuecomment-78336613

gvilarino:
Hey just to be shure: should I manually uninstall the cache on my existing deis?

PR #3277, opened 2015-03-11
fix(controller): use same API call to determine loadstate

bacongobbler:
sometimes the unit load state and the fleet api disagree due to some discrepancies between the dbus API and fleet. Re-using the same API call to determine the loadState will help better determine the current state of the job.

### DIFF
--- a/controller/scheduler/fleet.py
+++ b/controller/scheduler/fleet.py
@@ -337,16 +337,16 @@ class FleetHTTPClient(object):
             "deactivating": "down",
         }
         try:
-            # NOTE (bacongobbler): this call to ._get_unit() also acts as a pre-emptive check to
+            # NOTE (bacongobbler): this call to ._get_unit() acts as a pre-emptive check to
             # determine if the job no longer exists (will raise a RuntimeError on 404)
-            unit = self._get_unit(name)
+            self._get_unit(name)
             state = self._wait_for_container_state(name)
             activeState = state['systemdActiveState']
             # FIXME (bacongobbler): when fleet loads a job, sometimes it'll automatically start and
             # stop the container, which in our case will return as 'failed', even though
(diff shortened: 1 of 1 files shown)


carmstrong:
Code LGTM, I think.

controller/scheduler/fleet.py
```diff
@@ -337,16 +337,16 @@ def state(self, name):
             "deactivating": "down",
         }
         try:
-            # NOTE (bacongobbler): this call to ._get_unit() also acts as a pre-emptive check to
+            # NOTE (bacongobbler): this call to ._get_unit() acts as a pre-emptive check to
             # determine if the job no longer exists (will raise a RuntimeError on 404)
-            unit = self._get_unit(name)
+            self._get_unit(name)
```
carmstrong: I assume this is doing something other than returning a value which we are no longer using?

mboersma: Looks like calling `_get_unit` will still make the request to the units/ endpoint and raise an error if there was no entry (after retries), but we no longer trust the value it returns for this case.

mboersma:
Code LGTM.

controller/scheduler/fleet.py
bacongobbler: Correct. We have a check in the models [when the model is destroyed](https://github.com/deis/deis/blob/master/controller/api/models.py#L286-L288). Calling fleet.destroy() removes the unit from fleet. Instead of polling for 20 minutes to wait for the request to eventually time out polling for the unit state, `_get_unit()` raises a RuntimeError if the [request fails 3 times](https://github.com/deis/deis/blob/master/controller/scheduler/fleet.py#L58-L65). It's an optimization as noted in the comment :)

carmstrong:
Code LGTM.

PR #3281, opened 2015-03-12
fix(router): allow customization of connection limits

### LINKED ISSUE #3279: 503's from nginx
https://gist.github.com/dylanz/c1c61793fa6b352bf619

We're running at about 80% memory usage across our cluster (CPU usage is low), and everything was humming along fine.  At one point however, performance degraded, and we started getting 503's from nginx when trying to communicate with our cluster.

I found this:  http://serverfault.com/questions/412323/nginx-503-error-in-high-traffic
... which lead me to this:  https://github.com/deis/deis/blob/master/router/image/templates/nginx.conf

I'm assuming we need to customize our worker_connections, which I don't think is a customizable option in the Router yet:  http://docs.deis.io/en/latest/customizing_deis/router_settings/

We're going to try a custom router in the mean time.

If this sounds like I'm moving in the wrong direction, any guesses would be appreciated (above is my best guess =)).  If this ends up being the issue, I'd definitely be game to make this a configurable setting in Router and issue a PR for it.

aledbf:
closes #3279

### DIFF
--- a/router/image/templates/nginx.conf
+++ b/router/image/templates/nginx.conf
@@ -2,11 +2,11 @@
 daemon off;
 
 user www-data;
-worker_processes auto;
+worker_processes {{ or (.deis_router_workerProcesses) "auto" }};
 pid /run/nginx.pid;
 
 events {
-    worker_connections 768;
+    worker_connections {{ or (.deis_router_maxWorkerConnections) "768" }};
     # multi_accept on;
 }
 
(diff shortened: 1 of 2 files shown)


router/image/templates/nginx.conf
```diff
@@ -2,11 +2,11 @@
 daemon off;
 
 user www-data;
-worker_processes auto;
+worker_processes {{ or (.deis_router_workersProcesses) "auto" }};
```
bacongobbler: Could you please change this to `deis_router_workerProcesses` for consistency?

aledbf:
Jenkins, test this please.

aledbf:
Jenkins, test this please.

mboersma:
Tests passed.

carmstrong:
Code LGTM.

mboersma:
This lets users change the per-worker nginx connection limits, which should help with some high-load situations. Code LGTM.

PR #3283, opened 2015-03-12
fix(client): Older dates are incorrectly parsed as today and yesterday.

glogiotatidis:


mboersma:
This replaces #3181. Code LGTM.

bacongobbler:
:shipit:

PR #3286, opened 2015-03-12
feat(deis client): add ssl-verify parameter to deis register command

rvaralda:
I've just copied and pasted from deis login command.

### DIFF
--- a/client/deis.py
+++ b/client/deis.py
@@ -707,8 +707,11 @@ class DeisClient(object):
             provide a password for the new account.
           --email=<email>
             provide an email address.
+          --ssl-verify=false
+            disables SSL certificate verification for API requests
         """
         controller = args['<controller>']
+        ssl_verify = True
         if not urlparse.urlparse(controller).scheme:
             controller = "http://{}".format(controller)
         username = args.get('--username')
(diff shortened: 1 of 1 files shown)


deis-admin:

carmstrong:
I'd really like to see if we can pull this outside of these functions and make this a global flag, so we don't maintain the code in two places. Does that seem feasible?

rvaralda:
@carmstrong where should you think put this?
I checked the deis script and found that the methods is straight called, like auth_register. (correct me if I am wrong)

Do you think to call auth method, include all these global flags, and then call its sub method like auth:register inside it?

carmstrong:
@bacongobbler @mboersma what do you guys think? Maybe we *do* really only need this on login/register? I would have thought that we want to hook in anywhere we're building a request.

bacongobbler:
Once you login, `ssl_verify` is read from ~/.deis/client.json so it is used for every other authenticated request. The only one that's missing are the requests which do not require authentication (in this case, anything under `/v1/auth`). I think it's only necessary on login/register :)

carmstrong:
> We could eventually clean up some of the code in here and in register at a later date (using docopt to specify default behaviour etc.) but this LGTM.

Fair enough. Thanks @rvaralda!

mboersma:
This slipped through with trailing whitespace that broke `make test-style`. Fixed in 0b4b5eef4b8601d4665a8cd7399e32a71ed8cad2, just noted here to justify that commit.

bacongobbler:
Thanks @mboersma

rvaralda:
Thanks guys

PR #3295, opened 2015-03-13
fix(controller): retry querying fleet unit state for about 30 sec

mboersma:
This is an attempt to work around the "Failed to retrieve state - 500 internal [server error](https://ci.deis.io/job/test-ec2/26/consoleFull#21457490765502085f60b2236c09cc9bfe)" that sometimes kills integration tests by retrying the `_get_state` method. The timing uses exponential backoff over 5 attempts:
```console
$ python test.py
elapsed: 4.31537628174e-05
elapsed: 2.00020813942
elapsed: 6.00302195549
elapsed: 14.0057561398
elapsed: 30.0096731186
final: 30.00989604
```


carmstrong:
Code LGTM, although it looks like it didn't pass :(

bacongobbler:
test-ec2 passed. Code LGTM!

mboersma:
I'm not convinced this retry code is needed after we merged #3311. Let's hold off merging this until we see the original error happen again, if we do.

mboersma:
We've had many test runs since #3311 was merged, but none of the failure cases have shown the specific fleet 500 error that motivated this PR. I'm going to close this, but we can obviously try this approach again if it recurs.

PR #3299, opened 2015-03-14
feat(setup-node.sh): remove user intervention setting up new nodes

aledbf:
And remove wget progress output


mboersma:
Nice, this should streamline things the next time we set up a node. Code LGTM.

carmstrong:
Code LGTM.

PR #3300, opened 2015-03-14
feat(cloud-init): temporal fix for ntpd error and clock skew

aledbf:
replaces #3222 

References:
- https://github.com/coreos/bugs/issues/304
- https://github.com/coreos/coreos-overlay/pull/1145

**This should be temporal until the overlay PR is available in CoreOS stable**

### DIFF
--- a/contrib/coreos/user-data.example
+++ b/contrib/coreos/user-data.example
@@ -171,3 +171,8 @@ write_files:
     content: |
       [Coredump]
       Storage=none
+  - path: /etc/systemd/system/ntpd.service.d/debug.conf
+    content: |
+      [Service]
+      ExecStart=
+      ExecStart=/usr/sbin/ntpd -g -n -f /var/lib/ntp/ntp.drift


gvilarino:
:+1:

contrib/coreos/user-data.example
```diff
@@ -159,3 +159,8 @@ write_files:
     content: |
       [Coredump]
       Storage=none
+  - path: /etc/systemd/system/ntpd.service.d/debug.conf
+    content: |
+      [Service]
+      ExecStart=
```
ianblenke: I'm not sure what the implications of an empty ExecStart= line is in systemd.

contrib/coreos/user-data.example
```diff
@@ -159,3 +159,8 @@ write_files:
...
+      ExecStart=/usr/sbin/ntpd -g -n -f /var/lib/ntp/ntp.drift
```
ianblenke: This is only going to work for HVM by default. To work with PV, you'll need to "sysctl -w xen.independent_wallclock=1" in order to allow the PV clock to be settable independently from the underlying Xen host (which is already running NTP, providing the wallclock that the PV sees). PV maps the time to a hypervisor call to ask the underlying host.  HVM, and raw-iron linux clocks, keep track of the time using linux kernel jiffy counters, which can drift wildly when cycles are stolen by other VMs on the same Xen host machine.

contrib/coreos/user-data.example
aledbf: @ianblenke please check the last comment in coreos/bugs#304

aledbf:
ping @carmstrong

carmstrong:
Code LGTM, although I'd like to find a way for it to work with Xen instances too, based on @ianblenke's suggestion. But I'm okay with shipping if we can't find a clean way to detect virtualization type.

aledbf:
> find a clean way to detect virtualization type.

I need a little help to do that. 
If anyone can execute and publish a gist with the output of  `sudo dmesg | grep Xen` and indicate which type of instance is running I can add a check and run the sysctl modification in case if needed

Reference:
http://wiki.xen.org/wiki/Xen_Linux_PV_on_HVM_drivers#Verify_Xen_Project_PVHVM_drivers_are_working_in_the_Linux_HVM_guest_kernel

carmstrong:
@ianblenke Could you help platform-proof the sysctl command?

ianblenke:
This is a nice guide to Xen instance types and detection:

http://www.brendangregg.com/blog/2014-05-07/what-color-is-your-xen.html
http://www.brendangregg.com/blog/2014-05-09/xen-feature-detection.html
https://github.com/brendangregg/Misc/blob/master/xen/xen-features.pl

We have instances of both kinds, so I can compare the two here.

If I `cat /sys/hypervisor/properties/features`, HVM returns 00000705, and PV returns 000000f0.

There is no mention of "HVM" in dmesg, so a `dmesg | grep HVM` will return a non-zero errno, that is until the kernel kmsg buffer rotates the boot stuff out.

Running `dmidecode` will return nothing on PV, but will show a BIOS DMI output on HVM:

    docker run --privileged alpine /bin/sh -c 'apk update && apk add dmidecode && dmidecode'

The suggestion to `grep hvm /sys/hypervisor/properties/capabilities` actually returns the same thing for both PV and HVM, so we can't rely on that.

Additionally, we can rely on the meta-data discovery API:

    curl -Ls http://169.254.169.254/latest/meta-data/profile

This will return 'default-paravirtual' for PV, and 'default-hvm' for HVM.

This works on AWS. I'm pretty sure this, or something very similar to this, will also work with OpenStack, Eucalyptus, as well as CloudStack (I think), as they expose the meta-data API at 169.254.169.254. Pretty sure the other cloud platforms do not, but I've not seen anyone do an official poll of this.

So those are the methods we can use.

bacongobbler:
Hmm, if there's no clean cross-platform solution and it's just a temporal fix, I think we should stick on the side of "hey, here's a general fix but CoreOS stable has an official patch incoming". What do you guys think about that?

johanneswuerbach:
Looks like this one is also related: https://github.com/coreos/coreos-overlay/pull/1142 no time sync in VMs.

aledbf:
> I think we should stick on the side of "hey, here's a general fix but CoreOS stable has an official patch incoming"

That could be an option if the default option were HVM and not PV (https://github.com/deis/deis/blob/master/contrib/ec2/deis.template.json#L77) because with this PR is possible to assume that the clock issue is gone in ec2.

aledbf:
Just as informative this slide is interesting http://www.slideshare.net/brendangregg/performance-tuning-ec2-instances/30

carmstrong:
> That could be an option if the default option were HVM and not PV

We could consider that change, but I think @ianblenke was seeing issues on HVM?

ianblenke:
I codified that AWS EC2 Tuning presentation in this fleet unit right after I watched that presentation a number of weeks ago:

https://github.com/ianblenke/coreos-vagrant-kitchen-sink/blob/master/tested/ec2.cloud-init

We've been running that for quite a while and have had problems both before and after having done so.

ianblenke:
Also, to @carmstrong's question: we were running HVM and seeing this problem. Since then, there have been folks who have presented problems with PV as well (an 8 second drift on one node report in the past couple of days). When a problem is encountered with PV clock drift, my current thinking is that the underlying Xen host is having NTP issues, and PV guests inherit this clock drift through the hypervisor as a result.

Running ntp inside a VM with an independent wallclock is asking for drift, as jiffies get stolen by other bursty bad actors on the same Xen host.

Trusting the underlying Xen host will always have correct time with no drift seems to have been proven wrong at least occasionally as well. The two solutions here are:

1. To enable an independent wallclock on PV and run ntpd (which conflicts with the point above), which we know to be problematic with HVM today.

2. Shutdown/stop the PV instance and start it back up, so that the instance gets moved to another Xen host in the AWS farm that potentially has the correct NTP time.

3. Fiddle with `mon-clock-drift-allowed` with the guidance of the Ceph project and see if we can't allow a little more wiggle room than 0.05 seconds for ntp and stolen jiffies in a VM:

http://ceph.com/docs/master/rados/troubleshooting/troubleshooting-mon/#clock-skews

johanneswuerbach:
@ianblenke systemd decided to re-enable clock sync in VMs https://github.com/coreos/coreos-overlay/pull/1142/files#diff-89420173d5fd9378745941ea580769f2R7 and it made it already into coreos

mboersma:
Code LGTM. This fixes many cases, if not all, and it will probably be a while before the timesyncd patch makes it to stable.

carmstrong:
We're going to merge this as-is, as having some clock sync is better than none, and this seems to have resolved the skew issue for folks like @gvilarino.

We are really waiting on the upstream fix in coreos/coreos-overlay#1142 to make its way to stable (it's currently in 626.0.0 in the alpha channel).

bacongobbler:
I'm okay with :ship:ping something that sorta works for the general case

gvilarino:
Thanks all for putting so much work into this!

PR #3301, opened 2015-03-14
fix(database): add requirements for wal-e build

mboersma:
Builds just began [failing](https://ci.deis.io/job/test-ec2/46/console) with `fatal error: ffi.h: No such file or directory` when building [wal-e](https://github.com/wal-e/wal-e) for deis/database. A python library required by wal-e probably changed its requirements; this fixes it. Also updates wal-e to the [most recent tag](https://github.com/wal-e/wal-e/releases/tag/v0.8c2), which differs from the version Deis was using by [two minor commits](https://github.com/wal-e/wal-e/compare/c16e58a...master).

### DIFF
--- a/database/build.sh
+++ b/database/build.sh
@@ -18,9 +18,12 @@ apt-get update && apt-get install -yq \
                                       daemontools \
                                       file \
                                       gcc \
+                                      g++ \
                                       git \
+                                      libffi-dev \
                                       libxml2-dev \
                                       libxslt1-dev \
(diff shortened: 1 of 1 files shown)


carmstrong:
Code LGTM.

mboersma:
This did pass test-ec2 regression tests, twice. And I think nothing else will build successfully until it is rebased from this.

aledbf:
Code LGTM.

PR #3306, opened 2015-03-16
fix(contrib): open port 443 in ELB security group

bacongobbler:
When following the instructions for installing an SSL cert on EC2, the load balancer will drop the connection because it does not allow connections from port 443 (HTTPS). Enabling port 443 on the load balancer allows users to connect to a SSL-enabled controller.

This was found after working with @gvilarino on his cluster. After opening port 443 on the DeisWebELB security group, he was able to connect.


gvilarino:
Yeah, this happened and I had to add the exception manually to the security group. Quoting an AWS team member:

> Hey Guido,
>
> I was just cruising through the Deis source code, specifically contrib/ec2/deis.template.json, that essentially created the infrastructure your Deis cluster is running on (I’m assuming you followed the Installing Deis on AWS guide on Deis website).
>
>This bit in that file defines the AWS security group your balancer belongs to:
>
>"DeisWebELBSecurityGroup": {
>     "Type": "AWS::EC2::SecurityGroup",
>     "Properties": {
>       "GroupDescription": "Deis Web ELB SecurityGroup",
>       "SecurityGroupIngress": [
>         {"IpProtocol": "tcp", "FromPort": "80", "ToPort": "80", "CidrIp": "0.0.0.0/0"},
>         {"IpProtocol": "tcp", "FromPort": "2222", "ToPort": "2222", "CidrIp": "0.0.0.0/0"}
>       ],
>       "VpcId": { "Ref" : "VPC" }
>     }
>   },
>
>Notice the absence of 443?
>
>The Deis guys were correct in telling you that doing SSL termination on the ELB is a best practice, they may have forgotten to tell you that you would need to add inbound 443 to the ELB’s security group. It’s not that ELB isn’t serving your cert, its that the security group is preventing inbound 443 from the internet to your ELB because you are not explicitly allowing it.  I would really recommend you make this change in your source code too, but you can modify the security group in-place. Find the elastic load balancer in question in the AWS EC2 console, look on the details tab for “Source Security Group” and grab the security group name. Then click on the Security Group section on the left bar and filter by that source security group name. Go to the inbound tab for that security group, click the edit button, and add HTTPS over 443 for 0.0.0.0/0. You should be good to go.

contrib/ec2/deis.template.json
```diff
@@ -298,6 +298,12 @@
             "Protocol": "HTTP"
           },
           {
+            "InstancePort": "443",
+            "InstanceProtocol": "HTTPS",
+            "LoadBalancerPort": "443",
+            "Protocol": "HTTPS"
+          },
```
gvilarino: Maybe the instance ports here should be `80` and `HTTP`? I mean, if you're looking to terminating the HTTPS chain there then back-end routing should be HTTP

contrib/ec2/deis.template.json
```diff
@@ -332,6 +339,7 @@
         "SecurityGroupIngress" : [
           {"IpProtocol": "tcp", "FromPort" : "22",  "ToPort" : "22",  "CidrIp" : { "Ref" : "SSHFrom" }},
           {"IpProtocol": "tcp", "FromPort": "80", "ToPort": "80", "SourceSecurityGroupId": { "Ref": "DeisWebELBSecurityGroup" } },
+          {"IpProtocol": "tcp", "FromPort": "443", "ToPort": "443", "SourceSecurityGroupId": { "Ref": "DeisWebELBSecurityGroup" } },
           {"IpProtocol": "tcp", "FromPort": "2222", "ToPort": "2222", "SourceSecurityGroupId": { "Ref": "DeisWebELBSecurityGroup" } }
```
gvilarino: I didn't add `443` to this secudrity group, just the other one.

contrib/ec2/deis.template.json
bacongobbler: Good catch. I was thinking in the context of installing ssl on the routers via https://github.com/deis/deis/pull/2194 which listens on port 443 but in EC2's case it'd be better to install the certs on the load balancer instead.

gvilarino: Yeah, actually that's [exaclty what the official Deis docs suggest](http://docs.deis.io/en/latest/managing_deis/ssl-endpoints/#installing-the-ssl-certificate)

bacongobbler:
fixed. thanks for the comments, @gvilarino.

carmstrong:
Personally, I'd prefer to have this remain a note rather than enabled in the security group across the board. I know many folks like to lock down their deployments, and if they're not serving HTTPs, they're not going to want this port open.

carmstrong:
At the very least, we should make this disabled by default, and an optional setting to enable it, like we do with: https://github.com/deis/deis/blob/d1cdd57e4df2d692ece05b0c662a507bc3c04913/contrib/ec2/deis.template.json#L128-L137

gvilarino:
I agree with @carmstrong in the concern; the thing is that it wasn't clear for me from the docs. I may have missed it; the AWS UI is quite confusing at times.

bacongobbler:
Good point. Maybe we should make it more clear in the docs on enabling SSL on EC2? Not sure how to spell it out if users are tripping up on this one (this isn't the first time this has occurred)

gvilarino:
I just realized that the issue is mentioned in [the official deis docs](http://docs.deis.io/en/latest/managing_deis/ssl-endpoints/#installing-the-ssl-certificate)

> To enable SSL, you will need to open port 443 on the load balancer and forward it to port 80 on the routers. For EC2, you’ll also need to add port 443 in the security group settings for your load balancer.

The thing is that, at least for me (and maybe other people out there), the usual flow is:
1. Follow deis docs
2. If external link is quoted, follow instructions on that link
3. Go back to deis docs and keep following from that step on.

Now, since the note about enabling the listener on ELB and inbound traffic on the security group is *before* the link to AWS's docs, I missed them completely. Maybe just emphasizing those two things *after* the link to AWS or by providing a bulleted/numbered list in the docs instead of a plain, long paragraph would suffice.

mboersma:
This appears to cause the CloudFormation template to rollback (destroying everything) somehow. The event logs have this:
	The following resource(s) failed to create: [VPCSecurityGroup, PublicRouteTable, DeisWebELB]. . Rollback requested by user.

carmstrong:
> Maybe just emphasizing those two things after the link to AWS or by providing a bulleted/numbered list in the docs instead of a plain, long paragraph would suffice.

I think a doc fix here is warranted rather than changing the template. We could have a note callout specific to EC2 users.

bacongobbler:
Okay. Considering the concerns that @carmstrong has, let's just emphasize the note in the docs more and we should be good.

gvilarino:
:+1:

bacongobbler:
closing. Will re-PR with a doc fix.

PR #3309, opened 2015-03-16
chore(tests): add debug-etcd log output to integration tests

mboersma:
Unexplained failures in the test suite seem to point to etcd communication issues currently. This turns on our [`debug-etcd.service`](https://github.com/deis/deis/blob/master/contrib/coreos/user-data.example#L60) and dumps those logs from each host into the integration test tarball for forensic analysis.


carmstrong:
Code LGTM, although this service was mainly intended for benchmarking etcd performance. I'd suspect that simply dumping the etcd journal would point to the failure here.

mboersma:
> simply dumping the etcd journal would point to the failure

We can try that first, sure. Although that `sed` stuff took me half an hour to figure out. :disappointed_relieved:

mboersma:
Closing in favor of #3310.

PR #3310, opened 2015-03-16
chore(tests): add etcd log output to integration tests

mboersma:
Unexplained failures in the test suite seem to point to etcd communication issues currently. This dumps etcd logs from each host into the integration test tarball for forensic analysis.

PR #3173, opened 2015-03-03
fix(azure): add support for Azure affinity groups

gvilarino:
This PR adds two fixes that will make a 5-vm Deis cluster stable on Windows Azure:

#### Enhancements
1. Support for specifying `affinity-group` as a valid argument to the `azure-coreos-cluster` script.
  * Overall platform startup speed reduced by about 15 minutes (ballpark) for a 5-vm cluster on Azure.
  * Overall platorm stability and inter-fleet latency reduced to an average of 59ms. [See full cluster stats](https://gist.github.com/gvilarino/aa20c427b5e08b959083)
2. Specific settings for `etcd.service` that will prevent the cluster from dying because of reaching the file-open limit
  * Cluster becomes stable and stops throwing "too many open files" error anymore (possibly related to #1240).

#### TODO
* Add setting an `affinity-group` option as a **very recommended** option in the official Deis docs for an Azure setup.

#### Acknowlegements
Enormous thanks to @aledbf , @sedouard and @carmstrong for helping me with the cluster setup and debugging my situation.

Issue #1240, opened 2014-07-02
Too many open files


ddollar:
Left a Deis cluster running on Amazon for a little over a week. When I came back, etcd was thrashing CPU on all hosts and spitting these errors:

```
Jul 02 12:22:32 ip-10-0-3-246 etcd[668]: 2014/07/02 12:22:32 http: Accept error: accept tcp [::]:7001: too many open files; retrying in 1s
Jul 02 12:22:32 ip-10-0-3-246 etcd[668]: 2014/07/02 12:22:32 http: Accept error: accept tcp [::]:4001: too many open files; retrying in 1s
Jul 02 12:22:33 ip-10-0-3-246 etcd[668]: 2014/07/02 12:22:33 http: Accept error: accept tcp [::]:7001: too many open files; retrying in 1s
```

PR #3311, opened 2015-03-16
Use LimitNOFILE for etcd service

mboersma:
Dumping etcd logs showed "too many open files" errors from etcd, which seems related to #1240 and https://github.com/coreos/coreos-overlay/commit/9d4e63bec512eb9ca97844e47afe93be8b2aea1a. This adapts the Azure user-data change from #3173 for the coreos/user-data file, which overrides the etcd service to increase file handles.

Also includes the commit from #3310 since I need to see etcd logs after this change.

### DIFF
--- a/contrib/coreos/user-data.example
+++ b/contrib/coreos/user-data.example
@@ -20,6 +20,18 @@ coreos:
   units:
   - name: etcd.service
     command: start
+    content: |
+      [Unit]
+      Description=etcd
+      [Service]
+      User=etcd
+      PermissionsStartOnly=true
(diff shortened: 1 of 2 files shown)


carmstrong:
Code LGTM.

bacongobbler:
Considering that this is a stopgap until it makes its way into the stable channel, this LGTM.

mboersma:
This passed test-ec2 although it failed test-vagrant (but *not* with "too many open files"). I think it did fix that specific failure, and this change is coming soon from CoreOS upstream anyway, so I'm merging it after discussion with other maintainers.

PR #3312, opened 2015-03-17
Proposal: feat(controller): Add token based registration

Joshua-Anderson:
Right now, deis has two options for user registration, enabled or disabled. I propose adding a third method as a middle ground, token based registration.

An admin can generate a one time use token that they can give to somebody to allow them to register a user. Ideally this token would be emailed to the prospective user, but for now the admin could copy and paste the token into a email or write it down. 

Enabling or disabling this feature would be done using `ectd` via the same methods used before to enable or disable auth.

Example usage:

``` console
$ deis register:token
abc124
$ deis register:token --delete abc123
$ deis register:token --delete-all
$ deis register http://deis.local3.deisapp.com
username: test
email: test@test.com
password: testing
password (confirm): testing
# Deis controller returns a token missing error
token: abc123
```

What do you guys think about this?

### DIFF
--- a/docs/reference/api-v1.1.rst
+++ b/docs/reference/api-v1.1.rst
@@ -48,7 +48,8 @@ Optional Parameters:
 
     {
         "first_name": "test",
-        "last_name": "testerson"
+        "last_name": "testerson",
+        "token" : "abc123"
     }
 
 Example Response:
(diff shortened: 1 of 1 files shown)


carmstrong:
I think this is a bit confusing, as we're already using tokens for API authentication. A token for a token seems a bit odd. 

I agree, though, that we need to solve the issue of resetting a user's password and creating a user without a password.

Joshua-Anderson:
This is for user registration, not passwrod resetting. You don't want to leave your deis user registration open, but you want to let selected people sign up. A admin creates a invitation token that can be used to register an account.

bacongobbler:
I agree that different registration strategies would be nice, so :+1: on the idea.

However, I think the UI should work like so:

```
$ # in this case, you must send your auth token to the initial auth:register or else you'll see...
$ deis auth:register https://deis.example.com --strategy token
403 FORBIDDEN
$ deis auth:login https://deis.example.com --username admin --password admin --confirm admin
$ deis auth:register https://deis.example.com --strategy token
abc123
$ deis logout
deis auth:register https://deis.example.com --strategy token
token: abc123
username: bacongobbler
email: matthewf@fishworks.io
password: supersecurepassword
password (confirm): supersecurepassword
```

Does this seem reasonable?

Joshua-Anderson:
How do you tell the difference between requesting a token and registering? You seemed to use the same command for both.

``` console
$ deis auth:register https://deis.example.com --strategy token
abc123
```

``` console
$ deis auth:register https://deis.example.com --strategy token
token: abc123
username: bacongobbler
email: matthewf@fishworks.io
password: supersecurepassword
password (confirm): supersecurepassword
```

How do you delete or clear old unused tokens?

I think the idea in general is a good one, just wanted to make sure we were clear before creating it :wink:.

bacongobbler:
> How do you tell the difference between requesting a token and registering? You seemed to use the same command for both.

If the request contains an HTTP_AUTHORIZATION header, then the user is requesting a token. Otherwise, someone's trying to register.

> How do you delete or clear old unused tokens?

I'd imagine that there may be some expiry date set to 10-15 minutes in the future from the time of authorizing a new token. Once the token is consumed, the token is destroyed on the server side. :)

Joshua-Anderson:
>I'd imagine that there may be some expiry date set to 10-15 minutes in the future from the time of authorizing a new token. Once the token is consumed, the token is then destroyed on the server side. :)

That would be a good use for redis, but you're only using that with the registry right now. :smile:

>If the expiry date's up, I guess it'd be up to the administrator to drop the row from the db table but they should be unique enough

I think that this should be exposed as a api endpoint to make admins lives easier, but possibly not given a deis command.

> If the request contains an HTTP_AUTHORIZATION header, then the user (in this case, an administrator) is requesting a token. Otherwise, someone's trying to register. I tried to make that clear with the auth:logout that the auth token is cleared from the user's settings but I guess it wasn't totally clear.

I think that it would be confusing to have the two commands so similar. I could see admins accidentally generating tokens.  How about this: If the server has token auth on and somebody tries to register, the server returns a error and specifies that a token is required. Deis would then prompt the user for the token and resend the registration command.

Example:

``` console
$ deis auth:register https://deis.example.com
username: bacongobbler
email: matthewf@fishworks.io
password: supersecurepassword
password (confirm): supersecurepassword
# Deis sees a 403 unauthorized with data containing token required.
# Deis then prompts for token and resubmits the request
token: abc123
```

You would use the token argument to generate a token.

carmstrong:
@Joshua-Anderson We discussed this in our PR review today, and filed #3343 - the root motivation here really seems to be that admin users need a way to create user accounts when self-service registration is disabled. Adding token auth seems to be an unnecessarily complex solution to this goal. 

We'd love to see you tackle #3343 if you're up for it - t should be a fairly straightforward implementation.

Issue #3317, opened 2015-03-17
git push returns 0 on docker error


mboersma:
This devicemapper error just happened, but the test suite kept on going, probably because we didn't propagate an error status back up to `git push`.
```console
Step 4 : CMD while true; do echo "HTTP/1.1 200 OK\n\nPowered by $POWERED_BY" | nc -l -p 1500; done
 ---> Running in bb71d98f49bd
time="2015-03-17T20:32:55Z" level="info" msg="Driver devicemapper failed to create image rootfs e7a54290a5b207228b6f0055d0475303dd5db5c724102c8c039f03fc6f0b8d21: Unknown device 17e66aff3e4691114da63bdf96bde6a5c8d425d7a53302c4e13c8e96d2fcfd3d" 
To ssh://git@deis.local3.deisapp.com:2222/pssample.git
 * [new branch]      master -> master
ok
/Users/matt/Projects/src/github.com/deis/deis/tests
ps:scale cmd=2 --app=pssample
Scaling processes... but first, coffee!
```

carmstrong:
@bacongobbler Didn't we fix this pipe issue?

bacongobbler:
Yes this has been fixed with https://github.com/deis/deis/pull/3457

PR #3319, opened 2015-03-17
chore(*): replace btrfs with overlayfs

### LINKED ISSUE #3294: (builder) Can't set task name /dev/mapper/docker-202:25-3667816-pool
If the builder is restarted on the same node the following (fatal) error is thrown:
``Can't set task name /dev/mapper/docker-202:25-3667816-pool ``

I'm a bit unfamiliar with how the device mapper works. I suspect the error may be occurring because a loopback device is being setup in ExecStartPre but not cleaned up in ExecStopPost.
https://github.com/deis/deis/blob/master/deisctl/units/deis-builder.service#L10

This is likely related:
https://github.com/docker/docker/issues/4639

carmstrong:
This includes the commits from #3316.

We switch from btrfs to ext4 to support overlayfs in Docker. Note that we also format the device with a particular number of inodes. This is what CoreOS does to the root volume to prepare it to run Docker, and since we're running Docker with a separate volume, we must prepare it in the same way. 

Since we're no longer using devmapper in the Docker-in-Docker builder component, this PR also:
closes #3294 
refs #3317

~~TODO: Write wrapper script in builder's /bin/boot to override overlayfs if it's supported.~~

/cc @marineam

### DIFF
--- a/builder/image/bin/boot
+++ b/builder/image/bin/boot
@@ -53,8 +53,15 @@ CONFD_PID=$!
 # remove any pre-existing docker.sock
 test -e /var/run/docker.sock && rm -f /var/run/docker.sock
 
+# force overlay if it's supported
+mkdir --parents --mode=0700 /
+fstype=$(findmnt --noheadings --output FSTYPE --target /)
+if [[ "$fstype" == "overlay" ]]; then
+	DRIVER_OVERRIDE="--storage-driver=overlay"
+fi
+
 # spawn a docker daemon to run builds
-docker -d --bip=172.19.42.1/16 --insecure-registry 10.0.0.0/8 --insecure-registry 172.16.0.0/12 --insecure-registry 192.168.0.0/16 --insecure-registry 100.64.0.0/10 &
+docker -d --bip=172.19.42.1/16 $DRIVER_OVERRIDE --insecure-registry 10.0.0.0/8 --insecure-registry 172.16.0.0/12 --insecure-registry 192.168.0.0/16 --insecure-registry 100.64.0.0/10 &
 DOCKER_PID=$!
 
 # wait for docker to start
--- a/contrib/azure/azure-user-data-template
+++ b/contrib/azure/azure-user-data-template
@@ -11,7 +11,7 @@ coreos:
         Type=oneshot
         RemainAfterExit=yes
         ExecStart=/usr/sbin/wipefs -f /dev/sdc
-        ExecStart=/usr/sbin/mkfs.btrfs -f /dev/sdc
+        ExecStart=/usr/sbin/mkfs.ext4 -i 4096 -b 4096 /dev/sdc
         ExecStart=/bin/touch /etc/azure-formatted
     - name: var-lib-docker.mount
       command: start
@@ -24,4 +24,4 @@ coreos:
         [Mount]
         What=/dev/sdc
         Where=/var/lib/docker
-        Type=btrfs
+        Type=ext4
--- a/contrib/coreos/user-data.example
+++ b/contrib/coreos/user-data.example
@@ -88,6 +88,14 @@ coreos:
       Type=oneshot
       ExecStartPre=/usr/sbin/modprobe nf_conntrack
       ExecStart=/bin/sh -c "sysctl -w net.netfilter.nf_conntrack_max=262144"
+  - name: load-overlay-module.service
+    content: |
+      [Unit]
+      Description=Load overlay module before docker start
+      Before=docker.service
+
+      [Service]
+      ExecStart=/bin/bash -c "modprobe overlay"
 write_files:
   - path: /etc/deis-release
     content: |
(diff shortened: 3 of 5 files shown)


marineam:
Just so you know what you are getting into with switching to ext4/overlay. Yes the btrfs issues are gone but overlay provides odd filesystem semantics which may harm some workloads. A number of our users have been hitting this issue with rpm based containers: https://github.com/docker/docker/issues/10180

carmstrong:
> Just so you know what you are getting into with switching to ext4/overlay.

Ugh, okay. Maybe I'll revert the overlayfs module enabling commit, and instead let Docker continue to choose the best graphdriver.

mboersma:
Do we also need to add back a `--storage-driver=overlay` arg to builder? It's still [marked as experimental](https://github.com/docker/docker/blob/82f390e139e611b97911c19dd1be2a98264cf4b9/daemon/graphdriver/driver.go#L33-L45) post-1.5.0, so needs to be asked for explicitly.

carmstrong:
I spoke with @gabrtv and we want to switch to overlayfs, so we'll put this through the gauntlet. 

> Do we also need to add back a --storage-driver=overlay arg to Docker-in-builder?

Doing that would break upgrades, as folks with old CoreOS versions on btrfs couldn't upgrade to this release. We really want Docker to use overlayfs if the docker graph is ext4 and the overlay kernel module is loaded, otherwise btrfs.

I don't think we want devicemapper.

carmstrong:
https://github.com/coreos/coreos-overlay/blob/master/app-emulation/docker/files/dockerd should take care of this on the CoreOS host, but we'll also need to do this check within the builder component.

mboersma:
>  we'll also need to do this check within the builder component

Yes, unfortunately. (We should really look into how feasible it is to bind-mount the host's docker into deis-builder instead of the Docker-in-Docker approach, someday.)

aledbf:
> We should really look into how feasible it is to bind-mount the host's docker into deis-builder instead of the Docker-in-Docker approach, someday.

Already done in https://github.com/deis/deis/pull/2915
(the thing is that with that you ends using the docker running in the node)

carmstrong:
> (We should really look into how feasible it is to bind-mount the host's docker into deis-builder instead of the Docker-in-Docker approach, someday.)

I think @gabrtv's main objection to that approach is security - anything deployed has access to the host's Docker graph.

carmstrong:
Note that having Jenkins pass this actually tests **both** forcing overlayfs (which is what will happen on EC2) and letting Docker choose its own storage driver (since tests locally likely use aufs).

ibukanov:
I also hit with docker/docker#10180 . Are there any workarounds like disabling overlay module?

carmstrong:
> I also hit with docker/docker#10180 .

Did you hit this with Deis? We haven't merged this PR yet, so our builder component isn't yet using overlay (although the CoreOS host will).

ibukanov:
@carmstrong I meant I hit this bug with coreos.

carmstrong:
We need to write some testing around this so we have some confidence we won't see more errors than we already do on devmapper. 

Ideally we can modify `test-acceptance` to run several integration runs and a stress test on a cluster before we cut a release.

gabrtv:
I move that we merge this now for a few reasons:

* CoreOS upstream has moved to `overlay` in 607, which we are now shipping
* The `overlay` graphdriver is no longer marked experimental in the [Docker docs](https://docs.docker.com/reference/commandline/cli/#daemon-storage-driver-option)
* Besides the issue @marineam pointed out, most anecdotal evidence I've seen points to positive experiences (decreased memory, increased speed)

Happy to see us switch Docker volumes to ext4 as well.  Code LGTM!

aledbf:
Code LGTM.

mboersma:
> The overlay graphdriver is no longer marked experimental in the Docker docs

Although it still is in current master [Docker code](https://github.com/docker/docker/blob/master/daemon/graphdriver/driver.go#L33-L45), but point taken.

I'm ok with testing this in the field before our next release. Code LGTM.

carmstrong:
Okay, tested this builder container on CoreOS 557.2.0, and it still uses btrfs:

```console
core@deis-01 ~ $ docker info
Containers: 10
Images: 59
Storage Driver: btrfs
 Build Version: Btrfs v3.17.1
 Library Version: 101
Execution Driver: native-0.2
Kernel Version: 3.18.1
Operating System: CoreOS 557.2.0
CPUs: 1
Total Memory: 1.959 GiB
Name: deis-01
ID: MB2W:UPO6:METE:ZXAT:LBOK:LK63:WXJ4:MJJH:ZXCH:Y4LA:KDTH:WALD
```

Inside builder:
```console
root@a9ac27333dbf:/app# docker info
Containers: 1
Images: 15
Storage Driver: btrfs
Execution Driver: native-0.2
Kernel Version: 3.18.1
Operating System: Ubuntu 14.04.1 LTS (containerized)
CPUs: 1
Total Memory: 1.959 GiB
Name: a9ac27333dbf
ID: 4BLW:WI2A:4SBF:M3SQ:S7EE:O4YF:3L5W:ORJJ:LWRT:7HDK:D7FU:TMU2
```

Haven't tested an upgrade, but we'll do that as part of the release testing. I'm pretty confident that this is a no-op for existing clusters.

carmstrong:
The only logic we're changing here is forcing overlayfs **if the host uses overlayfs** - otherwise, builder is still choosing its own graph driver based on the same logic as before, so this is safe. Merging...

PR #3325, opened 2015-03-18
Random test fixes

mboersma:
Multiple simultaneous runs of `make -C controller/ test-unit` have collided on our CI server, so this uses Django's [`TEST_NAME`](https://docs.djangoproject.com/en/1.6/topics/testing/overview/#the-test-database) setting to add a psuedo-random suffix to the PostgreSQL database name. Also changes the ec2 test script to use `openssl rand` instead of `/dev/urandom`, which allows us to `set -eo pipefail` again.

### DIFF
--- a/controller/deis/settings.py
+++ b/controller/deis/settings.py
@@ -4,6 +4,8 @@ Django settings for the Deis project.
 
 from __future__ import unicode_literals
 import os.path
+import random
+import string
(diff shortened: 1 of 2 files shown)


carmstrong:
Code LGTM.

bacongobbler:
LGTM!

PR #3339, opened 2015-03-19
chore(controller): update Django to 1.6.11 security release

mboersma:
See https://www.djangoproject.com/weblog/2015/mar/18/security-releases/

It looks unlikely that Deis' controller code was affected by either bug, but we keep up with Django security releases anyway.


carmstrong:
Code LGTM.

aledbf:
Code LGTM.

bacongobbler:
:ship:

PR #3348, opened 2015-03-21
fix(publisher): use TTL in services application etcd directories

### LINKED ISSUE #3337: bug(services) deis apps:destroy doesn't delete root directory
1. Let's say I create app  `mytest`
2. I deploy via `deis pull someImage --app mytest` to it.
3. I run it and use it.
4. I `ssh` into a node and `/opt/bin/deis-debug-logs | grep /deis/services/mytest` lists:
  `/deis/services/mytest`
  `/deis/services/mytest/mytest_v3.cmd.1`
5. Logout and `deis destroy --app mytest --confirm mystest`
6. Wait for a bit
7. `ssh` back into a node and `/opt/bin/deis-debug-logs | grep /deis/services/mytest` lists:
    `/deis/services/mytest`

Am I getteing this wrong or this should be removed as well by `deis destroy`?

I think I'm being unable to create apps with a previously used name because of this (I get `500`s from the deis api)

aledbf:
closes #3337

### DIFF
--- a/publisher/server/publisher.go
+++ b/publisher/server/publisher.go
@@ -98,11 +98,13 @@ func (s *Server) publishContainer(container *docker.APIContainers, ttl time.Dura
 		appName := match[1]
 		appPath := fmt.Sprintf("%s/%s", appName, containerName)
 		keyPath := fmt.Sprintf("/deis/services/%s", appPath)
+		dirPath := fmt.Sprintf("/deis/services/%s", appName)
 		for _, p := range container.Ports {
 			port := strconv.Itoa(int(p.PublicPort))
 			hostAndPort := host + ":" + port
 			if s.IsPublishableApp(containerName) && s.IsPortOpen(hostAndPort) {
 				s.setEtcd(keyPath, hostAndPort, uint64(ttl.Seconds()))
+				s.updateDir(dirPath, uint64(ttl.Seconds()))
 				safeMap.Lock()
 				safeMap.data[container.ID] = appPath
 				safeMap.Unlock()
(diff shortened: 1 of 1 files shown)


publisher/server/publisher.go
```diff
@@ -103,6 +103,8 @@ func (s *Server) publishContainer(container *docker.APIContainers, ttl time.Dura
 			hostAndPort := host + ":" + port
 			if s.IsPublishableApp(containerName) && s.IsPortOpen(hostAndPort) {
 				s.setEtcd(keyPath, hostAndPort, uint64(ttl.Seconds()))
+				dirPath := fmt.Sprintf("/deis/services/%s", appName)
```
bacongobbler: nitpick: let's move this outside of the inner `for` loop on line 101 so we're not allocating memory every iteration

bacongobbler:
do we know for sure that this fixes the issue? Code LGTM but would be good to check with @gvilarino to see if this solves his problem.

mboersma:
I tested this interactively using the steps described in #3337. This works as advertised. LGTM!
```console
core@deis-01 ~ $ etcdctl ls --recursive /deis/services
/deis/services/ttltest
/deis/services/ttltest/ttltest_v2.cmd.1
core@deis-01 ~ $ etcdctl ls --recursive /deis/services  # immediately after destroying the app
core@deis-01 ~ $ 
```

gvilarino:
Sorry for being late: it works!

Grindizer:
Hi everyone. 

This merge was actually reverted here: https://github.com/deis/deis/commit/a11775d
I actually failed to see why exactly, and I wanted to know if someone has an idea about what motivated this revert ? 

Thank you in advance.

PR #3361, opened 2015-03-23
chore(controller): bump confd to v0.8

bacongobbler:


mboersma:
This new warning showed up in the functional test:
```
deis-controller running...
2015-03-23T19:40:18Z 40cafdfbebf1 confd[84]: WARNING Skipping confd config file.
[2015-03-23 19:40:18 +0000] [86] [INFO] Starting gunicorn 19.3.0
```

bacongobbler:
yeah that's due to confd expecting a config file yet not respecting it in recent versions, hence the reliance on flags instead. I'm fixing that up.

bacongobbler:
@mboersma kelseyhightower/confd#235

bacongobbler:
FYI because the check script is gone with this PR, we should no longer be seeing "config check failed: exit status 1" any more. We should see more verbose information... Hopefully.

bacongobbler:
I also filed kelseyhightower/confd#236 to track the `--config-file` issue/workaround.

mboersma:
This does make the templates a bit uglier, but I gather there's no alternative if we want to catch up with newer `confd`, which we do. Code LGTM.

bacongobbler:
Also note kelseyhightower/memkv#4 should help make the "get value or default" calls here a lot cleaner.

aledbf:
Code LGTM.

Issue #3332, opened 2015-03-19
tests: Could not find token


carmstrong:
https://ci.deis.io/job/test-ec2/128/console

```
/var/lib/jenkins/jobs/test-ec2/workspace/src/github.com/deis/deis/tests
Could not find token. Use `deis login` or `deis register` to get started.
error at command wait
--- FAIL: TestBuilds (86.49s)
	builds_test.go:64: exit status 1
```

bacongobbler:
to note for documentation purposes, this seems to be a concurrency issue when running multiple jobs on the same host.

PR #3367, opened 2015-03-24
feat(client): Add multiple profile support to the client

a-saad:
Profile names map to deis client configuration file. The profile
name is picked up from the DEIS_PROFILE environment variable if it exists.
The default is 'client' which points to 'client.json'. To run deis
client under a different profile use: DEIS_PROFILE=dev deis

### DIFF
--- a/client/deis.py
+++ b/client/deis.py
@@ -172,7 +172,8 @@ class Settings(dict):
         # Create the $HOME/.deis dir if it doesn't exist
         if not os.path.isdir(path):
             os.mkdir(path, 0700)
-        self._path = os.path.join(path, 'client.json')
+        filename = '%s.json' % os.environ.get('DEIS_PROFILE', 'client')
+        self._path = os.path.join(path, filename)
         if not os.path.exists(self._path):
             settings = {}
             with open(self._path, 'w') as f:
--- a/docs/using_deis/install-client.rst
+++ b/docs/using_deis/install-client.rst
@@ -78,3 +78,21 @@ To get help on subcommands, use ``deis help [subcommand]``:
 
 .. _pip: http://www.pip-installer.org/en/latest/installing.html
 .. _Python: https://www.python.org/
+
+Multiple Profile Support
+------------------------
+
+The Deis client supports running commands against multiple installations
+and/or accounts by setting the ``$DEIS_PROFILE`` environment variable
+before logging in and running any subsequent commands. If not set, all
+commands will default to the ``client`` profile which maps to
+a configuration file at ``$HOME/.deis/client.json``. Here's an example
+of running the ps command against an app with the same name from two profiles:
+
+
+.. code-block:: console
+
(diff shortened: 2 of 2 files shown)


deis-admin:

mboersma:
This would probably fix #3332 just by setting `DEISPROFILE=${STACK_NAME}`. Code LGTM.

Jenkins, add to whitelist.

bacongobbler:
IMO this case is solved in #3177. https://github.com/fishworks/deis-accounts allows you to switch profiles and plugs right into ~/.deis/client.json for multi-account management. Is there a use case which is missed in the proposed plugin PoC?

mc-team:
This breaks something I attempted to work around by relying on environment variables (and not having to mess up with all commands adding a --profile option.) How do I simultaneously run a series of commands targeting different clusters using the accounts plugin?

a-saad:
@bacongobbler  This breaks something I attempted to work around by relying on environment variables (and not having to mess up with all commands adding a --profile option.) How do I simultaneously run a series of commands targeting different clusters using the accounts plugin?

mboersma:
@bacongobbler #3177 does the `deis-accounts` plugin manage multiple `client.json`-type files as this change would? My POV here is to find a way to keep concurrent integration tests from stomping on each  other. I'm concerned that `deis logout` wouldn't play nice without separate files.

bacongobbler:
@a-saad https://github.com/fishworks/deis-accounts#usage shows how you can switch between different accounts on different clusters via `deis accounts:set <username>`. Essentially it works the same as if you were checking out a separate git branch.

> does the deis-accounts plugin manage multiple client.json-type files as this change would?

The plugin manages only the main client.json which we use today. It probably won't work in a concurrent sense, so I guess that solves one use case

a-saad:
The question was not about how one can switch accounts. The emphasis is on running commands *simultaneously* against different clusters, a use case which I happen to need on a daily basis. It also seems that such a use case is needed for testing Deis against different stacks. 

Also prefixing the command with the profile/account name (if you use the suggested form) makes it clear which cluster is going to be changed by running potentially uptime-sensitive commands.

bacongobbler:
Yes, I misunderstood and amended my comment. I agree that this is a good change.

To cover all of our bases, would you please add documentation around how a user would use $DEISPROFILE in http://docs.deis.io/en/latest/using_deis/install-client/?

a-saad:
@bacongobbler Done. Thank you for the suggestion.

mboersma:
Code LGTM.

bacongobbler:
Small nitpick: I'd change the envvar to DEIS_PROFILE for conistency with other platform envvars like DEIS_APP, DEIS_RELEASE etc.

Barring @mboersma's comments above, code lgtm

a-saad:
@mboersma @bacongobbler Done.

bacongobbler:
Awesome!

Merging as this has gone green. Thanks @a-saad!

PR #3374, opened 2015-03-25
feat(client): add deis config:push

### LINKED ISSUE #1532: Push configuration from existing environment file

bacongobbler:
This new feature reads from a local .env and pushes the environment variables to Deis.

closes #1532

### DIFF
--- a/client/deis.py
+++ b/client/deis.py
@@ -984,6 +984,7 @@ class DeisClient(object):
         config:set         set environment variables for an app
         config:unset       unset environment variables for an app
         config:pull        extract environment variables to .env
+        config:push        set environment variables from .env
 
         Use `deis help [command]` to learn more.
         """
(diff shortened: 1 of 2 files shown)


bacongobbler:
I figured that there are enough users requesting this feature in the main client that this warrants first-party support rather than a plugin. It's also a pretty small feature with **very** few moving parts, so I doubt there's much maintenance work to keep this feature in the core client.

mboersma:
Nice work, with tests. Code LGTM.

gabrtv:
Code LGTM as well.  Happy to see this land.

PR #3389, opened 2015-03-27
feat(tests): use m3.medium instances for ec2 integration tests

mboersma:
This change [modifies `cloudformation.json`](http://docs.deis.io/en/latest/installing_deis/aws/#customize-cloudformation-json) temporarily to use `m3.medium` instances during integration tests. If it works reliably, we can reduce our EC2 charges for [CI](https://ci.deis.io/). :moneybag: 

screen shot 2015-03-27 at 12 53 17 pm

Deis still recommends (and defaults to) `m3.large` instances on EC2.

### DIFF
--- a/tests/bin/test-integration-ec2.sh
+++ b/tests/bin/test-integration-ec2.sh
@@ -67,11 +67,29 @@ if ! aws ec2 describe-key-pairs --key-names "deis" >/dev/null ; then
 fi
 
 make discovery-url
+
+# customize cloudformation.json to use m3.medium instances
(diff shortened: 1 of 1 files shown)


bacongobbler:
Okay, let's keep this as a trial run since it's just test infra. Previous experience when we lowered our vagrant instances down to 2GB per VM doesn't inspire me, but less :moneybag: spent on test infra is nice.

bacongobbler:
that's promising. Code LGTM!

mboersma:
It's passed three times in a row, so I think this will work for tests.

aledbf:
Code LGTM.

PR #3392, opened 2015-03-29
Proposal ref(database): rewrite boot scripts in go

### LINKED ISSUE #2741: bug(database): the backup should not interfere the etcd publication
If the backup time is bigger than the etcd timeout the database becomes unavailable.
A side effect of this is that the service status is ok (active - running) without any flapping.

This is the output from a node with this problem:
```
  PID TTY      STAT   TIME COMMAND
24393 ?        S      0:00 -bash
24501 ?        R+     0:00  \_ ps axf
    1 ?        Ss     0:00 /bin/bash /app/bin/boot
  110 ?        S      0:00 sudo -i -u postgres /usr/lib/postgresql/9.3/bin/postgres -c config-file=/etc/postgresql/9.3/main/postgresql.conf -c listen-addresses=*
  117 ?        S      0:00  \_ /usr/lib/postgresql/9.3/bin/postgres -c config-file=/etc/postgresql/9.3/main/postgresql.conf -c listen-addresses=*
  120 ?        Ss     0:01      \_ postgres: checkpointer process
  121 ?        Ss     0:00      \_ postgres: writer process
  122 ?        Ss     0:00      \_ postgres: wal writer process
  123 ?        Ss     0:00      \_ postgres: autovacuum launcher process
  124 ?        Ss     0:00      \_ postgres: archiver process   last was 000000020000000400000044
  125 ?        Ss     0:00      \_ postgres: stats collector process
  111 ?        Sl     1:03 confd -node 192.168.2.201:4001 -config-file /app/confd.toml
 4867 ?        S      0:00 sudo -u postgres envdir /etc/wal-e.d/env wal-e delete --confirm retain 5
 4868 ?        Sl     0:05  \_ /usr/bin/python /usr/local/bin/wal-e delete --confirm retain 5
```

How is possible to reproduce this? 
Any timeout during the upload of the backup ( `wal-e`) produces this output:
https://gist.github.com/aledbf/43e2c19344f812f7411f
wal-e retry the upload and it waits forever

This behavior avoids the periodic publication of the host/port in etcd
`https://github.com/deis/deis/blob/master/database/bin/boot#L147`

#2701 solves this behavior

aledbf:
closes #2741


database/bindata/bindata.go
```diff
@@ -0,0 +1,265 @@
+package bindata
```
johanneswuerbach: I think this file should be removed. I creates merge conflicts, churn and can be easily regenerated. We went the same way on `herokuish` https://github.com/gliderlabs/herokuish/pull/11

aledbf:
@johanneswuerbach you are right (in fact the file is regenerated during build, line 9 in the Makefile)
I will remove the file if this is accepted to be merged.

Godeps/Godeps.json
```diff
@@ -1,6 +1,6 @@
 {
 	"ImportPath": "github.com/deis/deis",
-	"GoVersion": "go1.4.1",
+	"GoVersion": "go1.4",
```
bacongobbler: can you please bump this back to 1.4.1?

aledbf:
This could the deis-cache boot with this approach https://gist.github.com/aledbf/dc4a483adb063ec76122

aledbf:
This could the deis-router boot with this approach https://gist.github.com/aledbf/8b215c6dd0b660440905

pkg/log/log.go
```diff
@@ -0,0 +1,20 @@
+package log
+
+import (
+	"os"
+
+	"github.com/Sirupsen/logrus"
+)
+
+var Log = logrus.New()
```
bacongobbler: Let's create a `log.New()` function instead to initialize a logger. Specific packages could then tweak their logger without modifying another goroutine's logger (as would be the case by sharing a global variable).

pkg/log/log.go
```diff
@@ -0,0 +1,20 @@
...
+
+func init() {
+	Log.Formatter = new(StdOutFormatter)
+
+	logLevel := os.Getenv("LOG")
```
bacongobbler: let's change this to `LOG_LEVEL` to be more verbose

bacongobbler:
In order to properly test these new packages, you'll also want to add them to https://github.com/deis/deis/blob/ec1f8f632faa49744f03f770838a8238d1fd42f2/Makefile#L10 so the unit/style tests are run against them

database/boot.go
```diff
@@ -0,0 +1,71 @@
+package main
+
+import (
+	"github.com/robfig/cron"
```
bacongobbler: :+1: I should look into this for #2911

pkg/boot/types.go
```diff
@@ -0,0 +1,19 @@
+package boot
+
+import (
+	"net"
+	"time"
+
+	"github.com/coreos/go-etcd/etcd"
+)
+
+// Boot boot struct
```
bacongobbler: this comment's a little confusing. Would you mind elaborating more on how someone would use this structure? Maybe a little more clarification is required on what `Boot` actually is would be good :)

pkg/log/stdout_formatter.go
```diff
@@ -0,0 +1,20 @@
...
+import (
+	"bytes"
+	"fmt"
+	"strings"
+
+	"github.com/Sirupsen/logrus"
+)
+
+// StdOutFormatter struct
+type StdOutFormatter struct {
```
bacongobbler: this is a straight rip of https://github.com/deis/deis/blob/master/router/logger/stdout_formatter.go. If we're shifting this over here, could we please remove the other formatter and update the calls to this one?

pkg/boot/types.go
aledbf: this struct should not be public. I will change this

pkg/net/net.go
```diff
@@ -0,0 +1,16 @@
...
+import (
+	"net"
+	"time"
+)
+
+// WaitForPort wait for successful network connection
+func WaitForPort(proto string, ip string, port string, timeout time.Duration) {
+	for {
+		_, err := net.DialTimeout(proto, ip+":"+port, timeout)
+		if err == nil {
```
bacongobbler: won't the error be non-nil if we hit the timeout, therefore making the timeout useless?

pkg/os/os.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	_, err := bash.Run("main", nil)
+	if err != nil {
+		log.Printf("command finished with error: %v", err)
+		signalChan <- syscall.SIGTERM
+	}
+}
+
+// RunCommand run a command and return. I
```
bacongobbler: There's a trailing "I" here

aledbf:
I will open another PR using go-extpoints to provide a cleaner interface

PR #3305, opened 2015-03-16
feat(builder): add ssh key variable to support private repositories

rvaralda:
Just reopened because a merge mess.

Ref: #3259

PR #3398, opened 2015-03-31
ref(docs): move "private repos" section after "custom buildpacks"

mboersma:
It was suggested that the "Optional: Add SSH key to use private repositories" section detours too much from the basic workflow in the "Using Buildpacks" doc, and that it would fit better after "Custom Buildpacks." This does that as well as shortens some language and adds a basic usage example. ping @gabrtv.

Refs #3305.


bacongobbler:
huuuuuuuuge :+1: from me. +:100:, in fact!

Docs LGTM

gabrtv:
LGTM, as this was my ask originally.

PR #3399, opened 2015-03-31
feat(controller): Allow adminOnly registration

### LINKED ISSUE #3343: Proposal: Admins can create new users
We currently have the ability to [disable user registration](http://docs.deis.io/en/latest/managing_deis/operational_tasks/#disabling-user-registration) entirely, but a common scenario is that admin users want to be able to create users, but to disallow just anyone from creating an account.

The proposal is to have three registration permission modes:
* Users can register themselves (the default)
* No users can register (`registrationEnabled`, which we currently have)
* Only admins can create user accounts (`registrationByAdminOnly` or something similar)

These could even be collapsed into one setting called something like `registrationMode`.

Joshua-Anderson:
Use etcd key `/deis/controller/registrationMode` to manage registration mode, which can now be "enabled", "disabled", or "adminOnly".

If a user is logged into the deis client and tries to register, the token is now sent to the controller so that the controller can check if the current user is a admin and allow them to register if the registration mode is "adminOnly".

I also created a migration from `/deis/controller/registrationEnabled` to `/deis/controller/registrationMode`.

Fixes #3343

### DIFF
--- a/controller/api/authentication.py
+++ b/controller/api/authentication.py
@@ -1,5 +1,6 @@
 from django.contrib.auth.models import AnonymousUser
 from rest_framework import authentication
+from rest_framework.authentication import TokenAuthentication
 
 
 class AnonymousAuthentication(authentication.BaseAuthentication):
@@ -9,3 +10,15 @@ class AnonymousAuthentication(authentication.BaseAuthentication):
         Authenticate the request for anyone!
         """
         return AnonymousUser(), None
+
+
+class AnonymousOrAuthenticatedAuthentication(authentication.BaseAuthentication):
+
+    def authenticate(self, request):
+        """
+        Authenticate the request for anyone or if a valid token is provided, a user.
+        """
+        try:
+            return TokenAuthentication.authenticate(TokenAuthentication(), request)
+        except:
+            return AnonymousUser(), None
--- a/controller/api/permissions.py
+++ b/controller/api/permissions.py
@@ -97,7 +97,22 @@ class HasRegistrationAuth(permissions.BasePermission):
     Checks to see if registration is enabled
     """
     def has_permission(self, request, view):
-        return settings.REGISTRATION_ENABLED
+        """
+        If settings.REGISTRATION_MODE does not exist, such as during a test, return True
+        Return `True` if permission is granted, `False` otherwise.
+        """
+        try:
+            if settings.REGISTRATION_MODE == 'disabled':
+                return False
+            if settings.REGISTRATION_MODE == 'enabled':
+                return True
+            elif settings.REGISTRATION_MODE == 'admin_only':
+                return request.user.is_superuser
+            else:
+                raise Exception("{} is not a valid registation mode"
+                                .format(settings.REGISTRATION_MODE))
+        except AttributeError:
+            return True
 
 
 class HasBuilderAuth(permissions.BasePermission):
--- a/client/deis.py
+++ b/client/deis.py
@@ -756,8 +756,14 @@ Make sure that the Controller URI is correct and the server is running.
             email = raw_input('email: ')
         url = urlparse.urljoin(controller, '/v1/auth/register')
         payload = {'username': username, 'password': password, 'email': email}
+        headers = {}
+
+        token = self._settings.get('token')
+        if token:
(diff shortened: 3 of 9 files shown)


bacongobbler:
> Users who currently have registration disabled would find it enabled when upgrading, as I changed etcd key "registrationEnabled" to "registrationMode" as the key is no longer a bool as we have to support more than two modes of registration.

IMO we should keep it available such that Deis is backwards-compatible, then deprecate it somewhere down the line i.e. if `registrationEnabled` is set, then that should be used to stay compatible then remove it in v1.6. Ye Savvy?

Haven't reviewed this yet, just fly-by comments :)

Joshua-Anderson:
@bacongobbler Could we possibly transparently migrate `registrationEnabled` to `registrationMode` and delete the key, so that everybody is eventually up to date?

I agree there is sense to respecting it, though, but it is hard to tell if the user is using the registration enabled setting. For example, a user used to have registration enabled in a old version of deis, so`registrationEnabled` is set to `False`. Later, on the current version of deis, he decided to renable registration, and sets it to `registraionEnabled` to `enabled`. How can we tell which to respect?

To make matters worse, whether or not anybody is using them, `registrationEnabled` and `registrationMode` both have default values. If the user upgrades, both of them set and it will be really tricky to tell which one the user intentds us to use.

carmstrong:
> Could we possibly transparently migrate registrationEnabled to registrationMode and delete the key, so that everybody is eventually up to date?

I like that approach. It seems like we don't need to bother with deprecating it if we can migrate the setting automatically, right?

Joshua-Anderson:
I added a migration based on @bacongobbler's new etcd migration system, so this should be ready to be reviewed.

controller/api/permissions.py
```diff
@@ -97,7 +97,19 @@ class HasRegistrationAuth(permissions.BasePermission):
     Checks to see if registration is enabled
     """
     def has_permission(self, request, view):
-        return settings.REGISTRATION_ENABLED
+        """
+        If settings.REGISTRATION_MODE does not exist, such as during a test, return True
```
bacongobbler: I think that a user could accidentally write out a typo for the registration mode. In that case we should be raising an argument error. We could use "open_season" to accept all requests.

controller/api/permissions.py
```diff
@@ -97,7 +97,19 @@ class HasRegistrationAuth(permissions.BasePermission):
...
+        Return `True` if permission is granted, `False` otherwise. Defaults to True
+        """
+        try:
+            if settings.REGISTRATION_MODE == 'disabled':
+                return False
+            elif settings.REGISTRATION_MODE == 'adminOnly':
```
bacongobbler: I'd prefer `admin_only` as the type here. It's more pythonic to use underscores rather than camelcase.

controller/api/authentication.py
```diff
@@ -9,3 +10,15 @@ def authenticate(self, request):
...
+
+
+class AnonymousOrAuthenticatedAuthentication(authentication.BaseAuthentication):
+
+    def authenticate(self, request):
+        """
+        Authenticate the request for anyone or if a valid token is provided, a user.
+        """
+        try:
+            return TokenAuthentication.authenticate(TokenAuthentication(), request)
```
bacongobbler: Hmm, this might not work when LDAP auth is enabled. We should support that somehow.

Joshua-Anderson: As far as I can tell, LDAP users are instructed to leave their cluster in `disabled` mode and never register users (The LDAP data given by the LDAP server is translated into a user), so it shouldn't be a problem.

bacongobbler: Yeah I guess it's fine since we're hooking into just Token auth for registration right now but we will hit a scalability problem in the future as we add more pluggable auth strategies.

carmstrong: > as we add more pluggable auth strategies.

Let's cross that bridge when we come to it. :+1:

Joshua-Anderson:
I addressed all comments and the tests are now passing :smile:.

mboersma:
LGTM, code looks good and I tested manually with variations on `deisctl config controller set registrationMode=`.

PR #3403, opened 2015-03-31
fix(controller): read from SSH channel before checking exit status

### LINKED ISSUE #3035: deis run cannot handle large output
I need to get a large generated text file from deis. I do the following

`deis run -- cat foo`

If foo is large enough, in my case is (7468 lines   26438 words  186989 bytes) , then `deis run` hangs and no output gets displayed.

Running `deis logs` afterwards shows the full output of the cat command in the logs.

If I run `deis run -- "cat foo | head -n 20"` everything works OK.

mboersma:
Controller code was checking the SSH channel's exit status *before* reading data from it, which surprisingly worked--up to a certain size in the neighborhood of [2182300 bytes](https://github.com/deis/deis/issues/3035#issuecomment-73433740). The similar issue paramiko/paramiko#448 suggested calling `recv_exit_status()` first was the problem, and changing the call order does fix large file output from `deis run` for the largest test files I've used, which previously failed.

```console
$ deis run -- "cat pg3008.txt | wc -c"
Running `cat pg3008.txt | wc -c`...
2474769
$ deis run -- "cat pg3008.txt"
Running `cat pg3008.txt`...
...
15331. jarg422h.htm#The%20Cuckoo's%20Egg
15332. jarg422h.htm#wizard
15333. jarg422h.htm#wannabee




End of The Project Gutenberg Etext of The New Hacker's Dictionary version
4.2.2

$
```

Closes #3035.

### DIFF
--- a/controller/scheduler/fleet.py
+++ b/controller/scheduler/fleet.py
@@ -268,7 +268,8 @@ class FleetHTTPClient(object):
                 chan.get_pty()
                 out = chan.makefile()
                 chan.exec_command(cmd)
-                rc, output = chan.recv_exit_status(), out.read()
+                output = out.read()
+                rc = chan.recv_exit_status()
                 return rc, output
 
             # wait for container to launch


gabrtv:
Wow, nice find.  Color me impressed.  Code LGTM.

bacongobbler:
that's a heisenbug if I ever saw one. One gold :star: for @mboersma 

code LGTM

PR #3411, opened 2015-04-01
fix(builder): retrieve process types from slugbuilder's Procfile

### LINKED ISSUE #3404: 'Container type {} does not exist in application'.format(container_type))
I have a Node.js app with a `package.json` file and a start script (`node ./server/index.js`). When I `git push deis`, I receive the following error:

```
-----> Launching...
2015/04/01 15:31:32 failed retrieving config from controller: &{%!s(*http.body=&{0xc20800ae10 0xc208058120 0xc20805a420 true {0 0} false}) {%!s(int32=0) %!s(uint32=0)} %!s(bool=false) <nil> %!s(func(error)=0x49c180) %!s(func() error=0x49c110)}
```

Following the advice [given here](https://github.com/deis/deis/issues/3290#issuecomment-78965073) by @aledbf , I checked the logs of deis-controller and am getting this:

```
Apr 01 15:18:54 ip-10-21-1-80.ec2.internal sh[4183]: 'Container type {} does not exist in application'.format(container_type))
Apr 01 15:18:54 ip-10-21-1-80.ec2.internal sh[4183]: EnvironmentError: Container type web does not exist in application
```

How can I fix this? Is this a problem with deis or my app?

bacongobbler:
/bin/release was unofficially deprecated for declaring default process
types a while back, which is why some buildpacks don't supply any
default process types in /bin/release. Instead, the buildpack will
generate a Procfile if none is supplied.

Before: https://gist.github.com/bacongobbler/0a530ebaf08685b20d09
After: https://gist.github.com/bacongobbler/6ae9fc07f2fd7b248953

closes #3404

### DIFF
--- a/builder/image/templates/builder
+++ b/builder/image/templates/builder
@@ -138,7 +138,14 @@ echo
 if [ -f Procfile ]; then
     PROCFILE=$(cat Procfile | /app/bin/yaml2json-procfile)
 elif [ -f $TMP_DIR/slug.tgz ]; then
-    PROCFILE=$(tar --to-stdout -xf $TMP_DIR/slug.tgz ./.release | /app/bin/extract-types)
+    # Sometimes, the buildpack will generate a Procfile instead of populating /bin/release
+    # /bin/release was unofficially deprecated for declaring default process types
+    if tar -tf $TMP_DIR/slug.tgz ./Procfile &> /dev/null;
+    then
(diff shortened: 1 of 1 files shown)


mboersma:
Nice sleuthing. Code LGTM!

smothiki:
Code LGTM.

PR #3408, opened 2015-04-01
fix(database): stub out reload script before starting confd

mboersma:
deis-database startup logs a couple of `ERROR exit status 127` messages. Adding `-debug=true -verbose=true` and removing `--quiet` from `confd` flags showed that we had a chicken-and-egg situation where our check_cmd `/usr/local/bin/reload` script was called before it existed, since it is also templated by `confd`.

This PR installs a no-op shell script to `/usr/local/bin/reload` to make `confd` quieter during initial startup.

```console
--- Run deisci/database:git-5246753 at 192.168.59.103:58203
Error response from daemon: No such container: deis-database-git-5246753
2015-04-01T19:52:55Z c5022545f3b2 confd[49]: WARNING Skipping confd config file.
database: no existing database found.
database: no backups found. Initializing a new database...
2015-04-01T19:52:55Z c5022545f3b2 confd[82]: WARNING Skipping confd config file.
2015-04-01 19:52:55 UTC LOG:  database system was shut down at 2015-03-26 17:25:01 UTC
2015-04-01 19:52:55 UTC LOG:  autovacuum launcher started
2015-04-01 19:52:55 UTC LOG:  database system is ready to accept connections
ALTER ROLE
CREATE ROLE
...
```

Closes #3381.

PR #3412, opened 2015-04-01
Proposal feat(go): refactor boot script using go with go-extpoints

### LINKED ISSUE #2597: Go-based /bin/boot scripts output warnings for "Key already exists"
In the new Go-based /bin/boot binaries, when we make sure etcd directories exist, we don't swallow the error if they do exist. Users see output like this:

```
[WARNING] - 105: Key already exists (/deis/controller) [15]
[WARNING] - 105: Key already exists (/deis/services) [15]
[WARNING] - 105: Key already exists (/deis/domains) [15]
[WARNING] - 105: Key already exists (/deis/builder) [15]
```

We'd really like to swallow these errors, like we do in the bash-based /bin/boot scripts:

```
function etcd_safe_mkdir {
	set +e
	etcdctl --no-sync -C $ETCD mkdir $1 >/dev/null 2>&1
	if [[ $? -ne 0 && $? -ne 4 ]]; then
		echo "etcd_safe_mkdir: an etcd error occurred. aborting..."
		exit 1
	fi
	set -e
}
```

Note here we interpret the exit code of `etcdctl`. It should be fairly easy to do this with the vendored code, as we can check the type of error directly.

### LINKED ISSUE #2701: Rewrite controller, database, registry, store bin/boot scripts in Go
@aledbf Did a great job rewriting some components in Go, but we should move all components to a Go-based /bin/boot script. @gabrtv also prefers this method to calling confd, as we can vendor the code.

### LINKED ISSUE #2889: Proposal: allow custom log level in components
For example in `deis-controller` is not possible to customize the log level. 
In this case the value is hardcoded in `/bin/boot` as `--log-level info` and only using a custom image is possible to change this.

aledbf:
**TODO:**
- [X] incorporate #3408
- [X] remove variable declaration
- [ ] cleanup interface?
- [ ] choose better names?
- [X] custom logging level (using LOG_LEVEL env)
- [ ] terminate execution if more than X confd errors occur in one minute (to avoid the endless loop)
 
**Component refactoring TODO:**
- [x] builder
- [X] cache
- [ ] database
- [ ] pass database restauration test
- [X] controller
- [X] registry
- [x] router

~~-  store - base~~ (no need)

- [X] store - daemon
- [X] store - gateway
- [X] store - metadata
- [X] store - monitor

replaces #3392

closes #2701
closes #2597
closes #2889


aledbf:
@johanneswuerbach right. With this will work :) https://github.com/deis/deis/pull/3415

aledbf:
ping @carmstrong @bacongobbler @mboersma 

This PR need a rebase but before that I need to know if this approach is the "correct" one to remove the boot shell scripts to go and homogenize the start of the deis components using go.

bacongobbler:
From my perspoective, this is the way we'd want to move forward with the boot scripts. I think it's great! There's always going to be nitpicks here and there to get this merged, maybe pulling out some of the changes into a separate PR to make this easier to review but it's a :+1: from me.

mboersma:
Needs a rebase. :recycle: :baseball:

aledbf:
@mboersma rebased and passing ci :)

One small detail: `deis-database` is not using this refactoring because I cannot pass/fix the restauration test, so for now is using the boot.sh shell script. The rest is using this refactoring.

builder/bash/build-slugbuilder-slugrunner.bash
```diff
@@ -0,0 +1,41 @@
+set -eo pipefail
+
+# set debug based on envvar
+[[ $DEBUG ]] && set -x
+
+main() {
+  DOCKER="/usr/bin/docker"
+  DRIVER_OVERRIDE="$(cat /etc/docker.env)"
```
bacongobbler: I believe the best way to override defaults with an initialization script is to source `/etc/default/docker`. It may be better to write out a file with all file flags like so (similar to how the daemon is set up in Ubuntu):

    EXTRA_ARGS="--insecure-registry 10.0.0.0/8 --insecure-registry 172.16.0.0/12 --insecure-registry 192.168.0.0/16 --insecure-registry 100.64.0.0/10"

aledbf: @bacongobbler that file is only to replicate this https://github.com/deis/deis/blob/master/builder/image/bin/boot#L56

aledbf:
I don't want to disable it. Just I can pass the test with the refactoring. 
That's why `deis-database` uses the current boot shell script

pkg/os/os.go
```diff
@@ -0,0 +1,114 @@
...
+import (
+	"crypto/rand"
+	"errors"
+	"fmt"
+	"os"
+	"os/exec"
+	"strings"
+	"syscall"
+
+	Log "github.com/deis/deis/pkg/log"
```
eMxyzptlk: Upper-case package name? Why not name the package something like `logger` and just import it here?

builder/bin/main.go
```diff
@@ -0,0 +1,123 @@
...
+	"io/ioutil"
+	"os/exec"
+	"strconv"
+	"time"
+
+	"github.com/deis/deis/builder/bindata"
+
+	"github.com/deis/deis/pkg/boot"
+	Log "github.com/deis/deis/pkg/log"
+	. "github.com/deis/deis/pkg/os"
```
eMxyzptlk: dot import in Go is not very Go-idiomatic, see [here](https://github.com/golang/go/wiki/CodeReviewComments#import-dot). Maybe name the package `oswrap` and simply import it here?

aledbf:
@eMxyzptlk the use of `Log` is because of a comment from @bacongobbler to not use logger as package, that could be confused with `deis-logger` (the component).
I will update the PR to address your comments

PR #3415, opened 2015-04-02
ref(router): update confd to 0.90

aledbf:


aledbf:
ping @bacongobbler

bacongobbler:
Seems like it needs a rebase, but I'm happy if we can get this into v1.6 (we're in code freeze at the moment)!

aledbf:
@bacongobbler done

carmstrong:
@aledbf Did you have a chance to investigate the failure? It's failed twice on the same test, so I'm thinking there's a problem with the template.

aledbf:
@carmstrong no but I think that is an issue with the flag watch in confd.

aledbf:
@carmstrong ok, my mistake was that I didn't update the template after @bacongobbler added the ssl cert per app.

aledbf:
@carmstrong this passed twice (I don't know why only shows 1)

mboersma:
> (I don't know why only shows 1)

GitHub started displaying different test contexts together, rather than their previous behavior of "last test wins." Somehow we run afoul of that if the very first test run fails--"test-ec2" goes green but "default" never will.

...until I run my status overwrite script. :hear_no_evil: See, I can too write Ruby! There, fixed.

aledbf:
ping @bacongobbler @carmstrong

carmstrong:
Should we disable watch for all components, or is the router one which is particularly busy? I think router was one of the components we were most excited to use watch for... :(

router/image/templates/nginx.conf
```diff
@@ -137,38 +144,38 @@ http {
...
-    upstream {{ Base $service.Key }} {
-        {{ if $affinityArg }}hash $arg_{{ $affinityArg }} consistent;
+    {{ range $app := lsdir "/deis/services" }}
+    {{ $upstreams := printf "/deis/services/%s/*" $app}}
+    upstream {{ $app }} {
+        {{ if exists "/deis/router/affinityArg" }}
+        hash $arg_{{ getv "/deis/router/affinityArg" }} consistent;
         {{ end }}
-        {{ range $upstream := $service.Nodes }}server {{ $upstream.Value }};
+        {{range gets $upstreams }}server {{.Value}};
```
mboersma: This line should add back the whitespace for consistency with the rest of the `confd` template IMHO.

carmstrong:
I think we dropped the `--prefix=/deis` with the removal of confd watch, and we'll have to manually test the commit which fixed #3483 on confd 0.90.

aledbf:
@carmstrong  I will update this one after #3516 lands

aledbf:
includes #3519

carmstrong:
@aledbf Could you rebase again please?

router/image/templates/nginx.conf
```diff
@@ -141,47 +146,44 @@ http {
...
-        server_name {{ Base $domain.Key }};
+        server_name {{ $app_domain }};
         {{/* if a SSL certificate is installed for this domain, use SSL */}}
-        {{/* NOTE (bacongobbler): domains are separate from the default platform domain, */}}
+        {{/* Note (bacongobbler): domains are separate from the default platform domain, */}}
         {{/* so we can't rely on deis.conf as each domain is an island */}}
-        {{/* FIXME (bacongobbler): confd turns hyphens to dashes, so we need to account */}}
-        {{/* for that in domains */}}
-        {{ if index $root (printf "deis_certs_%s_cert" (Replace (Base $domain.Key) "-" "_" -1)) }}
+        {{ range $cert := $certs }}{{ if eq $app_domain $cert }}
```
bacongobbler: this behaviour was reversed with the last rebase (from #3519). Would you mind fixing that up please?

aledbf:
@bacongobbler done

aledbf:
ping @bacongobbler

mboersma:
This looks good. Could you squash some of the commits together? At least the first one with confd 0.8 and the last one with "3574" as the subject.

router/image/templates/nginx.conf
```diff
@@ -15,43 +15,47 @@ http {
     sendfile on;
     tcp_nopush on;
     tcp_nodelay on;
-
     # The Timeout value must be greater than the front facing load balancers timeout value.
     # Default is the deis recommended timeout value for ELB - 1200 seconds + 100s extra.
-    {{ $defaultTimeout := or (.deis_router_defaultTimeout) "1300" }}
```
bacongobbler: we use this value in a few places so we should only have the `if exists` logic in one place as we once did. Makes it easier to make changes in the future :)

router/image/templates/nginx.conf
```diff
@@ -15,43 +15,47 @@ http {
...
-    keepalive_timeout {{ $defaultTimeout }};
+    keepalive_timeout "{{ if exists "/deis/router/defaultTimeout" }}{{ getv "/deis/router/defaultTimeout" }}{{ else }}1300{{ end }}";
```
bacongobbler: does quoting the keepalive timeout make any changes to the value? We should probably leave it unquoted for feature parity

router/image/templates/nginx.conf
aledbf: It doesn't works because the scope of the variable is defined in the if or else section. 
I tried to define the variable outside the if and just change the value but it doesn't work

router/image/templates/nginx.conf
aledbf: ok

router/image/templates/nginx.conf
```diff
@@ -15,43 +15,47 @@ http {
...
+    gzip_min_length "{{ if exists "/deis/router/gzipMinLength" }}{{ getv "/deis/router/gzipMinLength" }}{{ else }}256{{ end }}";
+    gzip_types {{ if exists "/deis/router/gzipTypes" }}{{ getv "/deis/router/gzipTypes" }}{{ else }}application/atom+xml application/javascript application/json application/rss+xml application/vnd.ms-fontobject application/x-font-ttf application/x-web-app-manifest+json application/xhtml+xml application/xml font/opentype image/svg+xml image/x-icon text/css text/plain text/x-component{{ end }};
+    gzip_proxied "{{ if exists "/deis/router/gzipProxied" }}{{ getv "/deis/router/gzipProxied" }}{{ else }}any{{ end }}";
+    gzip_vary {{ if exists "/deis/router/gzipVary" }}{{ getv "/deis/router/gzipVary" }}{{ else }}on{{ end }};
     {{ end }}
 
-    {{ $useFirewall := or .deis_router_firewall_enabled "false" }}{{ if eq $useFirewall "true" }}# include naxsi rules
+    {{ $useFirewall := "false" }}
+    {{ if exists "/deis/router/firewall/enabled" }}
+    {{ $useFirewall := getv "/deis/router/firewall/enabled" }}
```
bacongobbler: Because of how [go templates work with variables](http://golang.org/pkg/text/template/#hdr-Variables), this value will revert to "false" after the `{{ end }}`.

> A variable's scope extends to the "end" action of the control structure ("if", "with", or "range") in which it is declared, or to the end of the template if there is no such control structure. A template invocation does not inherit variables from the point of its invocation.

router/image/templates/nginx.conf
bacongobbler: > (I hate the if exists change)

Me too. Maybe we can make this easier on ourselves by introducing a function like https://github.com/kelseyhightower/memkv/pull/4 into our confd fork, which would make the entire template **much** easier to grok. @mboersma @carmstrong would you be okay with this?

router/image/templates/nginx.conf
bacongobbler: after reading your comment on defaultTimeout, this seems like a potential blocker for confd. I'm thinking that introducing that function would be a good thing.

router/image/templates/nginx.conf
mboersma: > we can make this easier on ourselves by introducing a function like kelseyhightower/memkv#4 into our confd fork

+1. This really needs to be addressed soon upstream, but I have high confidence that useful changes like that will be merged. We really want to make the templates more readable, to lower the bar for contributors. Plus we are already using our own fork, so we have crossed that bridge. :bridge_at_night:

bacongobbler: Something like `get` which returns the key or an empty string if not found. That way we can go `{{ $defaultTimeout := or (get "/deis/router/defaultTimeout") "1300" }}`

router/image/templates/nginx.conf
aledbf: And changing the `$useFirewall` thing I see that I need to change all the template and use the `if exists` in a lot of places
Like this:
```
{{ if exists "/deis/router/firewall/enabled" }}{{ $useFirewall := getv "/deis/router/firewall/enabled" }}{{ if eq $useFirewall "true" }}
location /RequestDenied {
    return {{ if exists "/deis/router/firewall/errorCode" }}{{ getv "/deis/router/firewall/errorCode" }}{{ else }}400{{ end }};
}{{ end }}{{ end}}
```

which is horrible to read and understand

carmstrong:
@aledbf This needs another rebase :crying_cat_face:

aledbf:
@carmstrong rebased

carmstrong:
Code LGTM.

bacongobbler:
code LGTM. :shipi

PR #3418, opened 2015-04-06
ref(tests): move route53-wildcard.py script to contrib/ec2

mboersma:
The `route53-wildcard.py` script @gabrtv wrote is generally useful for setting up EC2 DNS, so this moves it out of tests/bin into contrib/ec2.


carmstrong:
Code LGTM.

bacongobbler:
LGTM. Agreed that this script does deserve a place in contrib/ec2.

PR #3419, opened 2015-04-06
fix(deisctl): omit cache from `deisctl stop platform"

mboersma:
`deisctl start platform` omits deis-cache as of #3273, but `deisctl stop platform` and `deisctl uninstall platform` weren't updated:
```console
$ deisctl stop platform
● ▴ ■
■ ● ▴ Stopping Deis...
▴ ■ ●

Routing mesh...
deis-router@1.service: inactive/dead                                 
deis-router@2.service: inactive/dead                                 
deis-router@3.service: inactive/dead                                 
Data plane...
deis-publisher.service: inactive/dead                                 
Control plane...
deis-builder.service: inactive/dead                                 
deis-controller.service: inactive/dead                                 
deis-database.service: inactive/dead                                 
deis-registry@1.service: inactive/dead                                 
could not find unit: deis-cache.service
```

### DIFF
--- a/deisctl/cmd/cmd.go
+++ b/deisctl/cmd/cmd.go
@@ -510,7 +510,7 @@ func uninstallAllServices(b backend.Backend, wg *sync.WaitGroup, outchan chan st
 	wg.Wait()
 
 	outchan <- fmt.Sprintf("Control plane...")
-	b.Destroy([]string{"controller", "builder", "cache", "database", "registry@*"}, wg, outchan, errchan)
+	b.Destroy([]string{"controller", "builder", "database", "registry@*"}, wg, outchan, errchan)
 	wg.Wait()
 
 	outchan <- fmt.Sprintf("Logging subsystem...")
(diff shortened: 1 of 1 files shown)


carmstrong:
Looks like we still have it in `uninstall` too.

mboersma:
Should have looked there; I'll clean up `deisctl uninstall platform` as well.

carmstrong:
Code LGTM.

bacongobbler:
:+1:

smothiki:
https://github.com/mboersma/deis/blob/release-1.5/registry/conf.d/config.toml#L8
We have to remove cache keys from registry conf.d templates

bacongobbler:
We discussed offline in the original PR: https://github.com/deis/deis/pull/3273#issuecomment-78353257

Basically we want users to still be able to install and start the cache if they *really* want to, but until we figure out the root problem then we just keep it uninstalled for now.

smothiki:
Ok got it. @mboersma is submitting a PR to fix it 
https://github.com/deis/deis/pull/3420/

PR #3422, opened 2015-04-06
chore(contrib/coreos): upgrade fleet to 0.9.2

carmstrong:
This commit introduces a systemd unit to upgrade fleet to 0.9.2
before the daemon is started. This is intended to address
https://github.com/coreos/fleet/issues/1158 which was a regression
from fixes introduced in fleet 0.9.1, which is included for the first
time in CoreOS 607.0.0.

PR #3426, opened 2015-04-07
fix(deisctl): add systemd reload hack as workaround

bacongobbler:
By installing this temporary hack, this should help prevent the
"no such file or directory" bug in fleet v0.9.1 without being too
invasive to the filesystem.

replaces #3422

### DIFF
--- a/deisctl/cmd/cmd.go
+++ b/deisctl/cmd/cmd.go
@@ -162,6 +162,9 @@ func startDefaultServices(b backend.Backend, wg *sync.WaitGroup, outchan chan st
 	var _wg sync.WaitGroup
 
 	// wait for groups to come up
+	outchan <- fmt.Sprintf("Workaround for Fleet v0.9.1...")
+	b.Start([]string{"reload-hack"}, wg, outchan, errchan)
+	wg.Wait()
 	outchan <- fmt.Sprintf("Storage subsystem...")
 	b.Start([]string{"store-monitor"}, wg, outchan, errchan)
 	wg.Wait()
@@ -283,6 +286,10 @@ func stopDefaultServices(b backend.Backend, wg *sync.WaitGroup, outchan chan str
 	b.Stop([]string{"store-daemon"}, wg, outchan, errchan)
 	wg.Wait()
 	b.Stop([]string{"store-monitor"}, wg, outchan, errchan)
+	wg.Wait()
+
+	outchan <- fmt.Sprintf("Workaround for Fleet v0.9.1...")
+	b.Stop([]string{"reload-hack"}, wg, outchan, errchan)
 	wg.Wait()
 }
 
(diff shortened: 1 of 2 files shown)


bacongobbler:
credit where credit is due: @sukrit007 https://gist.github.com/sukrit007/574bfb21fea7bdf1711f

carmstrong:
@bacongobbler See the discussion in #3422 - I think upgrading fleet is the way to go here.

carmstrong:
Ensuring that fleet works smoothly is something that should occur at the system level, and should be transparent to users. 

The fix should:
* Be transparent to users
* Require minimal changes

I don't think we should be adding support to deisctl for this - it's out of the realm of Deis entirely.

bacongobbler:
To address your concerns:

> - be transparent to users

IMO this change is a lot more transparent on what's going on, rather than trusting a compiled binary blob from OpDemand. If you meant about making the installation "invisible" then yes the other is more transparent, though there are ways we can fix that (remove display output, remove entirely from `deisctl list`, etc.). There's one problem with the alternative proposal...

> - require minimal changes

How will existing clusters get #3422's new unit file when upgrading? Any existing clusters will have to manually write out the conf files as well as modify `fleet.service`  on each host. That is a huge maintenance pain. This proposal requires significantly less maintenance work, in which it's just a simple hack that addresses the issue until fleet v0.9.2 rolls into stable. Once we've upgraded, we can simply remove the unit.

> I don't think we should be adding support to deisctl for this - it's out of the realm of Deis entirely.

It's not necessarily using `deisctl` moreso than it's using fleetctl to install a fix for fleet. I couldn't think of a better alternative which would upgrade both existing and new v1.5.0 clusters. We could silence its output and remove it entirely from `deisctl list` if you'd prefer, similar to deployed applications.

carmstrong:
> How will existing clusters get #3422's new unit file when upgrading?

I'm not too worried about that, since we disable CoreOS upgrades by default. Enabling CoreOS upgrades requires manual work, at which time users can be instructed to manually use the unit. Users can run Deis v1.5.0 on old CoreOS versions with no issues.

This is all, of course, temporary. As soon as a CoreOS stable release lands with fleet 0.9.2, this is no longer necessary, so upgrades will work without any changes.

> It's not necessarily using deisctl moreso than it's using fleetctl to install a fix for fleet.

Yes, but this proposal requires logic changes to `deisctl`, which is used to control the Deis platform. The fix has nothing to do with the platform itself. 

Additionally, this workaround is contributed by the community, whereas fleet 0.9.2 is supported by the fleet team. We don't know if this workaround really solves all of the possible failure cases.

carmstrong:
We discussed this internally, and are testing #3422 mainly for the reasons stated above. We're also confirming that fleet 0.9.2 resolves the issues with our stress testing, and exploring the possibility of a CoreOS stable patch release with fleet 0.9.2.

Issue #3429, opened 2015-04-08
Registry can't get BUCKET_NAME from etcd


Blystad:
This causes issues setting up registry on Amazon S3 instead of Ceph

Ref #2252 / #2812

krancour:
@Blystad are you certain?

This looks to me like it does source the bucket name from etcd:

https://github.com/deis/deis/blob/master/registry/templates/config.yml

Am I missing something?

Blystad:
It should, but it doesn't in the `boot` script. See: [https://github.com/deis/deis/blob/95746aadc8e2c77771174f2702dae6fc48951c4d/registry/bin/boot#L22,L54](https://github.com/deis/deis/blob/95746aadc8e2c77771174f2702dae6fc48951c4d/registry/bin/boot#L22,L54)

(Really just line 22, 45 and 54 which is interesting, but Github wouldn't let me mark those).

Blystad:
This cause issues with `create_bucket` which attempts to create the S3 bucket if it doesn't exist.

krancour:
Yup.  This is an issue alright.  I had looked at the boot script and really couldn't spot the problem in it.  _Now_ I see it... but only after banging my head against the wall for a while in #3431.  D'oh.  Sorry I doubted you, @Blystad.

I'd be happy to try and fix this issue, but I'd need a little guidance.  I'm not clear on why there's an environment variable involved here at all.  Any thoughts?  If I try to fix this, I want to get it right.

Ping @mboersma.  Any input?

mboersma:
I'm not entirely sure why `BUCKET_NAME` is there for registry, if that's the env var you mean. It doesn't appear to be referenced elsewhere--maybe it was copy-pasted from similar code in database?

I think it would be safe to use the `BUCKET_NAME` env var if it's set, otherwise fetch the bucket name from etcd, otherwise default to "registry".

krancour:
IMO, it's not worth even messing with a `BUCKET_NAME ` environment variable at all.  For someone to tune the value of an environment requires them to hack on a unit file and that just seems like a PIA compared to just setting the `bucketName` via deisctl.

How about this:

```
etcd_set_default bucketName registry
```

Then later in the boot script:

```
/app/bin/create_bucket $(etcdctl --no-sync -C ${ETCD} get ${ETCD_PATH}/registry)
```

Does that seem about right?  I did try this and I got a variety of errors, but I _think_ they are unrelated.  Trying it again now to be certain.

[...]

PR #3431, opened 2015-04-08
fix(database+registry): check for S3 bucket name before creating it

### LINKED ISSUE #3430: registry create_bucket - bucket exist check seems non-functional
This check doesn't seem to work as expected:

```
if not bucket in conn.get_all_buckets():
  conn.create_bucket(bucket)
```
[registry/templates/create_bucket#L15-L16](https://github.com/deis/deis/blob/95746aadc8e2c77771174f2702dae6fc48951c4d/registry/templates/create_bucket#L15-L16)

`bucket` is a String where as the items in `conn.get_all_buckets()` is of type `boto.s3.bucket`.

I changed it to this:
```
for b in conn.get_all_buckets():
  if b.name == bucket:
    print bucket + ' exists'
    sys.exit(0)

conn.create_bucket(bucket)
```

which seems to maybe work. Going to test it now.

mboersma:
The previous test was always returning `False` since it compared a string to a `boto.s3.bucket.Bucket` instance. So before this change we always tried to create the bucket and presumably hit errors if it already existed. 

I tested this change interactively in `ipython` and it fixes the comparison. I also ran `flake8` on the template scripts and cleaned them up. Thanks to @Blystad for catching this bug!

Closes #3430.

### DIFF
--- a/database/templates/create_bucket
+++ b/database/templates/create_bucket
@@ -1,14 +1,16 @@
 #!/usr/bin/env python
+
 import boto
 import sys
 from boto.s3.connection import OrdinaryCallingFormat
 
-conn = boto.connect_s3(host='{{ getv "/deis/store/gateway/host" }}',
-                       port={{ getv "/deis/store/gateway/port" }},
-                       is_secure=False,
-                       calling_format=OrdinaryCallingFormat(),
-                      )
-bucket = sys.argv[1]
 
-if not bucket in conn.get_all_buckets():
-  conn.create_bucket(bucket)
+conn = boto.connect_s3(
+    host='{{ getv "/deis/store/gateway/host" }}',
+    port={{ getv "/deis/store/gateway/port" }},
+    is_secure=False,
+    calling_format=OrdinaryCallingFormat())
+name = sys.argv[1]
+
+if name not in (bucket.name for bucket in conn.get_all_buckets()):
+    conn.create_bucket(name)
--- a/registry/templates/create_bucket
+++ b/registry/templates/create_bucket
@@ -1,16 +1,17 @@
 #!/usr/bin/env python
+
 import boto
 import sys
 from boto.s3.connection import OrdinaryCallingFormat
 
-conn = boto.connect_s3(aws_access_key_id='{{ getv "/deis/store/gateway/accessKey" }}',
-                       aws_secret_access_key='{{ getv "/deis/store/gateway/secretKey" }}',
-                       host='{{ getv "/deis/store/gateway/host" }}',
-                       port={{ getv "/deis/store/gateway/port" }},
-                       is_secure=False,
-                       calling_format=OrdinaryCallingFormat(),
-                      )
-bucket = sys.argv[1]
+conn = boto.connect_s3(
+    aws_access_key_id='{{ getv "/deis/store/gateway/accessKey" }}',
+    aws_secret_access_key='{{ getv "/deis/store/gateway/secretKey" }}',
+    host='{{ getv "/deis/store/gateway/host" }}',
+    port={{ getv "/deis/store/gateway/port" }},
+    is_secure=False,
+    calling_format=OrdinaryCallingFormat())
+name = sys.argv[1]
 
-if not bucket in conn.get_all_buckets():
-  conn.create_bucket(bucket)
+if name not in (bucket.name for bucket in conn.get_all_buckets()):
+    conn.create_bucket(name)


bacongobbler:
code LGTM

carmstrong:
Code LGTM.

krancour:
I built a registry image from this branch, and it appears I still run into the original issue:

```
Apr 08 19:00:38 ip-10-0-3-235.us-west-1.compute.internal systemd[1]: Started deis-registry.
Apr 08 19:00:51 ip-10-0-3-235.us-west-1.compute.internal sh[18170]: registry: waiting for confd to write initial templates...
Apr 08 19:00:56 ip-10-0-3-235.us-west-1.compute.internal sh[18170]: Traceback (most recent call last):
Apr 08 19:00:56 ip-10-0-3-235.us-west-1.compute.internal sh[18170]: File "/app/bin/create_bucket", line 17, in <module>
Apr 08 19:00:56 ip-10-0-3-235.us-west-1.compute.internal sh[18170]: conn.create_bucket(name)
Apr 08 19:00:56 ip-10-0-3-235.us-west-1.compute.internal sh[18170]: File "/usr/local/lib/python2.7/dist-packages/boto/s3/connection.py", line 621, in create_bucket
Apr 08 19:00:56 ip-10-0-3-235.us-west-1.compute.internal sh[18170]: response.status, response.reason, body)
Apr 08 19:00:56 ip-10-0-3-235.us-west-1.compute.internal sh[18170]: boto.exception.S3ResponseError: S3ResponseError: 403 Forbidden
Apr 08 19:00:56 ip-10-0-3-235.us-west-1.compute.internal sh[18170]: <?xml version="1.0" encoding="UTF-8"?>
Apr 08 19:00:56 ip-10-0-3-235.us-west-1.compute.internal sh[18170]: <Error><Code>AccessDenied</Code><Message>Access Denied</Message><RequestId>B2B7321500872B1F</RequestId><HostId>TnxtQ2IrJZE9dJSRZQVGh+S8vWvQgL9krfeFrNlxg8F/cJk52zpyBdnSVI5qkGSV</HostId></Error>
Apr 08 19:00:57 ip-10-0-3-235.us-west-1.compute.internal systemd[1]: deis-registry@1.service: main process exited, code=exited, status=1/FAILURE
```

Blystad:
@krancour Try my code in #3430 and see if that works

mboersma:
@krancour when I try to create an AWS S3 bucket that already exists, I get a 409 error, not a 403:
```python
in [10]: conn.create_bucket('mboersma-test', location='us-west-2')
---------------------------------------------------------------------------
S3CreateError                             Traceback (most recent call last)
<ipython-input-10-897b1d2205ad> in <module>()
----> 1 conn.create_bucket('mboersma-test', location='us-west-2')

/Users/matt/Projects/src/github.com/deis/deis/venv/lib/python2.7/site-packages/boto/s3/connection.pyc in create_bucket(self, bucket_name, headers, location, policy)
    614         if response.status == 409:
    615             raise self.provider.storage_create_error(
--> 616                 response.status, response.reason, body)
    617         if response.status == 200:
    618             return self.bucket_class(self, bucket_name)

S3CreateError: S3CreateError: 409 Conflict
<?xml version="1.0" encoding="UTF-8"?>
<Error><Code>BucketAlreadyOwnedByYou</Code><Message>Your previous request to create the named bucket succeeded and you already own it.</Message><BucketName>mboersma-test</BucketName><RequestId>36797942C013AFC6</RequestId><HostId>yx++8QHVl1DxrLxP2JykfPsRlUvCmZYJi8IEjbjZtwiExiN9UogYRol/9H+KgvqYbCKA5yB2lcE=</HostId></Error>
```
I think this PR prevents Deis from calling `conn.create_bucket()` when the bucket already exists (the code snippet in #3430 is functionally equivalent to this). I'm not sure that's the problem you're seeing though.

Blystad:
He could have created a IAM user without the permission to create buckets. I did that.

Use `nse` to enter the docker container and verify that all parameters are set as expected.

carmstrong:
@krancour As others have said, I'm pretty sure your error is due to using an incorrect access key, or an access key for a user without permissions to create a bucket.

krancour:
@Blystad hit the nail on the head... I'm using an IAM user that is not authorized to create buckets.  That is why the bucket create is failing with a 403 instead of 409.  So that 403 would be avoided if this weren't still, as far as I can determine, still attempting to create a bucket that already exists.

Blystad:
Actually, you make it sound like a user error but it could and probably is - intentional.

Here is the use case: Use s3 as backing data store I will create a S3 bucket and a IAM user with full permissions to this one bucket, and list all buckets.

I would then not expect deis-registry to create a bucket that already exista

krancour:
Yes... 100% intentional.

mboersma:
> if this weren't still, as far as I can determine, still attempting to create a bucket that already exists

I wonder if the restricted IAM user isn't able to list the existing buckets either? That would explain why the code is apparently trying to create an existing bucket for you--it didn't think it existed.

krancour:
@mboersma that's a good thought.  I will double check, but I think I am already in the habit of giving all my IAM users (or at least those used for S3 access) permission to list all my buckets.

krancour:
@mboersma just double-checked... IAM user has the following policy, which should be adequate:

```
{
    "Statement": [
        {
            "Action": "s3:ListAllMyBuckets",
            "Effect": "Allow",
            "Resource": "arn:aws:s3:::*"
        },
        {
            "Action": "s3:*",
            "Effect": "Allow",
            "Resource": [
                "arn:aws:s3:::<my bucket name>",
                "arn:aws:s3:::<my bucket name>/*"
            ]
        }
    ]
}
```

Blystad:
That IAM policy looks good to me, it's equals to what I use.

Blystad:
Use `nse` to enter the registry (SSH to server and then run `nse deis-registry` and `cat/app/bin/create_bucket` (exists after confd had created it).

Check that host is set to your correct regional URL. I can't recall if s3 bucket names are global or regional, but could be worth a check.

krancour:
Bucket names are global.  Here's what I see:

```
#!/usr/bin/env python

import boto
import sys
from boto.s3.connection import OrdinaryCallingFormat

conn = boto.connect_s3(
    aws_access_key_id='<redacted>',
    aws_secret_access_key='<redacted>',
    host='s3.amazonaws.com',
    port=80,
    is_secure=False,
    calling_format=OrdinaryCallingFormat())
name = sys.argv[1]

if name not in (bucket.name for bucket in conn.get_all_buckets()):
    conn.create_bucket(name)
```

mboersma:
That script looks correct, as do your IAM permissions. Hmm. You can test it in the container by calling it as `python /app/bin/create_bucket [bucket_name]`. Maybe add a line to see what buckets it does see?
```python
print [bucket.name for bucket in conn.get_all_buckets()]
```

krancour:
```
root@472e45e18b0e:/app# python /app/bin/create_bucket my-krancour-bucket
```

Output: 403 as expected, given that this IAM user can not create buckets:

```
  File "/app/bin/create_bucket", line 17, in <module>
    conn.create_bucket(name)
  File "/usr/local/lib/python2.7/dist-packages/boto/s3/connection.py", line 621, in create_bucket
    response.status, response.reason, body)
boto.exception.S3ResponseError: S3ResponseError: 403 Forbidden
<?xml version="1.0" encoding="UTF-8"?>
<Error><Code>AccessDenied</Code><Message>Access Denied</Message><RequestId>59E7617FDA0F665E</RequestId><HostId>oeOWVV2dQR0RNckEhCQZb47uXJgOcZCRqG3SOuFJSK4hquaeeYKkBkcQG56ZBDQY</HostId></Error>
```

krancour:
I'll try adding that additional debug statement.  Will be a while before I can get to it tho.

mboersma:
No problem, thanks for debugging this with us. I think this PR is valid on its own, but it's confusing why `create_bucket` is apparently still called in your case, and it would be good to pin that down before merging this.

krancour:
@mboersma, I added the debug line you suggested.  I get the same results as before, but before that, I now see a list of all my buckets.

The strange thing, however, is that each bucket name appears to be prefixed by a lowercase u _before_ the opening single quote around each bucket name.

Example (with fictitious bucket name):

```
[u'bucket-a', u'bucket-b', u'bucket-c', u'bucket-d']
```

I have to admit to not being well-acquainted with Python, but even so, that looks fishy to me.

Could this be the problem?

mboersma:
No, the `u'string'` notation is python's way of showing it's a Unicode string object rather than an old ASCII string. But `u'foo' == 'foo'` returns `True`.

I'm most curious if the intended bucket name shows up in that printed list. I assume it doesn't, otherwise we wouldn't run the `create_bucket` code inside the loop.

krancour:
Sorry... I should have mentioned that before.  The intended bucket name _is_ in the list.

mboersma:
>  The intended bucket name *is* in the list.

Well that's confusing. I can't imagine how `if name not in (bucket.name for bucket in conn.get_all_buckets())` would match then. 

The only other thing I can think of is to print out the argument to the script as well, to make sure we're comparing the strings we think we are. Perhaps add a line like this inside the loop?
```python
print name, type(name)
```
Either that line won't be reached, or it will show us we are still comparing apples and oranges somehow.

krancour:
@mboersma I was one step ahead... already trying that... and d'oh.

Intended name is showing up as `registry` (and _that_ ain't in the list!)

It's this-- which I was doubtful of: #3429

mboersma:
> Intended name is showing up as registry (and that ain't in the list!)

Ok good--at least we have our sanity back. Now we need to fix #3429. :hammer:

PR #3433, opened 2015-04-08
ref(docs): clean up app ssl documentation

bacongobbler:
"Using App SSL" belongs in "Using Deis". I've also cleaned up some of
the more confusing terms like "domain ssl" in favour of "app ssl".

### DIFF
--- a/docs/reference/index.rst
+++ b/docs/reference/index.rst
@@ -11,7 +11,6 @@ Reference Guide
 
     usage
     terms/index
-    domain-ssl
     self-signed-certs
     client
     server/index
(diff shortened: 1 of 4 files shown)


mboersma:
I didn't realize App SSL was hidden under the "Reference" section. Good cleanup. Code LGTM.

carmstrong:
LGTM.

bacongobbler:
note that this will break the link for app SSL at http://deis.io/deis-1-5-application-ssl-support/ so we'd probably want to co-ordinate a fix ASAP on our blog

PR #3435, opened 2015-04-08
feat(client): check controller before attempting to register. Fixes #3224

### LINKED ISSUE #3224: Proposal feat(cli): check the specified url before any command
This could be useful in `deis register <url>` and `deis login <url>` before any user input

Joshua-Anderson:
This sends a request to `\v1` and sees if it returns as a 401 and has the same api version before starting the registration request.

I chose to check the url `\v1` because it is the top level api and it returns a `401`, which is more rare than a `200` and therefore is less likely to be matched by accident. If you guys want to use another url to check I'm open to that.

Closes #3224.

### DIFF
--- a/client/deis.py
+++ b/client/deis.py
@@ -378,6 +378,34 @@ class DeisClient(object):
         self._settings = Settings()
         self._logger = logging.getLogger(__name__)
 
+    def check_connection(self, controller, ssl_verify):
+        url = urlparse.urljoin(controller, '/v1/')
+        error_message = """
+{} does not appear to be a valid Deis controller.
+Make sure that the Controller URI is correct and the server is running.
+""".format(controller)
+
+        try:
+            response = self._session.get(url, allow_redirects=False,
+                                         verify=ssl_verify)
+        except requests.exceptions.ConnectionError as err:
+            raise EnvironmentError(error_message + "\nSpecific Error: " + str(err.message))
+
+        if response.status_code != 401:
+            raise EnvironmentError(error_message)
+        self.check_api_version(response.headers.get('X_DEIS_API_VERSION'))
+
+    def check_api_version(self, server_api_version):
+        """
+        Check if the client is compatable with the api
+        """
+        if server_api_version is not None and server_api_version != __api_version__:
+            self._logger.warning("""
+  !    WARNING: Client and server API versions do not match. Please consider upgrading.
+  !    Client version: {}
+  !    Server version: {}
+""".format(__api_version__, server_api_version))
+
     def _dispatch(self, method, path, body=None, **kwargs):
         """
         Dispatch an API request to the active Deis controller
@@ -397,13 +425,7 @@ class DeisClient(object):
         }
         response = func(url, data=body, headers=headers, verify=ssl_verify)
         # check for version mismatch
-        server_api_version = response.headers.get('X_DEIS_API_VERSION')
-        if server_api_version is not None and server_api_version != __api_version__:
-            self._logger.warning("""
- !    WARNING: Client and server API versions do not match. Please consider upgrading.
- !    Client version: {}
- !    Server version: {}
-""".format(__api_version__, server_api_version))
+        self.check_api_version(response.headers.get('X_DEIS_API_VERSION'))
         return response
 
     def apps(self, args):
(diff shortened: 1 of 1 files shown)


client/deis.py
```diff
@@ -378,6 +378,27 @@ def __init__(self):
...
+            raise EnvironmentError("{} is not a valid deis controller url.".format(controller))
+        self.check_api_version(response.headers.get('X_DEIS_API_VERSION'))
+
+    def check_api_version(self, server_api_version):
+        """
+        Check if the client is compatable with the api
+        """
+        if server_api_version is not None and server_api_version != __api_version__:
+            self._logger.warning("""
+         !    WARNING: Client and server API versions do not match. Please consider upgrading.
```
bacongobbler: since this is in a quoted string, the whitespace is important. Could you please shift this back to the same format as it used to be?

Joshua-Anderson: Done!

mboersma:
LGTM. Tested locally with `deis register` and `deis login`:
```console
$ deis login http://foo.local3.deisapp.com
http://foo.local3.deisapp.com is not a valid deis controller url.
$ deis login http://deis.local3.deisapp.com
username:
...
$ deis login http://deis.local3.deisapp.not.com
Couldn't connect to the Deis Controller:
('Connection aborted.', gaierror(8, 'nodename nor servname provided, or not known'))
```
Checking the `v1` endpoint for a 401 seems like a reasonable approach to me.

mboersma:
@Joshua-Anderson could you `git commit --amend` the subject so it doesn't reference the GitHub issue? We try to keep that out--you never know, Deis could go bitbucket someday... :smiling_imp: And I would argue this is a feature, not a bug fix, so maybe:
```feat(client): check controller before login or register```

**Comment by bacongobbler** [comment]
I think we should re-word the error as it probably is a valid controller URL, but it could be down or unavailable.

**Comment by Joshua-Anderson** [comment]
@mboersma I amended the commit message
@bacongobbler I reworded the error message, what do you think of it now?

File: client/deis.py [hunk 2]
```diff
@@ -378,6 +378,30 @@ def __init__(self):
...
 
+    def check_connection(self, controller, ssl_verify):
+        url = urlparse.urljoin(controller, '/v1/')
+
+        response = self._session.get(url, allow_redirects=False,
+                                     verify=ssl_verify)
+
+        if response.status_code != 401:
+            raise EnvironmentError("""
+{} does not appear to be a valid deis controller.
```
bacongobbler: Deis should be capitalized

bacongobbler:
I agree on this being a feature rather than a fix. Other than that and the above comment, code LGTM :)

Joshua-Anderson:
I fixed the servname error, capitalized Deis and made the commit a feature.

client/deis.py
```diff
@@ -378,6 +378,34 @@ def __init__(self):
...
+        url = urlparse.urljoin(controller, '/v1/')
+        error_message = """
+{} does not appear to be a valid Deis controller.
+Make sure that the Controller URI is correct and the server is running.
+""".format(controller)
+
+        try:
+            response = self._session.get(url, allow_redirects=False,
+                                         verify=ssl_verify)
+        except:
```
mboersma: We should catch a specific exception class here instead and log its message too:
```console
try:
    response = self._session.get(url, allow_redirects=False,
                                 verify=ssl_verify)
except requests.exceptions.ConnectionError as err:
    raise EnvironmentError(error_message + '\n' + err.message)
```
(I tried this code interactively, and that's the error `type` thrown when the host can't be contacted. It wraps the `socket.gaierror`.)

Sorry to make you keep tweaking this PR--this is great stuff!

Joshua-Anderson: done!

mboersma:
LGTM.
```console
$ deis login http://deis.local3.deisapp.com
password: 
Logged in as test
$ deis login http://deis.local3.deisapp.doesntexist

http://deis.local3.deisapp.doesntexist does not appear to be a valid Deis controller.

Specific Error: ('Connection aborted.', gaierror(8, 'nodename nor servname provided, or not known'))
$ deis login http://doesntexist.local3.deisapp.com

http://doesntexist.local3.deisapp.com does not appear to be a valid Deis controller.
```

PR #3451, opened 2015-04-10
chore(*): bump confd to v0.9.0

### LINKED ISSUE #3436: bug(database): confd log error
Using `deis/database:v1.5.0` I see this error:

```
Apr 08 22:28:51 nodo-3 sh[13633]: 2015-04-08T22:28:51Z 55b2af686f67 confd[152]: ERROR 401: The event in requested index is outdated and cleared (the requested history has been cleared [298928281/298901368]) [298929280]
Apr 08 22:28:53 nodo-3 sh[13633]: 2015-04-08T22:28:53Z 55b2af686f67 confd[152]: ERROR 401: The event in requested index is outdated and cleared (the requested history has been cleared [298928362/298904039]) [298929361]
Apr 08 22:28:53 nodo-3 sh[13633]: 2015-04-08T22:28:53Z 55b2af686f67 confd[152]: ERROR 401: The event in requested index is outdated and cleared (the requested history has been cleared [298928362/298901368]) [298929361]
Apr 08 22:28:55 nodo-3 sh[13633]: 2015-04-08T22:28:55Z 55b2af686f67 confd[152]: ERROR 401: The event in requested index is outdated and cleared (the requested history has been cleared [298928451/298904039]) [298929450]
Apr 08 22:28:55 nodo-3 sh[13633]: 2015-04-08T22:28:55Z 55b2af686f67 confd[152]: ERROR 401: The event in requested index is outdated and cleared (the requested history has been cleared [298928451/298901368]) [298929450]
Apr 08 22:28:57 nodo-3 sh[13633]: 2015-04-08T22:28:57Z 55b2af686f67 confd[152]: ERROR 401: The event in requested index is outdated and cleared (the requested history has been cleared [298928547/298904039]) [298929546]
Apr 08 22:28:57 nodo-3 sh[13633]: 2015-04-08T22:28:57Z 55b2af686f67 confd[152]: ERROR 401: The event in requested index is outdated and cleared (the requested history has been cleared [298928547/298901368]) [298929546]
Apr 08 22:28:59 nodo-3 sh[13633]: 2015-04-08T22:28:59Z 55b2af686f67 confd[152]: ERROR 401: The event in requested index is outdated and cleared (the requested history has been cleared [298928631/298904039]) [298929630]
```

From the `etcd` (docs)[https://github.com/coreos/etcd/blob/master/Documentation/api.md] this could be a `confd` error using watch?

bacongobbler:
I've also cleaned up some of the flags like `--interval` which is only used when `--watch` is not present.

closes #3436

### DIFF
--- a/builder/image/bin/boot
+++ b/builder/image/bin/boot
@@ -41,13 +41,13 @@ function etcd_safe_mkdir {
 etcd_safe_mkdir $ETCD_PATH/users
 
 # wait for confd to run once and install initial templates
-until confd -onetime -node $ETCD --confdir /app --interval 5 --quiet; do
+until confd -onetime -node $ETCD --confdir /app --log-level error; do
 	echo "builder: waiting for confd to write initial templates..."
 	sleep $(($ETCD_TTL/2))  # sleep for half the TTL
 done
 
 # spawn confd in the background to update services based on etcd changes
-confd -node $ETCD --confdir /app --interval 5 --quiet --watch &
+confd -node $ETCD --confdir /app --log-level error --watch &
 CONFD_PID=$!
 
 # remove any pre-existing docker.sock
--- a/controller/Dockerfile
+++ b/controller/Dockerfile
@@ -10,7 +10,7 @@ RUN curl -sSL -o /usr/local/bin/etcdctl https://s3-us-west-2.amazonaws.com/opdem
     && chmod +x /usr/local/bin/etcdctl
 
 # install confd
-RUN curl -sSL -o /usr/local/bin/confd https://github.com/kelseyhightower/confd/releases/download/v0.8.0/confd-0.8.0-linux-amd64 \
+RUN curl -sSL -o /usr/local/bin/confd https://github.com/kelseyhightower/confd/releases/download/v0.9.0/confd-0.9.0-linux-amd64 \
     && chmod +x /usr/local/bin/confd
 
 # define execution environment
--- a/builder/image/Dockerfile
+++ b/builder/image/Dockerfile
@@ -10,7 +10,7 @@ RUN curl -sSL -o /usr/local/bin/etcdctl https://s3-us-west-2.amazonaws.com/opdem
     && chmod +x /usr/local/bin/etcdctl
 
 # install confd
-RUN curl -sSL -o /usr/local/bin/confd https://github.com/kelseyhightower/confd/releases/download/v0.8.0/confd-0.8.0-linux-amd64 \
+RUN curl -sSL -o /usr/local/bin/confd https://github.com/kelseyhightower/confd/releases/download/v0.9.0/confd-0.9.0-linux-amd64 \
     && chmod +x /usr/local/bin/confd
 
 # install docker-in-docker
--- a/builder/image/bin/check
+++ b/builder/image/bin/check
@@ -1,8 +1,0 @@
-#!/bin/bash
-#
-# Check that the configuration for deis-builder is valid.
-#
-
-if grep -q "<no value>" $1 ; then
-	exit 1
-fi
--- a/builder/image/conf.d/authorized_keys.toml
+++ b/builder/image/conf.d/authorized_keys.toml
@@ -7,4 +7,3 @@ mode  = "0600"
 keys = [
   "/deis/builder/users",
 ]
-check_cmd = "/app/bin/check {{ .src }}"
--- a/builder/image/conf.d/builder.toml
+++ b/builder/image/conf.d/builder.toml
@@ -9,4 +9,3 @@ keys = [
   "/deis/builder",
   "/deis/registry",
 ]
-check_cmd = "/app/bin/check {{ .src }}"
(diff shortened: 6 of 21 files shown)


logspout/image/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM flynn/busybox
+FROM alpine
```
carmstrong: I feel like this should be a separate PR, or at the very least, a separate commit. It's not really related to this PR, is it?

builder/image/bin/boot
```diff
@@ -41,13 +41,13 @@ function etcd_safe_mkdir {
 etcd_safe_mkdir $ETCD_PATH/users
 
 # wait for confd to run once and install initial templates
-until confd -onetime -node $ETCD --confdir /app --interval 5 --quiet; do
+until confd -onetime -node $ETCD --confdir /app --log-level warn --watch; do
```
carmstrong: Do `-onetime` and `--watch` conflict? Seems like `-onetime` would override here.

carmstrong:
Wasn't the motivation to move to `--log-level` to only output on error only? I thought the warnings were too verbose.

bacongobbler:
we can change it to error if that's what you want. `--quiet` is equivalent to `--log-level warn` so I was just doing a cut-over but I'm happy to make that change

builder/image/bin/boot
bacongobbler: yeah in this case `--watch` will be ignored. I'll remove it from the `--onetime` calls

mboersma:
I agree that the `confd` warnings are too verbose currently and should be silenced in our usage.

bacongobbler:
done!

carmstrong:
> we can change it to error if that's what you want. --quiet is equivalent to --log-level warn so I was just doing a cut-over but I'm happy to make that change

Could we get a summary of which messages are shown in each loglevel option so we have an idea of which ones we're suppressing?

bacongobbler:
From the user's perspective, both are complete silence on boot. No logs show up. https://gist.github.com/bacongobbler/3a7f84c345cd5f6557be

carmstrong:
> From the user's perspective, both are complete silence on boot. No logs show up.

Ok, but what about when a template can't be templated? Do we still see the missing keys in the logs? I think we want to see keys that aren't found, otherwise nothing.

bacongobbler:
Yes, we will still see `ERROR: could not find key` error messages and the like (or whatever the error message is now). I'll try and kill off my controller so you can see :)

controller/Dockerfile
```diff
@@ -10,7 +10,8 @@ RUN curl -sSL -o /usr/local/bin/etcdctl https://s3-us-west-2.amazonaws.com/opdem
     && chmod +x /usr/local/bin/etcdctl
 
 # install confd
-RUN curl -sSL -o /usr/local/bin/confd https://github.com/kelseyhightower/confd/releases/download/v0.8.0/confd-0.8.0-linux-amd64 \
+ENV CONFD_VERSION 0.9.0
```
mboersma: An `ENV` statement is another Docker image layer, with all the TCP latency that implies. And we don't need to know `CONFD_VERSION` elsewhere in the container environment. Could we make these one-liners?
```sh
RUN CONFD_VERSION=0.9.0 curl -sSL -o /usr/local/bin/confd https://github.com/kelseyhightower/confd/releases/download/v$CONFD_VERSION/confd-$CONFD_VERSION-linux-amd64 \
     && chmod +x /usr/local/bin/confd
```

bacongobbler: with the switch to parallel layer downloads, things are faster on pull these days so it's not much of an issue. I was just doing this to reduce typing of `0.9.0` to one place but I'd be happy to if this is a concern :)

mboersma: I realize it's a bit of a micro-optimization, but that's how `docker pull` makes me think. I think the one-line change accomplishes your goal of easier editing as well, but I won't hold up the PR for this change.

bacongobbler: I just saw your code example. I'll use that instead :)

mboersma: Also since we don't have signed images, we don't benefit from parallel downloads.

I wouldn't normally squawk about just TCP latency, but you've watched `docker pull` go--there is a startup cost to each layer download that is still noticeable. It's probably more than the network stack. Some of the latency seems to be in `docker` itself still, although I know they've improved that some.

carmstrong:
Code LGTM.

bacongobbler:
After playing around with this for a bit, I just found out that the inline envvar actually **breaks** the cache! This is why the controller always has to rebuild requirements.txt every time due to https://github.com/deis/deis/blob/master/controller/Dockerfile#L27

I'm just gonna opt out of CONFD_VERSION for the sake of preserving the `docker build` cache.

mboersma:
I'm fine with doing `ENV CONFD_VERSION 0.9.0` if you'd rather--my Docker layer rant really was a nitpick, especially if the envvar breaks the cache every time...

bacongobbler:
nah it's just a small nit, let's keep on :truck:ing the same as it was before

mboersma:
Code LGTM.

PR #3452, opened 2015-04-10
chore(logspout): switch to busybox

bacongobbler:
flynn/logspout has not been modified in 15 months, whereas busybox is
an officially supported image by docker.

### DIFF
--- a/logspout/image/Dockerfile
+++ b/logspout/image/Dockerfile
@@ -1,4 +1,4 @@
-FROM flynn/busybox
+FROM busybox
 MAINTAINER OpDemand <info@opdemand.com>
 
 ENV DOCKER_HOST unix:///tmp/docker.sock


bacongobbler:
also note that there was a glibc issue a while back so this is a security enhancement as well.

mboersma:
`logspout` isn't much smaller, but that wasn't the point here:
```console
$ docker images
REPOSITORY          TAG             IMAGE ID            CREATED             VIRTUAL SIZE
deis/logspout       after           c284a1c2f184        8 seconds ago       7.882 MB
deis/logspout       before          9c68e36ee26b        33 seconds ago      10.04 MB
```

bacongobbler:
yeah the difference between the 2 is like 3MB so not much to gain here. Just something that I forgot to update a while back

bacongobbler:
whoops, forgot to edit the commit title

carmstrong:
Code LGTM.

mboersma:
Code LGTM.

PR #3456, opened 2015-04-10
fix(*): remove DOCKER_BUILD=true from Dockerfile

bacongobbler:
This optimization busts the docker cache on every build, slowing
consequent builds significantly. Removing it fixes the root issue.

### DIFF
--- a/controller/Dockerfile
+++ b/controller/Dockerfile
@@ -24,7 +24,7 @@ ADD build.sh /app/tmp/build.sh
 
 ADD requirements.txt /app/requirements.txt
 
-RUN DOCKER_BUILD=true /app/tmp/build.sh
+RUN /app/tmp/build.sh
 
 ADD . /app
 
--- a/controller/build.sh
+++ b/controller/build.sh
@@ -3,13 +3,6 @@
 # fail on any command exiting non-zero
 set -eo pipefail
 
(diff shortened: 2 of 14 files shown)


carmstrong:
Code LGTM.

mboersma:
The tradeoff here is the few extra support questions we will get when users run these scripts, expecting them to do something useful outside of the `docker build` context. Also I'd like to know the specific reason `RUN` statements with initial vars break the docker cache...

Still, I'd much rather have the cache working. I rebuild these things twenty times a day. Code LGTM.

bacongobbler:
Hmm, this doesn't seem to be the case. It seems to be the ADD:

```
Step 9 : ADD requirements.txt /app/requirements.txt
 ---> ba6aa2ff42c1
Removing intermediate container 04c1c898d6d8
Step 10 : RUN /app/tmp/build.sh
```

I'm going to go ahead and close this until I nail down the true issue. I think it may be some form of difference between ADD and COPY, perhaps?

From the note just above http://docs.docker.com/reference/builder/#known-issues-run:

> The cache for RUN instructions can be invalidated by ADD instructions. See below for details.

mboersma:
> It seems to be the ADD

While at least that behavior is documented.

PR #3457, opened 2015-04-10
fix(builder): exit 1 when gitreceive captures a build

### LINKED ISSUE #3342: If buildpack crashes, deis-builder won't accept any new pushes.
We have a node.js buildpack that runs some gulp tasks. If one of those tasks crashes, it appears that deis-builder hangs. It won't accept any new builds and errors with "Another git push is ongoing. Aborting..."

Restarting the builder seems to fix for now but the builder shouldn't hang.

Buildpack: https://github.com/EQCO/heroku-buildpack-nodejs

### LINKED ISSUE #3427: Any user seems to be able to push to app
#### Update

Ok, this is maybe worse than I thought. I have managed to fix this, but the issue was that the user didn't have perms on the correct app (i have example-production and example-staging, it just had permission on example-production but i was pushing to example-staging).

So a user who does not have permission on an app, can push to that app.

#### Original Post 

Hey,

Trying to setup continuous integration deployments via CircleCI. I have the following to run:

```
deployment:
    staging:
      branch: master
      commands:
        - git push ssh://git@example:2222/example.com.git $CIRCLE_SHA1:master
```

I created a deploy user, and added the keys for them. 

The issue I am having is that the actual git push works perfectly, but it does not do the actual deployment process for some reason.

```
ubuntu@box1351:~/example.com$ git push ssh://git@example:2222/example.com.git 82143c0798a9c6dc973f76ef88bdad3ef9f15e76:master
Warning: Permanently added the ECDSA host key for IP address '[ip_address]:2222' to the list of known hosts.
Counting objects: 1, done.
Writing objects: 100% (1/1), 178 bytes | 0 bytes/s, done.
Total 1 (delta 0), reused 0 (delta 0)
To ssh://git@example:2222/example.com.git
   3ca779d..82143c0  82143c0798a9c6dc973f76ef88bdad3ef9f15e76 -> master
```

~~The only difference I can tell is the git client version. The one on CircleCI is 1.8.5.6, locally I am using 2.3.5.~~

Edit: Nope, not that. Upgraded git client and it still does not trigger hooks.

Edit: Could this be due to user permissions or something? The user has collab permissions on the application.

Thanks for any help.

bacongobbler:
The `exit $?` was capturing the call to `rm -rf`, which will almost always exit 0. As a fix, whenever gitreceive captures a failed build, it will exit with exit code 1 so that the commit will get rejected.

closes #3342
closes #3427

### DIFF
--- a/builder/image/templates/gitreceive
+++ b/builder/image/templates/gitreceive
@@ -44,7 +44,7 @@ EOF
     do
       LOCKFILE="/tmp/$RECEIVE_REPO.lock"
       if ( set -o noclobber; echo "$$" > "$LOCKFILE" ) 2> /dev/null; then
-        trap 'rm -f "$LOCKFILE"; exit $?' INT TERM EXIT
+        trap 'rm -f "$LOCKFILE"; exit 1' INT TERM EXIT
 
         # check for authorization on this repo
         $GITHOME/receiver "$RECEIVE_REPO" "$newrev" "$RECEIVE_USER" "$RECEIVE_FINGERPRINT"


mboersma:
Code LGTM.

bacongobbler:
does that look

mboersma:
+1

carmstrong:
Code LGTM.

PR #3134, opened 2015-02-23
feat(scheduler): add swarm scheduler support to deis controller

smothiki:
Install [deis](http://docs.deis.io/en/latest/) platform .
Follow the instructions to launch [controller]( http://docs.deis.io/en/latest/contributing/hacking/?highlight=hacking%20deis)
Deploy deis-swarm-node ```deisctl install swarm-node && deisctl start swarm-node```
Deploy deis-swarm-manager ```deisctl install swarm-manager && deisctl start swarm-manager```

set the scheduler config ```deisctl config controller set schedulerModule=swarm```

Swarm is in a very early stage of development . Doesn't support fail-over of swarm nodes and fail-over of containers .
Swarm follows bin packing scheduling strategy . We are working on providing a minimal HA strategy for App containers by using affinity flags .

PR #3467, opened 2015-04-13
chore(controller): update docker-py to 1.1.0

mboersma:
See https://github.com/docker/docker-py/compare/0.7.2-release...1.1.0-release

Deis still uses only the (unchanged) [`utils.parse_repository_tag`](https://github.com/docker/docker-py/blob/master/docker/utils/utils.py#L183) function, so this is purely housekeeping for now, although #3134 may want this.


carmstrong:
Code LGTM.

bacongobbler:
:shipit:

PR #3487, opened 2015-04-14
bug(contrib): check for cloudfront errors on stack creation

### LINKED ISSUE #3481: EC2 provisioning script failed with no error info
During the EY Deis install-a-thon, an EC2 provisioning failed after "Creating instances" with a return code of 1 but no printed error. We think this was due to specifying availability zones that weren't accessible, or possibly due to using an AWS account that didn't have CloudFormation access.

econnell:
add checks to provision script to make sure the stack is either in the creating or created state.  otherwise, error out and show all resources that failed

Closes #3481

### DIFF
--- a/contrib/ec2/provision-ec2-cluster.sh
+++ b/contrib/ec2/provision-ec2-cluster.sh
@@ -49,10 +49,18 @@ ATTEMPTS=45
 SLEEPTIME=10
 COUNTER=1
 INSTANCE_IDS=""
-until [ `wc -w <<< $INSTANCE_IDS` -eq $DEIS_NUM_INSTANCES ]; do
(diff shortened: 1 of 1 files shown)


deis-admin:

mboersma:
Jenkins, add to whitelist.

carmstrong:
Code LGTM.

mboersma:
Could you add the `CREATE_IN_PROGRESS` text to the second status polling line at https://github.com/deis/deis/blob/master/docs/installing_deis/aws.rst#run-the-provision-script? We have an example of EC2 provisioning in the docs, and that would keep it up-to-date with this change.

...which is awesome BTW, code LGTM otherwise.

mboersma:
I'm going to merge this and make the trivial doc change separately. Thanks again @econnell!

PR #3490, opened 2015-04-15
switch to CoreOS Stable 633.1: includes several Azure related commits

timfpark:

PR #3492, opened 2015-04-15
chore(*): bump CoreOS to 633.1.0

carmstrong:
https://coreos.com/releases/#633.1.0

Note that we really want ~~CoreOS >= 640.0.0 for Linux 3.19 for Ceph
and~~ fleet 0.9.2, but since at least Digital Ocean gets the latest
stable, we should do our best to bump to the latest stable
whenever possible.

replaces #3490

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -56,7 +56,7 @@ Vagrant.configure("2") do |config|
   config.ssh.insert_key = false
 
   config.vm.box = "coreos-%s" % $update_channel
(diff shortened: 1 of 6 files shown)


carmstrong:
https://github.com/coreos/bugs/issues/322 looks like it's worth investigating.

mboersma:
Code LGTM.

aledbf:
Code LGTM.

Issue #3475, opened 2015-04-14
Deis domains:remove removing all domains


babarinde:
Hello and Good day guys and thanks for the great work

Am experiencing a strange behaviour in v1.5
when i add two domains to my app using the 
```deis domains:add my1.example.com```
```deis domains:add my2.example.com```
When I delete one
```deis domains:remove my2.example.com```
both domains are deleted
Is this an expected behaviour?

carmstrong:
@babarinde What version of Deis are you running? This sounds like #3308 which should be fixed in Deis v1.5.0 (and v1.5.1).

babarinde:
@carmstrong Deis v1.5.0 upgraded from v1.4

mboersma:
Note to self: uncomment the skipped test we wrote to validate the fix I'm going to write.

PR #3499, opened 2015-04-15
test(controller): add unit test for deleting domains

bacongobbler:
.get_object() returns all domains instead of the requested domain, which
causes an issue when attempting to delete or modify a specific domain.

Left this unit test open in case someone wants to work on this until I get time to fix 'er up for v1.6

Refs #3475.

### DIFF
--- a/controller/api/tests/test_domain.py
+++ b/controller/api/tests/test_domain.py
@@ -7,6 +7,7 @@ Run the tests with "./manage.py test api"
 from __future__ import unicode_literals
 
 import json
+import unittest
 
 from django.contrib.auth.models import User
 from django.test import TestCase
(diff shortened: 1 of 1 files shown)


mboersma:
Code LGTM. Writing the tests before the code to be tested is always fine with me. :tropical_drink:

carmstrong:
Code LGTM, although is there much value in merging this before the implementation? We could easily forget to go back and remove the `skip`.

mboersma:
True, but we won't forget the related issue #3475, so it's probably ok to merge either way.

carmstrong:
> probably ok to merge either way.

Ok, ship it.

carmstrong:
Oops. Sorry @bacongobbler

PR #3500, opened 2015-04-16
feat(controller): add a option to limit the number of log lines

### LINKED ISSUE #1264: feat(client): add `deis logs -n`
how can I get more logs using `deis logs --app=<appname>`

i can see in the API there is settings.`settings.LOG_LINES`... can it be passed as a parameter somehow so we can get as much log as we want?

mainly wanted to give developers access to deis without having to give them ssh access.

Joshua-Anderson:
Usage: `deis apps:logs -n <number>`

I used a URL query parameter instead of a JSON body because that is generally considered better practice in a GET request. If you would like me to change to using a JSON body I can.

Do you want a API bump for this, or since 1.3 hasn't been released yet, add to that?

Closes #1264

### DIFF
--- a/client/deis.py
+++ b/client/deis.py
@@ -633,12 +633,20 @@ Make sure that the Controller URI is correct and the server is running.
         Options:
           -a --app=<app>
             the uniquely identifiable name for the application.
+          -n --lines=<lines>
+            the number of lines to display
         """
         app = args.get('--app')
         if not app:
             app = self._session.app
-        response = self._dispatch('get',
-                                  "/v1/apps/{}/logs".format(app))
+
+        url = "/v1/apps/{}/logs".format(app)
+
+        log_lines = args.get('--lines')
+        if log_lines:
+            url += "?log_lines={}".format(log_lines)
+
+        response = self._dispatch('get', url)
         if response.status_code == requests.codes.ok:
             # strip the last newline character
             for line in response.json().split('\n')[:-1]:
--- a/controller/api/models.py
+++ b/controller/api/models.py
@@ -337,12 +337,12 @@ class App(UuidAuditedModel):
 
         self.scale(user, structure)
 
-    def logs(self):
+    def logs(self, log_lines):
         """Return aggregated log data for this application."""
         path = os.path.join(settings.DEIS_LOG_DIR, self.id + '.log')
         if not os.path.exists(path):
             raise EnvironmentError('Could not locate logs')
-        data = subprocess.check_output(['tail', '-n', str(settings.LOG_LINES), path])
+        data = subprocess.check_output(['tail', '-n', log_lines, path])
         return data
 
     def run(self, user, command):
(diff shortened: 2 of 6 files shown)


mboersma:
> Do you want a API bump for this...?

I think it's fine to just fold it in to the 1.3 API since it hasn't been in an official release, but I defer to @bacongobbler.

bacongobbler:
I agree that it's fine to fold it into v1.3

bacongobbler:
looks like the integration tests need to be fixed

Joshua-Anderson:
I'm getting weird test failures now that seem to be related to ceph not binding to the monitor. Any advice?

bacongobbler:
yeah that's an odd issue. Let me re-run the build and see if that's okay. Jenkins has a temper these days

bacongobbler:
I shipped off job 576 on your PR but it failed on the recent integration test changes: https://ci.deis.io/job/test-ec2/576/consoleFull#682192135502066c60b2236c09cc9bf8

carmstrong:
> I'm getting weird test failures now that seem to be related to ceph not binding to the monitor. Any advice?

From the test logs, `ceph version 0.80.9` is Ceph Firefly, which is quite old. I'm not sure how your builds are using such an old version of Ceph. Have you rebased off master recently?

Joshua-Anderson:
I have rebased recently. On top of that, I believe that PR's are run from merge of the PR commits on top of master, so that shouldn't affect anything.

mboersma:
> PR's are run from merge of the PR commits on top of master

You would think so... But that's apparently not how the Jenkins PR builder plugin works, and this is defensible behavior I'm told. It confuses me, but in practice you have to rebase early and often. It's probably worth `git pull`ing down latest master and rebasing this against that once more. Otherwise we're a bit clueless. :grey_question:  Sorry!

Joshua-Anderson:
> You would think so... But that's apparently not how the Jenkins PR builder plugin works, and this is defensible behavior I'm told.

Good to know, thanks for the tip!

Joshua-Anderson:
I'm really confused with the way you guys are using your `utils.Execute` command.

For example in line 99 of the `apps_test.go` file:
```
	utils.Execute(t, cmd, params, false, "created initial release")
```

You're setting the fail flag to false but you are also setting an expectation for the result. However, looking at the execute function itself, if you set the fail flag to false, the expectation is never checked. 

Is the purpose of `utils.Execute` only to check if a command doesn't throw an error, or to check that  the error thrown is the expected one? If so, why are you passing an expect string with a sane looking value (which is deceptive, as the string looks realistic but often is not actually contained in the output at all), instead of nothing or just an empty string? What is the use case of `utils.Execute` vs `utils.CheckList` which seems to do almost exactly the same thing but also checks for a string is contained if the command passes.

bacongobbler:
I've always been confused, too. @smothiki, do you understand the rationale behind the function signatures?

mboersma:
Agreed that those test helper funcs are basically inscrutable. I have a refactoring in progress that discards them and starts over with more obvious code, and I will revisit that work soon.

Thanks for persevering @Joshua-Anderson. I'll re-review this today.

mboersma:
LGTM, reviewed and tested locally with vagrant.

carmstrong:
Code LGTM. Thanks so much @Joshua-Anderson!

PR #3507, opened 2015-04-16
fix(router): multiple certificate generation

### LINKED ISSUE #3483: Application SSL is not working when using multiple certs
Hi

I have 2 apps (lets call it staging.app.com & www.app.com).
Both are using custom domains (my Deis isn't hosted on app.com)

Both domains have attached a cert for that CN.

As a test I've attached a cert to staging.app.com, and that worked perfectly.
Afterwards I've done the same to www.app.com, that worked perfectly as well.

Now when accessing staging.app.com, I get unsafe connection, because the cert served is for www.app.com.

This is obviously incorrect, and renders using multiple certs useless...

Nathan

aledbf:
closes #3483


nathansamson:
While I am thankfull that someone provided a fix this fast I think the solution is a bit weird.

Using etcdctl just seems unnecessaty in a confd'd template file... Unless I am totally wrong ofcourse as I have no real experience with etcd & confd...

aledbf:
@nathansamson you are right, is not the "correct" way to solve this. That said in confd 0.5 is not possible to mix bash and go templates like is possible now (post confd 0.7 actually)

This is how it should be done https://github.com/aledbf/deis/blob/router_update_confd/router/image/templates/generate-certs
(but that can't be merged in releases only for bugfixes like 1.5.2)

bacongobbler:
@aledbf this has a few more changes than we'd want to cut in a patch release. I've manually tested #3516 and there are fewer moving parts (i.e. the cron changes here are an enhancement vs. a patch). Let's close this in favour of #3516 and roll this into the confd upgrade for the router (though we should really split out some of the smaller changes like the cron job stuff). We really appreciate the PR though! Is this ok with you?

aledbf:
@bacongobbler sure.

PR #3516, opened 2015-04-17
fix(router): write out only if cert matches the path

### LINKED ISSUE #3483: Application SSL is not working when using multiple certs

bacongobbler:
closes #3483 
closes #3507

What's happening is that we are iterating through each cert twice. This is necessary to work around some boot issues with confd. However with the current logic, we're writing the last cert in each iteration to every file. By checking the domain with the etcd key, only the correct one will be written to the cert/key path.

### DIFF
--- a/router/image/templates/generate-certs
+++ b/router/image/templates/generate-certs
@@ -27,13 +27,17 @@ while read etcd_path; do
     {{ range $cert := .deis_certs }}{{ if $cert.Nodes }}
     {{ range $certFields := $cert.Nodes }}
     {{ if eq (Base $certFields.Key) "cert" }}
+    if [[ "$(basename $etcd_path)" == "{{ Base $cert.Key }}" ]]; then
     cat << EOF > "$CERT_PATH/$(basename $etcd_path).cert"
 {{ $certFields.Value }}
 EOF
+    fi
     {{ else if eq (Base $certFields.Key) "key" }}
+    if [[ "$(basename $etcd_path)" == "{{ Base $cert.Key }}" ]]; then
     cat << EOF > "$KEY_PATH/$(basename $etcd_path).key"
 {{ $certFields.Value }}
 EOF
+    fi
     {{ end }}
     {{ end }}
     {{ end }}{{ end }}


bacongobbler:
@nathansamson please pull the following image to your own dev registry for testing once Jenkins has finished building the image: `deisctl config router set image=deisci/router:jenkins-test-ec2-574`

carmstrong:
Code LGTM, but I would like @nathansamson to run this image in his cluster and ensure that we haven't introduced any oddities.

mboersma:
Code LGTM.

carmstrong:
We're going to ship this even without @nathansamson's confirmation, as this seems to address the root cause and we'd like to get this addressed.

nathansamson:
After running `eisctl config router set image=deisci/router:jenkins-test-ec2-574`what do I need to do to make use of the new router image?

carmstrong:
> After running eisctl config router set image=deisci/router:jenkins-test-ec2-574what do I need to do to make use of the new router image?

`deisctl restart router@*`

nathansamson:
Figured something like that out myself :)

`for i in 1 2 3 4 5; do deisctl stop router@$i; deisctl start router@$i; done` :)

Router works like a charm now

nathansamson:
I assume that when upgrading to v1.5.2 (or 1.6) I need to unset my config change somhow, to get the latest supported router image?

carmstrong:
> Router works like a charm now

Glad to hear it!

/cc @bacongobbler

carmstrong:
> I assume that when upgrading to v1.5.2 (or 1.6) I need to unset my config change somhow, to get the latest supported router image?

Correct. You would unset this key (like `etcdctl rm /deis/router/image`) and then set the platform version like you would in any upgrade.

PR #3517, opened 2015-04-17
fix(*): use "confd --interval 5" instead of "--watch"

### LINKED ISSUE #3501: After upgrading to 1.5.1 deis uses lot more resources
Hi

I've upgraded to v1.5.1 a few days ago.

Since then I see on resource usage graphs (bandwidth & cpu) that all my nodes are a LOT more active.

I also noticed that deploying new apps seems to take more time than before (although this was a feeling that I can't proof).

This is even resulting in etcd timeouts & other impacting problems.

I believe this might be related to #3479 

One thing I noticed that some nodes are running more than one confd at the same time, I am not sure if that is normal.

### LINKED ISSUE #3505: confd: Limit scope of etcd --watch to /deis or remove it altogether
See coreos/etcd#2679. confd is refreshing keys whenever their TTLs change, which is quite frequently, likely due to:

- Watching the entire etcd keyspace rather than just `/deis`
- Each component's published keys (host/port) 

The etcd team is discussing adding support to filter out TTL updates from "normal" key changes, but in the meantime we need to investigate whether limiting watch to `/deis` reduces traffic substantially. If it doesn't, we'll need to disable watch and revert to a 5s polling interval. 

refs #3501
refs coreos/etcd#2679

mboersma:
*edit - carmstrong*
See the discussion in #3505 for why this was necessary. 

closes #3505
closes #3501 
*/edit*

### DIFF
--- a/builder/image/bin/boot
+++ b/builder/image/bin/boot
@@ -47,7 +47,7 @@ until confd -onetime -node $ETCD --confdir /app --log-level error; do
 done
 
 # spawn confd in the background to update services based on etcd changes
-confd -node $ETCD --confdir /app --log-level error --watch &
+confd -node $ETCD --confdir /app --log-level error --interval 5 &
 CONFD_PID=$!
 
 # remove any pre-existing docker.sock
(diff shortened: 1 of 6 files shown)


bacongobbler:
code LGTM

bacongobbler:
also closes #3501

carmstrong:
Code LGTM. We'll also have to double-check that router doesn't slip any of this back in with #3415.

PR #3519, opened 2015-04-17
fix(router): include deis.conf if no match with an SSL cert

### LINKED ISSUE #3470: SSL is broken after upgrade to 1.5.1
After upgrading deis from 1.4.1 to 1.5.1, which introduced application-level certificates (http://docs.deis.io/en/latest/using_deis/app-ssl/), my previous wildcard certificates (http://docs.deis.io/en/latest/managing_deis/ssl-endpoints/#router-ssl) no longer work. Deployed application work via http, but trying to access them via https always give 404. According to deis-router logs, it tries to route all https requests, whatever subdomain they are for, to deis-controller (in my case 192.168.0.1:8000), like:
```
Apr 13 18:15:13 staging-coreos-2 sh[5425]: [INFO] - [13/Apr/2015:18:15:13 +0000] - 46.219.37.191 - - - 301 - "HEAD / HTTP/1.1" - 183 - "-" - "curl/7.41.0" - "staging.example.com" - 192.168.0.2:49159 - staging.example.com - 0.000 - 0.000
Apr 13 18:15:23 staging-coreos-2 sh[5425]: [INFO] - [13/Apr/2015:18:15:23 +0000] - 46.219.37.191 - - - 404 - "HEAD / HTTP/1.1" - 214 - "-" - "curl/7.41.0" - "~^deis\x5C.(?<domain>.+)$" - 192.168.0.1:8000 - staging.example.com - 0.003 - 0.003
```
First request is http, seconds is https. Obviously `staging.example.com` doesn't match `^deis` wildcard, but this upstream is chosen, probably because it is first in nginx.conf.

bacongobbler:
The idea behind not including deis.conf was so that custom domains which
are different from the platform domain would not attach the platform SSL
certificate when it was enabled. However, this has the side effect of
not attaching domains which are subdomains of the platform domain, which
is intentional.

A better fix would be to check if the domain is a subdomain of the platform domain but text/template or confd doesn't have `strings.Contains` included as a function.

closes #3470

### DIFF
--- a/router/image/Dockerfile
+++ b/router/image/Dockerfile
@@ -5,8 +5,8 @@ ENV DEBIAN_FRONTEND noninteractive
 # install common packages
 RUN apt-get update && apt-get install -y curl net-tools sudo
 
-# install confd
-RUN curl -sSL -o /usr/local/bin/confd https://s3-us-west-2.amazonaws.com/opdemand/confd-git-b8e693c \
+# install confd from https://github.com/deis/confd/tree/deis
+RUN curl -sSL -o /usr/local/bin/confd https://s3-us-west-2.amazonaws.com/opdemand/confd-git-4c50136 \
     && chmod +x /usr/local/bin/confd
 
 # install common packages
--- a/router/image/templates/nginx.conf
+++ b/router/image/templates/nginx.conf
@@ -157,9 +157,11 @@ http {
     server {
         server_name {{ Base $domain.Key }};
         {{/* if a SSL certificate is installed for this domain, use SSL */}}
-        {{/* Note (bacongobbler): domains are separate from the default platform domain, */}}
+        {{/* NOTE (bacongobbler): domains are separate from the default platform domain, */}}
         {{/* so we can't rely on deis.conf as each domain is an island */}}
-        {{ if index $root (printf "deis_certs_%s_cert" (Base $domain.Key)) }}
+        {{/* FIXME (bacongobbler): confd turns hyphens to dashes, so we need to account */}}
+        {{/* for that in domains */}}
+        {{ if index $root (printf "deis_certs_%s_cert" (Replace (Base $domain.Key) "-" "_" -1)) }}
         server_name_in_redirect off;
         port_in_redirect off;
         listen 80;
--- a/router/image/templates/nginx.conf
+++ b/router/image/templates/nginx.conf
@@ -142,6 +142,7 @@ http {
     {{ $affinityArg := .deis_router_affinityArg }}
     {{ $certs := .deis_certs }}
     {{ $domains := .deis_domains }}
+    {{ $root := . }}
     {{ range $service := .deis_services }}{{ if $service.Nodes }}
     upstream {{ Base $service.Key }} {
         {{ if $affinityArg }}hash $arg_{{ $affinityArg }} consistent;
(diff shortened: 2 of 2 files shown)


bacongobbler:
filed kelseyhightower/confd#270 for this request

carmstrong:
We should manually test that the catchall router-level certificate takes precedence and is applied for all domains **unless** a domain has an SSL certificate attached, in which case that certificate is served only for that domain.

bacongobbler:
fixed:

```
><> deis info
=== go Application
{
  "updated": "2015-04-20T17:03:24UTC",
  "uuid": "c306b6e2-2192-469e-bf27-3126ed3ad6c8",
  "created": "2015-04-20T17:00:40UTC",
  "url": "go.bacongobbler.com",
  "owner": "bacongobbler",
  "id": "go",
  "structure": {
    "web": 1
  }
}

=== go Processes

--- web:
web.1 up (v2)

=== go Domains
go.23.253.174.228.xip.io
www.bacongobbler.com
foo.fishworks.io
><> deis certs
Common Name       Expires
----------------  ----------------------
foo.fishworks.io  2016-04-19T18:04:53UTC
```

Output of nginx.conf in the routers (only showing the relevant parts): https://gist.github.com/bacongobbler/28f61e38e16474ab9127

Note that every domain except for `foo.fishworks.io` is relying on the default deis.conf, whereas `foo.fishworks.io` is using app SSL.

mboersma:
Code LGTM.

carmstrong:
```console
$ openssl s_client -connect foo.fishworks.io:443
CONNECTED(00000003)
depth=0 /C=CA/ST=British-Columbia/L=Vancouver/O=Fishworks Development and Consulting/CN=*.bacongobbler.com/emailAddress=matthewf@fishworks.io
verify error:num=18:self signed certificate
verify return:1
depth=0 /C=CA/ST=British-Columbia/L=Vancouver/O=Fishworks Development and Consulting/CN=*.bacongobbler.com/emailAddress=matthewf@fishworks.io
verify return:1
---
Certificate chain
 0 s:/C=CA/ST=British-Columbia/L=Vancouver/O=Fishworks Development and Consulting/CN=*.bacongobbler.com/emailAddress=matthewf@fishworks.io
   i:/C=CA/ST=British-Columbia/L=Vancouver/O=Fishworks Development and Consulting/CN=*.bacongobbler.com/emailAddress=matthewf@fishworks.io
---
```

Looks to me like `foo.fishworks.io` is using the wildcard certificate?

bacongobbler:
I see that too when using the openssl client...

```
><> openssl s_client -connect foo.fishworks.io:443
CONNECTED(00000003)
depth=0 /C=CA/ST=British-Columbia/L=Vancouver/O=Fishworks Development and Consulting/CN=*.bacongobbler.com/emailAddress=matthewf@fishworks.io
```

but when I use Chrome:

screen shot 2015-04-20 at 3 11 24 pm

Same for Safari, which I **never** use:

screen shot 2015-04-20 at 3 12 50 pm

Is there something I'm doing wrong here?

carmstrong:
I dunno... are you running multiple routers, some of which may not have the fix? Maybe it's the luck of the draw for a request.

bacongobbler:
Just did a fresh deploy. All three have the same config :S

carmstrong:
Now that I think about it, maybe `s_client` doesn't send a Host header (just connects TCP on port 443), so it makes sense that the router serves the default cert. LGTM.

carmstrong:
https://www.sslshopper.com/ssl-checker.html#hostname=foo.fishworks.io agrees too.

carmstrong:
This has failed twice in a row in the same place:
```
domains:add test.test-7cc8a64e6a.deis.works --app domainsample
Adding test.test-7cc8a64e6a.deis.works to domainsample...  ...  o.. done
ok
certs:add /var/lib/jenkins/jobs/test-ec2/workspace/src/github.com/deis/deis/tests/test.test-7cc8a64e6a.deis.works.cert /var/lib/jenkins/jobs/test-ec2/workspace/src/github.com/deis/deis/tests/test.test-7cc8a64e6a.deis.works.key
Adding SSL endpoint...  ... done
test.test-7cc8a64e6a.deis.works
ok
sleeping for 60 seconds until certs are generated...
ok
SIGQUIT: quit
PC=0x43ec61
```

@bacongobbler I suspect something is wrong with this PR. Can you try running the integration tests locally and see if you see the same issue?

bacongobbler:
The common problem on Jenkins passes for me locally on both Rackspace and EC2:

```
sleeping for 60 seconds until certs are generated...
ok
Powered by Deis

certs:remove test.fishr.pw
Removing test.fishr.pw... Done.
```

bacongobbler:
Gotta love heisenbugs :bug:

mboersma:
Reviewed this again, plus deis/confd@4c50136a8c67a0cd020d93c015bfbb23eef934d6. Code LGTM.

carmstrong:
:ship:

PR #3524, opened 2015-04-18
feat(logger): expose config as flags

bacongobbler:
So I was trying to look around for a syslog server to store logs on some of my remote machines so I can test the Docker 1.6 changes and I thought to myself... "Hey, deis built one of those!" So this PR is a general enhancement to the logger itself. These enhancements are a no-op to the existing logger, but exposes flags to the end-user such that it's possible to extend the logger to the user's content.

Along with exposing common configuration as feature flags, this PR includes a feature flag for enabling/disabling publishing to service discovery (etcd). In some cases where you are running the logger in a local environment, you don't require the need to publish the logger to etcd. By default this is set to false, but I've enabled it in the Dockerfile so it's a no-op for Deis.

This doesn't need to be merged as it does not really affect Deis, but it does increase the value of the logger to be adopted by non-Deis platforms. Woo open source! :fireworks:

### DIFF
--- a/logger/README.md
+++ b/logger/README.md
@@ -10,19 +10,9 @@ the [Deis Project](https://github.com/deis/deis).
 
 ## Usage
 
-Please consult the [Makefile](Makefile) for current instructions on how to build, test, push,
-install, and start **deis/logger**.
-
-## Environment Variables
-
-* **DEBUG** enables verbose output if set
-* **ETCD_PORT** sets the TCP port on which to connect to the local etcd
-  daemon (default: *4001*)
-* **ETCD_PATH** sets the etcd directory where the logger announces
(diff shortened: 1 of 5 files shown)


bacongobbler:
Note that this is also relevant for work that I'd like to work on the logger for `deis logs -f`. I'll propose a PR but basically I'd like to add a http server on the logger which we can spawn sessions that will expose the logs as a stream, similar to [Heroku's Logplex server](https://github.com/heroku/logplex).

carmstrong:
I'm fine with this as long as it doesn't break anything. Code LGTM.

mboersma:
I like these changes, code LGTM.

PR #3527, opened 2015-04-19
fix(controller): return the correct domain from get_object()

### LINKED ISSUE #3475: Deis domains:remove removing all domains

mboersma:
The Django view helper method `get_object` was returning a queryset rather than the `Domain` model instance.

Thanks to @bacongobbler for starting this with a unit test. :green_heart: 

Closes #3475, refs #3499.

### DIFF
--- a/controller/api/tests/test_domain.py
+++ b/controller/api/tests/test_domain.py
@@ -7,7 +7,6 @@ Run the tests with "./manage.py test api"
 from __future__ import unicode_literals
 
 import json
-import unittest
 
 from django.contrib.auth.models import User
 from django.test import TestCase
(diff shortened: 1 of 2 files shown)


bacongobbler:
Awesome!  I knew the fix was going to be this, just couldn't get the right incantation. Code LGTM!

carmstrong:
Code LGTM.

PR #3253, opened 2015-03-10
fix(controller): Update CORS whitelisted request headers.

Joshua-Anderson:
Some of the CORS headers changes since I last looked at the spec so I updated them. This allows external domains, like a deis ui, to be permitted to send and receive these headers.

Issue #3512, opened 2015-04-17
enforceHTTPS=true causes client side and git issues


azurewraith:
[user@lavos app|master]$ git push -f deis master
    /usr/local/bin/gitreceive: line 33: /home/git/app.git/hooks/pre-receive: No such file or directory
    fatal: Could not read from remote repository.

    Please make sure you have the correct access rights
    and the repository exists.

So I tried to delete and recreate the app:

    [user@lavos app|master]$ deis apps:destroy --app app
    
     !    WARNING: Potentially Destructive Action
     !    This command will destroy the application: app
     !    To proceed, type "app" or re-run this command with --confirm=app

    > app
    Destroying app...
     ... /var/folders/2q/1zznrz4j3bs37qmd_159m_n00000gp/T/_MEIcsMCXL/requests/packages/urllib3/connection.py:251: SecurityWarning: Certificate has no `subjectAltName`, falling back to check for a `commonName` for now. This feature is being removed by major browsers and deprecated by RFC 2818. (See https://github.com/shazow/urllib3/issues/497 for details.)
    done in 4s
    [user@lavos app|master]$ deis apps:create app
    Creating application...  ... /var/folders/2q/1zznrz4j3bs37qmd_159m_n00000gp/T/_MEIpx2oZp/requests/packages/urllib3/connection.py:251: SecurityWarning: Certificate has no `subjectAltName`, falling back to check for a `commonName` for now. This feature is being removed by major browsers and deprecated by RFC 2818. (See https://github.com/shazow/urllib3/issues/497 for details.)
    200 OK
    {u'count': 1, u'previous': None, u'results': [{u'updated': u'2015-04-08T20:24:55UTC', u'uuid': u'9253b6c7-3706-4a14-be61-c4b687a8410c', u'created': u'2015-04-08T20:17:56UTC', u'url': u'other-app-name.domain.com', u'owner': u'<me>', u'id': u'other-app-name', u'structure': {u'web': 1}}], u'next': None}

The interesting thing is that the JSON returned references the first app created on this cluster (other-app-name)...

Once I set enforceHTTPS=false, things worked fine again...

    [user@lavos app|master]$ deis apps:destroy --app app
    
     !    WARNING: Potentially Destructive Action
     !    This command will destroy the application: app
     !    To proceed, type "app" or re-run this command with --confirm=app

    > app 
    Destroying app... 
    404 NOT FOUND
    Detail:
    Not found
    [user@lavos app|master]$ deis apps:create app
[...]

PR #3532, opened 2015-04-20
fix(client): use pyOpenSSL for improved security features

mboersma:
Recently urllib3 began [emitting these warnings](https://urllib3.readthedocs.org/en/latest/security.html#insecureplatformwarning) pointing to limitations in python's `ssl` that could create vulnerabilities. It's not easy to require Python 2.7.9 for all dev users and our build environment, so [using `pyOpenSSL`](https://urllib3.readthedocs.org/en/latest/security.html#pyopenssl) is the other recommended fix, which is what this PR does. 

This also moves `deis` CLI libraries to `client/requirements.txt` and cleans up a dead Makefile target and test script.

Closes #3253, refs #3512.

### DIFF
--- a/client/Makefile
+++ b/client/Makefile
@@ -1,6 +1,6 @@
 
 build: setup-venv
-	venv/bin/pip install docopt==0.6.2 ndg-httpsclient==0.3.3 pyasn1==0.1.7 pyOpenSSL==0.15.1 python-dateutil==2.4.2 PyYAML==3.11 requests==2.5.1 git+https://github.com/pyinstaller/pyinstaller@7413317 tabulate==0.7.4 termcolor==1.1.0 urllib3==1.10.2
+	venv/bin/pip install -r requirements.txt git+https://github.com/pyinstaller/pyinstaller@7413317
 	venv/bin/pyinstaller deis.spec
 	chmod +x dist/deis
 
@@ -16,9 +16,6 @@ uninstall:
 clean:
 	rm -rf build/ dist/ *.egg-info
 
-client:
-	pyinstaller deis.spec
-
 installer: build
 	@if [ ! -d makeself ]; then git clone -b single-binary https://github.com/deis/makeself.git; fi
 	PATH=./makeself:$$PATH BINARY=deis makeself.sh --bzip2 --current --nox11 dist \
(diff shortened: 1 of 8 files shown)


carmstrong:
Code LGTM.

client/Makefile
```diff
@@ -32,9 +29,9 @@ installer: build
 
 setup-venv:
 	@if [ ! -d venv ]; then virtualenv venv; fi
-	venv/bin/pip install -q flake8==2.4.0
 
 test-style: setup-venv
+	venv/bin/pip install -q flake8==2.4.0
```
bacongobbler: Shouldn't this stay as part of `setup-venv`?

docs/docs_requirements.txt
```diff
@@ -23,12 +23,17 @@ PyYAML==3.11
...
 python-ldap==2.4.19
 
-# Deis client requirements
+# Deis client requirements, copied from client/requirements.txt
 docopt==0.6.2
+ndg-httpsclient==0.3.3
+pyasn1==0.1.7
+pyOpenSSL==0.15.1
 python-dateutil==2.4.2
+#PyYAML==3.11
```
bacongobbler: how come we're commenting out PyYAML here? Just because we don't want to handle the hassle of installing PyYAML's dependencies?

client/Makefile
mboersma: :shrug: `setup-venv` creates the python virtualenv, and `test-style` is the only Makefile target that uses `flake8`, so it seemed more appropriate to localize it, but I can revert.

bacongobbler:
my q's are just questions. I know they don't really affect today's behaviour, just wondering why we're shifting 'em around.  Code LGTM!

docs/docs_requirements.txt
mboersma: Because we have PyYAML elsewhere in the same file, and `pip` will refuse and quit when it sees a duplicate requirement. So this is commented out to document that the client requires it (and so this copied section matches `client/requirements.txt` better.

(And `docs_requirements.txt` has to be maintained separately because Deis docs are published at ReadTheDocs.org, and its admin interface allows us to specify exactly one requirements file. So we still have some copy-and-pasting maintenance we have to do.)

carmstrong:
@mboersma What are the steps required in manually testing this?

mboersma:
I was just hoping for someone else to test the binary against their HTTPS installation. It seems to work but another validation would be good.

bacongobbler:
I don't have my cacert bundle set up properly to handle RapidSSL certificates, but I see the warning as expected when I disable SSL verification. Code LGTM!

bacongobbler:
Never mind! I got it working. Code LGTM

carmstrong:
:ship:

PR #3535, opened 2015-04-20
fix(controller): allow "*" wildcard in cert REST URLs

### LINKED ISSUE #3533: Error with `deis certs:remove`
Getting the following error when trying to remove a cert.

```
$ deis certs:remove '*.brandfolder.com'
Removing *.brandfolder.com... 405 METHOD NOT ALLOWED
Detail:
Method 'DELETE' not allowed.
```

mboersma:
One of the regular expressions for `Certificate` endpoints wasn't allowing `*` through, so wildcard certs could be created but not deleted. This relaxes the regex to allow such certs to be deleted. Thanks to @bacongobbler for the test! :green_heart: 

Note that wildcard certificates aren't fully supported yet, but users should still be able to delete them.

Closes #3533.


carmstrong:
Code LGTM.

bacongobbler:
code LGTM

carmstrong:
Looks like this passed the latest one. :ship:

Issue #3508, opened 2015-04-16
Affinity arg is not working correctly


ashaffer:
Right now if you set the affinity arg, I would expect that if a url does not contain that argument at all, it will be routed with no affinity.  But it appears that instead it is routed consistently to the same server, as if it were consistently hashing the undefined value.

This is a problem for us, because we only want to use sticky routing on one of our services (for xhr polling / sockets), but the rest we would like to be load balanced in the standard way.

bacongobbler:
I believe this was the intended behaviour, but I'm not knowledgeable enough about affinity args to refute that. @wenzowski, how do you feel about this one?

wenzowski:
Digging into the `$arg_name` construct, it does indeed return empty string when the argument is unset and therefore nginx is consistently hashing on empty string. This is indeed a bug.

Yes @ashaffer what you're doing does sound like the most common use case.

I'm exploring this, but completely open to suggestions.

carmstrong:
I guess I'm not sure why https://github.com/deis/deis/blob/master/router/image/templates/nginx.conf#L147 is being evaluated when `/deis/router/affinityArg` is unset?

wenzowski:
Oh was that the question? My understanding @carmstrong is that when affinityArg was set, eg to `'sticky'` and nginx received a request containing the arg, eg to `GET 'https://example.com/xhr?sticky=prngHash'` everything works as expected, but when a request comes in that doesn't contain an http get variable, eg to GET `'https://example.com/'` then nginx returns an empty string for `$arg_sticky` thus breaking load balancing.

wenzowski:
Off the top of my head, I'm thinking of something like this
```
set_random $prng 0 99;
set_if_empty $arg_{{ $affinityArg }} $prng;
```
Which [takes advantage] of c++ `rand()` rather than a csprng and stores the resulting integer in a string. I would assume this would always generate the random number, regardless of if it gets used or not, but seems like negligible overhead, and the fastest way to fix the problem.

[takes advantage]: http://wiki.nginx.org/HttpSetMiscModule#set_random

wenzowski:
It'll probably take me to Tuesday to get to this. Can you confirm I understand correctly @ashaffer or is @carmstrong on the money?

ashaffer:
@wenzowski Yes, your understanding is correct.  That should address the issue.

carmstrong:
Yep, this is something we should fix. Let us know how we can help, @wenzowski!

[...]

PR #3541, opened 2015-04-21
feat(router): allow sticky or persistent sessions in nginx

babarinde:
This allows clients to be tied to the same application server and allow
sessions to be maintained

Noticed that the apps that were scaled to more than one container were asking for login upon every requests.
This is just a modest attempt.
Thanks

### DIFF
--- a/router/image/templates/nginx.conf
+++ b/router/image/templates/nginx.conf
@@ -144,6 +144,7 @@ http {
     {{ $domains := .deis_domains }}
     {{ range $service := .deis_services }}{{ if $service.Nodes }}
     upstream {{ Base $service.Key }} {
+	ip_hash;
         {{ if $affinityArg }}hash $arg_{{ $affinityArg }} consistent;
         {{ end }}
         {{ range $upstream := $service.Nodes }}server {{ $upstream.Value }};


bacongobbler:
I don't think this is something we want by default. I think this is what the affinity argument is about in #3088. Note that there's currently an issue with it via https://github.com/deis/deis/issues/3508 but does tht work for you?

carmstrong:
The discussion around sticky sessions was in #2060, and the consensus was that since sticky sessions violate 12-factor methodology, it's not something we want. Does that make sense, @babarinde?

babarinde:
@bacongobbler 
#3508 does not work for me.
I think I might have used the wrong commit title on this. something like #3089  would have been fine

May we should see the 'ip_hash' as a load balancing algorithm wt affinity like #3089  rather than using the 'sticky sessison' term. It makes a lot of sense for a client to deal with the same server all through its lifetime, as some of us do have their session data on the server. 

From their docs [here](http://nginx.org/en/docs/http/ngx_http_upstream_module.html#ip_hash) the ip_hash:
> Specifies that a group should use a load balancing method where requests are distributed between servers based on client IP addresses. The first three octets of the client IPv4 address, or the entire IPv6 address, are used as a hashing key. The method ensures that requests from the same client will always be passed to the same server except when this server is unavailable. In the latter case client requests will be passed to another server. Most probably, it will always be the same server as well.

I believe that the algorithm runs based on the available upstream servers and if i scale containers to say 4, then the upstream is recomputed and the ip_hash also recomputed based on the number of available servers. And there are no session cookies involved here.

However when the container is trashed and new one comes up its ok to terminate and start a new session. Knowing that bringing up new version containers will most likely not be done at a very frequent rate (like say every hour/ 30 mins). Terminating a session even after such period is normal.

Passing an affinity arg on all our request will not be an ideal scenario for us.

bacongobbler:
The reason why we've been hesitant to support stucky sessions is due to the following statement from 12factor.net:

> Some web systems rely on “sticky sessions” – that is, caching user session data in memory of the app’s process and expecting future requests from the same visitor to be routed to the same process. Sticky sessions are a violation of twelve-factor and should never be used or relied upon. Session state data is a good candidate for a datastore that offers time-expiration, such as Memcached or Redis.

bacongobbler:
To elaborate a bit more why this precedence was set (I was on my phone):

Heroku is an AWS shop with their own scheduler for dynos. At any given time, they may decide to rebalance your dynos such that everything is properly distributed across multiple AZs. Because of this, you cannot expect the container to live persistently under regular operation. This is something which we may eventually support as we dive into more sophisticated schedulers.

Closing, as this is not something which we want to support on a large scale.

PR #3542, opened 2015-04-21
chore(router): update nginx to 1.8

johanneswuerbach:
Changes: http://nginx.org/en/CHANGES-1.8

### DIFF
--- a/router/build.sh
+++ b/router/build.sh
@@ -10,7 +10,7 @@ if [[ -z $DOCKER_BUILD ]]; then
   exit 1
 fi
 
-export VERSION_NGINX=nginx-1.7.10
+export VERSION_NGINX=nginx-1.8.0
 export VERSION_TCP_PROXY=f2156eff9f6621aaf601eaa8dee40c6820dea0b0
 export VERSION_NAXSI=0d53a64ed856e694fcb4038748c8cf6d5551a603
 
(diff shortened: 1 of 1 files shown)


carmstrong:
@aledbf Had mentioned waiting for 1.9 mainline, since TCP support will be pulled into nginx proper rather than relying on an external patch.

Thoughts?

bacongobbler:
> *) Bugfix: in the ngx_http_spdy_module

We use this module so I'm wondering what the relevant bugfix was for it.

johanneswuerbach:
Oh, I didn't know about that. Despite that I think it should still be bumped to 1.8 as 1.8 includes a few bug fixes and 1.9 might not work ootb.

mboersma:
My only issue is the "optics" of being on the 1.8 Release branch versus the 1.9 Dev branch (at least that's how I understand nginx to be). I know we switched off 1.6 to 1.7 at some point, and that real nginx users say there's little difference in stability between the two, but the even-numbered versions *should* be more stable.

So that's an indirect argument in favor of this PR versus waiting for 1.9.

carmstrong:
I'm fine with shipping this now and discussing the 1.9 bump when it's time. Code LGTM.

mboersma:
Code LGTM. Let's have @aledbf or @bacongobbler glance at this too before we merge it.

aledbf:
Code LGTM.

bacongobbler:
I've been silently following this issue. Code LGTM

PR #3548, opened 2015-04-22
feat(Makefile): Add a command to start a local postgresql container for running unit tests

Joshua-Anderson:
`make -C controller postgres` will start up a PostgreSQL container you can use to run unit tests. Useful for people without PostgreSQL installed or who don't want to mess with their local database. Exporting the variables printed by the command will allow the controller to connect to the specified database.

### DIFF
--- a/controller/Makefile
+++ b/controller/Makefile
@@ -50,6 +50,14 @@ deploy: build dev-release restart
 runserver:
 	python manage.py runserver
 
+postgres:
+	docker start postgres || docker run --restart="always" -d -p 5432:5432 --name postgres postgres:9.4.1
+	docker exec postgres createdb -U postgres deis 2>/dev/null || true
+	@echo "To use postgres for local development:"
+	@echo "    export PGHOST=`boot2docker ip 2>/dev/null || echo 127.0.0.1`"
+	@echo "    export PGPORT=5432"
+	@echo "    export PGUSER=postgres"
+
 db:
 	python manage.py syncdb --migrate --noinput
 


bacongobbler:
I think this is convenient, but I wonder if this would be more fitting in the [hack docs](http://docs.deis.io/en/latest/contributing/hacking/) as a suggestion rather than an actual feature.

Joshua-Anderson:
> Some users may have different setups

This is the primary reason I created this PR. PostgreSQL setups can vary greatly, so I wanted a "universal" setup that could be used on every computer to run the deis unit tests.

This way people can run:

``` console
$ make -C controller postgres
$ make -C controller test
```

And know that they are running their tests in a setup that will work, rather than try to install and configure their local PostgreSQL database to work with the tests.

I created this based with the same idea as `make dev-registry`, a simple way to get a complicated component working.

This could also simply the current CI setup, which currently has a complicated procedure for installing and configuring PostgreSQL.

I understand this is not a straightforward decision, and I don't have any hard feelings if this isn't accepted :heart::smile:. I found it useful for development and thought others might as well.

bacongobbler:
Indeed. I can see the usefulness with this. Any simplification of bootstrapping a dev environment gets a +1 from me. I'm just proposing alternative solutions so the other devs understand what I'm thinking about. :)

To be clear, this has my LGTM.

carmstrong:
> To be clear, this has my LGTM, but my own personal opinion says that this should live as docs or a blog post so there's less maintenance on our end but that's just me :D

I'm personally okay with this - there's really not much of a maintenance burden, and it seems like it would be useful for people. Code LGTM.

mboersma:
I like this very much--PostgreSQL setup is the fiddliest part of getting ready to run all the Deis tests. If this works easily with `make -C controller/ test-unit`, I'm in favor of it. I'll test it.

controller/Makefile
```diff
@@ -50,6 +50,14 @@ deploy: build dev-release restart
 runserver:
 	python manage.py runserver
 
+postgres:
+	docker start postgres || docker run --restart="always" -d -p 5432:5432 --name postgres postgres:9.4.1
+	docker exec postgres createdb -U postgres deis 2>/dev/null || true
+	@echo "To use postgres for local development:"
+	@echo "    export export PGHOST=192.168.1.97"
```
mboersma: The word "export" is repeated here.

Instead of hard-coding a particular local IP, this should call `boot2docker` like [`make dev-registry`](https://github.com/deis/deis/blob/master/Makefile#L23)--then at least on Macs we get the right IP. Or else use `[container IP address]`.

mboersma:
This worked great for me: `make -C controller/ postgres test-unit`. (Note that you still need `psql` and Python's `psycopg2` package built locally to run the tests.)

The commit message is [too long](http://docs.deis.io/en/latest/contributing/standards/#subject-line). Could you change it to something like:
```
feat(controller): offer DB container for unit tests
```

Joshua-Anderson:
@mboersma  Done!

bacongobbler:
Once @mboersma's comment on the double export being addressed, this has my LGTM

controller/Makefile
Joshua-Anderson: :sob: Sorry about that, I missed the line from development. The IP address should be `127.0.0.1`, which works on linux and should work on OSX.

Joshua-Anderson:
@mboersma's comment has been resolved, but it probably should be tested on OSX to confirm that the container ports are being exposed as expected there.

controller/Makefile
```diff
@@ -50,6 +50,14 @@ deploy: build dev-release restart
 runserver:
 	python manage.py runserver
 
+postgres:
+	docker start postgres || docker run --restart="always" -d -p 5432:5432 --name postgres postgres:9.4.1
+	docker exec postgres createdb -U postgres deis 2>/dev/null || true
+	@echo "To use postgres for local development:"
+	@echo "    export PGHOST=127.0.0.1"
```
mboersma: 127.0.0.1 doesn't work for `boot2docker` unfortunately: ```OperationalError: could not connect to server: Connection refused``` when running the controller unit tests.

Maybe change it to:
```sh
@echo "    export PGHOST=`boot2docker ip 2>/dev/null || echo 127.0.0.1`"
```

That's not perfect, but works for me&trade; on my Mac and on a Linux box without `boot2docker`.

mboersma:
Hard-coding `PGHOST=127.0.0.1` won't work--I pointed to a solution [here](https://github.com/deis/deis/pull/3548/files#r29487912).

Joshua-Anderson:
@mboersma Used your solution to fix the host problem.

mboersma:
Code LGTM. Thanks!

Issue #3448, opened 2015-04-10
deis router requests hang after about an hour


azurewraith:
Slightly different from a recently reported error.

After about an hour, all connections to Deis' router hang in v1.5.0.  Once I restart each router, everything goes back to working fine...

    $ curl -v http://172.17.100.29/health-check
    * About to connect() to 172.17.100.29 port 80 (#0)
    *   Trying 172.17.100.29...
    * Adding handle: conn: 0x7fa212000000
    * Adding handle: send: 0
    * Adding handle: recv: 0
    * Curl_addHandleToPipeline: length: 1
    * - Conn 0 (0x7fa212000000) send_pipe: 1, recv_pipe: 0
    * Connected to 172.17.100.29 (172.17.100.29) port 80 (#0)
    > GET /health-check HTTP/1.1
    > User-Agent: curl/7.30.0
    > Host: 172.17.100.29
    > Accept: */*
    >
    ^C (more than 30s hang, so interrupted)

    $ deisctl restart router@1 && deisctl restart router@2 && deisctl restart router@3
    - deis-router@1.service: inactive/dead
    - deis-router@1.service: active/running
    - deis-router@2.service: inactive/dead
    - deis-router@2.service: active/running
    - deis-router@3.service: inactive/dead
    - deis-router@3.service: active/running

    $ curl -v http://172.17.100.29/health-check
    * About to connect() to 172.17.100.29 port 80 (#0)
    *   Trying 172.17.100.29...
    * Adding handle: conn: 0x7ff96480aa00
    * Adding handle: send: 0
    * Adding handle: recv: 0
    * Curl_addHandleToPipeline: length: 1
    * - Conn 0 (0x7ff96480aa00) send_pipe: 1, recv_pipe: 0
    * Connected to 172.17.100.29 (172.17.100.29) port 80 (#0)
    > GET /health-check HTTP/1.1
    > User-Agent: curl/7.30.0
    > Host: 172.17.100.29
    > Accept: */*
    >
    < HTTP/1.1 200 OK
    * Server nginx/1.7.10 is not blacklisted
    < Server: nginx/1.7.10
    < Date: Fri, 10 Apr 2015 01:46:58 GMT
    < Content-Type: text/plain
    < Content-Length: 0
    < Connection: keep-alive
    <
    * Connection #0 to host 172.17.100.29 left intact

Nothing out of the ordinary in the global journalctl, or the router logs.

[...]

PR #3554, opened 2015-04-22
fix(router): Unify timeout values

Blystad:
The router was having different timeout values for keepalive_timeout
and for proxy timeouts.

This can cause issues with front facing load balancers in front of
deis-router (like Amazon ELB), and could be the cause of issue #3448

Timeout values should be greater than the timeout value of the front facing
load balancer. 

**Ref:**

> The keepalive timeout value on your backend server must be higher than that
> of your ELB connection timeout. If it is lower, the ELB will re-use the idle
> connection when your server has already dropped the connection, resulting in
> the client being served up a blank response.

http://engineering.chartbeat.com/2014/02/12/part-2-lessons-learned-tuning-tcp-and-nginx-in-ec2/

> The ELB timeout should be set to less then the backend server idle timeout.

https://forums.aws.amazon.com/thread.jspa?threadID=104688

### DIFF
--- a/docs/customizing_deis/router_settings.rst
+++ b/docs/customizing_deis/router_settings.rst
@@ -39,6 +39,7 @@ setting                                      description
 /deis/domains/*                              domain configuration for applications (set by controller)
 /deis/router/affinityArg                     for requests with the indicated query string variable, hash its contents to perform session affinity (default: undefined)
 /deis/router/bodySize                        nginx body size setting (default: 1m)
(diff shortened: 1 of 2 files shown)


carmstrong:
Thank you for your work in debugging this, @Blystad!

Blystad:
My pleasure. Anything which can help my peers and me sleep better at night :smile:

bacongobbler:
code LGTM. Nice sleuthing

carmstrong:
Code LGTM. Thanks @Blystad!

PR #3555, opened 2015-04-22
fix(registry/issue-3429): registry boot script can now get bucket name from etcd

krancour:


bacongobbler:
Looks like this also exists in the docs, so no changes required there :)

Could you please amend your commits such that it looks like this:

```
fix(registry): retrieve bucket name from etcd
```

We want the scope to be a component name so the changelog shows which components are affected by this issue. e.g. http://docs.deis.io/en/latest/contributing/standards/#example

carmstrong:
We may want to continue to support the environment variable in case anyone's using it. 

We've also been bitten in the past trying to read an etcd key right after it's been set - we have no guarantees that the write has persisted and is there by the time we need it. 

Maybe a solution is to move the bucket name into `/app/bin/create_bucket`, since it's being managed by confd anyway.

bacongobbler:
> We've also been bitten in the past trying to read an etcd key right after it's been set - we have no guarantees that the write has persisted and is there by the time we need it.

Good point. That's why we switched to `etcdctl mk` to create the keys instead of `etcdctl set/get` in an older PR. I like the idea of moving it over to `create_bucket` so we can rely on confd.

krancour:
Bear with me... confd is something new for me.

I am trying as you have suggested and am updating create_bucket, line 14 as so:

```
name = '{{ getv "/deis/registry/bucketName" }}'
```

Oddly, this is giving me an error like the following even though this key _is_ set:

```
sh[3313]: 2015-04-22T22:50:38Z 435e216bdabc confd[41]: ERROR template: create_bucket:14:11: executing "create_bucket" at <getv "/deis/registry...>: error calling getv: key does not exist
```

Being new to this, I am sure there is something here that I simply don't understand.

krancour:
Altho in my case, that key has a value set because _I_ set it.  (Still strange that that's not resolving.) But I'm not sure how it would be set in the average case.  I guess confd needs to run to _create_ `bin/create_bucket` before `bin/boot` can possibly succeed and `bin/boot` is where the logic for setting an initial value for `deis/registry/bucketName` currently resides...

Long story short... I guess I am very confused on how the proposed alternative was supposed to work... though I'd like to try and understand it.

krancour:
Nevermind... I see now... `bin/boot` runs, gets at least as far as setting a default value (if not existing?) for `/deis/registry/bucketName` and then _waits_ for confd to run once.  So now I get how it's supposed to work... still don't get why it isn't.  :-/

krancour:
I can `nse` to the floundering registry and run `etcdctl -C $HOST:4001 get /deis/registry/bucketName` and I _do_ see the key is set with the value I expect.  Not sure why confd wouldn't also see it.

krancour:
Had to rtf confd m.  I had not updated the create_bucket.toml.  Knowing is half the battle.

Got this working.  I am going to close this PR and open a new one.

PR #3556, opened 2015-04-22
feat(controller): support SAN certificates

### LINKED ISSUE #3441: Application SSL does not support SAN certificates
When I register a SAN certificate using `deis certs:add` only the common name is picked up, not the other SANs (which can be entierly different domain names)

An example certificate can have: `CN=example.com` and `SAN=DNS:example.co.uk,DNS:example.no,DNS:example.io,DNS:my-aweseome-example.com,DNS:other-website.com`

See this for a guide on how to create SAN certificates: http://apetec.com/support/GenerateSAN-CSR.htm

My (hopefully temporary) workaround is to just use platform SSL certificates for now and remove the application SSL solution from my `deis-router`.

bacongobbler:
This feature allows a user to specify a list of SubjectAltNames (SANs)
for a specified certificate. This has the effect of creating multiple
certificate entries in the database (one for each SAN), but it gives the
user the benefit to revoke the certficate for each custom domain.

By default, SANs are **not** added as new entries with `certs:add`. The
user will need to explicitly give each subject alt name they wish to add
as a certificate.

Usage:

```
$ deis certs:add ~/.openssl/star.fishworks.io.cert ~/.openssl/private.key.nopass --subject-alt-name foo.fishworks.io --subject-alt-name bar.fishworks.io
Adding SSL endpoint... done
*.fishworks.io
Adding SSL endpoint foo.fishworks.io...done
Adding SSL endpoint bar.fishworks.io...done
$ deis certs
Common Name       Expires
----------------  ----------------------
*.fishworks.io    2015-08-29T08:17:48UTC
foo.fishworks.io  2015-08-29T08:17:48UTC
bar.fishworks.io  2015-08-29T08:17:48UTC
```

In order for users to update their certificate, they will need to run
`deis certs:remove` for each entry, then re-run `deis certs:add`.

Docopt does not respect ellipsis when '[options]' is present, so I had
to explicitly write out the option in the usage in order for the flag to
work.

Additionally, a new --common-name flag has also been introduced. This is
a temporary workaround for users to add their wildcard certificates to
their custom domain endpoints. This acts the same way where a database
entry is created for each call to `certs:add`. If users want to update
their wildcard certificate, they'll have to update each entry they've
added. This is not the optimal solution, but it provides a way for us to
support wildcard certificates for custom domains.

Usage:

```
$ deis certs:add ~/.openssl/star.fishworks.io.cert ~/.openssl/private.key.nopass --common-name foo.fishworks.io
Adding SSL endpoint foo.fishworks.io...done
$ deis certs
Common Name       Expires
----------------  ----------------------
foo.fishworks.io  2015-08-29T08:17:48UTC
```

closes #3441

### DIFF
--- a/client/deis.py
+++ b/client/deis.py
@@ -1043,18 +1043,35 @@ Make sure that the Controller URI is correct and the server is running.
         """
         Binds a certificate/key pair to an application.
 
-        Usage: deis certs:add <cert> <key>
+        Usage: deis certs:add <cert> <key> [--subject-alt-name=<san>...] [options]
 
         Arguments:
           <cert>
             The public key of the SSL certificate.
           <key>
             The private key of the SSL certificate.
+
+        Options:
+          --common-name=<cname>
+            The common name of the certificate. If none is provided, the controller will
+            interpret the common name from the certificate.
+          --subject-alt-name=<san>...
+            The subject alternate names (SAN) of the certificate. This will create multiple
+            Certificate objects in the controller, one for each SAN.
         """
         cert = args.get('<cert>')
         key = args.get('<key>')
+        self._certs_add(cert, key, args.get('--common-name'))
+        for san in args.get('--subject-alt-name'):
+            self._certs_add(cert, key, san)
+
+    def _certs_add(self, cert, key, common_name=None):
         body = {'certificate': file(cert).read().strip(), 'key': file(key).read().strip()}
-        sys.stdout.write("Adding SSL endpoint... ")
+        if common_name:
+            body['common_name'] = common_name
+            sys.stdout.write("Adding SSL endpoint {}...".format(common_name))
+        else:
+            sys.stdout.write("Adding SSL endpoint... ")
         sys.stdout.flush()
         try:
             progress = TextProgress()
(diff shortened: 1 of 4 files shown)


client/deis.py
```diff
@@ -1043,18 +1043,35 @@ def certs_add(self, args):
         """
         Binds a certificate/key pair to an application.
 
-        Usage: deis certs:add <cert> <key>
+        Usage: deis certs:add <cert> <key> [--subject-alt-name=<san>...] [options]
```
mboersma: I see what you mean here about `docopt` swallowing the options. This is a reasonable workaround.

carmstrong: Is it possible to specify multiple SANs with one flag, separated by commas or something? Just a thought - some certs support up to 2000 SANs or some large number, and it may be a pain to specify them all by repeating the flag. Nitpick.

bacongobbler: That's possible. We could just make it a single flag called `--subject-alt-names=foo.example.com,bar.example.com` and split them up by the comma. There doesn't seem to be any arguments against that in http://pubs.opengroup.org/stage7tc1/basedefs/V1_chap12.html

carmstrong: :+1: If that's an easy change, I'd recommend that. I have a feeling the multiple flags will get unruly very quickly.

bacongobbler:
Please let me know if I completely missed the target with this feature -- finding information on installing certificates with SANs on nginx is very sparse so I thought that it's as simple as installing the cert/key for the specified domain.

carmstrong:
Code LGTM.

mboersma:
Code LGTM.

carmstrong:
@bacongobbler let's make sure this works for Blystad before we ship v1.6.0.

The router image is `deisci/controller:jenkins-test-ec2-658` but we should also build a custom client.

PR #3561, opened 2015-04-23
feat(contrib): optional graceful shutdown

iancoffey:
This is an attempt to add logic to gracefully remove a node from a cluster when halt, shutdown or reboot units are called on the host.

- [x] Where should the graceful shutdown script actually live? Its just a gist right now for convenience. I guess it could be in write files, but I feel like that size script prob shouldnt be directly in user-data.
- [x] The graceful stop unit does indeed get called during shutdown, halt or reboot, but seems to be unable to complete. Im not sure why yet. It runs as expected when called any other way.
- [x] Refactor graceful stop unit to be less brittle.
- [ ] Are there any other steps we can take to make this smoother?

### DIFF
--- a/contrib/coreos/user-data.example
+++ b/contrib/coreos/user-data.example
@@ -43,6 +43,22 @@ coreos:
       Type=oneshot
       ExecStart=/usr/bin/systemctl stop update-engine.service
       ExecStartPost=/usr/bin/systemctl mask update-engine.service
+  - name: graceful-deis-shutdown.service
+    content: |
+      [Unit]
+      Description=Clean up
+      DefaultDependencies=no
+      After=fleet.service etcd.service docker.service docker.socket deis-store-admin.service deis-store-daemon.service deis-store-volume.service deis-store-monitor.service
+      Requires=fleet.service etcd.service deis-store-admin.service deis-store-daemon.service deis-store-volume.service docker.service docker.socket deis-store-monitor.service
+
+      [Install]
+      WantedBy=shutdown.target halt.target reboot.target
+
+      [Service]
+      ExecStop=/opt/bin/graceful-shutdown.sh --really
+      Type=oneshot
+      TimeoutSec=1200
+      RemainAfterExit=yes
   - name: install-deisctl.service
     command: start
     content: |
@@ -176,3 +192,61 @@ write_files:
     content: |
       [Coredump]
       Storage=none
+  - path: /opt/bin/graceful-shutdown.sh
+    permissions: '0755'
+    content: |
+      #!/usr/bin/bash
+      if [ "$1" != '--really' ]; then
+        echo "command must be run as: $0 --really"
+        exit 1
+      fi
+      # procedure requires the store-admin
+      ADMIN_RUNNING=$(docker inspect --format="{{ .State.Running }}" deis-store-admin)
+      if [ $? -eq 1 ] || [ "$ADMIN_RUNNING" == "false" ]; then
+        echo "deis-store-admin container is required for graceful shutdown"
+        exit 2
+      fi
+      set -e -x -o pipefail
+      # determine osd id
+      CURRENT_STATUS=$(/usr/bin/docker exec deis-store-admin ceph health | awk '{print $1}')
+      OSD_HOSTS=($(/usr/bin/etcdctl ls /deis/store/hosts/| awk -F'/' '{print $5}'))
+      for HOST in "${OSD_HOSTS[@]}"
+      do
+        PUBLIC_IP=$(fleetctl list-machines -fields="machine,ip" -full -no-legend| grep `cat /etc/machine-id` | awk '{print $2}')
+        if [ "$HOST" = "$PUBLIC_IP" ] ; then
+          OSD_ID=$(/usr/bin/etcdctl get /deis/store/osds/$PUBLIC_IP)
+          break
+        fi
+      done
+      # if we own an osd and its healthy, try to gracefully remove it
+      if [ ! -z "$OSD_ID" ] && [[ "$CURRENT_STATUS" == *"HEALTH_OK"* ]] && [ ${#OSD_HOSTS[@]} -gt "3" ]; then
+        /usr/bin/docker exec deis-store-admin ceph osd out $OSD_ID
+        sleep 30
+        TIMEWAITED=0
+        until [[ $(/usr/bin/docker exec deis-store-admin ceph health) == *"HEALTH_OK"* ]]
+        do
+          if [ $TIMEWAITED -gt "1200" ]
+          then
+            echo "ceph graceful removal timeout exceeded"
+            break
+          fi
+          echo "waiting" && sleep 5
+          TIMEWAITED=$((TIMEWAITED+5))
+        done
+        /usr/bin/docker stop deis-store-daemon
+        /usr/bin/docker exec deis-store-admin ceph osd crush remove osd.$OSD_ID
+        /usr/bin/docker exec deis-store-admin ceph auth del osd.$OSD_ID
+        /usr/bin/docker exec deis-store-admin ceph osd rm $OSD_ID
+        /usr/bin/etcdctl rm /deis/store/osds/$PUBLIC_IP
+        etcdctl rm /deis/store/hosts/$PUBLIC_IP && sleep 10
+        # remove ceph mon
+        /usr/bin/docker stop deis-store-monitor || true
+        /usr/bin/docker exec deis-store-admin ceph mon remove `hostname -f` # fixme
+        /usr/bin/docker stop deis-store-metadata || true
+      fi
+      NODE=$(curl -L http://127.0.0.1:7001/v2/admin/machines/`cat /etc/machine-id`)
+      # remove from etcd cluster
+      if [ $NODE != 'null' ]; then
+        /usr/bin/curl -L -XDELETE http://127.0.0.1:7001/v2/admin/machines/`cat /etc/machine-id`
+      fi
+manage_etc_hosts: localhost
--- a/docs/managing_deis/add_remove_host.rst
+++ b/docs/managing_deis/add_remove_host.rst
@@ -284,3 +284,28 @@ This can be achieved by making a request to the etcd API. See `remove machines`_
 .. _`remove machines`: https://coreos.com/docs/distributed-configuration/etcd-api/#remove-machines
 .. _`removing monitors`: http://ceph.com/docs/hammer/rados/operations/add-or-rm-mons/#removing-monitors
 .. _`removing OSDs`: http://docs.ceph.com/docs/hammer/rados/operations/add-or-rm-osds/#removing-osds-manual
+
+Automatic Host Removal
+======================
+
+The ``contrib/user-data.example`` provides a unit which provides some experimental logic to clean-up a Deis
+node's Ceph and etcd membership before reboot, shutdown or halt events. Currently, the procedure will
+also be triggered if it is manually stopped via systemctl.
+
+The unit requires that the optional ``deis-store-admin`` component is installed.
+
+.. code-block:: console
+
+    root@deis-1:/# deisctl install store-admin
+    root@deis-1:/# deisctl start store-admin
+
+To enable the feature you must enable and start the unit:
+
+.. code-block:: console
+
+    root@deis-1:/# systemctl enable graceful-deis-shutdown
+    root@deis-1:/# systemctl start graceful-deis-shutdown
+
+The unit is now active and will now be stopped ahead of the store components, Docker and etcd.
+The script ``/opt/bin/graceful-shutdown.sh`` will determine if the Ceph cluster is healthy, and attempt
+to remove this node from the Deis store components if there are greater than 3 Ceph nodes in the cluster.


deis-admin:

mboersma:
Jenkins, add to whitelist.

aledbf:
@iancoffey just a couple of questions:
- Why the removal of the node from etcd? (that will be done automatically by etcd if the node never joins the cluster again)
- The `ceph` steps are interesting but what's the costs of doing that?
  - you are not deleting the `ceph` data from the node
  - creating a new `osd` from the same node can return a different id and that means potential data duplication.
  - during the initial synchronization (of a new node) the `ceph` gateway could not be available

The final fleet stop is not necessary (and it could be the guilty for the not completion of the shutdown)

mboersma:
> Where should the graceful shutdown script actually live?

(It is [currently at this gist](https://gist.githubusercontent.com/iancoffey/7f2ac99cfe6b47397812/raw/4250f91a5ffdd5f53226883c8b728261ea9be193/gistfile1.sh) just to help reviewers.) Ideally we wouldn't `curl` it from elsewhere, since that fails sometimes when a machine is starting up, even with the `--retry` flag. And `user-data` is already unwieldy, but including it directly there would be more reliable. At worst, we can host it for S3 download like we do some other artifacts.

iancoffey:
good points @aledbf 
- Implicitly removing the node from etcd leaves no room for error - a dying server with filesystem issues could potentially hang around, making things weird.
- The cost of the ceph recovery is usually about 2-6 minutes of waiting for the cluster to recover before moving to shutdown.
- Im curious why we wouldnt want to be certain fleet is down. IMO, it would be even better to preemptively be able to tell fleet "this suckers going down, reschedule accordingly" (which we cant).

aledbf:
> Im curious why we wouldn't want to be certain fleet is down

Because you already remove the node from `etcd` cluster.

> The cost of the `ceph` recovery is usually about 2-6 minutes of waiting for the cluster to recover before moving to shutdown.

That is true if you don't remove the `ceph` data from the node and [this step](https://github.com/deis/deis/blob/master/store/daemon/bin/boot#L23) return the same id. If not this node will be considered a new "pristine" `ceph` node and it will replicate all the data (and the time is proportional to the current content).

*Just in case I think that this is a good idea :) and it could solve (at least partially) the autoscale issues with `ceph` in AWS.*

mboersma:
This needs rebasing now, sorry about that.

mboersma:
`make -C docs/ test` is very strict about formatting:
```
Warning, treated as error:
/var/lib/jenkins/jobs/test-ec2/workspace/src/github.com/deis/deis/docs/managing_deis/add_remove_host.rst:126: ERROR: Unexpected indentation.

make: *** [test] Error 1
make: Leaving directory `/var/lib/jenkins/jobs/test-ec2/workspace/src/github.com/deis/deis/docs'
```

carmstrong:
@iancoffey How did you solve the script not completing?

iancoffey:
@carmstrong Just pushed some updates, I changed the unit file around a lot based on some feedback in #systemd. It works but it has a couple flaws that Im not crazy about.

One is that the script will run if you just stop docker. Another is that accidentally stopping the graceful-deis-shutdown unit will also trigger the sequence start.

I have some ideas for cleaning that up, I just added a checkbox for it. Would love input if theres any.

iancoffey:
Seems this is kinda blocked on https://github.com/deis/deis/pull/3645 since the cluster is never healthy, so it gets skipped.

Before rebasing though, it did what I expected - take the osd out, block shutdown for about 2-3 mins and let the cluster get healthy before removing itself.

carmstrong:
> Seems this is kinda blocked on #3645 since the cluster is never healthy, so it gets skipped.

#3645 now suppresses the warning for old clusters that are upgraded (as well as new ones provisioned), so we should be good once it's merged.

iancoffey:
Looks like https://github.com/deis/deis/pull/3645 is merged - anything else we should do here?

docs/managing_deis/add_remove_host.rst
```diff
@@ -284,3 +284,29 @@ This can be achieved by making a request to the etcd API. See `remove machines`_
 .. _`remove machines`: https://coreos.com/docs/distributed-configuration/etcd-api/#remove-machines
 .. _`removing monitors`: http://ceph.com/docs/hammer/rados/operations/add-or-rm-mons/#removing-monitors
 .. _`removing OSDs`: http://docs.ceph.com/docs/hammer/rados/operations/add-or-rm-osds/#removing-osds-manual
+
+Graceful Shutdown
+~~~~~~~~~~~~~~~~~
+
+The contrib user-data provides a unit which provides some experimental logic to clean-up a Deis
+nodes ceph and etcd membership before reboot, shutdown or halt events. Currently, the procedure will
```
bacongobbler: s/nodes/node's

docs/managing_deis/add_remove_host.rst
```diff
@@ -284,3 +284,29 @@ This can be achieved by making a request to the etcd API. See `remove machines`_
...
+also be triggered if it is manually stopped via systemctl.
```
bacongobbler: What would happen if you gracefully shut down the node with `sudo shutdown 00:00` and didn't install deis-store-admin? Will this script still be run?

bacongobbler:
Jenkins failed this due to #3690 but I saw it go green earlier so I think we're good here. LGTM asides for my one comment. I'm happy this is optional so that we can test it before actually enabling it by default!

iancoffey:
Cool, updated. :beers: Agree about this needing to be optional - my hope is that we learn a lot about failover along the way.

contrib/coreos/user-data.example
```diff
@@ -176,3 +190,70 @@ write_files:
     content: |
       [Coredump]
       Storage=none
+  - path: /etc/systemd/system/ntpd.service.d/debug.conf
```
carmstrong: This unit and the following fleet custom binary should be removed - botched rebase, it looks like.

contrib/coreos/user-data.example
```diff
@@ -176,3 +190,70 @@ write_files:
...
+        echo "deis-store-admin container is required for graceful shutdown"
+        exit 2
+      fi
+      set -e -x -o pipefail
+      # determine osd id
+      CURRENT_STATUS=$(/usr/bin/docker exec deis-store-admin ceph health | awk '{print $1}')
+      OSD_HOSTS=($(/usr/bin/etcdctl ls /deis/store/hosts/| awk -F'/' '{print $5}'))
+      for HOST in "${OSD_HOSTS[@]}"
+      do
+        PUBLIC_IP=$(/usr/bin/etcdctl get /_coreos.com/fleet/machines/`cat /etc/machine-id`/object | /opt/bin/jq -r '.PublicIP')
```
carmstrong: Is this really the public IP? Can't we use `fleetctl list-machines` here, or the fleet HTTP API? https://github.com/coreos/fleet/blob/master/Documentation/api-v1.md#list-machines

contrib/coreos/user-data.example
iancoffey: yup, totally botched

docs/managing_deis/add_remove_host.rst
```diff
@@ -284,3 +284,29 @@ This can be achieved by making a request to the etcd API. See `remove machines`_
 .. _`remove machines`: https://coreos.com/docs/distributed-configuration/etcd-api/#remove-machines
 .. _`removing monitors`: http://ceph.com/docs/hammer/rados/operations/add-or-rm-mons/#removing-monitors
 .. _`removing OSDs`: http://docs.ceph.com/docs/hammer/rados/operations/add-or-rm-osds/#removing-osds-manual
+
+Graceful Shutdown
+~~~~~~~~~~~~~~~~~
+
+The contrib user-data provides a unit which provides some experimental logic to clean-up a Deis
```
carmstrong: Should probably be contrib/user-data wrapped in double ticks so it's formatted like code. Ceph should also be capitalized.

docs/managing_deis/add_remove_host.rst
```diff
@@ -284,3 +284,29 @@ This can be achieved by making a request to the etcd API. See `remove machines`_
...
+node's ceph and etcd membership before reboot, shutdown or halt events. Currently, the procedure will
+also be triggered if it is manually stopped via systemctl.
+
+The unit requires that the optional deis-store-admin component is installed.
```
carmstrong: We should wrap deis-store-admin in double backticks as well.

docs/managing_deis/add_remove_host.rst
```diff
@@ -284,3 +284,29 @@ This can be achieved by making a request to the etcd API. See `remove machines`_
...
+    root@deis-1:/# deisctl start deis-store-admin
+
+To enable the feature you must enable and start the unit:
+
+.. code-block:: console
+
+    root@deis-1:/# systemctl enable graceful-deis-shutdown
+    root@deis-1:/# systemctl start graceful-deis-shutdown
+
+The unit is now active and will now be stopped ahead of the store components, docker and etcd.
```
carmstrong: Docker should be capitalized

docs/managing_deis/add_remove_host.rst
```diff
@@ -284,3 +284,29 @@ This can be achieved by making a request to the etcd API. See `remove machines`_
...
+The script (/opt/bin/graceful-shutdown.sh) will determine if the ceph cluster is healthy, and attempt
```
carmstrong: /opt/bin/graceful-shutdown.sh should be wrapped in double-ticks, and Ceph should be capitalized in both places.

contrib/coreos/user-data.example
iancoffey: This and the docs are updated

carmstrong:
Code LGTM. I'd like to manually test this today, and then we should be good to go. :+1:

mboersma:
This code LGTM, although it needs more manual testing before it is merged. Really nice work @iancoffey. Thanks for sticking with this!

docs/managing_deis/add_remove_host.rst
```diff
@@ -284,3 +284,28 @@ This can be achieved by making a request to the etcd API. See `remove machines`_
 .. _`remove machines`: https://coreos.com/docs/distributed-configuration/etcd-api/#remove-machines
 .. _`removing monitors`: http://ceph.com/docs/hammer/rados/operations/add-or-rm-mons/#removing-monitors
 .. _`removing OSDs`: http://docs.ceph.com/docs/hammer/rados/operations/add-or-rm-osds/#removing-osds-manual
+
+Graceful Shutdown
```
carmstrong: Maybe we should call this "Automatic Host Removal" to be more in-line with the page? We could also make it a top-level heading.

iancoffey: Makes sense, updated.

docs/managing_deis/add_remove_host.rst
```diff
@@ -284,3 +284,28 @@ This can be achieved by making a request to the etcd API. See `remove machines`_
...
+
+The ``contrib/user-data.example`` provides a unit which provides some experimental logic to clean-up a Deis
+node's Ceph and etcd membership before reboot, shutdown or halt events. Currently, the procedure will
+also be triggered if it is manually stopped via systemctl.
+
+The unit requires that the optional ``deis-store-admin`` component is installed.
+
+.. code-block:: console
+
+    root@deis-1:/# deisctl install deis-store-admin
```
bacongobbler: this should be `deisctl install store-admin`. The `deis-` is implied by deisctl:

```
><> deisctl install store-admin
deis-store-admin.service: loaded                                 
><> deisctl start store-admin
deis-store-admin.service: activating/start-pre                                 
deis-store-admin.service: active/running
```

iancoffey: oh good catch

iancoffey: updated

contrib/coreos/user-data.example
```diff
@@ -176,3 +192,60 @@ write_files:
...
+      for HOST in "${OSD_HOSTS[@]}"
+      do
+        PUBLIC_IP=$(fleetctl list-machines -fields="machine,ip" -full -no-legend| grep `cat /etc/machine-id` | awk '{print $2}')
+        if [ "$HOST" = "$PUBLIC_IP" ] ; then
+          OSD_ID=$(/usr/bin/etcdctl get /deis/store/osds/$PUBLIC_IP)
+          break
+        fi
+      done
+      # if we own an osd and its healthy, try to gracefully remove it
+      if [ ! -z "$OSD_ID" ] && [[ "$CURRENT_STATUS" == *"HEALTH_OK"* ]] && [ ${#OSD_HOSTS[@]} -gt "3" ]; then
```
bacongobbler: how come we're checking for > 3 OSDs? Shouldn't we be checking for >= 3, or is this script only supposed to be run if we have more than 3 nodes?

bacongobbler: I guess it's just to maintain quorum?

iancoffey: Yep - the idea is to only run the operation when Ceph can return to full health before going down.

bacongobbler:
I just tried this with a 5-node cluster and it failed :(

logs from the failure: https://gist.github.com/bacongobbler/0d720ba7c51cdcdc9e81

iancoffey:
@bacongobbler what provider did you use to create this cluster? Ive actually only tested this on ec2.

bacongobbler:
I tested this on vagrant

bacongobbler:
Odd:

```
><> vs deis-01 -c "hostname -f"
hostname: Unknown host
Connection to 127.0.0.1 closed.
><> vs deis-01 -c "hostname"
deis-01
Connection to 127.0.0.1 closed.
```

bacongobbler:
Maybe in order to get this merged, we can document that this has been only tested on EC2 and other providers are unsupported at this time?

iancoffey:
Yeah, that probably makes sense, since I wont be able to test every provider thoroughly.

Id like this to work for Vagrant at least though, but Im having bad getting a working coreos vagrant cluster  at the moment - let me figure that our and adjust the docs, and then maybe this will be good to go.

bacongobbler:
I found the issue. It's because there's no /etc/hosts file in the vagrant hosts, which is why it wouldn't work. Derp!

```
core@deis-01 ~ $ cat /etc/hostname  
deis-01
core@deis-01 ~ $ hostname
deis-01
core@deis-01 ~ $ hostname -f
hostname: Unknown host
core@deis-01 ~ $ sudo vim /etc/hosts
core@deis-01 ~ $ cat /etc/hosts 
127.0.0.1	deis-01
core@deis-01 ~ $ hostname -f
deis-01
```

iancoffey:
I think this _should_ work now on Vagrant with the addition of the ``manage_etc_hosts`` entry.

I cant tell for sure because I have not been able to get a healthy Vagrant Deis cluster up. Ill bring that up as a separate issue.

carmstrong:
> I think this should work now on Vagrant with the addition of the manage_etc_hosts entry.

I'm not too worried about this working for Vagrant outside of us being able to easily test it. We don't recommend long-lived Vagrant clusters anyhow due to issues like clock drift on VM/laptop suspension, etc., so it's not a major concern.

iancoffey:
Cool - I retested this on ec2 with 4 nodes, went to plan.

carmstrong:
Code LGTM! Once @bacongobbler has also confirmed this works well in manual testing, let's get it out there and get people testing it.

bacongobbler:
My cluster is still recovering, but the OSD did indeed remove itself:

```
root@ip-10-21-1-73:/# ceph -s
    cluster 607ac12e-175b-4d3e-aaa2-04fa8da839eb
     health HEALTH_WARN
            11 pgs degraded
            1 pgs stuck unclean
            recovery 21/270 objects degraded (7.778%)
     monmap e5: 5 mons at {ip-10-21-1-72.us-west-1.compute.internal=10.21.1.72:6789/0,ip-10-21-1-73.us-west-1.compute.internal=10.21.1.73:6789/0,ip-10-21-1-74.us-west-1.compute.internal=10.21.1.74:6789/0,ip-10-21-2-196.us-west-1.compute.internal=10.21.2.196:6789/0,ip-10-21-2-197.us-west-1.compute.internal=10.21.2.197:6789/0}
            election epoch 14, quorum 0,1,2,3,4 ip-10-21-1-72.us-west-1.compute.internal,ip-10-21-1-73.us-west-1.compute.internal,ip-10-21-1-74.us-west-1.compute.internal,ip-10-21-2-196.us-west-1.compute.internal,ip-10-21-2-197.us-west-1.compute.internal
     mdsmap e7: 1/1/1 up {0=ip-10-21-1-73.us-west-1.compute.internal=up:active}, 3 up:standby
     osdmap e30: 4 osds: 4 up, 4 in
      pgmap v83: 704 pgs, 11 pools, 7624 kB data, 90 objects
            24673 MB used, 330 GB / 374 GB avail
            21/270 objects degraded (7.778%)
                 693 active+clean
                  11 active+degraded
  client io 562 B/s rd, 4874 B/s wr, 0 op/s
```

LGTM!

Issue #2812, opened 2014-12-29
Make store component optional by allowing external database, registry, and log collector


mgartner:
Related to #2252 and #2295.

It seems to me like the store and Ceph add a lot of complexity to the Deis cluster. The store component is obviously useful, maybe necessary for applications that require everything to be self-contained in the cluster. However, for someone like me with little Devops experience, it would be nice to have the option to offload some of this complexity to other services that require little or no management on my part.

For example, from an AWS perspective I could ideally replace the database component with RDS, remove the logger component and have the logspouts post logs to a service like Loggly or Papertrail, and have the the registry use S3 as storage. If I'm correct, then no components would rely on the store and deis-store-daemon, deis-store-gateway, deis-store-metdata, and deis-store-monitor could all be removed from the cluster.

Some benefits that I could see from my perspective:

1. No need for Ceph experience to debug issues with the store. (I took down a node in the cluster without fully removing it from Ceph, and managed to leave my cluster in an unrecoverable state.)
2. Nodes are stateless - which make it much easier to scale down the cluster. No need to go through the process of tearing-down Ceph - just shut down the server.
3. Core components use less of the cluster's resources.

This could potentially make upgrading the Deis platform easier as well, but I'm not sure.

I just wanted to throw out this idea to see what the potential downsides are for this strategy, and what the Deis team's opinion is on supporting something like this.

krancour:
I'm quite interested in this as well for some of the same reasons that the OP mentioned.  By keeping all components within the CoreOS cluster stateless, I assume both autoscaling and upgrades will be easier.

Like the OP, as long as I can use a (highly available) external DB (RDS in my case), S3 to back my registry, and if I am already draining logs to elsewhere (Splunk in my case), then I really wouldn't care too much about not being able to get logs via deis logs.

+1 for this.

I think this is technically already doable, but  comprehensive documentation on the process seems to be missing.

Blystad:
I'm working on looking into this. We want to get rid of Ceph since we're already running on Amazon infrastructure were we have everything which should be necessary.

I'll post my findings asap. Currently fighting with registry....

Blystad:
Alright! I have a Deis environment now running completely without deis-store.

I'll post my findings now so we can figure out the way ahead:

First of all, my environment is now running with:
* PostgreSQL running on Amazon RDS
* Amazon S3 bucket in the same region as Deis (eu-west-1) 
[...]

Issue #3262, opened 2015-03-10
Different behaviour between plain docker and git+Dockerfile


haf:
I've been banging my head against this for about 2 hours now without a solution in sight. Not to mention I can't reproduce the error locally with docker alone -- but since it's very easy to test, I figured you might be able to assist me?

I have created this repo https://github.com/haf/sample-hello-fsharp as an addition to your selection of supported languages.

---

**The core of the issue is this:**

When I build the dockerfile in docker - in vagrant or in prod, doesn't matter, I can start the app, and the web server starts successfully. However, when I deploy using git push, and let deis do the work, then it seems that the app doesn't have network connectivity.

---

Here's the successful output using plain docker:

```
docker build -t sample-hello-fsharp .
Sending build context to Docker daemon 3.072 kB
Sending build context to Docker daemon
Step 0 : FROM haaf/fsharp-http-app
Pulling repository haaf/fsharp-http-app
309b1c981129: Download complete
511136ea3c5a: Download complete
5b12ef8fd570: Download complete
88f9454e60dd: Download complete
0eb34c16531d: Download complete
b8ad31fbc0fc: Download complete
e95110772f15: Download complete
bdbf715346cb: Download complete
4df30cac092a: Download complete
e2177bd424cd: Download complete
bc194e136392: Download complete
97e8a70b61c7: Download complete
8289d37fee12: Download complete
6d46e1c24343: Download complete
1a84ae484b3f: Download complete
Status: Downloaded newer image for haaf/fsharp-http-app:latest
 ---> 309b1c981129
Step 1 : MAINTAINER Henrik Feldt <henrik@haf.se>
 ---> Running in 2a9d31832fe0
 ---> 03c9fb7df16b
Removing intermediate container 2a9d31832fe0
Step 2 : COPY . /app
 ---> e12636d354f3
Removing intermediate container f7b1c4eae7c5
Step 3 : EXPOSE 8083
 ---> Running in d27eefc9f028
 ---> e1a2d9654252
Removing intermediate container d27eefc9f028
Step 4 : ENTRYPOINT /usr/bin/fsharpi Program.fsx
 ---> Running in 850fa1fcd49f
 ---> 6304d4acb169
Removing intermediate container 850fa1fcd49f
Successfully built 6304d4acb169
core@deis-3 ~/tmp $ docker run -it sample-hello-fsharp
found: /app/paket.dependencies
Resolving packages:
 - Suave 0.26.0
 - FSharp.Core 4.0.0
[...]

PR #3565, opened 2015-04-23
feature(deisctl): add install wizard

krancour:
Addresses #2812

I would not consider actually merging this!  I'm putting it out there as a straw man for constructive criticism on the approach.  I could also use some help figuring out how to test this.

Documentation is missing.  I'll amend the PR with doc updates once we settle on an approach.


bacongobbler:
How do you want us to test this? Without proper documentation we don't have anything to nitpick or follow when testing these new features.

As for tests, I'd break out the `start/stop/install/uninstallServices()` into two separate functions:

 - `getServices()` which takes a supplied install config (from the wizard or from `start platform`) that will return the services requested to install
 - `startServices()` which tells the backend to boot said services

There won't be a way to test `startServices()` because it's hard to mock fleet, but we should be able to write unit tests for supplying a config and getting back a list of services to install.

krancour:
@bacongobbler Sorry... like I said... a straw man.  Didn't invest time in documentation knowing this was all likely to change.  Here's some micro doc:

Traditional install; works the same as it always has:
```
$ deisctl install platform
```

(Caveat to the above is that if you have previously run the wizard, then uninstalled, and are now re-installing, the options you selected the first time around will still be respected since they're stored in etcd.)

Wizard-based:
```
$ deisctl install platform --wizard
```

Or this works too:
```
$ deisctl install platform -w
```

Unattended:
```
$ deisctl install platform --config=<file>
```

File takes the following form:
```
/any/etcd/key=value
/another/etcd/key=value
```

Re: breaking functions like `start/stopServices()` into separate `getServices()` and `start/stopServices()` functions as you have suggested-- I totally like the idea (it sounds cleaner), but even just looking at existing code in the master branch, wouldn't such a change be complicated by the logic needed to start/stop things in a certain order and wait for certain things?  For example, if starting the whole platform, we couldn't just rip through a list of `[ store-monitor, store-daemon, ..., publisher, router@* ]` and start each in turn.  There is (existing) logic that knows where to wait and what things can be started in parallel.  Same goes for the order in which things comes down.

Maybe `getServices()` could return a two dimensional array?  We could then iterate over _groups_ of things to start / stop / etc., with `wg.Wait()` between groups.  If we did that, we'd have to find some clever way to also maintain the output channel messages about different subsystems.  It would call for a more complex data structure than the two dimensional array.  Something like the following:

```
[
  { subsystem: 'subsys-1', services: [ [ 'service-1' ], [ 'service-2', 'service-3' ] ] },
  { subsystem: 'subsys-2', services: [ [ 'service-4' ] ] }
]
```

Seems doable, but then knowing that the order to stop thing varies from the order in which to start things, requiring us to wait in different points, what we'd end up with is logic that looks almost identical to the existing logic, just in a different method.  So I wonder if we'd just end up with code very similar to what we started with.

Just thinking out lout.

krancour:
Cross posting: #3262 affects / is affected by this PR.

krancour:
@bacongobbler thinking more about your advice to separate some logic from the `start/stop/etc()` functions into a separate `getServices()` function.  I still think that complicates things a little, but I also think I get where you were going with it now.  You mentioned that it's tough to actually test against `fleet`, so was your line of thinking that the data structure produced by `getServices()` is something that we could more easily make assertions against without having to involve `fleet`?  Sorry if I didn't pick up on that right away.

What if there's another way to make unit testing easier wherever actual `fleet` integration is a concern we wish to avoid?  I come (most recently) from the Ruby world where its dynamic nature means we don't think twice about mocking/stubbing a dependency.  With Go being compiled and statically typed, that's a more difficult proposition here, but not impossible.  What if I were to invest the time in creating an alternate implementation of the `Backend` interface that's specifically aimed at enabling more isolated tests?  Such an implementation could simply keep track of what services are loaded, started, stopped, destroyed, etc.  This could be injected into any other code under test that's normally dependent on `fleet` and any such test can conclude by querying the test implementation of the `Backend` interface to assert that the things that were suppose to happen on the backend _would have_?

In this manner, we could _easily_ test that the right things get loaded and started in the various scenarios where a an operator opts into or out of a particular component or subsystem.

I think this would be useful beyond testing just this install wizard feature as well.  It seems it could be used in a lot of places where tests are maybe lighter than they should be, owing to their dependency on fleet.

What do you think?

P.S. Testing the interactive aspects of the wizard will still be a pain.  Still thinking through that.  Gexpect looked promising, initially, but now I have my doubts.

bacongobbler:
> was your line of thinking that the data structure produced by getServices() is something that we could more easily make assertions against without having to involve fleet? Sorry if I didn't pick up on that right away.

yes indeed. Some coverage is better than no coverage :)

> What if I were to invest the time in creating an alternate implementation of the Backend interface that's specifically aimed at enabling more isolated tests?

I'd be so happy if you could write out a backend that works for tests. If you can come up with one for deisctl so we can rely on them instead of having to rely on testing by hand, I'll give you all of my shiniest marbles. I once tried to do this but was sidetracked with other feature work so I never got around to it.

krancour:
If I do it, it won't be anything too crazy... just something to keep track of what would have happened on the backend if it were real.  Think that's adequate for now?

bacongobbler:
sure, that'd be adequate

mboersma:
I realize this is a WIP, but just to note this PR currently fails with a `gofmt` error:
```console
make[1]: Entering directory `/var/lib/jenkins/jobs/test-ec2/workspace@2/src/github.com/deis/deis/deisctl'
gofmt -l backend client cmd config utils deisctl.go deisctl_test.go
cmd/cmd.go
cmd/wizard/wizard.go
gofmt check failed.
make[1]: *** [test-style] Error 1
```

bacongobbler:
@krancour according to your latest comment on the bug, it seems like you're retracting from this proposal. In that case, should we close this?

krancour:
@bacongobbler I'm honestly not sure at this point.  I'm strongly of the opinion that it needs to be easier to customize Deis clusters.  #2182 started with a few folks wanting it to be easier to install Deis without the storage subsystem.  I think it's easy enough to make something like that optional by means of an install wizard.  I think such an approach could easily take the pain out of all the most _common_ configuration options (log drains, etc.)

On the other hand, I get the sense that the level of customization that some people do is far more extensive and a framework for permitting such people to orchestrate the launch of Deis with their customizations seems like it would be a useful addition.  (My own use case is Deis with customized unit files and images - storage subsystem + monitoring stack.)

I think at this point I'm convinced that the latter has more value for myself, but I'm also interested in what provides the most value to the _community_.  Maybe there is room for both approaches?  Idk.  I'm really interested in hearing others sound off on this.

carmstrong:
@krancour Does it make sense to close this until an implementation is agreed upon in #3635?

krancour:
@carmstrong I think they may be unrelated.  I think I see this issue as having to do with _what_ runs, while seeing #3635 as having more to do with _where_ things run (as a consequence of a cluster's architecture).

That being said, I have backed off on the notion of an install wizard a little bit, because I think a framework for customizing and orchestrating install and launch of units might be more valuable.  (Tangent: such a framework need not even be Deis-specific.)

Ultimately, I might come back to this, but for right now, I'm not thinking so.  It should come off the radar for now.

PR #3566, opened 2015-04-23
docs(using_deis): explain difference between web and cmd

### LINKED ISSUE #2861: surprising effect of changing from go-buildpack to Dockerfile
I was using the go buildpack and I added a Dockerfile to switch to using that style of build.  I did this in an existing project.

The surprising result is that my containers built fine but the container would not start,  No output from the logs at all.  I eventually found the problem when `deis run <start command>`  succeeded in starting the container with my program inside it but the "automatic" deployment did not.

It wasn't being started because the "dyno type" had gotten set to web because of the original go build pack and the Dockerfile expected a type of "cmd."  `deis ps:scale cmd=1` fixes the problem, but even a warning message might be helpful.

bacongobbler:
Also states that Deis does not automatically migrate from web to
cmd process types when switching contexts.

closes #2861

### DIFF
--- a/docs/using_deis/process-types.rst
+++ b/docs/using_deis/process-types.rst
@@ -92,4 +92,19 @@ Use ``deis scale web=3`` to increase ``web`` processes to 3, for example. Scalin
 process type directly changes the number of :ref:`Containers <container>`
 running that process.
 
+Web vs Cmd Process Types
+------------------------
+
+When deploying to Deis using a Heroku Buildpack, Deis boots the ``web`` process type to boot the
+application server. When you deploy an application that has a Dockerfile or uses :ref:`Docker
+images <using-docker-images>`, Deis boots the ``cmd`` process type. Both act similarly in that they
+are exposed to the router as web applications. However, The ``cmd`` process type is special because
+it is equivalent to running the :ref:`container` without any additional arguments. Every other
+process type is equivalent to running the relevant command that is provided in the Procfile.
+
+When migrating from Heroku Buildpacks to a Docker-based deployment, Deis will not convert ``web``
+process types to ``cmd``. To do this, you'll have to manually scale down the old process type and
+scale the new process type up.
+
+
 .. _`process model`: https://devcenter.heroku.com/articles/process-model


docs/using_deis/process-types.rst
```diff
@@ -92,4 +92,20 @@ Use ``deis scale web=3`` to increase ``web`` processes to 3, for example. Scalin
...
 running that process.
 
+Web vs Cmd Process Types
+------------------------
+
+When deploying to Deis using a Heroku Buildpack, Deis boots the ``web`` process type to boot the
+application server. When you deploy an application that has a Dockerfile or uses :ref:`Docker
+images <using-docker-images>`, Deis boots the ``cmd`` process type. Both are identical to
+each other in that they are exposed to the router as web applications. However, The ``cmd`` process
+type is special in which it is equivalent to running the :ref:`container` without any additional
```
mboersma: I would replace "in which" by "in that" or "because".

docs/using_deis/process-types.rst
```diff
@@ -92,4 +92,20 @@ Use ``deis scale web=3`` to increase ``web`` processes to 3, for example. Scalin
 process type directly changes the number of :ref:`Containers <container>`
 running that process.
 
+Web vs Cmd Process Types
+------------------------
+
+When deploying to Deis using a Heroku Buildpack, Deis boots the ``web`` process type to boot the
+application server. When you deploy an application that has a Dockerfile or uses :ref:`Docker
+images <using-docker-images>`, Deis boots the ``cmd`` process type. Both are identical to
```
mboersma: Maybe instead of "Both are identical to each other" say "Both act similarly"?

docs/using_deis/process-types.rst
```diff
@@ -92,4 +92,20 @@ Use ``deis scale web=3`` to increase ``web`` processes to 3, for example. Scalin
...
+each other in that they are exposed to the router as web applications. However, The ``cmd`` process
+type is special in which it is equivalent to running the :ref:`container` without any additional
+arguments. Every other process type is equivalent to running the relevant command that is provided
+in the Procfile.
+
+When migrating from Heroku Buildpacks to a Docker-based deployment, Deis does not explicitly
```
mboersma: Maybe rephrase to ",Deis will not convert ``web`` process types to ``cmd``. To do this,"

This is all good, just trying to reduce verbiage. :pencil2:

bacongobbler: All good input. I suck at articulating my thoughts, so this is all very appreciated

mboersma:
Code LGTM.

carmstrong:
Much cleaner with @mboersma's comments. Good work, gents.

:ship:

Issue #1027, opened 2014-05-22
running deis-builder with slow/unreliable Internet connectivity


jperville:
Disclaimer: I am living in Myanmar, a country that is great for many things but not for its Internet connectivity. Right now the best I can hope getting is 100kb/s but in practice I have unreliable 20kb/s download speed, with power cuts in the middle to reset my DSL router. To play with deis, my workflow is 1/ `docker pull` on a remote server 2/ `docker save` the images 3/ rsync over night 4/ `docker load` in Vagrant. Now let's describe my issue.

When bootstrapping deis with `make start`, one of the last services to be started is the builder. According to `fleetctl list-units` the `deis-builder.service` remains in `post-start` state for a long time. It is kind of expected because the container's init `docker pull` the nested `deis/slugbuilder` and `deis/slugrunner` images, which weigh around 800Mb of layers together.

Running the `deis-builder` container once is bad because have to download the 800Mb of layers, which is terribly slow and can fail (because of network connectivity issues, eg. if don't recover fast enough from a power cut).

Running the `deis-builder` container multiple times is much worse because the container is recreated from the `deis/builder` image every time the `deis-builder` service is restarted (the systemd service has `ExecStop="docker rm -f deis-builder"`).

How to make this more efficient?:
- is it possible to include the two nested images in the `docker build deis/builder` so that they are not redownloaded every time the `deis-builder` container is recreated
- is it possible to cache the downloaded layer using a local repository (see #952)
- is it possible to force docker to speak plain http instead of https, so that I can cache the layers with squid?

PR #3567, opened 2015-04-23
chore(builder): migrate to heroku's cedar stack

bacongobbler:
Heroku now officially maintains a Docker image for their cedar stack.
The maintainer of cedarish is deprecating it in favour of this image,
so we should migrate over such that we emulate their environment
perfectly.

Since we no longer build a root filesystem and package it into a tarball,
this speeds up development time when compiling builder dramatically
with the tradeoff of having to fetch heroku/cedar:14 on boot, which is
slightly slower than a `docker import`. However, the difference
between relying on your local connection to pull cedarish from S3
vs. relying on a cloud provider's bandwidth to pull the image on boot
is quite significant, so really it's a question of tomato vs. tomatoe.

### DIFF
--- a/builder/Makefile
+++ b/builder/Makefile
@@ -74,8 +74,3 @@ test-style:
 	@$(GOFMT) $(GO_PACKAGES) $(GO_FILES) | read; if [ $$? == 0 ]; then echo "gofmt check failed."; exit 1; fi
 	$(GOVET) $(repo_path) $(GO_PACKAGES_REPO_PATH)
 	$(GOLINT) ./...
-
-cedarish/build:
-	mkdir -p build
-	docker pull progrium/cedarish:latest
-	docker save progrium/cedarish:latest | gzip -9 > build/progrium_cedarish_$(shell date "+%Y_%m_%d").tar.gz
--- a/builder/image/Dockerfile
+++ b/builder/image/Dockerfile
@@ -38,11 +38,6 @@ RUN useradd -d $GITHOME $GITUSER
 RUN mkdir -p $GITHOME/.ssh && chown git:git $GITHOME/.ssh
 RUN chown -R $GITUSER:$GITUSER $GITHOME
 
-# HACK: import progrium/cedarish as a tarball
-# see https://github.com/deis/deis/issues/1027
-RUN curl -#SL -o /progrium_cedarish.tar.gz \
-    https://github.com/progrium/cedarish/releases/download/v3/cedarish-cedar14_v3.tar.gz
-
 # define the execution environment
 # use VOLUME to remove /var/lib/docker from copy-on-write for performance
 # we don't want to stack overlay filesystems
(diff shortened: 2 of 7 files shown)


bacongobbler:
I was thinking about https://github.com/deis/deis/issues/1027 when thinking of this PR but I felt like it's such a small subset of the user base running dev clusters on Vagrant.

aledbf:
Code LGTM.

mboersma:
This looks good but needs to be tested with our example apps before merging.

carmstrong:
> This looks good but needs to be tested with our example apps before merging.

:+1:

bacongobbler:
I deployed both example-go and https://github.com/bacongobbler/uuid-generator to Deis with this PR without any issues. Do we want to test every app with this change?

bacongobbler:
Other apps tested:

 - scala
 - java-jetty
 - nodejs-express
 - php
 - ruby-sinatra
 - clojure

All of them work great. Internally cedarish was always using the script form heroku/stack-images, so it's no surprise that the cut-over works as the filesystem is practically identical. :)

carmstrong:
Code LGTM. Let's see if @mboersma has any objections, but I'm fine to :ship: based on the testing @bacongobbler has done.

bacongobbler:
As for upgrades, this should be completely transparent to the users on buildpacks. Any new pushes will just use the new slugrunner image based on cedar, all the image layers still exist in the registry for slugrunner so rollbacks should work as well, but that might be something good to test during our release testing

mboersma:
I tested `git push deis master` and `for i in $(seq 1 5); do deis scale web=20; sleep 1; deis scale web=0; sleep 1; done` with:
- [example-go](https://github.com/deis/example-go)
- [example-play](https://github.com/deis/example-play)
- [example-python-django](https://github.com/deis/example-python-django)
- [example-python-flask](https://github.com/deis/example-python-flask)

They were all fine with `heroku/cedar:14`. But [example-perl](https://github.com/deis/example-perl) failed with this:
```console
$ git push deis master
Counting objects: 34, done.
Compressing objects: 100% (29/29), done.
Writing objects: 100% (34/34), 8.73 KiB | 0 bytes/s, done.
Total 34 (delta 10), reused 0 (delta 0)
-----> Unable to select a buildpack
To ssh://git@deis.local3.deisapp.com:2222/deluxe-original.git
 ! [remote rejected] master -> master (pre-receive hook declined)
error: failed to push some refs to 'ssh://git@deis.local3.deisapp.com:2222/deluxe-original.git'
```
That perl buildpack issue is probably not unique to this PR, but I will test against v1.5.2 to confirm. Otherwise this is looking good.

mboersma:
(I confirmed that [example-perl](https://github.com/deis/example-perl) fails the same way on Deis v1.5.2. I'll make an issue there so it can be fixed.)

LGTM.

PR #3571, opened 2015-04-23
chore(Godeps): bump libcontainer

ngpestelos:


bacongobbler:
other than for the sake of bumping versions, is there a specific reason you bumped libcontainer to that commit?

ngpestelos:
No other reason. I noticed the difference when I ran `godep save` in my fork.

bacongobbler:
We don't actually use these libraries. `fleet` does, however, which is something that we use in `deisctl`. Let's leave bumping libcontainer up to them. Why fix what ain't broken? :D

ngpestelos:
OK, thank you.

PR #3574, opened 2015-04-24
feat(router): customize error_log level via etcd

### LINKED ISSUE #3560: deis-router error_log should have etcd configurable error level
When debugging issues with `deis-router` it might be difficult to track the error down since we are relying on the default logging level in `error_log` - `error`. While there are log levels going all the way down to `debug`. [ngx_core_module#error_log](http://nginx.org/en/docs/ngx_core_module.html#error_log)

We already compile with `--with-debug`. So all we need to do is expose the ability to change `error_log` via a `etcd` key.

(This should be a quick 2 line fix. One in nginx.conf and one in docs.)

Blystad:
Add method to customize the nginx error_log level via etcd. This helps when
debugging issues in deis-router.

Fixes #3560

Pretty easy to test:
```
deisctl config router set errorLogLevel=debug
deisctl config router set errorLogLevel=error
```

And check the log.

### DIFF
--- a/docs/customizing_deis/router_settings.rst
+++ b/docs/customizing_deis/router_settings.rst
@@ -50,6 +50,7 @@ setting                                      description
 /deis/router/enforceHTTPS                    redirect all HTTP traffic to HTTPS (default: false)
 /deis/router/firewall/enabled                nginx naxsi firewall enabled (default: false)
 /deis/router/firewall/errorCode              nginx default firewall error code (default: 400)
+/deis/router/errorLogLevel                   nginx error_log level (default: error) Valid options: debug, info, notice, warn, error, crit, alert, emerg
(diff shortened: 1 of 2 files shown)


carmstrong:
Code LGTM.

mboersma:
Code LGTM. I double-checked the `error_log` values here: http://nginx.org/en/docs/ngx_core_module.html#error_log

We don't have integration tests that set `/deis/router/errorLogLevel`, so this needs some manual testing.

mboersma:
LGTM, tested interactively on vagrant as specified above. `debug` output is quite verbose--this will be useful in troubleshooting.

carmstrong:
> LGTM, tested interactively on vagrant as specified above. debug output is quite verbose--this will be useful in troubleshooting.

:ship:?

PR #3577, opened 2015-04-24
fix(publisher): customize error log level via etcd

### LINKED ISSUE #3552: deis-publisher should be less verbose -- add log_level setting
It appears that the deis-publisher unit is uselessly filling up my log with lines like these:

```
Apr 22 17:40:50 Deis-2 logger:  2015-04-22T15:40:50UTC deis-publisher[1]: 2015/04/22 15:40:50 updateDir /deis/services/adapter
Apr 22 17:40:51 Deis-2 logger:  2015-04-22T15:40:51UTC deis-publisher[1]: 2015/04/22 15:40:51 set /deis/services/cms/cms_v6.web.1 -> 172.17.8.102:49160
Apr 22 17:40:51 Deis-2 logger:  2015-04-22T15:40:51UTC deis-publisher[1]: 2015/04/22 15:40:51 set /deis/services/adapter/adapter_v16.web.1 -> 172.17.8.103:49156
Apr 22 17:40:51 Deis-2 logger:  2015-04-22T15:40:51UTC deis-publisher[1]: 2015/04/22 15:40:51 updateDir /deis/services/cms
Apr 22 17:40:51 Deis-2 logger:  2015-04-22T15:40:51UTC deis-publisher[1]: 2015/04/22 15:40:51 updateDir /deis/services/adapter
Apr 22 17:40:52 Deis-2 logger:  2015-04-22T15:40:52UTC deis-publisher[1]: 2015/04/22 15:40:52 set /deis/services/auth/auth_v5.web.1 -> 172.17.8.102:49159
Apr 22 17:40:52 Deis-2 logger:  2015-04-22T15:40:52UTC deis-publisher[1]: 2015/04/22 15:40:52 set /deis/services/panel/panel_v10.web.1 -> 172.17.8.103:49154
Apr 22 17:40:52 Deis-2 logger:  2015-04-22T15:40:52UTC deis-publisher[1]: 2015/04/22 15:40:52 updateDir /deis/services/panel
Apr 22 17:40:52 Deis-2 logger:  2015-04-22T15:40:52UTC deis-publisher[1]: 2015/04/22 15:40:52 updateDir /deis/services/auth
```

Apparently the publisher is constantly rewriting etcd keys for my apps even if they are already there.

eMxyzptlk:
closes #3552

[Review on Reviewable](https://reviewable.io/reviews/deis/deis/3577)

### DIFF
--- a/publisher/server/publisher.go
+++ b/publisher/server/publisher.go
@@ -25,6 +25,7 @@ type Server struct {
 	DockerClient *docker.Client
 	EtcdClient   *etcd.Client
 
+	mu       sync.RWMutex
 	logLevel string
 }
 
@@ -47,7 +48,7 @@ func (s *Server) Listen(ttl time.Duration) {
 	for {
 		select {
 		case <-t.C:
-			s.loadSettings()
+			go s.loadSettings()
 		case event := <-listener:
 			if event.Status == "start" {
 				container, err := s.getContainer(event.ID)
@@ -77,11 +78,41 @@ func (s *Server) Poll(ttl time.Duration) {
 
 // loadSettings loads all of the settings from Etcd.
 func (s *Server) loadSettings() {
-	response, err := s.EtcdClient.Get("/deis/publisher/logLevel", false, false)
-	if err == nil {
-		s.logLevel = response.Node.Value
-	} else {
-		s.logLevel = defaultLogLevel
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	var wg sync.WaitGroup
+
+	wg.Add(1)
+	logLevel := make(chan string)
+	go func() {
+		defer wg.Done()
+
+		var res string
+		response, err := s.EtcdClient.Get("/deis/publisher/logLevel", false, false)
+		if err == nil {
+			res = response.Node.Value
+		} else {
+			res = defaultLogLevel
+		}
+
+		select {
+		case logLevel <- res:
+			close(logLevel)
+		default:
+		}
+	}()
+
+	done := make(chan struct{})
+	go func() {
+		wg.Wait()
+		close(done)
+	}()
+
+	select {
+	case <-time.After(10 * time.Second):
+	case ll := <-logLevel:
+		s.logLevel = ll
+	case <-done:
 	}
 }
 
@@ -220,14 +251,21 @@ func max(n []int) int {
 	return val
 }
 
+func (s *Server) debugF(format string, v ...interface{}) {
+	s.mu.RLock()
+	logLevel := s.logLevel
+	s.mu.Unlock()
+	if logLevel == "debug" {
+		log.Printf(format, v...)
+	}
+}
+
 // setEtcd sets the corresponding etcd key with the value and ttl
 func (s *Server) setEtcd(key, value string, ttl uint64) {
 	if _, err := s.EtcdClient.Set(key, value, ttl); err != nil {
 		log.Println(err)
 	}
-	if s.logLevel == "debug" {
-		log.Printf("set %q -> %q. ttl: %d", key, value, ttl)
-	}
+	s.debugF("set %q -> %q. ttl: %d", key, value, ttl)
 }
 
 // removeEtcd removes the corresponding etcd key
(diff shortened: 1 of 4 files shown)


deis-admin:

bacongobbler:
I think we'd rather tackle this as a log level setting similar to https://github.com/deis/deis/pull/3524.

bacongobbler:
see also https://github.com/kelseyhightower/confd/pull/249

eMxyzptlk:
Should the debug level be cluster-wide? Or do you want to control on a per component-basis.

bacongobbler:
Since we deploy it as a global fleet unit any changes to the component are cluster-wide

eMxyzptlk:
@bacongobbler I can implement it as a flag or through a config in etcd like #3574. That way we could debug a running publisher by changing that config value.

```
deisctl config publisher set errorLogLevel=debug
deisctl config publisher set errorLogLevel=error
```

carmstrong:
> That way we could debug a running publisher by changing that config value.

That would be nice if we can pull that off :+1: - restarting publisher isn't something we typically want to do.

mboersma:
Jenkins, add to whitelist.

publisher/server/publisher.go
```diff
@@ -203,12 +204,24 @@ func max(n []int) int {
 	return val
 }
 
+// getErrorLevel returns the error level
+func (s *Server) getErrorLevel() string {
+	response, err := s.EtcdClient.Get("/deis/publisher/errorLogLevel", false, false)
```
aledbf: @eMxyzptlk what about the overhead of this?
I mean `setEtcd` is invoked from a for loop each 10 seconds so if you have just 10 app is 10 additional `etcd` calls just to get the log level.

Why not check the flag periodically from `etcd` (like 30 or 60 seconds) and "store" the value in a local variable instead of the retrieve the value from `etcd`?

eMxyzptlk: That makes sense, I'll update the pull request.

Blystad:
I would suggest naming it `logLevel` instead of `errorLogLevel`. The reason I used `errorLogLevel` in the router is that it is part of the nginx `error_log` (as opposed to the `access_log`). The `publisher` only deals with generic (status) logging :)

eMxyzptlk:
@aledbf done.
@Blystad done.

I am planning to refactor the publisher and unit test it in #3583

publisher/server/publisher.go
```diff
@@ -203,12 +251,21 @@ func max(n []int) int {
 	return val
 }
 
+func (s *Server) debugF(format string, v ...interface{}) {
```
bacongobbler: I'm not sure the server should handle logging. We should be using a separate package for log handling like [`Logrus`](https://github.com/Sirupsen/logrus) which supports shipping messages at certain log levels. We use this library on some of our other components as well.

What do you think?

bacongobbler:
The reason why I didn't want to rely so hard on etcd for configuration is because I'd prefer steering away from over-engineering a solution built specifically towards etcd. There may be a point where we'd want to run unit/functional tests against publisher to ensure specific logs are displayed after some certain event. If we rely on etcd for configuration then the unit tests will have a hard requirement to rely on that infrastructure. This is partially why I implemented the features shown in #3524.

publisher/server/publisher.go
eMxyzptlk: I was leaning towards a centralized setup, one logger entry point (a separate package). Unfortunately I am fairly new to the codebase so I don't know off the top of my head what's being used where. If you have a general preference, I can probably get it done.

eMxyzptlk:
@bacongobbler I can implement it as a `flag` but I thought the configuration through etcd is much nice as we can change it on the fly. I can however turn the defaultLogLevel constant into a flag so you can start it by default to debug. What do you think?

bacongobbler:
I think it's great that you can set it on the fly, but I believe that feature should be implemented in the init (which we loosely call `bin/boot`) rather than in publisher itself. The init system can read from etcd for changes and reboot the server gracefully. This will still allow us to change config on the fly, but ties in publisher a little less with etcd. See https://github.com/deis/deis/issues/2701 and https://github.com/deis/deis/blob/master/publisher/main.go

The rationale behind this thinking is that I believe the server itself should be *very* simple. Reflecting on pre-1.0 we generated a lot of technical debt by having these hard ties with third-party projects like etcd. We cannot reliably write unit tests against these components which makes features such as this one mandate a manual test-by-hand, so personally I'd like to refactor these out in the future for better test coverage.

bacongobbler:
@eMxyzptlk we really appreciate the effort you've put into this. Let's break this down into two separate PRs:

1) Add --log-level flag to publisher so we can silence the debug output
2) Refactor `main.go` to watch etcd for key changes and reload config for the server

Does that sound like something you want to tackle? :)

eMxyzptlk:
@bacongobbler Yes I can do that. I'll work on (1) tonight.

eMxyzptlk:
closing in favor of #3641 

@bacongobbler Any examples for watching etcd for key changes?

PR #3582, opened 2015-04-25
fix(store): change template to return ip address instead of :6789

aledbf:
The current template generates `mon addr = :6789` instead of a list of ip:port of each ceph node.

### DIFF
--- a/store/base/templates/ceph.conf
+++ b/store/base/templates/ceph.conf
@@ -2,7 +2,7 @@
 fsid = {{ getv "/deis/store/fsid" }}
 mon initial members = {{ getv "/deis/store/monSetupLock" }}
 mon host = {{ join (getvs "/deis/store/hosts/*") "," }}
-mon addr = {{ join (ls "/deis/store/hosts/*") ":6789 ," }}:6789
+mon addr = {{ range $index, $element := (gets "/deis/store/hosts/*") }}{{if $index}},{{end}}{{ base $element.Key }}:6789{{ end }}
 auth cluster required = cephx
 auth service required = cephx
 auth client required = cephx


bacongobbler:
If this still works when there's no mons available then LGTM, though AFAIK `gets` throws an error when nothing exists

aledbf:
@bacongobbler good point, but `getvs` is used in the line 4 so I think that the error never occurs

carmstrong:
> The current template generates mon addr = :6789 instead of a list of ip:port of each ceph node.

Sounds like the current template is quite broken, then, since the confd template rewrite. I wonder how things are working at all? Enumerating the monitors here enables the clients to connect to all of them, rather that just the one enumerated in `initial members`. That means that currently, if the single initial member monitor goes down or permanently leaves the cluster, no clients can connect...

carmstrong:
LGTM. I manually verified. Before:

```console
root@deis-01:/# cat /etc/ceph/ceph.conf
[global]
fsid = 851c7744-8b42-4c1b-8962-3c7b24078e3f
mon initial members = deis-03
mon host = deis-01,deis-03
mon addr = :6789
auth cluster required = cephx
auth service required = cephx
auth client required = cephx
osd pool default size = 3
osd pool default min_size = 1
osd pool default pg_num = 128
osd pool default pgp_num = 128
osd recovery delay start = 15
log file = /dev/stdout
rgw_frontends = "civetweb port=8888"
rgw_thread_pool_size = 512

[client.radosgw.gateway]
host = deis-store-gateway
keyring = /etc/ceph/ceph.client.radosgw.keyring
rgw socket path = /var/run/ceph/ceph.radosgw.gateway.fastcgi.sock
```

After:
```console
core@deis-01 ~ $ nse deis-store-monitor
root@deis-01:/# cat /etc/ceph/ceph.conf
[global]
fsid = 851c7744-8b42-4c1b-8962-3c7b24078e3f
mon initial members = deis-03
mon host = deis-01,deis-02,deis-03
mon addr = 172.17.8.100:6789,172.17.8.101:6789,172.17.8.102:6789
auth cluster required = cephx
auth service required = cephx
auth client required = cephx
osd pool default size = 3
osd pool default min_size = 1
osd pool default pg_num = 128
osd pool default pgp_num = 128
osd recovery delay start = 15
log file = /dev/stdout
rgw_frontends = "civetweb port=8888"
rgw_thread_pool_size = 512

[client.radosgw.gateway]
host = deis-store-gateway
keyring = /etc/ceph/ceph.client.radosgw.keyring
rgw socket path = /var/run/ceph/ceph.radosgw.gateway.fastcgi.sock
```

This is good to merge once we get another Code LGTM.

mboersma:
Code LGTM.

mboersma:
Code LGTM.

PR #3586, opened 2015-04-27
feat(controller): add ps:restart command

### LINKED ISSUE #467: Should be able to restart container processes

bacongobbler:
closes #467

### DIFF
--- a/controller/scheduler/states.py
+++ b/controller/scheduler/states.py
@@ -1,8 +1,13 @@
 import enum
 
 
-class TransitionNotAllowed(Exception):
+class TransitionError(Exception):
     """Raised when a transition from one state to another is illegal"""
+
+    def __init__(self, prev, next, msg):
+        self.prev = prev
+        self.next = next
+        self.msg = msg
 
 
 class JobState(enum.Enum):
--- a/client/deis.py
+++ b/client/deis.py
@@ -1747,12 +1747,14 @@ Make sure that the Controller URI is correct and the server is running.
         Arguments:
           <type>
             the process name as defined in your Procfile, such as 'web' or 'worker'.
+            To restart a particular process, use 'web.1'.
 
         Options:
           -a --app=<app>
             the uniquely identifiable name for the application.
         """
         app = args.get('--app')
+        procname = args.get('<type>')
         if not app:
             app = self._session.app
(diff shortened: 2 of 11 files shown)


nathansamson:
by the looks of it these are not rolling restarts. Would it be possible to make that an option to minimize downtime?

bacongobbler:
In a future PR, sure. I'm sure there's going to be many different implementations where users are going to want different restart policies, so this is just the simplest solution for now and we can work on a proposal for rolling restarts later.

bacongobbler:
I still need to implement `ps:restart web.2` as per https://devcenter.heroku.com/articles/dynos#cli-commands-for-dyno-management. Commit coming up

bacongobbler:
done. Once jenkins agrees with me this is ready for code review

controller/scheduler/states.py
```diff
@@ -1,9 +1,14 @@
 import enum
 
 
-class TransitionNotAllowed(Exception):
+class TransitionError(Exception):
```
mboersma: Is this more generic name better for this exception class? Just curious.

mboersma:
Code LGTM, nice work! It looks like there is good test coverage, but let me know if you think this needs manual testing as well.

controller/scheduler/states.py
bacongobbler: I saw documentation in python https://docs.python.org/2/tutorial/errors.html#user-defined-exceptions and thought it was better

bacongobbler:
Whoops, forgot to bump API docs as well

controller/scheduler/states.py
mboersma: > Most exceptions are defined with names that end in “Error,” similar to the naming of the standard exceptions.

Works for me. +1.

bacongobbler:
@mboersma I think the unit/integration tests speak for themself :)

carmstrong:
:ship:

bacongobbler:
Jenkins derped at the last minute, but this did pass all tests: https://ci.deis.io/job/test-ec2/722/console

Good to :shipit:

carmstrong:
> Good to ?

I'd say so.

PR #3589, opened 2015-04-28
feat(controller): allow user to scale gunicorn worker processes

mboersma:
The hard-coded default of 8 python worker processes was a stab in the dark, and may use more resources than required in general. This change uses gunicorn's recommended default of (CPUs * 2 + 1), and allows a user to override that with `deisctl config controller set workers=N`.

### DIFF
--- a/controller/conf.d/gconf.toml
+++ b/controller/conf.d/gconf.toml
@@ -1,0 +1,10 @@
+[template]
+src   = "gconf.py"
+dest  = "/app/deis/gconf.py"
+uid = 1000
+gid = 1000
+mode  = "0640"
+keys = [
+  "/deis/controller",
+]
+reload_cmd = "/app/bin/reload"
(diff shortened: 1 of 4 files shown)


carmstrong:
Good change. Code LGTM.

bacongobbler:
is it possible that `cpu_count()` could raise an exception if it cannot find the number of CPUs? If so then maybe we should have another case to fall back to 8 workers. Otherwise code LGTM

mboersma:
> is it possible that cpu_count() could raise an exception if it cannot find the number of CPUs?

In theory, [yes](https://docs.python.org/2/library/multiprocessing.html#multiprocessing.cpu_count). Although since we control the deis-controller container environment, we know that Python implements `multiprocessing.cpu_count()`. But it's better code to handle the potential error as you suggest--I will make that change. Thanks!

bacongobbler:
code LGTM

PR #3591, opened 2015-04-28
feat(router): nginx 1.9.0. Remove third party tcp module

aledbf:
http://nginx.org/en/docs/stream/ngx_stream_core_module.html

### DIFF
--- a/router/image/templates/nginx.conf
+++ b/router/image/templates/nginx.conf
@@ -292,22 +292,15 @@ http {
 
 ## start builder
 {{ if .deis_builder_host }}
-tcp {
-    access_log /opt/nginx/logs/git.log;
-    tcp_nodelay on;
-    timeout {{ or (.deis_router_builder_timeout_tcp) "1200000" }};
-
-    # same directive names, but these are in miliseconds...
-    proxy_connect_timeout       {{ or (.deis_router_builder_timeout_connect) "10000" }};
-    proxy_send_timeout          {{ or (.deis_router_builder_timeout_send) "1200000" }};
-    proxy_read_timeout          {{ or (.deis_router_builder_timeout_read) "1200000" }};
-
+stream {
     upstream builder {
         server {{ .deis_builder_host }}:{{ .deis_builder_port }};
     }
 
     server {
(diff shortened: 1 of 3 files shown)


router/image/templates/nginx.conf
```diff
@@ -292,22 +292,15 @@ http {
...
-
+stream {
     upstream builder {
         server {{ .deis_builder_host }}:{{ .deis_builder_port }};
     }
 
     server {
         listen 2222;
+        proxy_connect_timeout  {{ or (.deis_router_builder_timeout_connect) "10000" }};
+        proxy_timeout          {{ or (.deis_router_builder_timeout_tcp) "1200000" }};
```
bacongobbler: could you please update the docs to reflect this new value?

mboersma:
Much simpler without having to apply the external patch. Code LGTM.

carmstrong:
LGTM once @bacongobbler's comment is addressed. Nice cleanup.

carmstrong:
Sorry, misclick :disappointed:

router/image/templates/nginx.conf
aledbf: which one? (In fact I removed the send and read builder timeouts)

mboersma: (done in b5b3b2d332d2501659c993577ab7dd0309184879)

mboersma:
The [commit style](http://docs.deis.io/en/latest/contributing/standards/#commit-style) would be better served with something like:
```feat(router): update to nginx 1.9 for built-in stream module```
Start with a verb, keep it lower-case. Just nitpicking, not necessary. :smiling_imp:

router/image/templates/nginx.conf
carmstrong: I think @bacongobbler was referring to `/deis/router/builder/timeout/tcp`, but it already exists: http://docs.deis.io/en/latest/customizing_deis/router_settings/

bacongobbler:
Yep, I'm all happy :)

carmstrong:
:shipit: @aledbf

PR #3594, opened 2015-04-28
fix(user-data): update fleetd 0.92 to /opt so it survives reboots

### LINKED ISSUE #3547: /run/deis/bin/fleetd-0.9.2 disappears on host reboot
We added fleet 0.9.2 in #3422, since CoreOS ships fleet 0.9.1 with a known issue that affected the reliability of deploying apps on Deis.

@rjocoleman reported in https://github.com/deis/deis/pull/3422#issuecomment-94619495 that upon rebooting his hosts, the binary went missing. 

Is this happening on all providers? Could it be due to the way CoreOS mounts filesystems?

mboersma:
This workaround should be removed when a [newer CoreOS](https://coreos.com/releases/#647.0.0) makes it into the stable channel, but for now putting the `fleetd-0.9.2` binary and conf file under /opt ensures they will survive a reboot.

Tested with `make discovery-url && vagrant up`, then `vagrant reload deis-01 && vagrant ssh deis-01 -c 'ps faux | grep fleetd'`, which fails on Deis master but succeeds when provisioned from this PR. Also checked by hand that the `fleetd` binary disappeared before, but remains now after reboot.

Closes #3547.

### DIFF
--- a/contrib/coreos/user-data.example
+++ b/contrib/coreos/user-data.example
@@ -29,7 +29,7 @@ coreos:
       ConditionPathIsSymbolicLink=!/etc/systemd/system/fleet.service.d/99-upgrade-fleet-091.conf
 
       [Service]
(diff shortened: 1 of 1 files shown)


bacongobbler:
nicely done. Code LGTM

carmstrong:
Code LGTM.

PR #3597, opened 2015-04-29
fix(router): return 503 for non-existing app

### LINKED ISSUE #3502: Non existing app returns http 200
Hi

Relaed to: #1624

I do not want customization right now, but more importantly this is now HTTP 200, while in fact this should be HTTP 404 or HTTP 500.

I encountered this issue when my cluster was unstable & my app wasn't served.
My monitoring did not raise errors as a 200 request was returned.
I could strengthen up my monitoring by looking for certain words, but that is cumbersome and should not be forced.

Can we change the default nginx page to return a 500 or any other error code that is sensible?

ngpestelos:
fixes #3502

### DIFF
--- a/router/image/templates/nginx.conf
+++ b/router/image/templates/nginx.conf
@@ -265,9 +265,7 @@ http {
             proxy_pass                  http://{{ Base $service.Key }};
         }
         {{ else }}
-        location / {
             return 503;
-        }
         {{ end }}
         {{ if eq $useFirewall "true" }}location /RequestDenied {
             return {{ $firewallErrorCode }};
@@ -286,6 +284,7 @@ http {
         }
         location /router-nginx-status {
             stub_status on;
+            return 200;
         }
     }
 }


johanneswuerbach:
Hm, it think this should be a 404.

503: http://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html
> The implication is that this is a temporary condition which will be alleviated after some delay

bacongobbler:
@johanneswuerbach it is a temporary condition. The 503 will revert back to a normal HTTP status code when the application is up and responding. Basically we're using it as a "server is down, please wait", which is what a 503 error message was intended for.

> The server is currently unable to handle the request due to a temporary overloading or maintenance of the server

johanneswuerbach:
A not existing document is usually a 404 even when it can exist later again.

I think it would make the nginx conf more readable to just have something like this:
```
server {
        listen 80 default_server;
        location /health-check {
            default_type 'text/plain';
            access_log off;
            return 200;
        }
        location /router-nginx-status {
            stub_status on;
            return 200;
        }
        location / {
            return 404;
        }
    }
```

bacongobbler:
However, I thought the resoution here was to respond with a `return 404;` under [this line](https://github.com/deis/deis/blob/master/router/image/templates/nginx.conf#L281) as per @carmstrong's comment.

bacongobbler:
@johanneswuerbach sorry, I was mistaken. I thought you were referring to the existing 503 in the router. My bad! Completely agree that the default response should be a 404.

ngpestelos:
Thank you for your quick feedback. However, I noticed my branch name still refers to 503, so I created a new PR #3598 and will close this one.

PR #3598, opened 2015-04-29
fix(router): return 404 for non-existing app

### LINKED ISSUE #3502: Non existing app returns http 200

ngpestelos:
fixes #3502

### DIFF
--- a/router/image/templates/nginx.conf
+++ b/router/image/templates/nginx.conf
@@ -302,6 +302,10 @@ http {
         }
         location /router-nginx-status {
             stub_status on;
+            return 200;
+        }
+        location / {
+            return 404;
         }
     }
 }
--- a/tests/apps_test.go
+++ b/tests/apps_test.go
@@ -3,6 +3,7 @@
 package tests
 
 import (
+	"fmt"
 	"math/rand"
 	"os"
 	"testing"
@@ -108,6 +109,7 @@ func appsLogsTest(t *testing.T, params *utils.DeisTestConfig) {
 
 func appsOpenTest(t *testing.T, params *utils.DeisTestConfig) {
 	utils.CurlApp(t, *params)
+	utils.CurlWithFail(t, fmt.Sprintf("http://%s.%s", "this-app-does-not-exist", params.Domain), true, "404 Not Found")
 }
 
 func appsRunTest(t *testing.T, params *utils.DeisTestConfig) {


router/image/templates/nginx.conf
```diff
@@ -264,10 +264,6 @@ http {
 
             proxy_pass                  http://{{ Base $service.Key }};
         }
-        {{ else }}
```
bacongobbler: We still need this :)

bacongobbler:
other than that, LGTM

carmstrong:
Do we have integration tests for apps that don't exist? If not, we should as part of this PR.

router/image/templates/nginx.conf
bacongobbler: what I meant is that this should not have been removed. We still need this code block. No modification is required here. Sorry about the confusion

ngpestelos:
I would like to write an integration test. Can you point me to an example?

eMxyzptlk:
Can we include this in 1.6 milestone?

bacongobbler:
@eMxyzptlk 1.6 code freeze was this morning so it probably won't make it in.

carmstrong:
> Can we include this in 1.6 milestone?

The code freeze for 1.6 is end-of-day Thursday. If this is mergeable by then, we will review it and include it in the release.

carmstrong:
> I would like to write an integration test. Can you point me to an example?

The integration tests for apps are mostly: https://github.com/deis/deis/blob/master/tests/apps_test.go

@mboersma is more familiar with the test code than I am, but that should hopefully give you something to go on!

ngpestelos:
@carmstrong @mboersma I've updated the PR to include a test.

router/image/templates/nginx.conf
```diff
@@ -265,9 +265,7 @@ http {
             proxy_pass                  http://{{ Base $service.Key }};
         }
         {{ else }}
-        location / {
-            return 503;
-        }
```
johanneswuerbach: We still need this. So only the `return 404` in line 290 should be enough.

ngpestelos: Thanks, I've put these lines back.

eMxyzptlk:
> The code freeze for 1.6 is end-of-day Thursday. If this is mergeable by then, we will review it and include it in the release.
I guess we missed it :(

tests/apps_test.go
```diff
@@ -108,6 +108,9 @@ func appsLogsTest(t *testing.T, params *utils.DeisTestConfig) {
 
 func appsOpenTest(t *testing.T, params *utils.DeisTestConfig) {
 	utils.CurlApp(t, *params)
+	cfg := utils.GetGlobalConfig()
+	cfg.AppName("ghostapp")
+	CurlWithFail(t, fmt.Sprintf("http://%s.%s", cfg.AppName, cfg.Domain), true, "404 Not Found")
```
bacongobbler: change to `CurlWithFail(t, fmt.Sprintf("http://%s.%s", "this-app-does-not-exist", cfg.Domain), true, "404 Not Found")`

tests/apps_test.go
```diff
@@ -108,6 +108,9 @@ func appsLogsTest(t *testing.T, params *utils.DeisTestConfig) {
 
 func appsOpenTest(t *testing.T, params *utils.DeisTestConfig) {
 	utils.CurlApp(t, *params)
+	cfg := utils.GetGlobalConfig()
+	cfg.AppName("ghostapp")
```
bacongobbler: with the other comment this is not required

tests/apps_test.go
```diff
@@ -108,6 +108,7 @@ func appsLogsTest(t *testing.T, params *utils.DeisTestConfig) {
 
 func appsOpenTest(t *testing.T, params *utils.DeisTestConfig) {
 	utils.CurlApp(t, *params)
+	CurlWithFail(t, fmt.Sprintf("http://%s.%s", "this-app-does-not-exist", cfg.Domain), true, "404 Not Found")
```
bacongobbler: The function is located in the `utils` package, so you need to call `utils.CurlWithFail`. Also, you need to add the `fmt` package to the top of this file and change `cfg` to `params`. Thanks!

bacongobbler:
LGTM

mboersma:
Code LGTM.

PR #3603, opened 2015-04-30
Update available datacenters for Digital Ocean

ghostbar:
ams2 and sfo1 now do have metadata and private-networking support.


deis-admin:

mboersma:
Right you are--thanks for noticing this. I double-checked in the DO "create droplet" screen, and nyc2 looks like the only region available without metadata (user-data) and private networking support. (nyc1 and ams1 are disabled for creating new droplets currently, but I assume they are still limited.)

Code LGTM.

ghostbar:
Indeed @mboersma, that's what `docl regions --metadata --private-networking` shows and nyc1 nor ams1 are shown.

carmstrong:
Code LGTM after the commit is amended.

mboersma:
@ghostbar we want this in the next Deis release, but we are strict about commit style. I'm going to amend the subject of your commit and open another PR so we can get this fixed ASAP since we are trying to get to a code freeze. I hope that's ok with you--thanks again!

PR #3607, opened 2015-05-01
ref(controller): simplify thread and set comparison code

mboersma:
This change makes some scheduler-related code shorter and easier to read. Certain parts are also marginally more efficient since Python's [`any()`](https://docs.python.org/2/library/functions.html#any) will return on the first match, whereas the `set() != set()` comparisons iterated through all elements.

### DIFF
--- a/controller/api/models.py
+++ b/controller/api/models.py
@@ -14,7 +14,7 @@ import os
 import re
 import subprocess
 import time
-import threading
+from threading import Thread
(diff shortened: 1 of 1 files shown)


bacongobbler:
Hah, snuck this in right before code freeze. Code LGTM

carmstrong:
Code LGTM.

PR #3609, opened 2015-05-01
fix(Makefile): make dev-registry target more compatible

mboersma:
`make dev-registry` on platforms without `boot2docker` returns this:
```console
$ make dev-registry
registry

To use local boot2docker registry for Deis development:
    export DEV_REGISTRY=:5000
```
The command used in #3548 for a similar purpose falls back to `127.0.0.1`, which will work on most Linux boxen with Docker. This appropriates that change to improve compatibility.


eMxyzptlk:
Not sure if it's a good idea to hardcode `boot2docker ip`. I use `docker-machine` instead of boot2docker which is replacing boot2docker anyway. You could try to find `$DOCKER_HOST` first.

```
$ docker-machine env dev
export DOCKER_TLS_VERIFY=1
export DOCKER_CERT_PATH="/Users/wmn/.docker/machine/machines/dev"
export DOCKER_HOST=tcp://192.168.99.100:2376
```

Joshua-Anderson:
@mboersma I might be completely misinterpreting this, but I think this may not work because the VM pulls the local registry, so you need the IP address of your computer so that the VM can access the registry, since 127.0.0.1 will point to a very different place on the VM and the local machine. I set my DEV_REGISTRY variable to the static private IP address of my comptuer.

bacongobbler:
> Not sure if it's a good idea to hardcode boot2docker ip. I use docker-machine instead of boot2docker which is replacing boot2docker anyway

boot2docker is still the de-facto way to deploy docker on OSX, and is still referenced widely in docs, blog posts etc. I don't imagine it's moving away any time soon, but we can switch over when the time comes to pass the torch :fire: 

@Joshua-Anderson good point. `make set-image` relies on DEV_REGISTRY, so unless you're running the entire deis stack locally (:trollface:) then this wouldn't work. I just grab the public ip address using `ip addr` and use that.

mboersma:
> unless you're running the entire deis stack locally then this wouldn't work

Yes that's probably the case. I'm leaving well enough alone then and closing this. If someone has a more reliable way to improve `make dev-registry`, let's make a new PR. Although I expect most Linux hackers did not find this difficult to work around anyway.

bacongobbler:
As a Linux haxx0r, I just have a script:

```
#!/bin/bash
# provision platform off some contrib script
deisctl config platform set domain=$DEIS_TEST_DOMAIN
deisctl config platform set sshPrivateKey=~/.ssh/id_rsa
deisctl install platform
deisctl start platform
```

Then again, for me my dev clusters aren't on vagrant because my Internet upload speed sucks (that will change tomorrow though :D). If I find a way to get your public up address, I'll try to contribute a patch to this. Thanks @mboersma :)

Issue #3293, opened 2015-03-13
feat(controller): add deis ps:stop command


pjvds:
Is there a way to stop an application beside executing `deis destroy`?

bacongobbler:
No, but a `deis ps:stop` command should be relatively easy to implement :)

pjvds:
Lets make that my goal for this weekend then. Any pointers?

bacongobbler:
You'll probably want to start [on a proposal](https://github.com/deis/deis/blob/master/CONTRIBUTING.md#design-proposals) similar to [this](https://github.com/deis/deis/pull/3110) PR. After that, you'll want to take a look at the fleet scheduler and implement `.stop()` and hook that into the API.

https://github.com/deis/deis/blob/master/controller/scheduler/fleet.py#L207-L209

bacongobbler:
@pjvds could you please elaborate on why this feature would be useful? I know for heroku this allows you to stop your dynos, effectively stopping from heroku billing you but I'm not sure how a `ps:stop` would be helpful in Deis.

ianblenke:
How about a `deis ps:scale web=0` ?

mboersma:
I agree with @ianblenke that `deis scale web=0` is the way to accomplish this. `deis ps:stop` could be implemented in terms of that.

Not to stifle the proliferation of ideas, but since this hasn't seen any activity in three months I'm closing it. Please do re-open it or just submit a PR @pjvds if you get re-energized about this. :electric_plug:

PR #3612, opened 2015-05-01
feat(controller): add swarm scheduler tech preview

mboersma:
Replaces #3134, see there for previous discussion.


deisctl/units/deis-controller.service
```diff
@@ -6,8 +6,10 @@ After=deis-store-volume.service
 [Service]
 EnvironmentFile=/etc/environment
 TimeoutStartSec=20m
+ExecStartPre=-/usr/bin/etcdctl mkdir /deis/scheduler >/dev/null 2>&1
 ExecStartPre=/bin/sh -c "IMAGE=`/run/deis/bin/get_image /deis/controller` && docker history $IMAGE >/dev/null 2>&1 || docker pull $IMAGE"
 ExecStartPre=/bin/sh -c "docker inspect deis-controller >/dev/null 2>&1 && docker rm -f deis-controller || true"
+ExecStartPre=-/bin/sh -c "etcdctl get /deis/scheduler/module >/dev/null 2>&1 || etcdctl mk /deis/scheduler/module 'fleet' "
```
carmstrong: Using the `-` here means that this can fail. Do we want the unit to succeed if we can't set this key?

mboersma: > Do we want the unit to succeed if we can't set this key?

No, probably not--the user will see errors later in the controller. @smothiki?

smothiki: @mboersma can you remove the line  ```ExecStartPre=-/bin/sh -c "etcdctl get /deis/scheduler/module >/dev/null 2>&1 || etcdctl mk /deis/scheduler/module 'fleet' "``` we are not using the key any where

deisctl/units/deis-controller.service
```diff
@@ -6,8 +6,10 @@ After=deis-store-volume.service
 [Service]
 EnvironmentFile=/etc/environment
 TimeoutStartSec=20m
+ExecStartPre=-/usr/bin/etcdctl mkdir /deis/scheduler >/dev/null 2>&1
```
carmstrong: Same thing here - we probably want to ensure this succeeds.

swarm/Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM golang:1.4
+WORKDIR /go/src/github.com/docker
+RUN git clone https://github.com/smothiki/swarm
```
carmstrong: We cloned `swarm` to `deis/swarm` so we could use that codebase instead. There shouldn't be any `smothiki`s here.

deisctl/units/deis-controller.service
mboersma: What if the etcd directory already exists? I think this is how we meant to ignore that error.
```console
core@deis-01 ~ $ /usr/bin/etcdctl mkdir /deis/foobar
Error:  105: Key already exists (/deis/foobar) [67559]
core@deis-01 ~ $ echo $?
4
```

carmstrong:
We'll also need to rejigger test-acceptance or whatever to push `deis/swarm` to Docker Hub.

swarm/Dockerfile
mboersma: Good catch--will fix. We will also change the branch name to something more appealing than "fail" (which means "failover" in this case).

mboersma:
> We'll also need to rejigger test-acceptance or whatever to push deis/swarm to Docker Hub.

Currently that image is being built and uploaded out-of-band to Docker Hub, then tagged to match Deis releases. We need to bring that fork into the CI process soon.

deisctl/units/deis-controller.service
carmstrong: Ok, but if it failed due to etcd errors, don't we want to exit there? This is how we accommodate that elsewhere, but it'd make for a funky one-liner: https://github.com/deis/deis/blob/b3a352d8722219f04daccc5c262c0d711d88c098/controller/bin/boot#L27-L35

Not sure what to do about that.

carmstrong:
> We need to bring that fork into the CI process soon.

Ok. Should we document that process in the release docs as part of this PR?

swarm/Dockerfile
carmstrong: > We will also change the branch name to something more appealing than "fail"

:wink:

deisctl/units/deis-swarm-manager.service
```diff
@@ -0,0 +1,15 @@
+[Unit]
+Description=deis-swarm-manager
+
+[Service]
+EnvironmentFile=/etc/environment
+TimeoutStartSec=20m
+ExecStartPre=/bin/sh -c "IMAGE=`/run/deis/bin/get_image /deis/swarm` && docker history $IMAGE >/dev/null 2>&1 || docker pull $IMAGE"
```
carmstrong: Shouldn't this be `/run/deis/bin/get_image /deis/swarm-manager` or are they the same Docker image?

deisctl/units/deis-swarm-node.service
```diff
@@ -0,0 +1,18 @@
+[Unit]
+Description=deis-swarm-node
+
+[Service]
+EnvironmentFile=/etc/environment
+TimeoutStartSec=20m
+ExecStartPre=/bin/sh -c "IMAGE=`/run/deis/bin/get_image /deis/swarm` && docker history $IMAGE >/dev/null 2>&1 || docker pull $IMAGE"
```
carmstrong: Shouldn't this be `/run/deis/bin/get_image /deis/swarm-node` or are they the same Docker image?

deisctl/units/deis-swarm-manager.service
mboersma: They are the same Docker image, just with different command lines.

deisctl/units/deis-swarm-node.service
mboersma: They are the same Docker image, just with different command lines.

carmstrong: :+1:

deisctl/units/deis-swarm-manager.service
carmstrong: :+1:

deisctl/units/deis-controller.service
mboersma: > if it failed due to etcd errors, don't we want to exit there?

Ideally yes, but as you pointed out it's hard to see a clean way to do that. Maybe in this case we have to allow `confd` to report the problem later. :disappointed:

carmstrong: Yeah, I've been chewing on this and can't think of a better solution.

swarm/Dockerfile
mboersma: > We will also change the branch name to something more appealing than "fail"

[nodefailover](https://github.com/deis/swarm/tree/nodefailover) it is.

carmstrong:
LGTM. Nice work, gents.

swarm/swarm.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+func setEtcd(client *etcd.Client, key, value string, ttl uint64) {
+	_, err := client.Set(key, value, ttl)
+	if err != nil {
+		log.Println(err)
+	}
+}
+func main() {
+	etcdproto := "etcd://" + getleaderHost() + swarmpath
```
aledbf: this should be os.Getenv("HOST") not the etcd leader ip address. If that node goes down after deis-swarm-* is running in the cluster it throws "501: All the given peers are not reachable (Tried to connect to each peer twice and failed) [0]"

aledbf:
How the failover works?
I tried this:
- use deis/example-go as app
- scaled the app to 1
- scaled the app to 10 (it's really fast)
- the spread ended with most of the apps running in 1 node
- killed that node node (virtualbox pause)
- nothing happened

docs/customizing_deis/choosing-a-scheduler.rst
```diff
@@ -0,0 +1,143 @@
...
+    - It is not yet possible to change the default affinity filter.
+    - If swarm can't create all the containers requested, Deis returns an error
+      and leaves some containers in the "created" state. Until this behavior is fixed,
+      be mindful of resource limitations on your cluster.
+
+To test the Swarm Scheduler backend, first install and start the swarm components:
+
+.. code-block:: console
+
+    $ deisctl install swarm && deisctl start swarm
```
aledbf: this didn't work.
I had to run this:
```
deisctl install swarm-manager && deisctl start swarm-manager
deisctl install swarm-node && deisctl start swarm-node
```

docs/customizing_deis/choosing-a-scheduler.rst
```diff
@@ -0,0 +1,143 @@
...
+
+Then set the controller's ``schedulerModule`` to "swarm":
+
+.. code-block:: console
+
+    $ deisctl config controller set schedulerModule=swarm
+
+The Swarm Scheduler is now active. Commands such as ``deis destroy`` or
+``deis scale web=9`` will use `swarm`_ to manage app containers.
```
aledbf: What's the `fleetctl list-units` equivalent? This could be confusing for users.

swarm/Dockerfile
```diff
@@ -0,0 +1,24 @@
+FROM golang:1.4
+
+WORKDIR /go/src/github.com/docker
+RUN git clone https://github.com/deis/swarm
+WORKDIR /go/src/github.com/docker/swarm
+RUN git fetch origin nodefailover
+RUN git checkout nodefailover
+ENV GOPATH /go/src/github.com/docker/swarm/Godeps/_workspace:$GOPATH
+RUN CGO_ENABLED=0 go install -v -a -tags netgo -ldflags "-w -X github.com/docker/swarm/version.GITCOMMIT `git rev-parse --short HEAD`"
```
aledbf: this should not be using godep?

aledbf:
Why the docker image is so huge? Maybe it could use the same trick than `deis-logspout`?

aledbf:
I see multiple versions of `deis/swarm` in the docker registry. Which one is the latest?

docs/customizing_deis/choosing-a-scheduler.rst
mboersma: > this didn't work

Could you be more specific? `deisctl install swarm` is what we've been using--did you rebuild `deisctl` from this PR branch?

swarm/Dockerfile
mboersma: Yes it should--setting `$GOPATH` to include Godeps/_workspace kind of defeats the purpose.

swarm/swarm.go
smothiki: I  remember trying etcd protocol with host ip and didn't work. I think etcd protocol works only with leader host

docs/customizing_deis/choosing-a-scheduler.rst
smothiki: you can do docker -H <swarm manager ip>:2395 ps

smothiki:
@aledbf  deis/swarm:v1.5.2 is the latest.

smothiki:
@aledbf  how much time did you wait for after node failover. Manager waits for approximately 60 seconds to see if the node comes back and then starts rescheduling

aledbf:
@smothiki 5 minutes

swarm/swarm.go
aledbf: > I remember trying etcd protocol with host ip and didn't work. I think etcd protocol works only with leader host

ok, but then is necessary to check periodically if the etcd leader is alive to avoid this error.

smothiki:
@aledbf I have been trying node failover and its working. I haven't paused any VM I did vagrant halt every time and in docker events you can see an engine disconnect event and in less than two minutes you can see create and start events. I will try with Vagrant Pause now

swarm/swarm.go
smothiki: etcd leader Host is used by swarm nodes and helps manager for discovery. I haven't hit the use case of leader host going down yet. I have to check and see if some thing abnormal is happening

carmstrong:
Some feedback we've received so far:
* We should add some additional logging so it's clear what the scheduler is doing (fleet, even when it's misbehaving, shows us what it's doing)
* We need to clarify what "preview" means, so that it's clear that these other scheduler demos are not feature complete, and not to be considered a replacement for fleet

/cc @smothiki @mboersma @gabrtv

mboersma:
> We should add some additional logging so it's clear what the scheduler is doing

`fleet` doesn't actually write to controller logs though, so from that point of view the swarm and fleet backends act similarly, it's just that we're used to switching back to `fleetctl list-units` and the like for troubleshooting, and with the swarm backend you'd use `docker` itself to monitor things as we pointed out.

Unless I'm missing something, I think this is covered by the [To monitor swarm operations...](https://github.com/deis/deis/pull/3612/files#diff-db82a3ff5cc12946f85e305a4a170dcbR123) section. But if that's not helpful enough or if there is somewhere we could actually add more log output, suggestions are welcome.

bacongobbler:
> But I am glad to add additional verbiage if there are suggestions.

Other than the obvious "THIS IS A TECH PREVIEW" in the release notes it seems like it's spelled out to me

docs/customizing_deis/choosing-a-scheduler.rst
```diff
@@ -0,0 +1,143 @@
...
+The scheduler creates, starts, stops, and destroys your app containers. For example,
+a command such as ``deis scale web=3`` tells the scheduler to run three containers
+from the Docker image for your app.
+
+The scheduler must decide which machines are eligible to run these container jobs.
+Scheduler backends vary in the details of their job allocation policies and whether
+or not they are resource-aware, among other features.
+
+Deis defaults to using the `Fleet Scheduler`_. A tech preview of the `Swarm Scheduler`_
+is available for testing. And work is ongoing on `Kubernetes`_ and `Mesos`_-based
```
bacongobbler: "Work is ongoing"

docs/customizing_deis/choosing-a-scheduler.rst
```diff
@@ -0,0 +1,143 @@
...
+:description: How to choose a scheduler backend for Deis.
+
+
+.. _choosing_a_scheduler:
+
+Choosing a Scheduler
+====================
+
+The scheduler creates, starts, stops, and destroys your app containers. For example,
+a command such as ``deis scale web=3`` tells the scheduler to run three containers
```
bacongobbler: we should reference [Container](http://docs.deis.io/en/latest/reference/terms/container/) here as a defined term. We should probably do the same here for a scheduler.

swarm/Dockerfile
bacongobbler: IMO we should be using the same build pipeline that we use for logger, logspout, builder etc. where the binary is built locally and then injected into the Docker image.

contrib/coreos/user-data.example
```diff
@@ -32,6 +32,19 @@ coreos:
       ExecStart=/usr/bin/bash -c 'if fleetd --version | grep -q 0.9.1; then curl -sSL --retry 5 --retry-delay 2 -o /opt/bin/fleetd-0.9.2 https://s3-us-west-2.amazonaws.com/opdemand/fleetd-v0.9.2 && chmod +x /opt/bin/fleetd-0.9.2 && mkdir -p /etc/systemd/system/fleet.service.d/ && ln -sf /opt/conf/fleetd-092-custom-binary.conf /etc/systemd/system/fleet.service.d/99-upgrade-fleet-091.conf; else rm -f /etc/systemd/system/fleet.service.d/99-upgrade-fleet-091.conf; fi'
       RemainAfterExit=yes
       Type=oneshot
+  - name: docker-tcp.socket
+    command: start
+    enable: true
```
bacongobbler: Ah, so this works now: https://github.com/coreos/coreos-cloudinit/pull/127

controller/scheduler/swarm.py
```diff
@@ -0,0 +1,136 @@
...
+        self.target = settings.SWARM_HOST
+        # single global connection
+        self.registry = settings.REGISTRY_HOST+":"+settings.REGISTRY_PORT
+        self.docker_cli = Client(base_url='tcp://'+self.target+':'+"2395",
+                                 timeout=1200, version='1.17')
+
+    def create(self, name, image, command='', template=None, **kwargs):
+        """Create a container"""
+        cimage = self.registry+"/"+image
+        cname = name
```
bacongobbler: Why are we assigning a new variable without any modification? If it's just for visual clarity then we should just change the signature names. Doesn't make sense to waste space in the private heap, IMO

controller/scheduler/swarm.py
```diff
@@ -0,0 +1,136 @@
...
+        """
+        Stop a container
+        """
+        self.docker_cli.stop(name)
+
+    def destroy(self, name):
+        """
+        Destroy a container
+        """
+        self.docker_cli.stop(name)
```
bacongobbler: this should be part of `stop`. `.destroy()` should just remove the container.

controller/scheduler/swarm.py
```diff
@@ -0,0 +1,136 @@
...
+
+    def run(self, name, image, entrypoint, command):
+        """
+        Run a one-off command
+        """
+        cimage = self.registry+"/"+image
+        cname = name
+        ccommand = command
+        affinity = "affinity:container!=~/"+(re.split("_v\d.", cname)[0]) + "*/"
+        self.docker_cli.create_container(image=cimage, name=cname,
```
bacongobbler: instead of duplicating code, we should be calling `self.create()`.

controller/scheduler/swarm.py
```diff
@@ -0,0 +1,136 @@
...
+                                         command=ccommand.encode('utf-8'),
+                                         environment=[affinity],
+                                         entrypoint=[entrypoint])
+        time.sleep(2)
+        self.docker_cli.start(cname)
```
bacongobbler: same here; should be calling `self.start()`

controller/scheduler/swarm.py
```diff
@@ -0,0 +1,136 @@
...
+        rc = 0
+        while (True):
+            if self._get_container_state(name) == JobState.created:
```
bacongobbler: shouldn't we be checking if the container is `up`, not `created`?

controller/scheduler/swarm.py
```diff
@@ -0,0 +1,136 @@
...
+        except Exception:
+            rc = 1
+            return rc, output
+
+    def _get_container_state(self, name):
+        try:
+            if self.docker_cli.inspect_container(name)["State"]["Running"]:
+                return JobState.up
+            else:
+                return JobState.created
```
bacongobbler: Is there other states which we should be checking for? We should also be checking for `down` (which is what `ps:restart` would check for), `fail`, etc.

swarm/Dockerfile
mboersma: It's also being created manually without the same review process we have for Deis, and that needs to be fixed.

mboersma:
> Why the docker image is so huge?

I fixed it so that `make build` now also builds `deis/swarm:$BUILD_TAG`, and the Docker image is 14.15MB. Please take another look at this PR when time permits.

carmstrong:
Latest commit looks great. Code LGTM.

controller/scheduler/swarm.py
smothiki: yes we are waiting till the container finishes executing the Job. I don't think docker API has a call to check whether a container finished executing or not. We do inspect container and check for container running state for True or False

controller/scheduler/swarm.py
smothiki: Docker API returns error if create fails but a different process will be getting this error than the process that checks get_state. So right now I haven't figured any way to get crashed state. If we could't find a container its just destroyed.

bacongobbler:
I just tried this. I found the initial `deis scale` to take about 5 minutes (i guess for pulling the image), but after that it worked great.

However, why do we ask the container to `die` on a `ps:restart`? Can't we just call [`docker restart`](https://github.com/docker/docker/blob/91fb4d672a8536d2f0437aeeb2dfc7b9e51aac16/api/server/server.go#L1564) directly?

```
core@deis-1 ~ $ docker -H localhost:2395 events
2015-05-05T06:16:38.000000000Z c7526f5e81e3ad88d4e0d6b093bb31777a3a91bf05ca240e230a86d6e095b3f5: (from 10.21.12.2:5000/go:v2 node:deis-1) die
2015-05-05T06:16:38.000000000Z c7526f5e81e3ad88d4e0d6b093bb31777a3a91bf05ca240e230a86d6e095b3f5: (from 10.21.12.2:5000/go:v2 node:deis-1) stop
2015-05-05T06:16:38.000000000Z 5292230cafcdd45bc508fadabb4e0c3fc9b74c158ef01fc771c8d43aac9d8857: (from 10.21.12.2:5000/go:v2 node:deis-2) die
2015-05-05T06:16:38.000000000Z 5292230cafcdd45bc508fadabb4e0c3fc9b74c158ef01fc771c8d43aac9d8857: (from 10.21.12.2:5000/go:v2 node:deis-2) stop
2015-05-05T06:16:38.000000000Z cbe69d659d2b9e919fc20fa2b789c83b99e7456bfb69145c4d991f6e481629b1: (from 10.21.12.2:5000/go:v2 node:deis-3) die
2015-05-05T06:16:38.000000000Z 9273cbbd59c4ad41c6dc1a94e07837bea5b80f259441de123ae9f5ac798fa561: (from 10.21.12.2:5000/go:v2 node:deis-3) die
2015-05-05T06:16:38.000000000Z 4bd1cff9637f8cbdfc2423cc19c80f1af33fc863c4299104cde1183996475800: (from 10.21.12.2:5000/go:v2 node:deis-3) die
2015-05-05T06:16:38.000000000Z cbe69d659d2b9e919fc20fa2b789c83b99e7456bfb69145c4d991f6e481629b1: (from 10.21.12.2:5000/go:v2 node:deis-3) stop
2015-05-05T06:16:38.000000000Z 9273cbbd59c4ad41c6dc1a94e07837bea5b80f259441de123ae9f5ac798fa561: (from 10.21.12.2:5000/go:v2 node:deis-3) stop
2015-05-05T06:16:38.000000000Z 4bd1cff9637f8cbdfc2423cc19c80f1af33fc863c4299104cde1183996475800: (from 10.21.12.2:5000/go:v2 node:deis-3) stop
2015-05-05T06:16:38.000000000Z c7526f5e81e3ad88d4e0d6b093bb31777a3a91bf05ca240e230a86d6e095b3f5: (from 10.21.12.2:5000/go:v2 node:deis-1) start
2015-05-05T06:16:38.000000000Z 5292230cafcdd45bc508fadabb4e0c3fc9b74c158ef01fc771c8d43aac9d8857: (from 10.21.12.2:5000/go:v2 node:deis-2) start
2015-05-05T06:16:38.000000000Z 4bd1cff9637f8cbdfc2423cc19c80f1af33fc863c4299104cde1183996475800: (from 10.21.12.2:5000/go:v2 node:deis-3) start
2015-05-05T06:16:38.000000000Z 9273cbbd59c4ad41c6dc1a94e07837bea5b80f259441de123ae9f5ac798fa561: (from 10.21.12.2:5000/go:v2 node:deis-3) start
2015-05-05T06:16:38.000000000Z cbe69d659d2b9e919fc20fa2b789c83b99e7456bfb69145c4d991f6e481629b1: (from 10.21.12.2:5000/go:v2 node:deis-3) start
```

controller/scheduler/swarm.py
```diff
@@ -0,0 +1,134 @@
...
+        """
+        Stop a container
+        """
+        self.docker_cli.stop(name)
+
+    def destroy(self, name):
+        """
+        Destroy a container
+        """
+        self.stop(name)
```
bacongobbler: is there a specific reason why we call `stop` here? the fleet scheduler just destroys the unit, so why does swarm rely on `stop`?

bacongobbler:
Initial testing looks good from my end, so code LGTM. My comments are more or less questions on some of the behavioural decisions.

controller/scheduler/swarm.py
mboersma: I'm not sure exactly why, but things failed when I removed this call to `stop()` as you had suggested, so I left it in.

mboersma:
@aledbf we tried to fix or address all your comments. Let me know what you think if you have time to re-review this PR.

controller/scheduler/swarm.py
smothiki: We cannot remove a running container. we have to stop the running container and then destroy the container. Some thing similar to ```docker rm -f```

bacongobbler: Okay, sounds reasonable, but what if we wanted to try destroying a container that is currently stopped such as when something like #3293 rolls around? Should we attempt to stop, but if that fails then we just destroy? Basically a

```
try:
    self.stop()
except:
    pass
finally:
    self.docker_cli.remove_container(name)
```

swarm/swarm.go
smothiki: I just tried swarm join with etcd and I think there is bug in the latest swarm image from docker repository.Only the last syntax works with etcd protocol
I have tried swarm join with the following  etcd protocol syntax only the last syntax worked. 
```
docker run --name swarm-node swarm join --addr=172.17.8.103:2375 etcd://172.17.8.100/mod/swarm
time="2015-05-05T16:29:10Z" level=fatal msg="501: All the given peers are not reachable (Tried to connect to each peer twice and failed) [0]" 
```
```
$ docker run --name swarm-node swarm join --addr=172.17.8.103:2375 etcd://172.17.8.100,172.17.8.101,172.17.8.102,172.17.8.103/mod/swarm
time="2015-05-05T16:33:36Z" level=fatal msg="501: All the given peers are not reachable (Tried to connect to each peer twice and failed) [0]"
```
```
$ docker run --name swarm-node swarm join --addr=172.17.8.103:2375 etcd://172.17.8.100:4001,172.17.8.101:4001,172.17.8.102:4001,172.17.8.103:4001/mod/swarm
time="2015-05-05T16:40:12Z" level=fatal msg="501: All the given peers are not reachable (Tried to connect to each peer twice and failed) [0]"
```
```
$ docker run --name swarm-node swarm join --addr=172.17.8.103:2375 etcd://172.17.8.100:4001/mod/swarm
time="2015-05-05T16:32:44Z" level=info msg="Registering on the discovery service every 25 seconds..." addr="172.17.8.103:2375" discovery="etcd://172.17.8.100:4001/mod/swarm" 
```

172.17.8.100 is the leader host.
I'll open an issue with docker/swarm

smothiki: @aledbf for next version I'll try using static files which avoids running swarm-node containers on each machine and bootstrapping swarm with custom deis wrapper. Right  now things we are very close to release and I dont want to change anything.
https://docs.docker.com/swarm/discovery/

controller/scheduler/swarm.py
smothiki: stopping an already stopped container and starting an already started container doesn't throw any errors. no need for exception handling cli.stop() and cli.start() .

bacongobbler: :+1:

Issue #3620, opened 2015-05-05
bug(router): remote etcd hosts not properly configurable


krancour:
`deis-router` sources its _own_ address _and_ the address of a machine running etcd from the _same_ environment variable:

https://github.com/deis/deis/blob/master/router/boot.go#L39-L49

```
host := getopt("HOST", "127.0.0.1")

// ...

hostEtcdPath := getopt("HOST_ETCD_PATH", "/deis/router/hosts/"+host)

// ...

client := etcd.NewClient([]string{"http://" + host + ":" + etcdPort})
```

This creates the unfortunate limitation of not being able to run the `deis-router` on any machine that isn't also running etcd.

According to CoreOS' own cluster architecture recommendations, it's not advisable to run etcd on more than nine nodes.  This means that large clusters, should run etcd on a small (probably fixed) number of nodes that run "common services," with a more elastic pool of worker nodes reserved for everything else.  In the case of a large Deis cluster, most Deis components probably run along with the "common services" as well.  Notable exceptions include `deis-logspout` and `deis-publisher`, which must run on the worker nodes.  (`deis-logspout` actually needs to run on both.)  I would argue that in a _large_ cluster where the `deis-router` component may need to scale to many instances, an admin might reasonably elect to schedule and scale that component on worker nodes.  But if the worker nodes aren't running etcd (and we've established they would not in such a case) then the code above is problematic.

Reference: https://coreos.com/docs/cluster-management/setup/cluster-architectures/

`deis-logspout` and `deis-publisher` both seem to address this issue by use of a dedicated `ETCD_HOST` environment variable.  The same approach could be used here, although I would argue that all three would benefit from having a comma delimited set of `ETCD_PEERS` specified instead, which can be parsed to an array and passed to `etcd.NewClient(...)` as that would improve their overall resilience.  Today, there would be problems using the host specified by `ETCD_HOST`if that host went offline-- even if the etcd cluster as a whole managed to maintain quorum/health.

krancour:
This is no longer relevant.  As of etcd 0.5.0, any machine that's not hosting etcd can host an etcd proxy instead.  We're now using etcd 2.1.2 and this feature is included.  Those who isolate etcd to a small, fixed number of nodes should be running the etcd proxy on all other nodes.  This is already adequately addressed in the documentation.

This means it's reasonable to _always_ connect to etcd at 127.0.0.1 and that makes this issue moot.

PR #3621, opened 2015-05-05
fix(router): make etcd host configurable by ETCD_HOST env var

krancour:
Addresses #3620


krancour:
I may follow with a separate PR to make ETCD_PEERS an opt-in improvement for this (and other components that may rely on a remote etcd host) to eliminate a potential SPOF.

bacongobbler:
I think every component asides from publisher relies on a remote etcd host, FYI, so it might make sense todo the same for the rest of the components as well

LGTM though :)

krancour:
@bacongobbler not sure I understood your comment correctly.  This PR brings `deis-router` in line with where other components already are.  See example from `deis publisher`:

https://github.com/deis/deis/blob/master/publisher/main.go#L29

Or was your comment specifically about the `ETCD_PEERS` idea?  If so, I agree, all the components could probably stand to receive that treatment (as an opt-in).

carmstrong:
> I think every component asides from publisher relies on a remote etcd host, FYI, so it might make sense to do the same for the rest of the components as well

:+1: I'd like to see them have this change, but it's probably something we can implement when we bring them all in line in #2701 (/cc @bacongobbler)

PR #3622, opened 2015-05-05
docs(components): clarify that deis-cache is optional

mboersma:


docs/understanding_deis/components.rst
```diff
@@ -27,7 +27,9 @@ platform state. Backups and WAL logs are pushed to :ref:`Store`.
 
 Cache
 -----
-The cache is an instance of `Redis`_ used by the registry.
+The cache component is an optional instance of `Redis`_. If the cache is running,
+it will be used by the :ref:`registry`. The ``deisctl`` provisioning tool does
```
carmstrong: Should we explain why one might choose to use the cache, or why it isn't used by default? It may be confusing for new users.

mboersma: How about:
> it will be used by the registry to speed up access to small files. This may be helpful when using a remote storage backend such as S3."

Reference: https://github.com/deis/deis/blob/master/registry/templates/config.yml#L28-L41

bacongobbler: We should probably clarify why we disabled it as well, which in this case is because we are running a modified fork of the registry whose endpoints don't play nicely with a cache. However, if the user is not sing `deis pull` then this is a nice optimization.

carmstrong: Yeah, I'd add a brief sentence about why we disabled it, and why one might want to enable it.

mboersma:
Ok, how about:
> The cache component is an optional instance of Redis. If the cache is running, it will be used by the :ref:`registry`.
Caching can cause problems for `deis pull`, but may be helpful when using a remote storage backend such as S3. The `deisctl` provisioning tool does not install or start the cache by default.

bacongobbler:
above docs LGTM

carmstrong:
LGTM.

Issue #1973, opened 2014-09-30
builder: Bad protocol version identification 'dummy-value' from 10.240.190.22


carmstrong:
When reading logs from builder, users may see something like:
```
Bad protocol version identification 'dummy-value' from 10.240.190.22
```

This is due to the changes made in #1940 - we had to refactor the ExecStartPost loops to use `ncat` instead of `cat /dev/tcp/...` because of the bash vulnerability fixed in [CoreOS 452.0.0](https://coreos.com/releases/#452.0.0) (specifically, bash no longer has direct network access). We need to pass something to ncat to send over the wire, so we use 'dummy-value'. We don't care that it's an invalid string - the import thing is that ncat can connect.

So, this is innocuous, but ideally we shouldn't need a string at all - maybe we can just send an end of transmission character?

mboersma:
After #4273 this no longer shows up in the logs.

PR #3625, opened 2015-05-05
fix(builder): silence pipefail errors on boot

mboersma:
Starting deis-builder shows lots of these warnings in the log:
```
/bin/sh: line 0: echo: write error: Broken pipe
```
Reworking the `ExecStartPost` line not to use a piped command silences this. Lots of users stare at the builder log, so it's worth cleaning it up.

### DIFF
--- a/deisctl/units/deis-builder.service
+++ b/deisctl/units/deis-builder.service
@@ -9,7 +9,7 @@ ExecStartPre=/bin/sh -c "IMAGE=`/run/deis/bin/get_image /deis/builder` && docker
 ExecStartPre=/bin/sh -c "docker inspect deis-builder >/dev/null 2>&1 && docker rm -f deis-builder || true"
 ExecStartPre=-/bin/sh -c "/sbin/losetup -f"
 ExecStart=/bin/sh -c "IMAGE=`/run/deis/bin/get_image /deis/builder` && docker run --name deis-builder --rm -p 2223:22 --volumes-from=deis-builder-data -c 800 -e EXTERNAL_PORT=2223 -e HOST=$COREOS_PRIVATE_IPV4 --privileged -v /etc/environment_proxy:/etc/environment_proxy $IMAGE"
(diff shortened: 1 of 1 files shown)


carmstrong:
Good find - didn't know `ncat` had an `--exec` option. Code LGTM.

This closes #1973, right?

mboersma:
> This closes #1973, right?

No, we're still sending a bogus 'dummy-value' string just to see if builder is alive, so we will get one "Bad Protocol" message when we start up.

carmstrong:
> No, we're still sending a bogus 'dummy-value' string just to see if builder is alive, so we will get one "Bad Protocol" message when we start up.

Yeah, bummer - that'd be nice to clean up.

mboersma:
If I change the command to:
```
 ncat localhost 2223 --exec '/usr/bin/echo -e SSH-2.0-deis probing builder\r\n'
```
Then it looks more like an SSH client, and you see one of two warnings in the builder logs (randomly, apparently):
```
May 05 21:04:17 deis-02 sh[3913]: Connection closed by 10.1.42.1 [preauth]
May 05 21:04:21 deis-02 sh[3913]: Write failed: Broken pipe [preauth]
```
I'm not sure that's any better WRT fixing #1973. We're probing for an SSH server here, and when `ncat` connects but isn't a proper SSH client, the server is going to squawk about that.

We could change the echo string so the log says:
```
May 05 21:10:33 deis-02 sh[3913]: Bad protocol version identification 'builder is running' from 10.1.42.1 port 59179
```

carmstrong:
> We're probing for an SSH server here, and when ncat connects but isn't a proper SSH client, the server is going to squawk about that.

Yeah, I don't think changing the message is worth the effort - it'll still be just as confusing.

carmstrong:
LGTM.

bacongobbler:
code LGTM :)

PR #3627, opened 2015-05-05
ref(controller): move mkdir to bin/boot

bacongobbler:
This is a mini optimization. We have a function and a place which we
set up directories which components rely upon. Since the controller now
relies on /deis/scheduler, this moves it over there.

I found this because I wasn't following the upgrade docs to a tee, but it made more sense to put it where all the rest are initialized anyways :)

### DIFF
--- a/controller/bin/boot
+++ b/controller/bin/boot
@@ -52,9 +52,10 @@ etcd_set_default webEnabled 0
 etcd_set_default unitHostname default
 
 # safely create required keyspaces
-etcd_safe_mkdir /deis/services
 etcd_safe_mkdir /deis/domains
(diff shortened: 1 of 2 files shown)


carmstrong:
Code LGTM.

mboersma:
Code LGTM--that is a more appropriate place than the systemd unit, and `etcd_safe_mkdir` is preferable.

smothiki:
code LGTM

carmstrong:
Jenkins passed - @mboersma are you okay with this going into 1.6 and rebuilding the images?

mboersma:
Yes please, this is more robust and should go into v1.6.

PR #3631, opened 2015-05-05
feat(contrib/azure/azure-coreos-cluster): automatically create discovery url

sedouard:
Added a the convenience of recreating azure-user-data for coreos with a new discovery
url. This prevents accidental cluster provisioning without creating a new url which
the user will only realize they did after the coreos has been provisioned and
attempts to connect with deisctl.

### DIFF
--- a/contrib/azure/azure-coreos-cluster
+++ b/contrib/azure/azure-coreos-cluster
@@ -55,6 +55,9 @@ parser.add_argument('--data-disk', action='store_true',
                    help='optional, attaches a data disk to each VM')
 parser.add_argument('--nohttps', action='store_true',
                    help='optional, disables the creation of the https load balanced endpoint')
+parser.add_argument('--no-discovery-url', action='store_true',
+                   help='optional, disables the creation of a new coreos discovery url')
+
 cloud_init_template = """#cloud-config
 
 coreos:
@@ -84,6 +87,11 @@ if not args.ssh_cert and not args.ssh_thumb:
       args.ssh_thumb = thumbprint.split('=')[1].replace('\n', '')
       args.ssh_cert = './ssh-cert.cer'
   print 'Generated SSH certificate with thumbprint ' + args.ssh_thumb
+
+# generate coreos discovery url
+if not args.no_discovery_url:
+  print 'Generating new CoreOS discovery URL for the cluster'
+  subprocess.call(['bash', '-c', 'python ./create-azure-user-data $(curl -s https://discovery.etcd.io/new)'])
 
 # Setup custom data
 if args.custom_data:
@@ -163,8 +171,7 @@ def network_config(subnet_name=None, port='59913', public_ip_name=None):
         if not args.nohttps:
           network.input_endpoints.input_endpoints.append(lb_endpoint_config('https', '443', load_balancer_probe(None, '443', 'tcp')))
         # create builder endpoint TCP probe check and extended timeout
-        network.input_endpoints.input_endpoints.append(lb_endpoint_config('builder', '2222',
-load_balancer_probe(None, '2222', 'tcp',), 20))
+        network.input_endpoints.input_endpoints.append(lb_endpoint_config('builder', '2222', load_balancer_probe(None, '2222', 'tcp',), 20))
     return network
 
 def data_hd(target_container_url, target_blob_name, target_lun, target_disk_size_in_gb):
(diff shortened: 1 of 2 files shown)


deis-admin:

sedouard:
same here - @timfpark gave this a look over on another thread. this one is just cuts out the coreos discovery url step.

carmstrong:
@sedouard Could you elaborate on when you would *not* want to generate a user-data on provision? It seems like the default case, to always generate a new discovery URL, is preferred.

I'm not sure I follow the motivation for this PR.

sedouard:
@carmstrong the current script does **not** create a new discovery url automatically. In the [doc](http://deis.io/get-deis/) the first thing to create a coreos cluster is to create the new url. This PR is to skip that step and just do it in the coreos script. There's almost never a case you would not and the flag, `--no-discovery-url` is simply there as an override. 99.99% of the time the flag won't be used and the *new* behavior will be to automatically create a new url.

The motivation is because I always forget to do this when I create a new cluster and only find out after the machines are provisioned.

carmstrong:
@sedouard I get the motivation behind this now, but I think it's a bit confusing to have this undocumented  `--no-discovery-url` but in the documentation we show passing a discovery URL to the script: http://docs.deis.io/en/latest/installing_deis/azure/#create-cluster-cloud-config

We should have these two agree - if we by default want the script to use a fresh discovery URL unless `--no-discovery-url`, the docs should omit the new URL as a parameter, and should mention this fact.

Does that make sense? Also, this PR needs a rebase - I think your other changes to the Azure scripts necessitate a rebase.

Thanks for your efforts on this!

contrib/azure/azure-coreos-cluster
```diff
@@ -85,6 +88,11 @@ if not args.ssh_cert and not args.ssh_thumb:
       args.ssh_cert = './ssh-cert.cer'
   print 'Generated SSH certificate with thumbprint ' + args.ssh_thumb
 
+# generate coreos discovery url
+if not args.no_discovery_url:
+  print 'Generating new CoreOS discovery URL for the cluster'
+  subprocess.call(['bash', '-c', 'python ./create-azure-user-data $(curl -s https://discovery.etcd.io/new)'])
```
carmstrong: Does it make sense to use the fully-qualified path to `create-azure-user-data` here, so we can run this script from any directory?

docs/installing_deis/azure.rst
```diff
@@ -71,6 +57,9 @@ With the management certificate and cloud config in place, we are ready to creat
 * Choose a cloud service name for your Deis cluster for the script below. The script will automatically create this cloud service for you.
 * Create an `affinity group`_ if you already don't have one. Supply it in quotes with the ``--affinity-group`` parameter. Although *using an affinity group is not mandatory*, it is **highly recommended** since it tells the Azure fabric to place all VMs in the cluster physically close to each other, reducing inter-node latency by a great deal. If you don't want ot use affinity groups, specify a `region`_ for Azure to use with a ``--location`` parameter. The default is ``"West US"``. If you specify both parameters, ``location`` will be ignored. Please note that the script *will not* create an affinity group by itself; it expects the affinity group exists.
 
+This script calls the ``./create-azure-user-data`` script which takes the stock cluster instance config in ``../coreos/user-data.example``, customizes it for Azure, and inserts a unique cluster discovery
+endpoint. It will then use the newly created coreos config on the newly provisioned cluster.
```
carmstrong: Can we change to "CoreOS" here?

carmstrong:
This is coming along - thanks for sticking with it, @sedouard! Your commits can also be squashed into one, and should follow the convention of using a colon instead of a dash after the first section.

A single commit like `feat(contrib/azure/azure-coreos-cluster): automatically create discovery url` would work. Thanks!

contrib/azure/azure-coreos-cluster
sedouard: Not really IMO, the script won't work today if you try to run it outside of `contrib/azure`, first thing it'll fail on is the call to `openssl` to create the ssh certificates. If you specify one on the command line it should work but by default (what's on the docs) is that it'll create a new ssh cert for the cluster. The doc also says "change directory into contrib/azure while following this documentation."

sedouard:
sorry I was on travel for a bit :-). Thanks for the words of encouragement :-). I'll make the fixes and squash.

carmstrong:
Code LGTM. Thanks @sedouard!

mboersma:
Code LGTM.

PR #3637, opened 2015-05-06
Suggest checking out a tag when getting the source

djmitche:
This would have saved me a bit of heartache; the issue was finally identified by @bacongobbler .

As it stands, this tag name will very soon be out of date, leaving people blithely copy/pasting (I would *never* do that, right?) running an old version of Deis.  But I don't see an easy fix for that..

It looks like `|release|` is always the release from which the docs are built, which will not correspond to the most recently released tag.  I don't know if there's any other automated way to generate the correct tag name.

### DIFF
--- a/docs/_includes/_get-the-source.rst
+++ b/docs/_includes/_get-the-source.rst
@@ -6,6 +6,9 @@ this documentation. Download an archive file from the `releases page`_, or use
 
     $ git clone https://github.com/deis/deis.git
     $ cd deis
+    $ git checkout v1.6.0
+
+Check out the latest Deis release, rather than using the default (master).
 
 If you contribute to Deis or build components locally, use ``go get`` instead to
 clone the source code into your `$GOPATH`_:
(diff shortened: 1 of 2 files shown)


deis-admin:

mboersma:
Good idea. Maybe it would also help to rename the GitHub "releases" branch as "old-releases" so as to avoid confusion there?

mboersma:
This adds another version string to update with each release. @djmitche could you also add `docs/_includes/_get-the-source.rst` just before `docs/installing_deis/install-deisctl.rst` in the three places that appears in `docs/contributing/releases.rst`? If this file is listed in those instructions, we'll remember to keep this version number in sync with releases.

djmitche:
Oo, cool!  done..

docs/_includes/_get-the-source.rst
```diff
@@ -6,6 +6,9 @@ this documentation. Download an archive file from the `releases page`_, or use
 
     $ git clone https://github.com/deis/deis.git
     $ cd deis
+    $ git checkout v1.5.2
```
carmstrong: This should be v1.6.0 now :)

djmitche: fixed

mboersma:
Jenkins, add to whitelist.

carmstrong:
Code LGTM.

Issue #3069, opened 2015-02-12
Proposal: Deis should support multiple ports for applications


carmstrong:
This was attempted in #2095, but the implementation is incomplete. Specifically, the PR is lacking:
* Router template logic to handle an application having multiple ports
* Tests wherever possible

Thanks for the great work in starting this, @gust1n! Also pinging @mattapperson as he expressed interest in tackling this.

refs #2044

juniorjbn:
Hi, this issue have any milestone ? or maybe a date to be fixed ? Dies didn't suport multiple ports make de things a lot harder.

bacongobbler:
there's a large design discussion in #2095 if you want to try tackling it, but there is currently no milestone for this.

mxk1235:
are there any plans to support this?  i'm willing to help, but having done no digging through the code, i would require some direction.

wrto #2095 , i think the limitation of supporting only one port is worse than having users provide explicit port configurations for their second/third port.  for example, the graphite dockerfile is a pretty commonly used one, but it is impossible to deploy here.

bacongobbler:
There are some fundamental changes with the routing mesh that would need to be replaced, so I think this would have to be done in a v2 so we don't have to worry too much about backwards compatibility. However, @mxk1235 if you're up for it and think you can do it in a backwards-compatible fashion, a [design document](http://docs.deis.io/en/latest/contributing/design-documents/) describing the architectural changes to the platform would be really helpful. It doesn't necessarily need to show code, but some form of how you'd expect we could hypothetically roll over users from the previous version to the next (as well as the expected user experience with this feature), then we can talk more in-depth on the implementation details in the proposal.

mxk1235:
thanks, i'll try to carve out some time to look into it.  is v2 the next planned major version, or are you working on 1.9 at the moment?

bacongobbler:
we're working on v1.9 at the moment :)

robeferre:
+1

bacongobbler:
Now that Deis Workflow has released a [stable v2.0.0 release](https://deis.com/blog/2016/workflow-stable/), we can start to tackle these problems with Kubernetes. Closing this in favour of deis/workflow#324 since we are not planning on backporting this feature to v1 and to keep the conversation going on over there, where new v2 users can see the feature request. Thanks everyone for all your support! :)

Issue #1315, opened 2014-07-14
TCP support for Deis applications


Xe:
This issue obsoletes #1061 and #18.

Also see #535.

Can an admin close #1061 and #18 and inform conversations to be directed here?

nekinie:
Is there any news on tcp support in the router?

bacongobbler:
We need to spec it out, make a proposal and hash out the details on this feature request. We need to figure out the UX as well as how we can properly implement this before moving onto development so we're all on the same page.

gust1n:
Maybe it's a stupid question but doesn't nginx handle TCP? http://nginx.com/resources/admin-guide/tcp-load-balancing/

aledbf:
@gust1n that is the commercial version of nginx.

Xe:
Another big problem is that a lot of applications that use IP addresses to figure out who is connecting and if they are banned or not use the remote TCP address to figure this out. With any sort of TCP proxy this is broken.

bacongobbler:
TCP support for applications will most likely be written up as an addon through #231, so it won't break our current application/proxy stack. How it will be architected is a separate question which we'd probably need to take a look at #231 first, but I'd imagine that it'd be a separate proxy layer outside of the core component stack.

chendo:
+1 for this feature.

There's a TCP module for nginx (https://github.com/yaoweibin/nginx_tcp_proxy_module) but I haven't used it yet, but we're planning to investigate it as a replacement for haproxy in our Marathon/Mesos cluster.

carmstrong:
> There's a TCP module for nginx

We're actually already using this - https://github.com/deis/deis/blob/ce484e5524b8570e511c041b327bdc38a70ff76a/router/build.sh#L66

chendo:
Oh whoops, never mind me then.

Is the strategy to update the config, then reload nginx? I was planning on using nginx-mruby to have to dynamically provide routing.

carmstrong:
> Is the strategy to update the config, then reload nginx?

Yep, that's how it happens currently - confd will re-template the nginx configuration, and then sighup the process to reload.

chendo:
How often does this happen? We have issues where our script currently reloads every 15s, but our deploys can happen faster than that so we end up with stale backends which cause 503s

carmstrong:
> How often does this happen?

[...]

PR #3640, opened 2015-05-06
chore(controller): remove $PORT logic from fleet

### LINKED ISSUE #1156: controller exposes all container port bindings instead of first
tested on deis revision: df1c0edd7f9f44999da5d11d5f2c674bf0b17a78  Tue Jun 10 14:00:57 2014 -0700

CoreOS 343.0.0

deis executes the following code:
 core@core1 ~ $ docker inspect -f '{{range $i, $e := .HostConfig.PortBindings }}{{$p := index $e 0}}{{$p.HostPort}}{{end}}' wp_v2.cmd.1
4915349154

the result is a concatenation of the 2 ports which obiously doesn't work.

bacongobbler:
For buildpack apps, [$PORT is exposed][1] as part of slugrunner's job. For Dockerfile/Docker images, the internal port numbers are already known to the developer, so if they want to follow 12-factor best practices then they will have set up $PORT in their own image. I tried seeing if there was a way to work around the docker API's limitations but didn't find one, and then it dawned on me that this is only relevant for users which *do not have control over the exposed ports*. For Dockerfiles/Docker images, this is managed by them so this logic was over-engineered.

deis/publisher exposes the port with the lowest numerical value to the router (docker sorts the internal ports before exposing them in the API), which is why the documentation is reflected to show this.

[1]: https://github.com/deis/deis/blob/e6a078b4119887b4de73e8a6d04291b7cdeb823b/builder/image/slugrunner/Dockerfile#L10-L12

closes #1156. We've been able to run multi-port containers for a while, it's just not what users would expect. This bug was (apparently) gone a long time ago. This is just updating the docs to reflect how to use it in today's world.

Official multi-port support is still untouched: #2095 #3069

### DIFF
--- a/docs/using_deis/using-docker-images.rst
+++ b/docs/using_deis/using-docker-images.rst
@@ -31,7 +31,7 @@ Docker Image Requirements
 In order to deploy Docker images, they must conform to the following requirements:
 
  * The Docker image must EXPOSE only one port
- * The exposed port must be an HTTP service that can be connected to an HTTP router
+ * The port must be listening for a HTTP connection
  * A default CMD must be specified for running the container
 
 .. note::
--- a/controller/scheduler/fleet.py
+++ b/controller/scheduler/fleet.py
@@ -379,7 +379,7 @@ CONTAINER_TEMPLATE = [
     {"section": "Unit", "name": "Description", "value": "{name}"},
     {"section": "Service", "name": "ExecStartPre", "value": '''/bin/sh -c "IMAGE=$(etcdctl get /deis/registry/host 2>&1):$(etcdctl get /deis/registry/port 2>&1)/{image}; docker pull $IMAGE"'''},  # noqa
     {"section": "Service", "name": "ExecStartPre", "value": '''/bin/sh -c "docker inspect {name} >/dev/null 2>&1 && docker rm -f {name} || true"'''},  # noqa
-    {"section": "Service", "name": "ExecStart", "value": '''/bin/sh -c "IMAGE=$(etcdctl get /deis/registry/host 2>&1):$(etcdctl get /deis/registry/port 2>&1)/{image}; port=$(docker inspect -f '{{{{range $k, $v := .ContainerConfig.ExposedPorts }}}}{{{{$k}}}}{{{{end}}}}' $IMAGE | cut -d/ -f1) ; docker run --name {name} {memory} {cpu} {hostname} -P -e PORT=$port $IMAGE {command}"'''},  # noqa
+    {"section": "Service", "name": "ExecStart", "value": '''/bin/sh -c "IMAGE=$(etcdctl get /deis/registry/host 2>&1):$(etcdctl get /deis/registry/port 2>&1)/{image}; docker run --name {name} {memory} {cpu} {hostname} -P $IMAGE {command}"'''},  # noqa
     {"section": "Service", "name": "ExecStop", "value": '''/usr/bin/docker stop {name}'''},
     {"section": "Service", "name": "ExecStop", "value": '''/usr/bin/docker rm -f {name}'''},
     {"section": "Service", "name": "TimeoutStartSec", "value": "20m"},
--- a/docs/using_deis/using-dockerfiles.rst
+++ b/docs/using_deis/using-dockerfiles.rst
@@ -22,7 +22,7 @@ Dockerfile Requirements
 In order to deploy Dockerfile applications, they must conform to the following requirements:
 
  * The Dockerfile must EXPOSE only one port
- * The exposed port must be an HTTP service that can be connected to an HTTP router
+ * The port must be listening for a HTTP connection
  * A default CMD must be specified for running the container
 
 .. note::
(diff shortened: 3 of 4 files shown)


bacongobbler:
Note that I tested and confirmed this by adding `EXPOSE 5001` in my Dockerfile for example-dockerfile-python. Worked great!

docs/using_deis/using-docker-images.rst
```diff
@@ -30,14 +30,10 @@ Docker Image Requirements
 ^^^^^^^^^^^^^^^^^^^^^^^^^
 In order to deploy Docker images, they must conform to the following requirements:
 
- * The Docker image must EXPOSE only one port
- * The exposed port must be an HTTP service that can be connected to an HTTP router
+ * The Docker image must EXPOSE at least one port
```
bacongobbler: Not sure how to clarify that official multi-port images (i.e exposing two HTTP processes in a container, or a web and an SSH process once #1315 is tackled) are still unsupported. Any help on this would be appreciated :)

carmstrong:
While this updates the documentation to be more in-line with what's actually happening, it seems to reveal that publisher is doing something that doesn't really make sense. It's not hard to imagine a user asking "Why use the lowest port?" when reading this documentation.

Instead of making this change, perhaps we should reconsider what publisher is doing here? Is supporting multiple ports (#3069) something we should consider doing?

bacongobbler:
It'd be a nice feature for sure! Maybe a happy medium then would to *not* point out how publisher works at the time being (by taking out the note) and consider multi-port apps still unsupported? I was doing a fly-by cleanup more than anything. Multi-port support would be a larger overhaul but I'd be happy to hash out a formal proposal. This just cleans up a lot of the ugly scheduler logic that wasn't necessary here

eMxyzptlk:
Will this allow me to deploy an app that does not expose any ports (think of a data-manipulation micro-service)? Up until now, I had to run a Go server that has no handlers to get out of this problem.

I love that `$PORT` is going to get killed!

bacongobbler:
> Will this allow me to deploy an app that does not expose any ports (think of a data-manipulation micro-service)? Up until now, I had to run a Go server that has no handlers to get out of this problem.

Deis still deploys a `web` process by default, though it's not mandatory. You can always scale it down to 0 and your `worker` process up to 1. This has been the workaround that others have used to deploy apps which don't require exposing any endpoints.

> I love that $PORT is going to get killed!

$PORT's never going to get killed while we're keeping compatibility with Heroku :) There was some overlap between the scheduler and slugrunner (the image that runs Heroku apps) setting up $PORT, so I removed the former for the latter: https://github.com/deis/deis/blob/master/builder/image/slugrunner/Dockerfile#L11-L12

carmstrong:
> Maybe a happy medium then would to not point out how publisher works at the time being (by taking out the note) and consider multi-port apps still unsupported?

:+1:

> This just cleans up a lot of the ugly scheduler logic that wasn't necessary here

Yeah, that cleanup is definitely nice. Is that no longer necessary because publisher does the inspection?

I'm in favor of leaving the documentation as-is and cleaning up the app template, but I'd like to hear what @mboersma thinks.

bacongobbler:
Yes, publisher inspects the ports and exposes the first port it finds (not the best way to solve it, but it works for single-port apps)

mboersma:
I think the behavior of "lowest-numbered port wins" is acceptable for now if it's documented, as it is in this PR. And if I read correctly, the fleet.py unit template code removed here wasn't doing anything anyway (and was very hard to parse). So LGTM IMHO.

carmstrong:
> I think the behavior of "lowest-numbered port wins" is acceptable for now if it's documented, as it is in this PR

The question is *should* this be explicitly documented to be the behavior, or should the language stay that we don't support multiple-port Dockerfiles. I prefer the latter, as that's less confusing for users.

mboersma:
Since there isn't official multi-port support and Deis intends to do that properly at some point, maybe it is counterproductive to document the current behavior. Could we drop that change from the external docs, but add code comments at https://github.com/deis/deis/blob/master/publisher/server/publisher.go#L102 that "lowest-numbered port wins" is the current behavior? tl:dr I agree with @carmstrong now.

bacongobbler:
sure :)

eMxyzptlk:
> Deis still deploys a web process by default, though it's not mandatory. You can always scale it down to 0 and your worker process up to 1. This has been the workaround that others have used to deploy apps which don't require exposing any endpoints.

Interesting, I never knew that. I do not wish to derail the discussion here much but I have one question about this, aren't all of the apps with a Dockerfile tied to one process (the cmd)? So can you actually define multiple Dockerfiles for multiple process types?

> $PORT's never going to get killed while we're keeping compatibility with Heroku :) There was some overlap between the scheduler and slugrunner (the image that runs Heroku apps) setting up $PORT, so I removed the former for the latter: https://github.com/deis/deis/blob/master/builder/image/slugrunner/Dockerfile#L11-L12

I understand, but I meant killed from being in the environment automagically when we clearly have a Dockerfile. It caused me some getting used to that the port being exposed is going to be in $PORT and I can't own that variable.

bacongobbler:
> aren't all of the apps with a Dockerfile tied to one process (the cmd)?

You can supply a Procfile to define multiple process types for an application, even for Dockerfile-based applications: http://docs.deis.io/en/latest/using_deis/process-types/#deploying-to-deis

> So can you actually define multiple Dockerfiles for multiple process types?

All process types share the same container image, but they run in separate containers (again, a Heroku thing).

bacongobbler:
pushed the doc edit as requested

carmstrong:
Code LGTM.

mboersma:
Code LGTM.

PR #3645, opened 2015-05-07
fix(store): lower number of placement groups

### LINKED ISSUE #3638: Ceph reports "too many PGs per OSD" in Deis 1.6
In testing Deis 1.6, my cluster reports:

```
     health HEALTH_WARN
            too many PGs per OSD (1536 > max 300)
```

This seems to be a new warning in the Hammer release of Ceph, which we're shipping in Deis 1.6. This got me thinking - perhaps we aren't calculating placement group usage properly, and this is contributing to the various issues we're seeing with Ceph performance (particularly on platform start).

I posted a message to the Ceph mailing list asking for clarification: http://lists.ceph.com/pipermail/ceph-users-ceph.com/2015-May/001222.html

We're using the default 12 pools:
```
root at ca-deis-1:/# ceph osd lspools
0 rbd,1 data,2 metadata,3 .rgw.root,4 .rgw.control,5 .rgw,6 .rgw.gc,7 .users.uid,8 .users,9.rgw.buckets.index,10 .rgw.buckets,11 .rgw.buckets.extra,
```

So, I think the calculation is something like:
12 pools * 128 = 1536
12 pools * 32 = 384
12 pools * 16 = 192

It seems that the placement group documentation is unclear. If we should shoot for 300 PGs per OSD, then it seems like we should use a `pg_num` of 16, since even 32 would result in more than 300 per OSD.

However, if this computation is based on the resources of the host, it'll be harder to choose a default value.

carmstrong:
We were incorrectly configuring a large number of placement groups
for the Ceph storage pools, resulting in each store-daemon doing
more work than necessary.

This commit removes the logic to reconfigure the number of placement
groups for the default pools (since they are already larger than 64),
and lowers the default number of placement groups from 128 to 64,
resulting in 768 placement groups per daemon. We also reconfigure
the warning for number of placement groups per OSD to 800 so we
no longer warn in the typical case.

closes #3638


carmstrong:
After this change:
```console
root@deis-01:/# ceph -s
    cluster 82f06ec8-c094-499e-b1db-f4f9aab58daf
     health HEALTH_OK
     monmap e3: 3 mons at {deis-01=172.17.8.100:6789/0,deis-02=172.17.8.101:6789/0,deis-03=172.17.8.102:6789/0}
            election epoch 8, quorum 0,1,2 deis-01,deis-02,deis-03
     mdsmap e6: 1/1/1 up {0=deis-01=up:active}, 2 up:standby
     osdmap e21: 3 osds: 3 up, 3 in
      pgmap v29: 192 pgs, 9 pools, 5809 bytes data, 65 objects
            17180 MB used, 28124 MB / 47831 MB avail
                 192 active+clean
  client io 5733 B/s rd, 1842 B/s wr, 15 op/s
root@deis-01:/# ceph osd lspools
0 rbd,1 data,2 metadata,3 .rgw.root,4 .rgw.control,5 .rgw,6 .rgw.gc,7 .users.uid,8 .users,
root@deis-01:/# ceph osd dump
epoch 21
fsid 82f06ec8-c094-499e-b1db-f4f9aab58daf
created 2015-05-07 20:53:08.276818
modified 2015-05-07 20:54:19.301211
flags
pool 0 'rbd' replicated size 3 min_size 1 crush_ruleset 0 object_hash rjenkins pg_num 64 pgp_num 64 last_change 1 flags hashpspool stripe_width 0
pool 1 'data' replicated size 3 min_size 1 crush_ruleset 0 object_hash rjenkins pg_num 16 pgp_num 16 last_change 11 flags hashpspool crash_replay_interval 45 stripe_width 0
pool 2 'metadata' replicated size 3 min_size 1 crush_ruleset 0 object_hash rjenkins pg_num 16 pgp_num 16 last_change 10 flags hashpspool stripe_width 0
pool 3 '.rgw.root' replicated size 3 min_size 1 crush_ruleset 0 object_hash rjenkins pg_num 16 pgp_num 16 last_change 12 flags hashpspool stripe_width 0
pool 4 '.rgw.control' replicated size 3 min_size 1 crush_ruleset 0 object_hash rjenkins pg_num 16 pgp_num 16 last_change 14 flags hashpspool stripe_width 0
pool 5 '.rgw' replicated size 3 min_size 1 crush_ruleset 0 object_hash rjenkins pg_num 16 pgp_num 16 last_change 16 flags hashpspool stripe_width 0
pool 6 '.rgw.gc' replicated size 3 min_size 1 crush_ruleset 0 object_hash rjenkins pg_num 16 pgp_num 16 last_change 17 flags hashpspool stripe_width 0
pool 7 '.users.uid' replicated size 3 min_size 1 crush_ruleset 0 object_hash rjenkins pg_num 16 pgp_num 16 last_change 18 flags hashpspool stripe_width 0
pool 8 '.users' replicated size 3 min_size 1 crush_ruleset 0 object_hash rjenkins pg_num 16 pgp_num 16 last_change 20 flags hashpspool stripe_width 0
max_osd 3
osd.0 up   in  weight 1 up_from 7 up_thru 20 down_at 0 last_clean_interval [0,0) 172.17.8.102:6800/1 172.17.8.102:6801/1 172.17.8.102:6802/1 172.17.8.102:6803/1 exists,up 2a7d881a-17ce-4526-ac67-58d9809112f7
osd.1 up   in  weight 1 up_from 5 up_thru 20 down_at 0 last_clean_interval [0,0) 172.17.8.101:6800/1 172.17.8.101:6801/1 172.17.8.101:6802/1 172.17.8.101:6803/1 exists,up 47fe2c74-a882-49f5-bbd9-4ca491ed4804
osd.2 up   in  weight 1 up_from 5 up_thru 20 down_at 0 last_clean_interval [0,0) 172.17.8.100:6800/1 172.17.8.100:6801/1 172.17.8.100:6802/1 172.17.8.100:6803/1 exists,up c69fdf07-9adf-4a9e-84cd-8e0eac3cc820
```

mboersma:
You had me at "1 addition and 41 deletions." Code LGTM.

bacongobbler:
code LGTM. Are there any other side-effects other than a shorter boot time with this change? Are there any trade-offs for having less than 200 PGs per daemon? Could this have also been the cause of why some user's gateway fails to come up?

carmstrong:
There's still some discussion around the ideal number of placement groups. It seems others are confused by this as well: http://lists.ceph.com/pipermail/ceph-users-ceph.com/2015-May/001297.html

I may try a `pg_num` of 64 and then change the warning threshold. With the current implementation, 128 placement groups results in a replication time of 10 minutes when adding a new host to the cluster. It'll be interesting to see how this metric changes on similar hardware when tuning the `pg_num`.

> Are there any other side-effects other than a shorter boot time with this change? Are there any trade-offs for having less than 200 PGs per daemon?

There's some information on PG calculations and tradeoffs here: http://ceph.com/docs/master/rados/operations/placement-groups/#placement-groups-tradeoffs - basically, the more the better, but that is bounded by CPU/memory constraints, and results in more work during replication.

> Could this have also been the cause of why some users' gateway fails to come up?

From what I've seen, gateways failing to come up are resolved automatically with time. What seems to be happening is that the [logic](https://github.com/deis/deis/blob/d1f5019d2a801cdfdf46f9b4d1a7a449b646e43e/store/gateway/bin/boot#L34-L71) in the gateway to increase the number of placement groups for the default pools just takes some time (sometimes more than others). I've seen this complete by itself, and sometimes necessitating a daemon restart.

carmstrong:
> I may try a pg_num of 64 and then change the warning threshold.

I made this change based on the ongoing discussion in the Ceph mailing list. This halves the number of placement groups we have, reducing the workload on replication. Another fact supporting this change is that the initial pool to be created, `rbd`, has seemingly always defaulted to 64 placement groups, so it seems like a sane enough number.

Ready for re-review @deis-admin

bacongobbler:
code still LGTM

store/monitor/bin/boot
```diff
@@ -6,10 +6,13 @@ set -e
...
 ETCD="$HOST:$ETCD_PORT"
 ETCD_PATH=${ETCD_PATH:-/deis/store}
-NUM_STORES=${NUM_STORES:-3}
-PG_NUM=${PG_NUM:-128} # default for 3 OSDs
 HOSTNAME=`hostname`
 
+# These defaults are for 3 hosts
+NUM_STORES=${NUM_STORES:-3}
+PG_NUM=${PG_NUM:-64}
+PGS_PER_OSD_WARNING=${PGS_PER_OSD_WARNING:-800} # 12 pools * 64 pgs per pool = 768
```
aledbf: This should not be the current number (1536) to avoid the warning in an existing cluster?

carmstrong: Hmm...

I kind of appreciate the warning, but we should probably suppress it since we've addressed it for future installs, and existing clusters are no less correct than they were before upgrading.

I'll make that change.

carmstrong:
Changed the warning so we we don't warn on the old number of PGs.

mboersma:
Code LGTM.

carmstrong:
Jenkins passed last run (previous was an unrelated "App not started" error). Merging...

wenzowski:
:+1:

@carmstrong I'm missing something. Would we end up with the old number of PGs when upgrading an existing cluster?

carmstrong:
> Would we end up with the old number of PGs when upgrading an existing cluster?

Yes, so we suppress the warning for that number of placement groups. The smaller number takes effect on new clusters. There's no way to shrink the number of placement groups for a pool once the pool is initially created, so that's nothing we can do on upgrade.

wenzowski:
Gotcha. Thanks

PR #3655, opened 2015-05-08
fix(deisctl): prefix unit name with "deis-" before expanding

### LINKED ISSUE #3497: deisctl tries to destroy app named "registry" when uninstalling platform
```
$ deisctl stop platform
● ▴ ■
■ ● ▴ Stopping Deis...
▴ ■ ●

Routing mesh...
deis-router@1.service: inactive/dead                                 
deis-router@2.service: inactive/dead                                 
deis-router@3.service: inactive/dead                                 
Data plane...
deis-publisher.service: inactive/dead                                 
Control plane...
Could not parse target: registry_v4.web.1.service
```

We have a app named as "registry", maybe this is a problem. I tried to start platform and hangs exactly the same.

Now deis client is getting 503 service unavailable for any command that i try. I'd like scale 0 or even destroy this app.

My units status: https://gist.github.com/rogerleite/f18b05a2099b10529a23

mboersma:
Commands such as `deisctl stop platform` were doing wildcard searches for the short names of units, effectively telling fleet to find all `*registry*` units, for example. This changes that wildcard to `*deis-registry*` and adds a test to ensure `deisctl` won't find a random app unit with "registry" in its name in that case.

Closes #3497.

### DIFF
--- a/deisctl/backend/fleet/utils.go
+++ b/deisctl/backend/fleet/utils.go
@@ -88,6 +88,10 @@ func splitTarget(target string) (component string, num int, err error) {
 // expand a target to all installed units
 func expandTargets(c *FleetClient, targets []string) (expandedTargets []string, err error) {
 	for _, t := range targets {
+		// ensure unit name starts with "deis-"
+		if !strings.HasPrefix(t, "deis-") {
+			t = "deis-" + t
+		}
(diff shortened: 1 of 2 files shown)


bacongobbler:
code LGTM!

carmstrong:
Code LGTM.

PR #3656, opened 2015-05-08
fix(nonstring-default) numeric default for webEnabled

djmitche:
`/deis/controller/webEnabled` uses Python's `bool` to parse this
configuration value; `0` and `1` are handled correctly, but `false` and
`true` lead to NameError.

### DIFF
--- a/docs/customizing_deis/controller_settings.rst
+++ b/docs/customizing_deis/controller_settings.rst
@@ -41,7 +41,7 @@ setting                                   description
 ====================================      ======================================================
 /deis/controller/registrationMode         set registration to "enabled", "disabled", or "admin_only" (default: "enabled")
 /deis/controller/schedulerModule          scheduler backend (default: "fleet")
-/deis/controller/webEnabled               enable controller web UI (default: false)
+/deis/controller/webEnabled               enable controller web UI (default: 0)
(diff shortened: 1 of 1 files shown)


bacongobbler:
Looks consistent with https://github.com/deis/deis/blob/eae2cfe8b674475f190b24a3853026245cf6d9e4/controller/bin/boot#L51, so docs LGTM

mboersma:
The documentation was incorrect: it should have been python's `False` and `True`, although `0` and `1` will work:
```python
>>> print bool(False)
False
>>> print bool(True)
True
>>> print bool(false)
  File "<stdin>", line 1, in <module>
NameError: name 'false' is not defined
>>> print bool(0)
False
>>> print bool(1)
True
```

djmitche:
And in trying to set the value to a proper Python identifier, it seemed to downcase it on me:
```
core@ip-10-21-2-25 ~ $ etcdctl set /deis/controller/webEnabled False
false
```
so I think the numeric values are the only options.

mboersma:
> it seemed to downcase it on me

That's odd. But I agree that `0` and `1` are less error-prone in the first place. And the [controller `boot` script](https://github.com/deis/deis/blob/master/controller/bin/boot#L51) agrees with you.

Code LGTM.

PR #3660, opened 2015-05-10
fix(logger): avoid hitting etcd for each log line

### LINKED ISSUE #3657: high load + logger causes excessive traffic to etcd
Running some very small load generators while playing with sysdig uncovered the fact that the syslog module appears to be hitting up etcd for the drain key on every log line.

Not sure if you want to memoize with a TTL or config file + confd.

2015-05-08 at 17 11

2015-05-08 at 17 13

Edit: forgot the code!

https://github.com/deis/deis/blob/df00fb4e41ed1f825e6c210c637d2e4772326f64/logger/syslogd/syslogd.go#L87

kalbasit:
fixes #3657

[Review on Reviewable](https://reviewable.io/reviews/deis/deis/3660)

### DIFF
--- a/logger/main.go
+++ b/logger/main.go
@@ -17,6 +17,7 @@ import (
 var (
 	logAddr         string
 	logPort         int
+	drainURI        string
 	enablePublish   bool
 	publishHost     string
 	publishPath     string
@@ -28,6 +29,7 @@ var (
 func init() {
 	flag.StringVar(&logAddr, "log-addr", "0.0.0.0", "bind address for the logger")
 	flag.IntVar(&logPort, "log-port", 514, "bind port for the logger")
+	flag.StringVar(&drainURI, "drain-uri", "", "default drainURI, once set in etcd, this has no effect.")
 	flag.StringVar(&syslogd.LogRoot, "log-root", "/data/logs", "log path to store logs")
 	flag.BoolVar(&enablePublish, "enable-publish", false, "enable publishing to service discovery")
 	flag.StringVar(&publishHost, "publish-host", getopt("HOST", "127.0.0.1"), "service discovery hostname")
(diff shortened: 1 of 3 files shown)


mboersma:
Code LGTM.

carmstrong:
@bacongobbler Does this look like a reasonable solution?

logger/syslogd/syslogd.go
```diff
@@ -97,7 +101,7 @@ func (h *handler) mainLoop() {
 }
 
 // Listen starts a new syslog server which runs until it receives a signal.
-func Listen(signalChan chan os.Signal, cleanupDone chan bool, bindAddr string) {
+func Listen(exitChan, cleanupDone chan bool, drainChan chan string, bindAddr string) {
```
bacongobbler: I think we still want to listen for specific signals (SIGTERM, SIGKILL, etc) as we would want to handle those separately. Would you please change `exitChan` and the respective `select` loop to listen for those signals?

bacongobbler:
On initial thoughts:

 - `main.go` contains flag `--drain-uri` :heavy_check_mark:
 - `main.go` watches etcd for key changes :heavy_check_mark:
 - `main.go` updates drain URI when the keys change :heavy_check_mark: 

I still think this needs more reviewing/testing, though. There looks like there's a bit of feature creep in this PR which I'd like to inspect more closely.

logger/syslogd/syslogd.go
eMxyzptlk: If I do, then I'd have to create two separate signal notifiers for two separate channels. As I need to know in `main.go` when a signal has fired to shutdown the etcd watcher.

eMxyzptlk: one possible solution is to have a signal notifier inside `syslogd.Listen` and one in main for the etcd watcher but I believe that it's better to handle all signals from main. I'm not opposed to have one channel per signal.

bacongobbler: > If I do, then I'd have to create two separate signal notifiers for two separate channels. As I need to know in main.go when a signal has fired to shutdown the etcd watcher.

The etcd watcher and the log server are two separate events, and we may include infrastructure in the future to handle different signals. I'd rather keep it the way things are if we can.

logger/syslogd/syslogd.go
```diff
@@ -106,18 +110,24 @@ func Listen(signalChan chan os.Signal, cleanupDone chan bool, bindAddr string) {
...
-		cleanupDone <- true
+	for {
+		select {
+		case <-exitChan:
+			// Shutdown the server
+			fmt.Println("Shutting down...")
+			s.Shutdown()
+			cleanupDone <- true
+		case d := <-drainChan:
+			h.drainURI = d
```
bacongobbler: could you please add a test or at the least a TODO to add tests for this? We'd want to make sure that the drain URI is updated when we send some new value over the channel.

logger/main.go
```diff
@@ -42,26 +44,49 @@ func main() {
...
+
 	for {
-		setEtcd(client, etcdPath+"/host", host, ttl)
-		setEtcd(client, etcdPath+"/port", port, ttl)
-		time.Sleep(time.Duration(publishInterval))
+		select {
+		case <-t.C:
+			setEtcd(client, etcdPath+"/host", host, ttl)
+			setEtcd(client, etcdPath+"/port", port, ttl)
+		case <-exitChan:
```
bacongobbler: We should unset the etcd key when we're exiting. The TTL's more of a safety net :)

bacongobbler: actually since this isn't part of the PR let's just add another TODO here as well, if you wouldn't mind

bacongobbler:
code LGTM. There's some cleanup that we can address later but the implementation looks fine

carmstrong:
Code LGTM if it's passed Jenkins and @bacongobbler is satisfied.

bacongobbler:
Yeah I can address my own nitpicks in a future PR. Thanks @kalbasit!

kalbasit:
Thanks for merging @bacongobbler sorry I missed your comments.

bacongobbler:
no worries! They weren't breaking changes, just my own opinions on how the world should turn. :earth_americas:

PR #3661, opened 2015-05-10
fix(deis): validate the syntax of the environment file

kalbasit:
fixes #3658

PR #3662, opened 2015-05-10
fix(client): ignore blank lines in .env on config:push|pull

### LINKED ISSUE #3658: deis config:pull and config:push fail on empty lines
Whenever an empty line is present in the .env file (like at the end), the two commands ``deis config:pull`` and ``deis config:push`` fail in a cryptic way.

Just skip empty lines, please.

    $ deis --version
    1.6.0

It behaves like this:

    [cimnine@local myapp]$ cat .env
    bla
    
    [cimnine@local myapp]$ deis config:push
    Usage: deis config:push [options]
    [cimnine@local myapp]$ deis config:pull
      File "<string>", line 2466, in <module>
      File "<string>", line 2462, in main
      File "<string>", line 2395, in _dispatch_cmd
      File "<string>", line 1332, in config_pull
    IndexError: list index out of range

mboersma:
This makes `deis config:pull` ignore blank lines and remove invalid entries, and it makes `deis config:push` ignore blank lines and print an error on an invalid entry.

```console
$ cat .env
bla

$ deis config:push
not a key=val entry: "bla"
Usage: deis config:push [options]
$ deis config:pull
removed invalid entry: "bla"
$ cat .env
TEST=1
FOO=bar
```

That `config:push` errors still end with printing the `usage:` string is annoying, but I couldn't find a fix for that behavior.

Closes #3658.


mboersma:
Oops, wow--I did this before I saw #3661. :laughing:

eMxyzptlk:
funny that we ended up with a similar solution :)

client/deis.py
```diff
@@ -1302,6 +1302,19 @@ def config_unset(self, args):
         else:
             raise ResponseError(response)
 
+    def _load_env(self, envfile='.env'):
+        try:
+            # load from .env, if it exists
+            with open(envfile) as f:
+                for line in (l.strip().strip('\n') for l in f):
+                    if '=' in line:
```
eMxyzptlk: This is probably not enough `=hello` matches this but results in no key.

carmstrong:
Code LGTM. This seems slightly preferred to #3661, since ignoring the empty/invalid lines is preferred to rejecting the file outright, no?

eMxyzptlk:
@carmstrong in #3661 I ignore the line, I do not reject the file.

mboersma:
#3661 was here first and should be preferred if it behaves similarly.

PR #3665, opened 2015-05-11
fix(contrib/ec2): use EBS volume for etcd data

### LINKED ISSUE #3618: ec2: Consider using EBS volume for etcd
When running a 5 node cluster, stopping and starting a node causes the cluster to fail.

Using the cloud formation scripts do the following:

* Boot a 5 node cluster in amazon.
* Ensure etcd is running and has consensus.
* Stop a node (thus wiping out ephemeral storage).
* Start the Node.

Expected: Cluster Unaffected, Node Rejoin
Actual: Cluster-wide failure, Ephemeral disk fails to mount, All nodes crash etcd.

Solution?

Don't use ephemeral storage for etcd.

carmstrong:
The ephemeral volume doesn't persist through host stop/start cycles.

/cc @jwaldrip

closes #3618

### DIFF
--- a/contrib/ec2/deis.template.json
+++ b/contrib/ec2/deis.template.json
@@ -275,6 +275,10 @@
           {
             "DeviceName" : "/dev/xvdf",
             "Ebs" : { "VolumeSize" : "100", "VolumeType": { "Ref": "EC2EBSVolumeType" } }
+          },
+          {
+            "DeviceName" : "/dev/xvdg",
+            "Ebs" : { "VolumeSize" : "10", "VolumeType": { "Ref": "EC2EBSVolumeType" } }
           }
         ]
       }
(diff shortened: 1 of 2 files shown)


jwaldrip:
:+1:

bacongobbler:
Were you able to reproduce the original issue? Etcd falling flat on its face just because we rebooted one node seems odd (maybe the leader?). If we had some logs to justify why this change is necessary (i.e. before/after) then I'd be more confident in this fix.

carmstrong:
> Etcd falling flat on its face just because we rebooted one node seems odd (maybe the leader?).

Yes, that's likely the cause of that issue. This pull request doesn't address that at all. 

The issue we're solving here is a loss of etcd data on host stop/start on AWS. The ephemeral device is wiped on stop/start, so you can easily hose an etcd member by stopping it. While this isn't something that the average user does often, it does seem like a silly failure case.

mboersma:
Code LGTM.

bacongobbler:
alrighty than. Code LGTM

PR #3669, opened 2015-05-11
fix(client): add requirements.txt to pypi distribution

### LINKED ISSUE #3668: Not able to install deis on Travis.
I'm getting the following error on Travis:

```
Installing deploy dependencies
!!! Deis support is experimental !!!
pip install --user deis==1.6.0
You are using pip version 6.0.8, however version 6.1.1 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
Collecting deis==1.6.0
  Downloading deis-1.6.0.tar.gz
      File "<string>", line 20, in <module>
      File "/tmp/pip-build-AcvuUw/deis/setup.py", line 30, in <module>
        with open('requirements.txt') as f:
    IOError: [Errno 2] No such file or directory: 'requirements.txt'
    Complete output from command python setup.py egg_info:
    
      File "<string>", line 20, in <module>
    
      File "/tmp/pip-build-AcvuUw/deis/setup.py", line 30, in <module>
    
        with open('requirements.txt') as f:
    
    IOError: [Errno 2] No such file or directory: 'requirements.txt'
    
    ----------------------------------------
    Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-build-AcvuUw/deis
```

mboersma:
Tested locally with `python setup.py sdist && pip install dist/deis-1.7.0-dev.zip`: broken before, works now.

Closes #3668.

### DIFF
--- a/client/MANIFEST.in
+++ b/client/MANIFEST.in
@@ -1,1 +1,1 @@
-include *.rst LICENSE
+include *.rst LICENSE requirements.txt


bacongobbler:
oops. LGTM!

mboersma:
This may necessitate a v1.6.1 release if we want to fix `pip install deis`. I'm not sure how urgent that is, so I'd prefer to wait and see if other fixes arrive.

carmstrong:
> This may necessitate a v1.6.1 release if we want to fix pip install deis. I'm not sure how urgent that is, so I'd prefer to wait and see if other fixes arrive.

I personally don't think that warrants a patch release - we link to the self-extracting binaries by default, so the impact should be relatively minor, right? 

My fear is that users will upgrade their entire platforms from 1.6.0 to 1.6.1 for absolutely no changes other than the method by which the client is installed.

It may be dirty, but is there a way to force-update a build on PyPI?

eMxyzptlk:
so in all of our repositories we have this in `.travis.yml`:

```
before_deploy:
  # Workaround travis-ci/dpl#240
  - export PATH="$PATH:$HOME/.local/bin"
  - export DEIS_PLATFORM_VERSION=`curl -sD- deis.example.com | grep X_DEIS_PLATFORM_VERSION | cut -d\  -f2` && echo $DEIS_PLATFORM_VERSION
deploy:
  - provider: deis
    cli_version: $DEIS_PLATFORM_VERSION
```

So all of our apps are currently broken on Travis. Luckily, it's not as urgent so, please feel free to wait for other fixes to bundle in Deis v1.6.1

Otherwise, isn't it possible to release the client 1.6.0 but with the patched manifest?

carmstrong:
@eMxyzptlk Are you aware you can get the self-extracting client with `curl -sSL http://deis.io/deis-cli/install.sh | sh`? Is that a viable workaround for you?

mboersma:
>  is there a way to force-update a build on PyPI?

No, and that's by design. It's impossible to overwrite an existing artifact on pypi, you must release a new version.

bacongobbler:
@eMxyzptlk as a workaround, you can use `pip install git+git://github.com/deis/deis.git@7c8c60ba71a344c396cb66c54fb0cedd4792065d`, though the version number would change from v1.6.0 to v1.7.0-dev. I'd see if there's some way you can do that with travis-ci/dpl. If not then @carmstrong's solution should work.

eMxyzptlk:
Unfortunately, I do not control the code installing the client on travis, that'd be [here](https://github.com/travis-ci/dpl/blob/master/lib/dpl/provider/deis.rb#L6-L8). In any case, I'll just wait for the deis release since it's not urgent for me.

PR #3670, opened 2015-05-11
fix(contrib/coreos): start ntpd to synchronize clocks

### LINKED ISSUE #2900: sync system clock before starting etcd/fleet
In our cloud-config, we sync the system clock using NTP: https://github.com/deis/deis/blob/dbd42a7352759f9ac98cbd59bf20ac757672bc6d/contrib/coreos/user-data.example#L46-L58

The cloudinit service parses and starts these services in-order, meaning that currently we're changing the system time after etcd has been started. I'm wondering if this could potentially result in etcd getting out of sync between nodes. We should move this service to start before etcd and fleet.

### LINKED ISSUE #3386: PV clock drift causes more ceph havok
I am running nothing but CoreOS 607 instances now, in both of my environments.

While load testing our 3-node staging cluster, the cluster was a bit slow. Looking into the problem, I noticed that one of these PV based instances had drifted 10-15 seconds into the future. I turned up ntp and synced the times. Not more than 15 minutes later, two of our instances became non-responsive. After rebooting both instances from the AWS console, we stopped the load testing, and I spent a couple of hours piecing back together the fleet units that had fallen over - repairing ceph in the process.

### LINKED ISSUE #3634: Continued host clock skew issues
We tried to solve clock skew in #3300 by adding time synchronization to the cloud-config script, but this seems to have not done a whole lot to remedy the issue - users are still frequently reporting that their hosts are going out of sync (anecdotally, this seems to happen more often on EC2?).

We really need the expertise of our community to help us come up with a solution for continually syncing clocks via NTP.

carmstrong:
Starting ntpd before ntpdate enables the clocks to sync properly.
We don't appear to need the other logic for syncing clocks.

This commit also reorders the units in cloud-config to be more
thoughtful in start ordering -- we also sync the clocks before
starting etcd/fleet.

fixes #3634
fixes #3386
closes #2900


mboersma:
> Starting ntpd before ntpdate...

Does CoreOS' `ntp` service already have a line like `After=ntpdate.service`? It's not clear to me from the diff that this change will start ntpd before ntpdate, since they are listed in the opposite order (and the re-ordering makes the diff hard to parse).

carmstrong:
Changing our approach here based on an ongoing discussion in #3634. Will replace.

PR #3673, opened 2015-05-12
fix(controller): force uniqueness for default app name.

### LINKED ISSUE #2831: fix(controller): generate_app_name should re-roll on duplicate name
Lurking through old comments, I found https://github.com/deis/deis/issues/475#issuecomment-32907618. If we find an existing app name when generating a random app name, we should re-roll `generate_app_name()` until we find a new name for the application, otherwise we will hit a 500 error because the application name already exists.

Joshua-Anderson:
Fixes #2831

### DIFF
--- a/controller/api/models.py
+++ b/controller/api/models.py
@@ -129,6 +129,16 @@ class AuditedModel(models.Model):
         abstract = True
 
 
+def select_app_name():
+    """Select a unique randomly generated app name"""
+    name = utils.generate_app_name()
+
+    while App.objects.filter(id=name).exists():
+        name = utils.generate_app_name()
+
+    return name
+
+
 class UuidAuditedModel(AuditedModel):
     """Add a UUID primary key to an :class:`AuditedModel`."""
 
@@ -146,7 +156,7 @@ class App(UuidAuditedModel):
     """
 
     owner = models.ForeignKey(settings.AUTH_USER_MODEL)
-    id = models.SlugField(max_length=64, unique=True, default=utils.generate_app_name,
+    id = models.SlugField(max_length=64, unique=True, default=select_app_name,
                           validators=[validate_id_is_docker_compatible,
                                       validate_reserved_names])
     structure = JSONField(default={}, blank=True, validators=[validate_app_structure])


bacongobbler:
code LGTM. Another question would be how we're gonna handle this problem when we theoretically use up at least half of the available app names since we're more or less going to come up with name collisions fairly often.

Joshua-Anderson:
>Another question would be how we're gonna handle this problem when we theoretically use up at least half of the available app names 

Yeah, that could start to cause problems. Luckally I don't think too many people have a significant number of apps with random names. Worst case scenario we can add more weird adjectives and nouns ( or the generate name algorithm can be rewriten to limit itself to selecting names that already exist, but that's much less fun)! :smiling_imp:

Joshua-Anderson:
The CI seems to have timed out. Could somebody please restart it?

mboersma:
This doesn't solve the problem of "diminishing randomly-generated app namespace", but it avoids the occasional collision which I think is of higher concern.

I think it might be cleaner to have `select_app_name` implemented in `models.py` itself. The `utils.py` module was blissfully unaware of all the Django machinery until it imported the `api` module.

bacongobbler:
> I think it might be cleaner to have select_app_name implemented in models.py itself. The utils.py module was blissfully unaware of all the Django machinery until it imported the api module.

I agree. If we're using the models anyways, we should move it over there.

Joshua-Anderson:
I moved the function to `models.py` but the tests timed out again.

carmstrong:
> but the tests timed out again.

Not your PR - we're looking into an issue caused by leftover cruft.

carmstrong:
Code LGTM and Jenkins has passed.

mboersma:
Code LGTM.

bacongobbler:
> If we're using the models anyways, we should move it over there.

this is something we can clean up after we merge this :)

carmstrong:
> this is something we can clean up after we merge this :)

He already did it in this PR?

bacongobbler:
Oh! I meant by attaching it to the App model rather than as a standalone function. Not really a huge issue, just nitpicking.

PR #3678, opened 2015-05-12
fix(router): check if there is certificates to generate

aledbf:
Debugging an error I found this output from the router:
```
/bin/generate-certs: line 21: syntax error near unexpected token `done'
/bin/generate-certs: line 21: `done < /etc/ssl/deis_certs'
```


mboersma:
Code LGTM.

carmstrong:
Code LGTM. Good fix, @aledbf. This will necessitate a 1.6.1 release.

Issue #3635, opened 2015-05-06
Proposal: Make Deis compatible with large-scale, production cluster architecture


krancour:
Apologies... this is long and detailed, but (I think) comprehensive...

Because building etcd consensus can be time-consuming when the number of nodes in an etcd cluster is large, CoreOS recommends the following architecture for production and/or large scale deployments:

https://coreos.com/docs/cluster-management/setup/cluster-architectures/#production-cluster-with-central-services

In such an architecture, only a small, fixed number of nodes run etcd while all other nodes (an elastic pool of "workers") are merely configured with a list of etcd peers they can speak to.

Unfortunately, Deis doesn't play nicely with this cluster architecture fresh out of the box because Deis components generally expect that etcd is running locally.  Some components allow you to configure your way around this on some level, but generally do not go far enough in that regard.

Ultimately, I have overcome these challenges, but it required quite a significant investment in customizing images and unit files.  That's not tenable in the long-term for operators who wish to upgrade Deis frequently and don't want to have to re-apply their customizations with each release.

Here, I want to enumerate the few issues that comprise this (each on its own is pretty minor) and then propose a simple solution, which I would be happy to work on.

The following components are those that would run on worker nodes without etcd running locally:

* `deis-logspout`
* `deis-publisher`
* Arguably, `deis-router` (since it may need to scale significantly)
* Applications

Issues arising on such worker nodes:

1. `deis-logspout` and `deis-publisher`'s unit files require `etcd.service`.  This imposes a constraint that etcd be running locally.

2.  `get_image` script (and raw `etcdctl`) don't work from within unit files-- this can interfere with resolving custom image names for `deis-logspout`, `deis-publisher`, and `deis-router` (if applicable).  In the case of applications (with unit files created by `deis-controller`), image names cannot be resolved at all.

3. Most components (their containers) can be passed an `ETCD_HOST` environment variable that will allow them to speak to a remote etcd.  This is good, but does not go far enough.  It creates a single point of failure on that one host and also means unit files need to be modified to assign a specific IP address to the `ETCD_HOST` environment variable.  That sort of customization-per-cluster is something we should want to avoid.

4. `MachineMetadata` that articulates a Deis component's assignment to a "common services" node vs. a "worker" node isn't present in existing component unit files.  For an application, it's present only if explicitly added via `deis tags:set`.

Here are simple solutions to each of these, in turn:

[...]

Issue #3023, opened 2015-02-04
How can isolate Control Plane and Router Mesh 


nekrox:
I like only scale the Data Plane how can isolate the Control Plane and Router Mesh, i dont like the idea of repeat data on every single host of cluster

http://docs.deis.io/en/latest/understanding_deis/architecture/

In the last paragraph you mention about topologies but its not clear and/or missing documentation.

sforsman:
Currently you can do this manually by copying the default unit files from deisctl/units and then adding metadata restrictions under [X-Fleet] (e.g. MachineMetadata=plane=control). You can point the Deis installer to the new unit files by setting the DEISCTL_UNITS environment variable. Of course, you also need to add the appropriate metadata for the CoreOS hosts in your cloud-config (e.g. coreos/fleet/metadata: plane=control). Finally, you need to tell your applications you want to run them on the Data Plane. For this, you need to use deis tags:set (e.g. deis tags:set plane=data).

Please correct me if there is already an easier way.

bacongobbler:
nope, that sounds pretty accurate to me

sforsman:
Related to this topic, have you ever considered making it possible to define default tags for new applications (with something like `deisctl config controller set defaultApplicationTags=plane=data`)?

What currently happens - when the planes are isolated - is that fleet sometimes schedules new applications to the routing (or control) plane which is undesirable for several reasons. It would be great if the application's initial unit file could already have the X-Fleet -section up, so that applications are always deployed only to the data plane. While you can avoid this by adding conflicts for the routing/control plane unit files (e.g. *.web.*) , the default tags could also be useful for other default metadata restrictions (e.g. region, cabinet, disk type, environment).

carmstrong:
> have you ever considered making it possible to define default tags for new applications (with something like deisctl config controller set defaultApplicationTags=plane=data)?

We haven't, although off the top of my head, that implementation would make sense. You'd really only need to change https://github.com/deis/deis/blob/0dc6819f544c24ccca4e834f993504cc498991a1/controller/scheduler/fleet.py#L322-L331 and maybe the run template too. The `DEFAULT_APPLICATION_TAGS` could be a setting in https://github.com/deis/deis/blob/0dc6819f544c24ccca4e834f993504cc498991a1/controller/templates/confd_settings.py which is written to by that etcd key. Does that make sense? That's only my view, though - the other maintainers have a different approach.

bacongobbler:
[...]

PR #3679, opened 2015-05-12
feat(platform): add large cluster compatibility

krancour:
Addresses #3635

I will amend the PR with documentation if the approach is  judged satisfactory.


controller/scheduler/fleet.py
```diff
@@ -152,10 +152,11 @@ def _create_container(self, name, image, command, unit, **kwargs):
             f['value'] = f['value'].format(**l)
         # prepare tags only if one was provided
         tags = kwargs.get('tags', {})
-        if tags:
-            tagset = ' '.join(['"{}={}"'.format(k, v) for k, v in tags.items()])
-            unit.append({"section": "X-Fleet", "name": "MachineMetadata",
-                         "value": tagset})
+        tagset = ' '.join(['"{}={}"'.format(k, v) for k, v in tags.items()])
+        unit.append({"section": "X-Fleet", "name": "MachineMetadata",
+                     "value": tagset + ' "deisRole=general"'})
```
PierreKircher: worker is usualy sufficent you can add other tags via deis app if you need to be explicit

deisctl/units/deis-logger.service
```diff
@@ -15,3 +15,7 @@ RestartSec=5
 
 [Install]
 WantedBy=multi-user.target
+
+[X-Fleet]
+MachineMetadata="deisRole=general"
+MachineMetadata="deisRole=control"
```
PierreKircher: MachineMetadata="deisRole=control,general" ?

krancour: Does that have the same effect?  I had trouble finding where that is documented.

What I'm after here is "a unit can run on a 'general purpose node"_or_ a control node."

(Or alternatively, for certain other units, "a unit can run on a 'general purpose node"_or_ a worker node."

krancour:
After some discussion with @bacongobbler, instead of _only_ letting operators opt-in to a control / worker partitioning as in this PR, we can instead allow operators to opt into (or out of) more finely-grained metadata that states what components each machine is eligible to run.

@bacongobbler, what do you think that metadata looks like?  I am thinking something along the lines of `<component name>Eligible=true`.

Example:

```
  fleet:
    public-ip: $private_ipv4
    etcd_request_timeout: 3.0
    metadata: controllerEligible=true builderEligible=true appEligible=true ...
```

Also, do you think it's worthwhile to also maintain the notion of a "general purpose" role?  This way the included `user-data.example` could simply look like the following and not have to change every time a new component is introduced:

```
  fleet:
    public-ip: $private_ipv4
    etcd_request_timeout: 3.0
    metadata: deisRole=general
```

Units, could thus be as follows (using controller for example):

```
[X-Fleet]
MachineMetadata="controllerEligible=true"
MachineMetadata="deisRole=general"
```

Eager to hear thoughts...

deisctl/units/deis-logger.service
krancour: I tested the syntax you suggested and it does not have the same effect.  That is to say, it is not interpreted as a logical or.  In the example you gave, "control,general" is evaluated as the machine's `deisRole`.

krancour:
Actually, as it turns out, the following amounts to a logical _and_; not an _or_:

```
[X-Fleet]
MachineMetadata="controllerEligible=true"
MachineMetadata="deisRole=general"
```

The only way to achieve a logical or is by naming the _same_ key twice with different values on different lines as in the following example, directly from CoreOS documentation:

```
[X-Fleet]
MachineMetadata=region=us-east-1
MachineMetadata=region=us-west-1
```

Unfortunately, these limitations make it difficult, if not impossible to provide `MachineMetadata` that (for example) logically amounts to "run where controllerEligible=true or deisRole=general".

Does this mean we'd have to sacrifice the notion of a "general purpose" role?  Doing so would mean `user-data.example` would need lengthy list of metadata that explicitly makes each node eligible to run each component and that would have to be updated in the future if new components are added to Deis.

Anyone have ideas on a better way to do all this that I may be overlooking?

Ping @bacongobbler

bacongobbler:
> Does this mean we'd have to sacrifice the notion of a "general purpose" role?

In my thinking there wasn't a general purpose role, so the logical OR is what we want. But then again, this should still work:

```
[X-Fleet]
MachineMetadata=plane=control
MachineMetadata=plane=data
```

or:

```
[X-Fleet]
MachineMetadata=role=controller
```

I think we'd have to decide on one or the other. The former would be much easier for upgrades, but harder to get the more fine-grained control of where we'd want each component to run. Then again, we've always modeled them as control plane/data plane so it might not make sense otherwise.

iancoffey:
For what its worth, I had previously achieved a similar solution with just two metadata tags: controlplane=true/false and dataplane=true/false along with custom unit files. Nothing else was needed and separation worked pretty seamlessly and it allows for a node to be either or both (general). I was planning to work on adding a small bit of logic to deisctl to be able to achieve this without custom units, so Im pretty interested where this ends up.

krancour:
@bacongobbler thanks for weighing in.  There are potentially a couple problems with the two examples you gave.  In the first, if metadata for each machine takes the form `plane=x`, there would be no way to configure a machine such that it's eligible to host both planes.  Very similarly, in the second example you gave, if metadata where to take the form `role=x`, where x is a component, there would be no way to configure a machine such that it's eligible to host more than one component.

I am liking what @iancoffey has just recommended.

krancour:
In addition to what @iancoffey recommended, we should probably also have a routerMesh=true/false, yes?

bacongobbler:
I'd have to read through the MachineMetadata more closely, but I believe that's a safe assumption.

iancoffey:
In my implementation, router units were control-plane only - just for simplicities sake. The data-plane existed only to host app containers. Its interesting to think about routing being handled separately but its implementation is a bit murky to me. Maybe that could be punted.

krancour:
@iancoffey, I think valid arguments (dependent on use case and volumes) can probably be made for all three of:

1. router mesh runs where control plane runs
2. router mesh runs where data plane runs
3. router mesh is isolated from both control plane and data plane

Given that valid arguments could be made for each of those, I think it's good that it is its own thing, and I think the metadata should treat it as such.

krancour:
Amended as discussed.

bacongobbler:
@krancour would you mind supplying documentation on how someone would customize the hosts so we can close out #3023?

krancour:
@bacongobbler document it as part of / assuming this PR is merged?  Or a more general, "this is how you'd do this," that delves into customizing units files, etc.?

(This PR absolutely still needs docs added to it anyway.)

krancour:
One other thing I might consider adding to this... you'll see a number of the commits in this PR have to do with either enabling components to work with a remote etcd or improving components to work with _multiple_ remote etcds in order to avoid single points of failure.  My initial assumption had been that control plane components run on the machines that run etcd and data plane components (and potentially router mesh) run on machines that are _not_ running etcd, so those enhancements were initially applied _only_ to data plane and router mesh components.

With where we seem to be landing on this, control over what components run where is just slightly more fine-grained than I initially had in mind.  (And I think that's for the better.)  Maybe those same improvements should also be applied to control-plane components as well?

Then this cluster architecture could be supportable as well:

https://coreos.com/docs/cluster-management/setup/cluster-architectures/#easy-development/testing-cluster

laurrentt:
@krancour I also think the routers should be on isolated (on larger cluster) as they will probably run on a public subnet. I'm gonna play a bit with your PR, do you have a TODO list for it ? When (approx.) do you think it will be mature enough to merged ? I'm really eager to see your changes come to life, so if there's anything I can do to help, let me know.

krancour:
@laurrentt router isolation is already possible in the current state of this PR.

Right now, I am working to make sure control plane components can work with remote etcd in the event that one wishes to use the "easy development" CoreOS cluster architecture _or_ wishes for any other reason to run the control plane components on nodes _not_ running etcd (for whatever reasons they may have).

Previously, I had only extended such capability to the data plane and router mesh components, but it seems to me that this possibility should be open across the board.

laurrentt:
@krancour I have a concrete use case for having the control plane on nodes not running etcd. On AWS, we want the control plane to be on our private VPC that is only accessible by via our VPN. Although I guess I could run etcd on that private VPC without issue... Anyway +1 for making sure the control plane can work with a remote etcd.

controller/scheduler/fleet.py
```diff
@@ -377,6 +376,7 @@ def attach(self, name):
 
 CONTAINER_TEMPLATE = [
     {"section": "Unit", "name": "Description", "value": "{name}"},
+    {"section": "Service", "name": "EnvironmentFile", "value": "/etc/environment"},
```
carmstrong: What are we thinking `/etc/environment` will have that the app units will need? It may be easier for malicious apps to glean environment information if they're sourcing things like `COREOS_PRIVATE_IPV4`.

krancour: App units rely on etcd to resolve registry host and port.  If etcd isn't running locally, ideally `ETCDCTL_PEERS` is set to enable `etcdctl` to use _remote_ etcds for that purpose.

The idea was that anyone wanting to run their data plane in an elastic pool of machines _not_ running etcd (which seems a semi-advanced use case) could use cloud-config to poke a value for `ETCDCTL_PEERS` into `/etc/environment`.

Since there is possibly a security concern here, what if I kept this approach, but used cloud-config to poke `ETCDCTL_PEERS` into a _different_ file instead, like `/etc/app-environment`?  It would lack anything potentially sensitive or ripe for abuse like `COREOS_PRIVATE_IPV4`.

Forthcoming documentation would illustrate how `ETCDCTL_PEERS` can be poked into that file using cloud-config.

krancour:
Most recent updates should allow most control plane components to run with remote etcd.  Only store and swarm components still require this treatment.  Those updates will follow shortly.

controller/scheduler/fleet.py
krancour: @carmstrong I put some more thought into this over the weekend.  Now I'm not certain I see the potential security issue you were raising.  If the generated application units pull in environment context from `/etc/app-environment`, isn't that only available to _systemd_ as it starts/manages those units?  i.e. I don't believe any of that is available to an app running in container unless it is explicitly passed in via `-e` flags in the `docker run` command.

krancour:
Is it possible for someone to restart the tests?  The failure doesn't look related to the changes.

krancour:
I'm going to break this into two separate PRs.

The thing that is really required to be compatible with larger / production-worthy clusters is that components and applications be able to run using remote etcds.

The ability to control which nodes are eligible to host each of the control plane, data plane, and router mesh is _complimentary_ in the sense that in a large cluster, one probably does want to run the control plane on a small, fixed number of nodes (just as they would etcd), but that's not 100% necessary.

iancoffey:
@krancour :+1: that makes a lot of sense. ive been holding off PRing control/data plane isolation to see where this goes - if you want collab or even dump that part off on me, by all means. :)

krancour:
@iancoffey #3718 has the placement option stuff now.  Feel free to have a look at that, give feedback, advice, etc.

krancour:
If anyone wants to review the doc changes I added in the last commit, that would be helpful.  I'm hoping it's enough.  It really amounts to "this is an advanced setup, but here's how to enable it."

krancour:
@carmstrong we talked about this yesterday?  I think the plan is to merge this, yes?  If so, I'll rebase.

carmstrong:
> I think the plan is to merge this, yes? If so, I'll rebase.

Yep, I thought I had replied to this already.... Please rebase.

krancour:
@carmstrong the test failure doesn't look related to the changes in this PR.  This tests clean locally.  Could you maybe restart the tests for me?

carmstrong:
> Could you maybe restart the tests for me?

I think @mboersma did: https://ci.deis.io/job/test-ec2/1251/

controller/bin/boot
```diff
@@ -10,14 +10,21 @@ set -eo pipefail
...
-export ETCD_PORT=${ETCD_PORT:-4001}
-export ETCD="$HOST:$ETCD_PORT"
+# if ETCDCTL_PEERS isn't already set, set it
+if [ -z "$ETCDCTL_PEERS" ]; then
+	export ETCDCTL_PEERS="http://${ETCD_HOST:-${HOST:-127.0.0.1}}:${ETCD_PORT:-4001}"
+fi
+
+# ETCD is used by settings.py
+IFS=',' read -ra ETCD_PEERS_ARR <<< "$ETCDCTL_PEERS"
+export ETCD=$(echo ${ETCD_PEERS_ARR[0]} | awk -F/ '{print $3}')
```
bacongobbler: We don't use this variable in the script any more, IIRC. Let's remove this confusing array parse logic if we're not using it.

controller/migrations/data/0001.sh
```diff
@@ -1,19 +1,15 @@
 #!/usr/bin/env bash
 
-ETCD_PORT=${ETCD_PORT:-4001}
-ETCD="$HOST:$ETCD_PORT"
```
bacongobbler: This will probably need to still exist if $ETCDCTL_PEERS does not exist

controller/scheduler/fleet.py
bacongobbler: I think it would be ideal if you demonstrated how ETCDCTL_PEERS would be set up via cloud config in this PR. AFAICT users would have to manage this outside of cloud-config since it's a cluster-wide setup.

controller/migrations/data/0001.sh
bacongobbler: Looking through the rest of this PR, this will definitely need to stay the same for older clusters without ETCDCTL_PEERS set up.

controller/bin/boot
krancour: @bacongobbler controller uses the ETCD environment variable here:

https://github.com/deis/deis/blob/master/controller/deis/settings.py#L279

controller/migrations/data/0001.sh
krancour: @bacongobbler If I am not mistaken, at the time this migration script runs, the controller's boot script has already created the `ETCDCTL_PEERS` environment variable if it didn't already exist.

As proof, all integration tests are passing locally and in CI on an "old style" cluster, straight-out-of-the-box and not modified in any way.  In other words, I have not set `ETCDCTL_PEERS` explicitly on any host in the test cluster, so an empty value for `ETCDCTL_PEERS` gets passed by each system unit into its corresponding container and the modified logic in each boot script is constructing `ETCDCTL_PEERS` successfully since it hasn't been set.

By contrast, I _do_ set `ETCDCTL_PEERS` on each host that does not run etcd locally in my live clusters.  And things work great there too.

I'm pretty confident this achieves what I set out to do without breaking anything existing... but given the magnitude of the change, continued scrutiny is welcome and appreciated.

controller/scheduler/fleet.py
krancour: @bacongobbler It's covered in the documentation in the PR-- which was previously missing due to a botched rebase.  I've fixed that and the documentation is back now.

docs/managing_deis/production_deployments.rst
```diff
@@ -56,3 +56,84 @@ Enable TLS
...
+
+For example:
+
+.. code-block:: yaml
+
+  coreos:
+    fleet:
+      # allow etcd to slow down at times
+      etcd_request_timeout: 3.0
+      etcd_servers: "http://10.0.0.101:4001,http://10.0.0.102:4001,http://10.0.0.103:4001,http://10.0.0.104:4001,http://10.0.0.105:4001"
```
aledbf: @krancour this is not working (yet) in case of any error in the first server in the list. 
This will work after this https://github.com/coreos/fleet/issues/1157

bacongobbler: So I guess at this point we're waiting for fleet 0.10.3 or 0.11.0 to land?

mboersma: Yes, I checked that too and the commit wasn't in fleet 0.10.2. :(

krancour: Strictly speaking nothing in this PR depends on that.  The problem (and solution) that @aledbf pointed out affect anyone running a large fleet cluster-- not just folks running Deis.  This PR makes it _possible_ to run Deis on such a cluster (with only smaller, fixed number of nodes running etcd), but it doesn't force anyone down that path-- it's not a default.  Could we node in the documentation that this is possible to do now, but that YMMV until newer fleet drops?

bacongobbler: Sure, I'm fine with that.

router/boot.go
```diff
@@ -75,9 +84,10 @@ func main() {
 	// FIXME: have to launch cron first so generate-certs will generate the files nginx requires
 	go launchCron()
 
-	waitForInitialConfd(host+":"+etcdPort, timeout)
+	confdHost := strings.Split(etcdPeers[0], "://")[1]
```
laurrentt: This doesn't work for some reason with my load balanced etcd quorum. I guess this is because I use a DNS name instead of an IP. I had to get rid of the `Split` and only keep the `etcdPeers[0]` in order to get it to work.

krancour: @laurrentt what exactly are you setting `ETCDCTL_PEERS` to?  That env variable (normally used by etcd itself and re-used here is (or should be) a comma-delimited list of etcd hosts, but should also include the protocol (`http`) and port (4001).  I think it shouldn't matter whether you use IP addresses or DNS name/hostname, but unless `ETCD_PEERS` conforms to the format I described, the line you're referencing won't properly pluck an IP/name and port out of that env variable.

laurrentt: Here's my `ETCDCTL_PEERS`:
`ETCDCTL_PEERS=http://internal-MT26-05-C-EtcdQuor-16VMRR4RJ8LI8-143835863.us-east-1.elb.amazonaws.com:4001`
I can't explain you why exactly it doesn't work. I've set confd to be more verbose and passed the ETCDCTL_PEERS without the http:// to the `-node` argument, the logs said it was *unable to connect to internal-MT26-05-C-EtcdQuor-16VM*... Then I tried to prefix the argument with `http://` and it worked. As I said, I ended up getting rid of the `.Split`, reinstalled the platform, no more problems.

builder/rootfs/bin/boot
```diff
@@ -14,14 +14,16 @@ fi
 [[ $DEBUG ]] && set -x
 
 # configure etcd
-export ETCD_PORT=${ETCD_PORT:-4001}
-export ETCD="$HOST:$ETCD_PORT"
+# if ETCDCTL_PEERS isn't already set, set it
+if [ -z "$ETCDCTL_PEERS" ]; then
+	export ETCDCTL_PEERS="http://${ETCD_HOST:-${HOST:-127.0.0.1}}:${ETCD_PORT:-4001}"
```
bacongobbler: I know at some point we may want to be able to support etcd clusters that are also backed by SSL. Can we make the protocol (or the URL as a whole) configurable?

carmstrong: We'll probably need to do a significant refactor when that happens, so I'm fine shipping this PR without that IMO.

krancour: For backwards compatibility, I was trying to always fall back to how things have been done thus far if the `ETCDCTL_PEERS` env variable isn't set, and that meant relying on all the other env variables you see here.  It's not ideal and even a bit confusing to understand.  I do think that on our next major release we should have a big simplification / refactor of this where the URL as a whole can be configured as @bacongobbler suggests.

carmstrong:
Looks like this needs another rebase :(

krancour:
I'm looking into whether this PR may actually be unnecessary now that we have upgraded to etcd2...

From etcd docs:

> etcd can now run as a transparent proxy. Running etcd as a proxy allows for easily discovery of etcd within your infrastructure, since it can run on each machine as a local service. In this mode, etcd acts as a reverse proxy and forwards client requests to an active etcd cluster. The etcd proxy does not participant in the consensus replication of the etcd cluster, thus it neither increases the resilience nor decreases the write performance of the etcd cluster.

And:

> If you bootstrap a etcd cluster using discovery service with more than the expected number of etcd members, the extra etcd processes will fall back to being readwrite proxies by default. They will forward the requests to the cluster as described above. For example, if you create a discovery url with size=5, and start ten etcd processes using that same discovery url, the result will be a cluster with five etcd members and five proxies. Note that this behaviour can be disabled with the proxy-fallback flag.

krancour:
Some experimentation seems to show that when a leader or follower dies, no proxy is every promoted to fill the gap in the etcd cluster.  So I think that means the recommendation to run etcd (leader and followers) on a small, fixed number of nodes remains.  However, a proxy can be run everywhere else.  This allows services and containers on _every_ node the convenience of connecting to localhost.

Bottom line is this development doesn't invalidate the need for this PR, but it dramatically simplifies it by negating the need for the majority of code changes.

I might close this and open a new PR with an lighter, alternate implementation.

mboersma:
> this development doesn't invalidate the need for this PR, but it dramatically simplifies it

Simpler is certainly better. Thanks for updating this with your investigations.

jchauncey:
So with consul you have a consul cluster with

krancour:
Replaced by #4142

jedsmith:
I found this PR through a really weird Google search for something totally unrelated, but just FYI:

> Some experimentation seems to show that when a leader or follower dies, no proxy is ever promoted to fill the gap in the etcd cluster.

Node identity is very strong in etcd. It isn't a dynamic cluster like you're thinking. For that to happen, the proxy would have to be added to the cluster, and since etcd manages the configuration, that operation requires consensus so it isn't viable for getting out of a no-quorum situation. This has bit me a couple times as a *user* of etcd, so thought I'd share.

rimusz:
@jedsmith I'm running in the same way my development and production clusters too.
etcd cluster must run on dedicated nodes with nothing else there.
For production clusters I usually use 3 or 5 nodes clusters, for development ones one etcd node is enough. And worker nodes use etcd proxy. 

In some cases if I need to run a very tiny production cluster then one etcd node is set.

PR #3684, opened 2015-05-13
ref(builder): organize for docker best practices

bacongobbler:
This is mostly organizational shifts on where things in `builder/` should go. To explain some of the changes: 

When I originally PR'd the shift to building deps locally then injecting them into `image/`, `image/` was meant to reflect the final root filesystem of the container, so I renamed it to `rootfs/` to better reflect that.

Internally, things have shifted around to better follow the [Filesystem Hierarchy Standard](http://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard). A lot of things we were doing in `bin/boot` or in `Dockerfile` was just copying files to their proper location (`/sbin/apparmor_parser`, `/home/git/shim.dockerfile`), so I just moved them to the proper place in `rootfs`.

Some old cruft has been cleaned up (i.e. unused `.dockerignore`). I can move that to a separate PR.

Go source files in `builder/bin/` was moved to `builder/src/`. They are source files, not binaries! (this can be moved to another PR on request)

Because of moving things around, `/app` in the container no longer exists. Everything is in its proper place now :)


mboersma:
This needs rebasing now.

mboersma:
This looks simpler to me. Code LGTM.

carmstrong:
Code LGTM.

PR #3686, opened 2015-05-13
fix(database): run backups in the background

### LINKED ISSUE #2741: bug(database): the backup should not interfere the etcd publication

### LINKED ISSUE #3643: Controler couldn't connect to database after node crash
- deis 1.5.0
- 3 node setup on digital ocean toasters (4GB instances)

1. One node (with controller and database) went off.
2. Controller and database were moved to other node.
3. Controller was unable to connect to database, it was trying to use old ip ([log](https://gist.github.com/szymonpk/a7b365aa6649193b97a4)).
4. Controller restart helped.

I've waited about 1-2 minutes before controller restart, I've thought maybe fleet/etcd had some issues. None of services were down atm and previously down node was up and fully running.

bacongobbler:
Instead of running the backup in the publish loop, we can schedule the
backups on a timed interval. I tried to do this with cron, but I was
having trouble sending the output of the forked processes to
/dev/stdout, so instead I just forked a for loop in the background.
We can easily change this to remove the `for` loop and point cron at
/app/bin/backup if someone can figure that out :)

closes #2741
closes #3643

### DIFF
--- a/database/bin/backup
+++ b/database/bin/backup
@@ -1,0 +1,18 @@
+#!/bin/bash
+
+export BACKUP_FREQUENCY=${BACKUP_FREQUENCY:-3h}
+export BACKUPS_TO_RETAIN=${BACKUPS_TO_RETAIN:-5}
+
+while true; do
+    sleep "$BACKUP_FREQUENCY"
+    echo "database: performing a backup..."
+    if [[ -f /var/lib/postgresql/9.3/main/recovery.conf ]] ; then
+        echo "database: database is currently recovering from a backup. Will try again next loop..."
+    else
+        # perform a backup
+        envdir /etc/wal-e.d/env wal-e backup-push /var/lib/postgresql/9.3/main
+        # only retain the latest BACKUPS_TO_RETAIN backups
+        envdir /etc/wal-e.d/env wal-e delete --confirm retain $BACKUPS_TO_RETAIN
+        echo "database: backup has been completed."
+    fi
+done
--- a/database/bin/boot
+++ b/database/bin/boot
@@ -16,10 +16,6 @@ export ETCD_PATH=${ETCD_PATH:-/deis/database}
 export ETCD_TTL=${ETCD_TTL:-20}
 
 export BUCKET_NAME=${BUCKET_NAME:-db_wal}
-export BACKUPS_TO_RETAIN=${BACKUPS_TO_RETAIN:-5}
-
-# how many TTL/2 sleeps between backups -- 2160 is 3 hours
-export BACKUP_FREQUENCY=${BACKUP_FREQUENCY:-2160}
 
 # wait for etcd to be available
 until etcdctl --no-sync -C $ETCD ls >/dev/null 2>&1; do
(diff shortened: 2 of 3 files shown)


aledbf:
@bacongobbler no, the issue is here https://github.com/bacongobbler/deis/blob/unblock-database-publish/database/bin/boot#L149 (in the periodic backup)

bacongobbler:
Okay, I addressed your comment. Thanks @aledbf :)

carmstrong:
Code LGTM. @bacongobbler have you tested this by leaving a cluster up for longer than 3 hours to confirm that the backup is created? If not, we'll need to do that.

bacongobbler:
@carmstrong the [recovery test](https://github.com/deis/deis/blob/master/database/tests/recovery_test.go) covers that :)

mboersma:
Code LGTM.

carmstrong:
> I modified the BACKUP_FREQUENCY to 1s so the backup occurs instantaneously. Functional tests pass np

Well, we already kick off a backup at startup here: https://github.com/deis/deis/blob/4de2c904f35c2a4862d6ed0791581ea7bb3b8912/database/bin/boot#L122-L126

So I'm not sure how that backup and and the timed one work when they're firing at the same time, as seems to happen in the tests. 

Let's manually confirm this with a `BACKUP_FREQUENCY` of `10m` or something to confirm that the backups perform as they would in the real world. I'd ideally like to trigger a failover and recovery too, to get them to replay. 

I agree that things *should* work just fine, but getting this right is super important. :+1:

bacongobbler:
Failover and everything else seems fine after waiting for a backup to occur:

```
May 18 21:46:54 deis-03 sh[5064]: 2015-05-18 21:46:54 UTC LOG:  archive recovery complete
May 18 21:46:54 deis-03 sh[5064]: 2015-05-18 21:46:54 UTC LOG:  database system is ready to accept connections
May 18 21:46:54 deis-03 sh[5064]: 2015-05-18 21:46:54 UTC LOG:  autovacuum launcher started
```

Does this seem good to :shipit:?

carmstrong:
:ship:

PR #3688, opened 2015-05-14
feat(contrib/ec2): add support for custom volume sizes

### LINKED ISSUE #3682: contrib/ec2: Make volume size customizable
This should be a parameter users can override. Currently, we default to 100GB for the Docker volume and 10GB for the etcd volume.

carmstrong:
closes #3682

### DIFF
--- a/docs/installing_deis/aws.rst
+++ b/docs/installing_deis/aws.rst
@@ -79,7 +79,7 @@ Customize cloudformation.json
 -----------------------------
 
 Any of the parameter defaults defined in deis.template.json can be overridden by setting the value
-in `cloudformation.json`_. For example, to configure all of the optional settings:
+in `cloudformation.json`_. For example, to configure all of the options to non-default values:
 
 .. code-block:: console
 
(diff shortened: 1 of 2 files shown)


carmstrong:
I manually tested both the default path and customizing all of these.

kalbasit:
:+1:

docs/installing_deis/aws.rst
```diff
@@ -102,11 +102,24 @@ in `cloudformation.json`_. For example, to configure all of the optional setting
     {
         "ParameterKey":     "ELBScheme",
         "ParameterValue":   "internal"
+    },
+    {
+        "ParameterKey":     "RootVolumeSize",
+        "ParameterValue":   "100"
```
bacongobbler: When someone comes to the docs, they may be confused on what unit of measurement this is. Maybe we should add a note stating that this is in gigabytes?

bacongobbler:
code LGTM

docs/installing_deis/aws.rst
carmstrong: They may also be confused as to what "internal" means - these options all have detailed descriptions in deis.template.json, so I'd rather not repeat those here. I tried to call them to the other file with "The default values are defined in deis.template.json."

bacongobbler: Ah, I skipped over that. :+1:

mboersma:
Code LGTM.

PR #3692, opened 2015-05-14
fix(controller): Roll back scale on failure

### LINKED ISSUE #3605: Failure to create a container leaves others created
If `deis scale cmd=5` creates 4 containers but fails on the 5th, it raises an exception and leaves 4 containers in the "created" state, not running. The code in question is here: https://github.com/deis/deis/blob/master/controller/api/models.py#L272-L292

It would be more intuitive and useful behavior if instead Deis made a "best effort" to start whichever containers it was able to create. In this particular case we used `deis limits:set -c cmd=1024` to ensure some containers wouldn't be created due to resource limitations, then scaled up, and were left with this:
```console
$ deis info
=== zigzag-drumroll Application
{
  "updated": "2015-04-30T22:05:12UTC", 
  "uuid": "0c0dd591-5d9c-45c4-914d-52d703886cb3", 
  "created": "2015-04-30T21:53:21UTC", 
  "url": "zigzag-drumroll.local5.deisapp.com", 
  "owner": "test", 
  "id": "zigzag-drumroll", 
  "structure": {
    "cmd": 1
  }
}

=== zigzag-drumroll Processes

--- cmd: 
cmd.1 created (v3)
cmd.2 created (v3)
cmd.3 created (v3)
cmd.4 created (v3)
cmd.5 destroyed (v3)
```

smothiki:
fix(controller): roll back if scale fails
fixes #3605 

Log_event method was implemented in scheduler and app.id is passed to scheduler record logs to app logs. Rollback happens if any of the container state is not in created.  
```
$ deis scale cmd=7
503 Service Unavailable
Detail:
aborting, failed to create some containers rolling back
reshut2-82-75-dhcp:example-dockerfile-http smothiki$ deis logs
2015-05-14T23:13:45UTC deis[api]: smothiki created initial release
2015-05-14T23:14:34UTC deis[api]: smothiki deployed 6578692
2015-05-14T23:14:35UTC deis[api]: smothiki scaled containers cmd=1
2015-05-14T23:15:54UTC deis[api]: smothiki changed limits for memory
2015-05-14T23:45:46UTC deis[api]: smothiki scaled containers cmd=5
2015-05-14T23:46:38UTC deis[api]: smothiki scaled containers cmd=6
2015-05-14T23:46:47UTC deis[api]: smothiki scaled containers cmd=7
2015-05-14T23:46:47UTC deis[api]: 500 Server Error: Internal Server Error  no resources available to schedule container
2015-05-14T23:46:47UTC deis[api]: aborting, failed to create some containers rolling back
2015-05-14T23:46:47UTC deis[api]: smothiki scaled containers cmd=7
2015-05-14T23:46:47UTC deis[api]: 500 Server Error: Internal Server Error  no resources available to schedule container
2015-05-14T23:46:47UTC deis[api]: aborting, failed to create some containers rolling back
2015-05-14T23:46:47UTC deis[api]: smothiki scaled containers cmd=7
2015-05-14T23:46:47UTC deis[api]: 500 Server Error: Internal Server Error  no resources available to schedule container
2015-05-14T23:46:47UTC deis[api]: aborting, failed to create some containers rolling back
reshut2-82-75-dhcp:example-dockerfile-http smothiki$ deis info
=== luxury-neckwear Application
{
  "updated": "2015-05-14T23:46:38UTC", 
  "uuid": "a77c23e2-da7c-4077-b201-ef4bf7032097", 
  "created": "2015-05-14T23:13:45UTC", 
  "url": "luxury-neckwear.smothiki.io", 
  "owner": "smothiki", 
  "id": "luxury-neckwear", 
  "structure": {
    "cmd": 6
  }
}

=== luxury-neckwear Processes

--- cmd: 
cmd.1 up (v3)
cmd.2 up (v3)
cmd.3 up (v3)
cmd.4 up (v3)
cmd.5 up (v3)
cmd.6 up (v3)

=== luxury-neckwear Domains
No domains
```

### DIFF
--- a/controller/api/models.py
+++ b/controller/api/models.py
@@ -292,6 +292,7 @@ class App(UuidAuditedModel):
         if any(c.state != 'created' for c in to_add):
             err = 'aborting, failed to create some containers'
             log_event(self, err, logging.ERROR)
+            self._destroy_containers(to_add)
             raise RuntimeError(err)
         [t.start() for t in start_threads]
         [t.join() for t in start_threads]
--- a/controller/api/tests/test_scheduler.py
+++ b/controller/api/tests/test_scheduler.py
@@ -75,10 +75,7 @@ class SchedulerTest(TransactionTestCase):
         url = "/v1/apps/{app_id}/containers".format(**locals())
         response = self.client.get(url, HTTP_AUTHORIZATION='token {}'.format(self.token))
         self.assertEqual(response.status_code, 200)
-        self.assertEqual(len(response.data['results']), 20)
-        # make sure some failed
-        states = set([c['state'] for c in response.data['results']])
-        self.assertEqual(states, set(['error', 'created']))
+        self.assertEqual(len(response.data['results']), 0)
 
     def test_start_chaos(self):
         url = '/v1/apps'
@@ -304,7 +301,6 @@ class SchedulerTest(TransactionTestCase):
         response = self.client.get(url, HTTP_AUTHORIZATION='token {}'.format(self.token))
         self.assertEqual(response.status_code, 200)
         self.assertEqual(len(response.data['results']), 20)
-
         # make sure all old containers are still up
         states = set([c['state'] for c in response.data['results']])
         self.assertEqual(states, set(['up']))
--- a/controller/scheduler/swarm.py
+++ b/controller/scheduler/swarm.py
@@ -33,7 +33,8 @@ class SwarmClient(object):
                 mem = mem[:-1]
         cpu = kwargs.get('cpu', {}).get(l['c_type'])
         self.docker_cli.create_container(image=cimage, name=name,
-                                         command=command.encode('utf-8'), mem_limit=mem,
+                                         command=command.encode('utf-8'),
+                                         mem_limit=mem,
                                          cpu_shares=cpu,
                                          environment=[affinity],
                                          host_config={'PublishAllPorts': True})


controller/api/models.py
```diff
@@ -287,8 +288,9 @@ def _start_containers(self, to_add):
         [t.start() for t in create_threads]
         [t.join() for t in create_threads]
         if any(c.state != 'created' for c in to_add):
-            err = 'aborting, failed to create some containers'
+            err = 'aborting, failed to create some containers rolling back'
```
bacongobbler: isn't "aborting" supposed to imply that the operation is being rolled back?

controller/api/models.py
```diff
@@ -287,8 +288,9 @@ def _start_containers(self, to_add):
...
             log_event(self, err, logging.ERROR)
+            self._destroy_containers(to_add)
```
bacongobbler: we already make this call [here](https://github.com/deis/deis/blob/master/controller/api/views.py#L193-L195) which will destroy the containers, though I agree that it really should be here. Would you mind updating the logic I just pointed out so we don't call `_destroy_containers()` twice?

controller/api/tests/test_scheduler.py
```diff
@@ -305,10 +304,6 @@ def test_build_chaos(self):
         self.assertEqual(response.status_code, 200)
         self.assertEqual(len(response.data['results']), 20)
 
-        # make sure all old containers are still up
```
bacongobbler: this should still pass if the scheduler failed to create containers

controller/scheduler/swarm.py
```diff
@@ -1,11 +1,21 @@
...
+import os
 from django.conf import settings
 from docker import Client
 
 from .states import JobState
 
+logger = logging.getLogger('api')
+
+
+def log_event(appid, msg, level=logging.INFO):
```
bacongobbler: let's change any calls here to `app.log()` instead so we're not maintaining duplicate code

controller/scheduler/swarm.py
```diff
@@ -48,14 +63,20 @@ def stop(self, name):
         """
         Stop a container
         """
-        self.docker_cli.stop(name)
+        try:
+            self.docker_cli.stop(name)
+        except:
+            pass
```
bacongobbler: don't we want to raise RuntimeError if we fail to stop? I don't think we should be suppressing these errors

controller/scheduler/swarm.py
```diff
@@ -48,14 +63,20 @@ def stop(self, name):
...
         """
         Destroy a container
         """
-        self.stop(name)
-        self.docker_cli.remove_container(name)
+        try:
+            self.stop(name)
+            self.docker_cli.remove_container(name)
+        except:
+            pass
```
bacongobbler: same here

controller/scheduler/swarm.py
```diff
@@ -32,11 +43,15 @@ def create(self, name, image, command='', template=None, **kwargs):
...
-                                         environment=[affinity],
-                                         host_config={'PublishAllPorts': True})
+        try:
+            self.docker_cli.create_container(image=cimage, name=name,
+                                             command=command.encode('utf-8'),
+                                             mem_limit=mem,
+                                             cpu_shares=cpu,
+                                             environment=[affinity],
+                                             host_config={'PublishAllPorts': True})
+        except Exception as err:
```
bacongobbler: Is there a specific error we're trying to catch here?

controller/scheduler/swarm.py
```diff
@@ -32,11 +43,15 @@ def create(self, name, image, command='', template=None, **kwargs):
...
+            log_event(self.appid, str(err.message)+"  "+str(err.explanation), logging.ERROR)
```
bacongobbler: we should still re-raise or raise a more scheduler-specific error if the container failed to create

controller/scheduler/swarm.py
smothiki: It doesn't matter if we dont raise runtime error as we any ways check the state of the container and raise runtime error. But I'll change the code any ways as this code might be avoiding some corner cases

controller/scheduler/swarm.py
smothiki: I think these errors come under errors.DockerException But just reading from exception object is a nicer way of logging the error

controller/api/models.py
smothiki: I'll make the necessary changes to call delete containers only once

controller/api/models.py
mboersma: I think all of these messages should be either left alone or rephased as:
> rolling back after failure to create containers

controller/scheduler/swarm.py
mboersma: It's almost always a bad idea to have a catch-all `except` statement; we should always catch a specific exception type or types.

controller/api/models.py
smothiki: @bacongobbler  release.destroy() has no link t0 destroy containers it might be doing some  cleaning or deleting the database records ?? I haven't looked much. But interestingly I have no idea why there are three different threads trying to scale when something fails  . Do we have any retry settings if scale fails ??
https://gist.github.com/smothiki/cd30977234cf7d1d3c7d

controller/scheduler/swarm.py
smothiki: DockerException but its not specific to create container call

mboersma: We should catch `DockerException` rather than `Exception`--it's almost always bad Python form to catch every exception. (I will make this same comment every time I see a catch-all `except`.)

controller/scheduler/swarm.py
```diff
@@ -48,14 +52,20 @@ def stop(self, name):
         """
         Stop a container
         """
-        self.docker_cli.stop(name)
+        try:
+            self.docker_cli.stop(name)
+        except:
```
mboersma: This should catch a specific error class or classes. Also if we raise a new Exception here, we have lost the original exception information, and [`RuntimeError`](https://docs.python.org/2/library/exceptions.html#exceptions.RuntimeError) is an uninformative subclass that you use when nothing else fits.

It would be better IMHO to re-raise the native `DockerException` type and ensure that the calling code can deal with that, or at least to add the original exception message and type into the new exception.

controller/scheduler/swarm.py
```diff
@@ -48,14 +52,20 @@ def stop(self, name):
...
     def destroy(self, name):
         """
         Destroy a container
         """
-        self.stop(name)
-        self.docker_cli.remove_container(name)
+        try:
+            self.stop(name)
+            self.docker_cli.remove_container(name)
+        except:
```
mboersma: See above. References:
- https://www.python.org/dev/peps/pep-0020/ "The Bible" lines 10-11
- http://c2.com/cgi/wiki?CatchWhatYouCanHandle

mboersma:
This adds a new parameter (`appid` or `app`--not consistent) to each scheduler class __init__ call, but only the `swarm` scheduler does anything with it. If logging messages are added to swarm, shouldn't we be logging similar messages from fleet (and maybe mock and chaos)?

smothiki:
@bacongobbler @mboersma I have updated the code with docker exceptions. Please take a look

controller/scheduler/swarm.py
```diff
@@ -1,21 +1,22 @@
...
 
-
 MATCH = re.compile(
     r'(?P<app>[a-z0-9-]+)_?(?P<version>v[0-9]+)?\.?(?P<c_type>[a-z-_]+)?.(?P<c_num>[0-9]+)')
 
 
 class SwarmClient(object):
 
-    def __init__(self, target, auth, options, pkey):
+    def __init__(self, target, auth, options, pkey, app):
```
bacongobbler: The idea behind the scheduler client is that it's supposed to be a single global client connection for all apps. Instead of adding try/catch blocks here, maybe we should be capturing the APIError and call `log_event` when we call `.create()` in models.py.

smothiki: Catching API error in .create() is somewhat difficult as we call it in different threads. But if we could catch the api Error in .create() in models.py then passing the same error to the client is easy, which was actually discussed in #3605. This is a temporary fix just to make sure somehow users understand the reason for roll back

bacongobbler: I agree that this is supposed to be a quick fix, but there has to be a middle-ground between quick fixes (adding tech debt) and clean fixes (potentially removing tech debt). This release has been mostly about removing tech debt. @mboersma has the same concern with adding this to the signature with his comment at https://github.com/deis/deis/pull/3692#issuecomment-104003783, so it sounds like this is something that may need to be refactored. I'd be err'ing on the side of getting this done right rather than getting this done fast, considering that this is a QoL change.

smothiki:
Removed ```appid``` argument from schedulers. catching docker exceptions instead of logging events

controller/scheduler/swarm.py
```diff
@@ -48,14 +50,20 @@ def stop(self, name):
         """
         Stop a container
         """
-        self.docker_cli.stop(name)
+        try:
+            self.docker_cli.stop(name)
+        except errors.APIError:
+            raise errors.DockerException("cannot stop container " + name)
```
bacongobbler: I think we still need to capture these in models.py so we don't return a 500 response with no error message, if I'm not mistaken. I'll build this PR maually and test this out.

controller/api/models.py
```diff
@@ -290,8 +290,9 @@ def _start_containers(self, to_add):
         [t.start() for t in create_threads]
         [t.join() for t in create_threads]
         if any(c.state != 'created' for c in to_add):
-            err = 'aborting, failed to create some containers'
+            err = 'aborting, rolling back after failure to create containers'
```
bacongobbler: personally I'm for keeping the error message the same unless we're actually exposing **why** the scheduler failed to create the containers.

smothiki: Sure. I'll revert back. But users can use deis logs and find out why exactly creation of containers failed

carmstrong:
@smothiki Let's see what we can do to get this merged soon. I fear letting it hang about will lead to a lot of refactoring and rebasing.

smothiki:
Removed exception from stop and destroy methods revert back the error message

controller/scheduler/swarm.py
```diff
@@ -1,12 +1,10 @@
 import re
 import time
-
```
mboersma: These blank lines shouldn't be removed. From [PEP8](https://www.python.org/dev/peps/pep-0008/#imports):

> Imports should be grouped in the following order:
- standard library imports
- related third party imports
- local application/library specific imports

> **You should put a blank line between each group of imports.**

controller/scheduler/swarm.py
```diff
@@ -32,11 +30,15 @@ def create(self, name, image, command='', template=None, **kwargs):
...
-                                         host_config={'PublishAllPorts': True})
+        try:
+            self.docker_cli.create_container(image=cimage, name=name,
+                                             command=command.encode('utf-8'),
+                                             mem_limit=mem,
+                                             cpu_shares=cpu,
+                                             environment=[affinity],
+                                             host_config={'PublishAllPorts': True})
+        except errors.APIError as err:
+            raise errors.DockerException(str(err.message)+"  "+str(err.explanation))
```
mboersma: I'm confused why we're catching an `APIError` and re-raising it as a `DockerException`. They're both from docker-py's `errors` package, but I can't find any code that looks for those types specifically. How does this change any behavior?

bacongobbler: I think it's better that we don't add a try/catch block for this issue. [`.create()`](https://github.com/deis/deis/blob/68415ae735c46131e48f818eb049da6bad357fd0/controller/api/models.py#L492-L495) will catch the original error and report it to the user, so we don't need to do any try/catch magic here.

controller/api/models.py
```diff
@@ -292,6 +292,7 @@ def _start_containers(self, to_add):
         if any(c.state != 'created' for c in to_add):
             err = 'aborting, failed to create some containers'
             log_event(self, err, logging.ERROR)
+            self._destroy_containers(to_add)
```
bacongobbler: I think this should be part of .deploy() and .scale() instead. If I'm following the code correctly, what'll happen is that the new containers will be destroyed *as well as* the old containers. In other words, we should be catching the RuntimeError in .scale() and in .deploy(). If an error occurs while running _start_containers() we can destroy that list of containers, prevent .scale() or .deploy() from destroying the old containers and then re-raise the original exception to ensure that other parts of the system still work as intended. How does that sound?

bacongobbler: Oh wait... That's not how exceptions work. The old containers won't be destroyed because an exception was raised, causing the rest of the block to not be interpreted until the RuntimeError is dealt with. Still, I do believe that we should be handling this better in .scale() and in .deploy() rather than in ._start_containers().

smothiki: what ever the containers create objects that are there in to_add will get destroyed. Yes, I agree that this should be handled better in scale operation rather than start_containers. But this is a temporary fix just to make sure we are rolling back the containers if resources are not satisfied by scale requirements. Any ways there is plan to achieve this in https://github.com/deis/deis/issues/3850, but this wont be ready for at-least 1 month.

bacongobbler: OK, I'm fine with this for now. We can clean it up as we go along later.

smothiki:
Removed exceptions from swarm.py handling every thing in models.py. Adjusted blank lines as suggested by @mboersma 
```
$ deis scale cmd=9
 o.. 
  !    WARNING: Client and server API versions do not match. Please consider upgrading.
  !    Client version: 1.3
  !    Server version: 1.5

503 SERVICE UNAVAILABLE
Detail:
aborting, failed to create some containers
$ deis logs
2015-06-16T21:16:48UTC deis[api]: sm scaled containers cmd=5
2015-06-16T21:17:08UTC deis[api]: sm scaled containers cmd=4
2015-06-16T21:17:44UTC deis[api]: sm scaled containers cmd=9
2015-06-16T21:17:44UTC deis[api]: docile-caffeine_v3.cmd.9 (create): __init__() takes at least 3 arguments (2 given)
2015-06-16T21:17:44UTC deis[api]: aborting, failed to create some containers
2015-06-16T21:17:44UTC deis[api]: docile-caffeine_v3.cmd.9 (destroy): 404 Client Error: Not Found ("No such container: docile-caffeine_v3.cmd.9")
2015-06-16T21:38:53UTC deis[api]: sm scaled containers cmd=9
2015-06-16T21:38:53UTC deis[api]: docile-caffeine_v3.cmd.9 (create): 500 Server Error: Internal Server Error ("no resources available to schedule container")
2015-06-16T21:38:54UTC deis[api]: aborting, failed to create some containers
2015-06-16T21:38:54UTC deis[api]: docile-caffeine_v3.cmd.9 (destroy): 404 Client Error: Not Found ("No such container: docile-caffeine_v3.cmd.9")
$ deis info

  !    WARNING: Client and server API versions do not match. Please consider upgrading.
  !    Client version: 1.3
  !    Server version: 1.5

=== docile-caffeine Application
{
  "updated": "2015-06-16T21:17:18UTC", 
  "uuid": "e1ad2367-06ca-4a44-a6c8-10e2fa9a3325", 
  "created": "2015-06-16T21:13:19UTC", 
  "url": "docile-caffeine.local3.deisapp.com", 
  "owner": "sm", 
  "id": "docile-caffeine", 
  "structure": {
    "cmd": 4
  }
}


  !    WARNING: Client and server API versions do not match. Please consider upgrading.
  !    Client version: 1.3
  !    Server version: 1.5

=== docile-caffeine Processes

--- cmd: 
cmd.1 up (v3)
cmd.2 up (v3)
cmd.3 up (v3)
cmd.4 up (v3)


  !    WARNING: Client and server API versions do not match. Please consider upgrading.
  !    Client version: 1.3
  !    Server version: 1.5

=== docile-caffeine Domains
No domains

File: controller/scheduler/swarm.py [hunk 17]
```diff
@@ -6,7 +6,6 @@
 
 from .states import JobState
 
-
```
bacongobbler: Two lines are required here as per @mboersma's comment about PEP8. It recommends one line between import groupings and two lines separating class definitions, variables, methods and import statements from each other. I think we skip the scheduler package entirely from `make test-style` due to the longevitiy of some lines (the container templates in fleet.py, more specifically) but it's still good to follow best practices whenever possible.

smothiki: test-style uses flake8. I ran flake8 manually and didn't throw any errors for swarm.py.

bacongobbler:
Code LGTM.

mboersma:
Code LGTM.

PR #3703, opened 2015-05-15
[WIP] ref(deisctl): use docker create to set up containers

bacongobbler:
Inspiration taken from https://docs.docker.com/articles/host_integration/, this removes the "confusing" calls to `docker history/docker pull`. `docker create` creates the image but does not start it, and also pulls the remote image if it does not exist. Essentially, it's the same as `docker run` without starting the
container, but allows us to configure the image as we see fit *before* we run it.

Note that I've only done this for one component for demonstration purposes, but this does remove the annoying `could not remove container: id does not exist` warning we see after every time the container dies or is stopped, probably because we use the `--rm` flag (even though we have a lot of catches for those edge cases when nodes are suddenly shut down and the units are rebooted). Not much else is changed, save for making it a little more readable :)


mboersma:
> making it a little more readable

I think so--this is a nice refactoring. Code LGTM.

deisctl/units/deis-controller.service
```diff
@@ -6,10 +6,10 @@ After=deis-store-volume.service
 [Service]
 EnvironmentFile=/etc/environment
 TimeoutStartSec=20m
-ExecStartPre=/bin/sh -c "IMAGE=`/run/deis/bin/get_image /deis/controller` && docker history $IMAGE >/dev/null 2>&1 || docker pull $IMAGE"
 ExecStartPre=/bin/sh -c "docker inspect deis-controller >/dev/null 2>&1 && docker rm -f deis-controller || true"
-ExecStart=/bin/sh -c "IMAGE=`/run/deis/bin/get_image /deis/controller` && docker run --name deis-controller --rm -p 8000:8000 -e EXTERNAL_PORT=8000 -e HOST=$COREOS_PRIVATE_IPV4 -v /var/run/fleet.sock:/var/run/fleet.sock -v /var/lib/deis/store:/data $IMAGE"
-ExecStopPost=-/usr/bin/docker rm -f deis-controller
+ExecStartPre=/bin/sh -c "IMAGE=`/run/deis/bin/get_image /deis/controller` && docker create --name deis-controller -p 8000:8000 -e EXTERNAL_PORT=8000 -e HOST=$COREOS_PRIVATE_IPV4 -v /var/run/fleet.sock:/var/run/fleet.sock -v /var/lib/deis/store:/data $IMAGE"
+ExecStart=/usr/bin/docker start -a deis-controller
```
carmstrong: Does container stdout still get captured in the unit journals like it does with a `docker run`?

bacongobbler: that's the `-a` flag for ya. It attaches to the container, meaning stdout gets captured. It wouldn't work without that flag

carmstrong: Ah. Interesting...

bacongobbler:
Okay if this looks like a good approach then I'll start refactoring the other units to bring them in line as well

mboersma:
LGTM, tested on Vagrant. 

Note that moving the creation of `/deis/cache` to the registry boot script means that provisioning v1.6.1 using the master unit files will fail with a confd error, since earlier registry versions depend on the unit file to create it. That's not a supported provisioning config anyway.

carmstrong:
Does this affect anywhere we use `deis-store-admin` and run it with an exec? For example, in #3561 we do: `/usr/bin/docker exec deis-store-admin ceph auth del osd.$OSD_ID`

bacongobbler:
nope. `docker exec` will work the same as before. `docker create`/`docker start -a` is the same as `docker run`. To demonstrate an example:

```
><> docker create alpine:3.1 sh -c "while true; do echo hi; sleep 1; done"
a9a997417a3d4f4fba1e30d507798d40649f8ec6b325d1f0d7f3cf97166fc203
><> docker start -a a9a99
hi
hi
hi
hi
hi
hi
hi
[in another terminal]
><> docker exec a9a99 echo bye
bye
```

carmstrong:
Ok. So if we create and give it a name, we can exec that name just like we could before?

bacongobbler:
precisely :)

carmstrong:
Cool. Would like @gabrtv to take a look quickly in case we missed anything, but code LGTM with @mboersma's testing blessing.

deisctl/units/deis-builder.service
```diff
@@ -5,13 +5,13 @@ Description=deis-builder
...
 ExecStartPre=/bin/sh -c "docker inspect deis-builder >/dev/null 2>&1 && docker rm -f deis-builder || true"
+ExecStartPre=/bin/sh -c "IMAGE=`/run/deis/bin/get_image /deis/builder` && docker create --name deis-builder -p 2223:22 --volumes-from=deis-builder-data -c 800 -e EXTERNAL_PORT=2223 -e HOST=$COREOS_PRIVATE_IPV4 --privileged -v /etc/environment_proxy:/etc/environment_proxy $IMAGE"
 ExecStartPre=-/bin/sh -c "/sbin/losetup -f"
-ExecStart=/bin/sh -c "IMAGE=`/run/deis/bin/get_image /deis/builder` && docker run --name deis-builder --rm -p 2223:22 --volumes-from=deis-builder-data -c 800 -e EXTERNAL_PORT=2223 -e HOST=$COREOS_PRIVATE_IPV4 --privileged -v /etc/environment_proxy:/etc/environment_proxy $IMAGE"
+ExecStart=/usr/bin/docker start -a deis-builder
 ExecStartPost=/bin/sh -c "echo 'Waiting for builder on 2223/tcp...' && until ncat $COREOS_PRIVATE_IPV4 2223 --exec '/usr/bin/echo dummy-value' >/dev/null 2>&1; do sleep 1; done"
 ExecStartPost=/usr/bin/docker exec deis-builder /usr/local/bin/push-images
-ExecStopPost=-/usr/bin/docker rm -f deis-builder
+ExecStop=/usr/bin/docker stop -t 2 deis-builder ; /usr/bin/docker rm -f deis-builder
```
gabrtv: I could be wrong, but I don't believe the command chaining will work without the subshell.  IOW, I don't think `/usr/bin/docker rm -f deis-builder` will be called.  Can we confirm that?

gabrtv: If I'm right, we can probably add the other command in an `ExecStopPost` and move on.

bacongobbler: Coming straight from the [systemd docs](http://www.freedesktop.org/software/systemd/man/systemd.service.html), it appears to work:

> Example:

    ExecStart=/bin/echo one ; /bin/echo "two two"

> This will execute /bin/echo two times, each time with one argument: "one" and "two two", respectively. Because two commands are specified, Type=oneshot must be used.

registry/bin/boot
```diff
@@ -40,9 +40,20 @@ function etcd_set_default {
   set -e
 }
 
+function etcd_safe_mkdir {
```
gabrtv: What's this doing in a deisctl commit?

deisctl/units/deis-builder.service
bacongobbler: ...oh. Missed that last bit. I guess we can make it part of the ExecStopPost. Good catch

registry/bin/boot
bacongobbler: There was a call in one of the units to `etcdctl mkdir /deis/cache`. We do this in the other boot scripts, so I refactored that into registry's boot script. Since there was no call to `etcd_safe_mkdir` in there I just copy-pasta'd the function over.

deisctl/units/deis-builder.service
bacongobbler: I guess docker's docs need to be updated too, hehe

registry/bin/boot
gabrtv: OK, though we _really_ need to refactor these init scripts into Go.

mboersma:
I tested this again after the last update, and `deisctl stop platform` still fails on publisher:
```
core@deis-01 ~ $ systemctl status deis-publisher.service
● deis-publisher.service - deis-publisher
   Loaded: loaded (/run/fleet/units/deis-publisher.service; linked-runtime; vendor preset: disabled)
   Active: failed (Result: exit-code) since Thu 2015-05-21 19:54:57 UTC; 7s ago
  Process: 14058 ExecStopPost=/usr/bin/docker rm -f deis-publisher (code=exited, status=0/SUCCESS)
  Process: 14043 ExecStop=/usr/bin/docker stop -t 2 deis-publisher (code=exited, status=0/SUCCESS)
  Process: 2856 ExecStart=/usr/bin/docker start -a deis-publisher (code=exited, status=2)
  Process: 2691 ExecStartPre=/bin/sh -c IMAGE=`/run/deis/bin/get_image /deis/publisher` && docker create --name deis-publisher -v /var/run/docker.sock:/var/run/docker.sock $IMAGE --host=$COREOS_PRIVATE_IPV4 --etcd-host=$COREOS_PRIVATE_IPV4 (code=exited, status=0/SUCCESS)
  Process: 2675 ExecStartPre=/bin/sh -c docker inspect deis-publisher >/dev/null 2>&1 && docker rm -f deis-publisher || true (code=exited, status=0/SUCCESS)
 Main PID: 2856 (code=exited, status=2)

May 21 19:26:26 deis-01 sh[2691]: Status: Downloaded newer image for 192.168.59.103:5000/deisci/publisher:git-6f2ea51
May 21 19:26:26 deis-01 systemd[1]: Started deis-publisher.
May 21 19:26:26 deis-01 sh[2691]: 3e6b4affba0b903a24540c62976e3e2d3b233941a9812f9c8c5af45d15c55feb
May 21 19:54:57 deis-01 systemd[1]: Stopping deis-publisher...
May 21 19:54:57 deis-01 docker[14043]: deis-publisher
May 21 19:54:57 deis-01 systemd[1]: deis-publisher.service: main process exited, code=exited, status=2/INVALIDARGUMENT
May 21 19:54:57 deis-01 docker[14058]: deis-publisher
May 21 19:54:57 deis-01 systemd[1]: Stopped deis-publisher.
May 21 19:54:57 deis-01 systemd[1]: Unit deis-publisher.service entered failed state.
May 21 19:54:57 deis-01 systemd[1]: deis-publisher.service failed.
```

carmstrong:
I think the consensus was we're for these changes once @mboersma's issue is addressed.

bacongobbler:
I also had reservations against this PR after thinking about it for a while. If we wanted to run one-off commands in the same container *before* we started it (such as moving inital backups from the database's boot script as a standalone command), we have to resort to `docker run`. The implementation is fine as-is. Why break what isn't broken? ;)

Closing for now.

PR #3705, opened 2015-05-16
ref(controller): remove unused abstract exception class

mboersma:
`pylint` spit out some warnings for exceptions.py, and on further inspection I couldn't see anywhere the `AbstractDeisException` class was being called. less == more.


carmstrong:
> less == more.

Preach it, brother. Code LGTM.

bacongobbler:
At one point we had one exception utilizing this exception class, but I don't see it any more so it must've been cycled out. Code LGTM!

PR #3707, opened 2015-05-16
ref(controller): add urls.py docstring and remove unused arg

mboersma:
`pylint` pointed out this module lacked a docstring, and that a certs URL was ignoring a format arg. Also switched single- to double-quotes for format strings to follow an unenforced convention in the codebase.

### DIFF
--- a/controller/api/urls.py
+++ b/controller/api/urls.py
@@ -1,3 +1,7 @@
+"""
+URL routing patterns for the Deis REST API.
+"""
+
 from __future__ import unicode_literals
 
 from django.conf import settings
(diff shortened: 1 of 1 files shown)


controller/api/urls.py
```diff
@@ -13,59 +17,58 @@
...
     # apps sharing
-    url(r'^apps/(?P<id>{})/perms/(?P<username>[-_\w]+)/?'.format(settings.APP_URL_REGEX),
+    url(r"^apps/(?P<id>{})/perms/(?P<username>[-_\w]+)/?".format(settings.APP_URL_REGEX),
         views.AppPermsViewSet.as_view({'delete': 'destroy'})),
-    url(r'^apps/(?P<id>{})/perms/?'.format(settings.APP_URL_REGEX),
+    url(r"^apps/(?P<id>{})/perms/?".format(settings.APP_URL_REGEX),
         views.AppPermsViewSet.as_view({'get': 'list', 'post': 'create'})),
     # apps base endpoint
-    url(r'^apps/(?P<id>{})/?'.format(settings.APP_URL_REGEX),
+    url(r"^apps/(?P<id>{})/?".format(settings.APP_URL_REGEX),
```
kalbasit: why double quotes? Not consisting with the rest of the files.

mboersma: Python doesn't care which quotes you use, but previously a couple of us had maintained lots of CoffeeScript code alongside Python. It's not consistent in the codebase, but I have a preference for Ruby-style quoting even in Python code to indicate strings which are used for templating.

Obviously it doesn't matter either way; we used to think it was a bit helpful to indicate intent.

bacongobbler:
I'm not picky on the whole single/double quote argument (unless the string itself contains single/double quotes, then I start to care). Code LGTM!

carmstrong:
Code LGTM.

PR #3708, opened 2015-05-17
More efficient iteration

mboersma:
This change prefers Python 2.7 [dict views](https://docs.python.org/2/library/stdtypes.html#dictionary-view-objects), [`xrange`](https://docs.python.org/2/library/functions.html#xrange) to `range`, and iterating over dicts directly without calling [`keys()`](https://docs.python.org/2/library/stdtypes.html#dict.keys). These idioms avoid unnecessary copying and are generally more efficient.

### DIFF
--- a/client/deis.py
+++ b/client/deis.py
@@ -1357,7 +1357,7 @@ Make sure that the Controller URI is correct and the server is running.
             # write env_dict to .env
             try:
                 with open('.env', 'w') as f:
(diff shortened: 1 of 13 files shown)


bacongobbler:
TIL about `view*()` and `xrange()`. Both are good changes, and removing the `.keys()` calls are good too. Code LGTM!

carmstrong:
Code LGTM.

PR #3710, opened 2015-05-18
fix(*): suppress "Key already exists" errors

### LINKED ISSUE #2597: Go-based /bin/boot scripts output warnings for "Key already exists"

bacongobbler:
closes #2597

### DIFF
--- a/cache/main.go
+++ b/cache/main.go
@@ -105,7 +105,7 @@ func publishService(client *etcd.Client, host string, etcdPath string,
 
 func setEtcd(client *etcd.Client, key, value string, ttl uint64) {
 	_, err := client.Set(key, value, ttl)
-	if err != nil {
+	if err != nil && !strings.Contains(err.Error(), "Key already exists") {
 		log.Println(err)
 	}
 }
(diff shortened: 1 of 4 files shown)


mboersma:
Code LGTM.

carmstrong:
Code LGTM.

PR #3713, opened 2015-05-18
feat(controller): Add endpoint to regenerate tokens

Joshua-Anderson:
If a token was leaked, it could be regenerated with: `deis auth:regenerate`. Admin can regenerate a user's token with `deis auth:regenerate -u user` and could also regenerate everybody's auth token, in the event of a security breach, with `deis auth:regenerate --all=true`

### DIFF
--- a/client/deis.py
+++ b/client/deis.py
@@ -83,7 +83,7 @@ if hasattr(signal, 'SIGPIPE') and hasattr(signal, 'SIG_DFL'):
 __version__ = '1.8.0-dev'
 
 # what version of the API is this client compatible with?
-__api_version__ = '1.4'
+__api_version__ = '1.5'
 
 
 locale.setlocale(locale.LC_ALL, '')
@@ -737,6 +737,7 @@ Make sure that the Controller URI is correct and the server is running.
         auth:passwd            change the password for the current user
         auth:whoami            display the current user
         auth:cancel            remove the current user account
+        auth:regenerate        regenerate user tokens
 
         Use `deis help [command]` to learn more.
         """
@@ -962,6 +963,36 @@ Make sure that the Controller URI is correct and the server is running.
         else:
             self._logger.info(
                 'Not logged in. Use `deis login` or `deis register` to get started.')
+
+    def auth_regenerate(self, args):
+        """
+        Regenerates auth token, defaults to regenerating token for the current user.
+
+        Usage: deis auth:regenerate [options]
+
+        Options:
+          -u --username=<username>
+            specify user to regenerate. Requires admin privilages.
+          --all
+            regenerate token for every user. Requires admin privilages.
+        """
+        payload = {}
+
+        if args.get('--all'):
+            payload = {'all': True}
+        elif args.get('--username'):
+            payload = {'username': args.get('--username')}
+
+        response = self._dispatch('post', '/v1/auth/tokens/', json.dumps(payload))
+
+        if response.status_code == requests.codes.ok:
+            if '--username' not in args or '--all' not in args:
+                self._settings['token'] = response.json()['token']
+                self._settings.save()
+            self._logger.info('Token regenerated.')
+        else:
+            self._logger.info("Token regeneration failed: {}".format(response.text))
+            sys.exit(1)
 
     def builds(self, args):
         """
--- a/controller/api/permissions.py
+++ b/controller/api/permissions.py
@@ -129,3 +129,18 @@ class HasBuilderAuth(permissions.BasePermission):
         if not auth_header:
             return False
         return auth_header == settings.BUILDER_KEY
+
+
+class CanRegenerateToken(permissions.BasePermission):
+    """
+    Checks if a user can regenerate a token
+    """
+
+    def has_permission(self, request, view):
+        """
+        Return `True` if permission is granted, `False` otherwise.
+        """
+        if 'username' in request.data or 'all' in request.data:
+            return request.user.is_superuser
+        else:
+            return True
--- a/controller/api/tests/test_auth.py
+++ b/controller/api/tests/test_auth.py
@@ -278,3 +278,33 @@ class AuthTest(TestCase):
         response = self.client.post(url, json.dumps(submit), content_type='application/json',
                                     HTTP_AUTHORIZATION='token {}'.format(self.user1_token))
         self.assertEqual(response.status_code, 200)
+
+    def test_regenerate(self):
+        """ Test that token regeneration works"""
+
+        url = '/v1/auth/tokens/'
+
+        response = self.client.post(url, '{}', content_type='application/json',
+                                    HTTP_AUTHORIZATION='token {}'.format(self.admin_token))
+
+        self.assertEqual(response.status_code, 200)
+        self.assertNotEqual(response.data['token'], self.admin_token)
+
+        self.admin_token = Token.objects.get(user=self.admin)
+
+        response = self.client.post(url, '{"username" : "autotest2"}',
+                                    content_type='application/json',
+                                    HTTP_AUTHORIZATION='token {}'.format(self.admin_token))
+
+        self.assertEqual(response.status_code, 200)
+        self.assertNotEqual(response.data['token'], self.user1_token)
+
+        response = self.client.post(url, '{"all" : "true"}',
+                                    content_type='application/json',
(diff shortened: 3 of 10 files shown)


carmstrong:
Not sure introducing a `deis token` command is the best way to expose this, as a. Maybe something like `deis auth:regenerate`? A user could do it to himself, or an admin could invoke the command on any user. 

I also wonder if `deis auth:logout` could suffice -- an admin could supply a username to the command, and it would log the user out by regenerating their access token.

bacongobbler:
`deis auth:token --new`, perhaps?

Joshua-Anderson:
I'm fine with either of those client commands. Is creating a new endpoint for the controller the right place to put this, or would you rather see it under, say, `/v1/auth/token/`?

mboersma:
I think `v1/auth/tokens` looks cleaner--we grouped the other auth URLs there (although they aren't proper REST endpoints). `v1/tokens` is acceptable too.

bacongobbler:
IOW what @mboersma is trying to say that the API implementation is really up to you, so use your own discretion. We're mostly concerned with getting the UX nailed down. Refactoring APIs is easy; refactoring UX is "impossible" unless we release v2.0, so that's what we're mostly concerned with :)

Joshua-Anderson:
I added actual code now. For now, I went with `auth:regenerate`. I also added `auth:regenerate --all=true` so a admin can regenerate everybody's token in the event of a security breach and `auth:regenerate -u user` so that a admin can regenerate a specific user.

docs/managing_deis/operational_tasks.rst
```diff
@@ -52,28 +52,22 @@ Re-issuing User Authentication Tokens
...
 .. code-block:: console
 
-    $ fleetctl ssh deis-controller
-    $ docker exec -it deis-controller python manage.py shell
-    >>>
+    $ deis auth:regenerate
 
-At this point, let's re-issue an auth token for this user. Let's assume that the name for the user
-is Bob (poor Bob):
+A admin can also regenerate the token of another user like this:
```
bacongobbler: An*

We should alsouse the fully-formed "administrator" here, since this is formal documentation. :necktie:

docs/managing_deis/operational_tasks.rst
```diff
@@ -83,14 +77,10 @@ token:
     Detail:
     Invalid token
 
-For Bob to be able to use the API again, he will have to authenticate against the controller to be
-re-issued a new token:
+They will need to log back in to use their new auth token.
+
+If there is a cluster wide security breach, a admin can regenerate everybody's auth token like this:
```
bacongobbler: an administrator*

docs/reference/api-v1.4.rst
```diff
@@ -117,6 +118,40 @@ Example Response:
...
+
+.. note::
+
+    This command could require administrative privileges
+
+Example Request:
+
+.. code-block:: console
+
+    DELETE /v1/auth/tokens/ HTTP/1.1
```
bacongobbler: Hmm... Maybe we should make this a POST instead of a DELETE. We're not actually deleting a token here, just generating a new one.

carmstrong: :+1:

carmstrong:
> For now, I went with auth:regenerate. I also added auth:regenerate --all=true so a admin can regenerate everybody's token in the event of a security breach and auth:regenerate -u user so that a admin can regenerate a specific user.

:+1: Awesome work so far, @Joshua-Anderson!

bacongobbler:
note to self: investigate implications with external auth (we have support for LDAP)

Joshua-Anderson:
@bacongobbler I fixed the documentation and changed from DELETE to POST.

carmstrong:
@Joshua-Anderson Looks like we need a rebase now :(

Joshua-Anderson:
@carmstrong Done!

carmstrong:
Code LGTM. We should see what effect this has on the LDAP auth integration, if any.

mboersma:
Talking with @bacongobbler, we're confident LDAP isn't hampered by this.That's on the backend and ends up using the same token method where the client is concerned, so this should not matter.

Code LGTM.

client/deis.py
```diff
@@ -941,6 +942,36 @@ def auth_whoami(self, args):
...
+    def auth_regenerate(self, args):
+        """
+        Regenerates auth token, defaults to regenerating token for the current user.
+
+        Usage: deis auth:regenerate [options]
+
+        Options:
+          -u --username=<username>
+            specify user to regenerate. Requires admin privilages.
+          --all=<all>
```
bacongobbler: We should change this to `--all` so it'll end up being true if set, otherwise users will have to use `--all=foo` in order to use this feature flag.

controller/api/permissions.py
```diff
@@ -129,3 +129,18 @@ def has_permission(self, request, view):
...
+    """
+
+    def has_permission(self, request, view):
+        """
+        Return `True` if permission is granted, `False` otherwise.
+        """
+        if 'username' in request.data or 'all' in request.data:
+            return request.user.is_superuser
+        else:
+            return True
```
bacongobbler: we should always be checking if the user is a superuser. In that case we can re-use the IsAdmin class instead.

controller/api/views.py
```diff
@@ -51,6 +52,40 @@ def passwd(self, request, **kwargs):
...
+        return User.objects.filter(pk=self.request.user.pk)
+
+    def get_object(self):
+        return self.get_queryset()[0]
+
+    def regenerate(self, request, **kwargs):
+        obj = self.get_object()
+
+        if 'all' in request.data:
+            for user in User.objects.exclude(username='AnonymousUser'):
```
bacongobbler: rather than checking for the username, we can check `.is_anonymous()` instead to achieve the same effect. That may be more portable.

controller/api/views.py
```diff
@@ -51,6 +52,40 @@ def passwd(self, request, **kwargs):
...
+                token = Token.objects.get(user=user)
+                token.delete()
+                token = Token.objects.create(user=user)
```
bacongobbler: let's remove the assignment since we're not doing anything with it.

controller/api/views.py
```diff
@@ -51,6 +52,40 @@ def passwd(self, request, **kwargs):
...
+            return Response("",
+                            status=status.HTTP_204_NO_CONTENT,
```
bacongobbler: this should be a 200 OK response to confirm that the operation succeeded

controller/api/permissions.py
bacongobbler: actually I see the reason behind this now. Disregard!

controller/api/tests/test_auth.py
```diff
@@ -278,3 +278,33 @@ def test_change_user_passwd(self):
...
+                                    content_type='application/json',
+                                    HTTP_AUTHORIZATION='token {}'.format(self.admin_token))
+
+        self.assertEqual(response.status_code, 200)
+        self.assertNotEqual(response.data['token'], self.user1_token)
+
+        response = self.client.post(url, '{"all" : "true"}',
+                                    content_type='application/json',
+                                    HTTP_AUTHORIZATION='token {}'.format(self.admin_token))
+        self.assertEqual(response.status_code, 204)
```
bacongobbler: same as other comment, we should be testing for a 200 OK response

Joshua-Anderson:
@bacongobbler Thanks for the feedback! I think I addressed it.

Joshua-Anderson:
The tests seemed to time out, could somebody please rerun this PR.

mboersma:
@Joshua-Anderson could you rebase this off current master? I think #3771 was killing most of the tests until we fixed it. Apologies for the constant rebasing. :hammer:

Joshua-Anderson:
@mboersma Rebased!

carmstrong:
Code LGTM.

bacongobbler:
Much better! Sadly we're in code freeze atm for v1.7 so I'll defer the merge to @carmstrong but code LGTM :)

Joshua-Anderson:
I bumped the api version since v1.4 was released.

mboersma:
Code LGTM after updating the API to v1.5.

carmstrong:
Code LGTM.

Issue #3698, opened 2015-05-15
deis CLI: `deis releases prune` command


carmstrong:
@haf has proposed a new command in #3693 which would purge all old releases and their images, reclaiming disk space.

croemmich:
I'd be a big fan of this. I'm running on Linode and have a somewhat limited supply of disk space.

The hardest part is removing the images from the registry, but my https://github.com/myriadmobile/docker-s3-cleaner project seems to be working properly for such. The only other steps required would be to delete the release from the database and untag the image in the registry.

**Possible Syntax:**
Destroy single release: `deis releases:destroy v5`
Destroy releases that are older than 10 days: `deis releases:prune age=10d`
Destroy all releases except the latest 3: `deis releases:prune keep=3`

**Questions:**
What is the expected behavior if the current version is pruned?
What is the expected behavior of rollback if the previous release has been deleted?

bacongobbler:
While I agree with the motivation behind this, I'm not too sure if we should approach it through the API. The release ledger is an append-only ledger. While you can remove the release from the database (which we should disable), the [views don't allow you to remove a specific release](https://github.com/deis/deis/blob/master/controller/api/tests/test_release.py#L110-L111), so we'd be breaking [one of our core tenets](http://12factor.net/build-release-run):

> Releases are an append-only ledger and a release cannot be mutated once it is created. Any change must create a new release.

That being said, I'd be happy if we introduced a deis-specific cleanup component to remove images from the hosts and removing images from the registry would be preferable, IMO. Since these are artifacts of the release being distributed across the scheduler, these can be cleaned up.

carmstrong:
> That being said, I'd be happy if we introduced a deis-specific cleanup component to remove images from the hosts and removing images from the registry. Since these are artifacts of the release being distributed across the scheduler, these can be cleaned up.

Yeah, I wasn't thinking we'd remove anything from the ledger - rather clean up the images themselves. This is also related to what we're discussing in #3675

croemmich:
Yeah, #3675 should take care of cleaning up the hosts, which just leaves the images in the registry.

What's the thought when the image backing the release is removed? At that point the release is nearly useless. I don't see an issue with soft-deleting the release or marking the build as deleted, as a row in the database isn't really an issue, however I really need a way to delete the backing images.

[...]

PR #3716, opened 2015-05-19
feat(user-data): add disk cleanup script

### LINKED ISSUE #3675: Does deis have disk self-cleaning mechanism?
My cluster is filling up, most of space is taken by /var/lib/docker, currently I've only 4 apps. After running `docker images -aq | xargs -l10 docker rmi` almost nothing is removed. But `docker images -a` shows many of previous deployments. Is there some kind of logic which will keep disk space sane?

croemmich:
Add a script that cleans up docker images, exited containers, and orphaned volumes.

Closes #3675.


contrib/coreos/user-data.example
```diff
@@ -176,3 +176,10 @@ write_files:
     content: |
       [Coredump]
       Storage=none
+  - path: /run/deis/bin/docker-cleanup
+    permissions: '0755'
+    content: |
+      #!/bin/bash
+      docker rm `docker ps -a --no-trunc | grep '\sExited\s' | grep -v '\sdeis-.*$' | awk '{print $1 }'`
+      docker run -v /var/run/docker.sock:/var/run/docker.sock -v /var/lib/docker:/var/lib/docker --rm martin/docker-cleanup-volumes:1.5.0
```
PierreKircher: martin/docker-cleanup-volumes:1.5.0 sourcecode please and it might as well be a deis repo for trust

croemmich:
@PierreKircher The source code can be found at:
https://github.com/chadoe/docker-cleanup-volumes

I am not the author, but it is MIT.

contrib/coreos/user-data.example
kalbasit: I agree, if this is merged it should reference a deis-maintained image. Otherwise :+1:

croemmich:
@carmstrong: Do you want to pull the martin/docker-cleanup-volumes:1.5.0 image to the deis repo or should I remove it from the script?

mboersma:
I made an `alpine:3.1` branch of [docker-cleanup-volumes](https://github.com/mboersma/docker-cleanup-volumes). Maybe that could be `deis/docker-cleanup-volumes` and we could switch this PR to that?

I'd like to hear what other folks think about this approach, and i'd also like to test this.

mboersma:
My PR to base [docker-cleanup-volumes](https://github.com/chadoe/docker-cleanup-volumes/pull/9) off `alpine:3.1` got merged, so that should play nicer with the rest of Deis' components if we decide to go this way.

mboersma:
This seems to work, but the script is pretty aggressive: it removes the `docker-cleanup-volumes` image itself, so every time you execute `/run/deis/bin/docker-cleanup` it must refetch from Docker Hub. Maybe we could just use `docker-cleanup-volumes.sh` directly rather than requiring it to be in a Docker container?
```console
core@deis-01 ~ $ /run/deis/bin/docker-cleanup 
docker: "rm" requires a minimum of 1 argument. See 'docker rm --help'.
Unable to find image 'mboersma/docker-cleanup-volumes:1.5.0' locally
Pulling repository mboersma/docker-cleanup-volumes
3bc6d3dd230f: Download complete 
a1dd7097a8e8: Download complete 
2d04186a421e: Download complete 
6fbcd934805b: Download complete 
e6ee654e8832: Download complete 
26ca66589ae8: Download complete 
3c9a683a8cf9: Download complete 
4f223e05cedd: Download complete 
Status: Downloaded newer image for mboersma/docker-cleanup-volumes:1.5.0

Delete unused volume directories from /var/lib/docker/volumes
In use 784eebd795936ad803399949cee5f5d11a9b975c44e65b050d1993f94c95b2d1
In use 7b08be91dc3832d7dc2493e1637f3d4a79bd47c6b514358a3618dbadeb4dda41
In use ea82c5331290f24d8033b3a1f9c3cd75ab42bb5c1d574989a8bf55e77fb12384
In use fba14b3f55041c550345fe527254dd65db296cfa996bcf306ffc1f118f11a0a7

Delete unused volume directories from /var/lib/docker/vfs/dir
In use ea82c5331290f24d8033b3a1f9c3cd75ab42bb5c1d574989a8bf55e77fb12384
In use fba14b3f55041c550345fe527254dd65db296cfa996bcf306ffc1f118f11a0a7
Error response from daemon: Conflict, 4f223e05cedd wasn't deleted
Untagged: mboersma/docker-cleanup-volumes:1.5.0
Deleted: 3bc6d3dd230f9708aa3fc41308a5f9123438a6f7e786d083d213f7c02e775eef
Deleted: 4f223e05cedde055be57fb9063844ca630f8274b1db3edd07b10c2028bca9746
Deleted: 3c9a683a8cf99e6471bf43f1213f28a3cde02e34cfd1d743d7d663fe691b471f
Deleted: 26ca66589ae865793da80075a08cc6f176be24efa0ef030ec446d41f7a74a9be
Deleted: e6ee654e8832b11a57366e4efdb79979a2343f0c3671f6a470306d1e032e7173
Deleted: 6fbcd934805bcf6d786d00983159c23881f238876775ab153a6280ef736c11d8
Deleted: 2d04186a421e4eac0eca49155d612ae6f36130a174a637d5dda2e4a76c67f5e8
Error response from daemon: No such image: 3c9a683a8cf9
Error response from daemon: No such image: 26ca66589ae8
Error response from daemon: No such image: e6ee654e8832
Error response from daemon: No such image: 2d04186a421e
Error response from daemon: No such image: 6fbcd934805b
Untagged: alpine:3.1
Untagged: alpine:latest
Deleted: a1dd7097a8e87f3f1459bf737f4a6f4fa867190a7069417ab7bf6bbf0ae038a9
Error response from daemon: No such image: a1dd7097a8e8
FATA[0000] Error: failed to remove one or more images
core@deis-01 ~ $
```

PierreKircher:
agree its a simple sh anyway which we could put in the writefile part of the cloud-init and ignore the 2 store volumes

croemmich:
@PierreKircher My only issue is that the cloud-init file is already getting long. I was going to open a ticket for precisely that issue last week, but never got around to it.

bacongobbler:
at this point we should be shipping our own Chef cookbook for Deis on CoreOS :trollface:

PierreKircher:
well could just download if from a public gist maintained by deis .. so its not that BRUTALY long .. and lol @bacongobbler "wasnt that how deis started ?"

gabrtv:
Today we rely on `cloud-init` so this PR is fine by me, but I can't resist... Lesson 7 from my talk at CoreOS Fest: http://gabrtv.github.io/10-lessons-learned-using-coreos/#/41

Over time we need to move what's in `cloud-init` into "agents" running in containers, with policy enforced via the Update Service referenced on #3728.  We'll know we've done our job when the only thing required in cloud-init besides normal sysadmin stuff is the installation of a Deis agent tied to a channel (alpha, beta, stable).

mboersma:
This could still be a container-based agent, we would just need to find a way to keep it from deleting itself. :)

croemmich:
Alright, it's a tad dirty, but the script will now maintain Deis images for the current platform version and the docker-cleanup-volumes image.

carmstrong:
Thanks for your hard work on this, @croemmich - you will be revered in the halls of Deis users for generations to come.

carmstrong:
@croemmich We just discussed this in this week's issue/PR review. We're wondering if it's a bit clunky to have a Docker container clean up other Docker containers (and even shoot itself in the head). We're debating moving the shell script in the `martin/docker-cleanup-volumes` onto the CoreOS host itself. One downside is that it's something we'd have to maintain, but I think for something this important the maintenance burden is worth it.

What do you think, @croemmich?

croemmich:
@carmstrong We certainly can. I just hate to add another 90 lines to the cloud-init.

carmstrong:
> I just hate to add another 90 lines to the cloud-init.

Me too.. I was toying around with the idea of splitting the user-data up into separate files and then having a Makefile target combine them, but not sure how that would work with the provision scripts.

I like that approach because we could isolate each unit service to its own file, making it *much* easier to maintain.

carmstrong:
> I was toying around with the idea of splitting the user-data up into separate files

Filed as #3751

carmstrong:
@croemmich If you wanted to take a stab at that, I would be forever grateful... :heart:

krancour:
Couldn't the new script be hosted in S3 instead of adding 90+ lines to cloud config?  Isn't that done for a handful of assets today?  (Not sure which ones exactly, but I recall it being mentioned recently.)

bacongobbler:
we only do that for some of the static binaries (like fleetd v0.9.2), but not for scripts.

carmstrong:
> Couldn't the new script be hosted in S3 instead of adding 90+ lines to cloud config? Isn't that done for a handful of assets today?

I don't like having another external dependency for platform startup - we already require access to the internet for Docker Hub pulls, and some users would like to package everything locally.

krancour:
Makes sense.  I did see some concern expressed above over an additional 90 lines in the cloud-config-- which means another 90 lines in the CloudFormation template.  Not sure about other providers, but CF has a limit on template size.  The existing template probably isn't close to it, but knowing that people also provision resources for Deis using custom templates that may be larger-- those extra 90 lines could be a tipping point for some.

But I totally see and concede to your point.  :smile:

mboersma:
Instead of adding to `user-data` at all, maybe the cleanup script is something we could just reference in documentation? It doesn't seem much more work to download and run the script, or it could still be a one-line command if run as a Docker container.

That would avoid the Docker image vs. raw script debate, and not require another potentially troublesome download during cloud-config processing. Deis could provide the image at a known Docker Hub repo, or the script at an S3 url, and this PR would just document how to access and use it.

@croemmich whatcha think?

croemmich:
@mboersma I think that's a reasonable solution, thanks!

carmstrong:
> maybe the cleanup script is something we could just reference in documentation?

I think we wanted something in Deis itself, a more or less "official" method to cleaning up images.

> The Deis project could provide the image at a known Docker Hub repo, or the script at an S3 URL, and this PR would just document how to access and use it.

My concern with that is there's less visibility. If it's an S3 script we'll have to maintain it somehow, and git seems like the best place to version something like that.

The goal here is to provide a safe, easy-to-use, and supported method for preventing users from running out of disk space.

Wherever the script lives, I think we have to take on the burden of its maintenance considering how these disk space issues impact our users. Note that #3698 is also similar to this, and we might be able to address these with one fell swoop. Maybe we should take a step back and think about how best to approach this.

bacongobbler:
So to keep the ball rolling, I guess there's a few things needed to be done here from my point of view to merge this:

1. Introduce a new component called `deis/waste-collector` (just throwing out a name) that is an agent like publisher and logspout that runs on all hosts
2. `deis/waste-collector` contains the script from https://github.com/chadoe/docker-cleanup-volumes, running it periodically (daily?)
3. introduce this new component to `deisctl` so that this new component is installed when a user upgrades to v1.8
4. Add documentation on this new component, including what it does, how often it does it, how to tweak/uninstall it, etc.

Since this is something that seems pretty important and that we'd need to support, installing a new component seems to be the most ideal solution such that users don't have to re-provision their platform from scratch, and it does give users who do not want their artifacts to be cleaned up a way to opt out.

Does that seem to cover your use cases/concerns, @croemmich and @carmstrong?

kalbasit:
:+1: @bacongobbler

croemmich:
:thumbsup: @bacongobbler

krancour:
:+1: @bacongobbler _and_ hot off the dockercon press... with some further polish, might this be similar to what you're describing?  https://github.com/rancherio/sherdock

bacongobbler:
Yes, that sounds pretty close to what I was envisioning. :)

carmstrong:
Perhaps it's time to close this and move the discussion back to #3675 until a solution is agreed upon? Sherdock looks like a good candidate, but we'll have to ensure that it's not removing any volumes or things we want.

bacongobbler:
Okay, let's do that. @croemmich you're more than welcome to take a hack at it if you're up for it, but let's open a new PR with the proposed solution. Thanks!

PR #3718, opened 2015-05-19
feat(platform): support placement options for each plane and router mesh

krancour:
Addresses #3023 

Documentation to follow if the approach is satisfactory.

This updates units (and code that generates units) such that Deis platform
components belonging to the control plane can be scheduled only where
machine metadata includes controlPlane=true.  Similarly, data plane components
(including applications) can be scheduled only where machine metadata includes
dataPlane=true and router mesh components can be scheduled only whrere machine
metadata includes routerMesh=true.

user-data.example is modified such that machines configured from that
cloud-config are eligible to host both planes and the router mesh.  This
permits small clusters composed entirely of general purpose nodes to be
constructed as easily as ever before and permits the platform to be installed
to such a cluster without complication, but it additionally creates the
the ability for advanced operators using custom cloud-config to easily partition
the nodes of a larger cluster by role without needing to modify/customize unit
files.  This is especially useful for anyone wishing to implement CoreOS' own
recommended large / production-worthy cluster architecture documented here:

https://coreos.com/docs/cluster-management/setup/cluster-architectures/#production-cluster-with-central-services

~~BREAKING CHANGE: Requires that all machines hosting Deis components
be assigned one or more of controlPlane=true, dataPlane=true, or
routerMesh=true via cloud-config.  This represents a breaking
change for anyone upgrading an existing clutser.~~

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,6 +27,7 @@ dev-cluster: discovery-url
 	ssh-add ~/.vagrant.d/insecure_private_key
 	deisctl config platform set sshPrivateKey=$(HOME)/.vagrant.d/insecure_private_key
 	deisctl config platform set domain=local3.deisapp.com
+	deisctl config platform set enablePlacementOptions=true
 	deisctl install platform
 
 discovery-url:
--- a/contrib/coreos/user-data.example
+++ b/contrib/coreos/user-data.example
@@ -7,6 +7,7 @@ coreos:
     public-ip: $private_ipv4
     # allow etcd to slow down at times
     etcd_request_timeout: 3.0
+    metadata: controlPlane=true,dataPlane=true,routerMesh=true
   units:
   - name: etcd.service
     command: start
--- a/controller/scheduler/fleet.py
+++ b/controller/scheduler/fleet.py
@@ -150,10 +150,10 @@ class FleetHTTPClient(AbstractSchedulerClient):
             f['value'] = f['value'].format(**l)
         # prepare tags only if one was provided
         tags = kwargs.get('tags', {})
-        if tags:
-            tagset = ' '.join(['"{}={}"'.format(k, v) for k, v in tags.viewitems()])
+        tagset = ' '.join(['"{}={}"'.format(k, v) for k, v in tags.viewitems()])
+        if settings.ENABLE_PLACEMENT_OPTIONS in ['true', 'True', 'TRUE', '1']:
             unit.append({"section": "X-Fleet", "name": "MachineMetadata",
-                         "value": tagset})
+                         "value": tagset + ' "dataPlane=true"'})
         # post unit to fleet
         self._put_unit(name, {"desiredState": "loaded", "options": unit})
 
--- a/controller/templates/confd_settings.py
+++ b/controller/templates/confd_settings.py
@@ -24,6 +24,8 @@ SSH_PRIVATE_KEY = """{{ if exists "/deis/platform/sshPrivateKey" }}{{ getv "/dei
 # platform domain must be provided
 DEIS_DOMAIN = '{{ getv "/deis/platform/domain" }}'
 
+ENABLE_PLACEMENT_OPTIONS = """{{ if exists "/deis/platform/enablePlacementOptions" }}{{ getv "/deis/platform/enablePlacementOptions" }}{{ else }}false{{end}}"""
+
 # use the private registry module
 REGISTRY_MODULE = 'registry.private'
 REGISTRY_URL = '{{ getv "/deis/registry/protocol" }}://{{ getv "/deis/registry/host" }}:{{ getv "/deis/registry/port" }}'  # noqa
--- a/deisctl/backend/fleet/create.go
+++ b/deisctl/backend/fleet/create.go
@@ -3,6 +3,7 @@ package fleet
 import (
 	"fmt"
 	"io"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
@@ -96,7 +97,15 @@ func (c *FleetClient) createServiceUnit(component string, num int) (name string,
 	if err != nil {
 		return "", nil, err
 	}
-	uf, err = NewUnit(component, c.templatePaths)
+	decorateStr, err := c.configBackend.GetWithDefault("/deis/platform/enablePlacementOptions", "false")
+	if err != nil {
+		return "", nil, err
+	}
+	decorate, err := strconv.ParseBool(decorateStr)
+	if err != nil {
+		return "", nil, err
+	}
+	uf, err = NewUnit(component, c.templatePaths, decorate)
 	if err != nil {
 		return
 	}
--- a/deisctl/backend/fleet/create_test.go
+++ b/deisctl/backend/fleet/create_test.go
@@ -6,6 +6,9 @@ import (
 	"sync"
 	"testing"
 
+	"github.com/deis/deis/deisctl/config/model"
+	"github.com/deis/deis/deisctl/test/mock"
+
 	"github.com/coreos/fleet/schema"
 )
 
@@ -30,7 +33,9 @@ func TestCreate(t *testing.T) {
 	testFleetClient := stubFleetClient{testUnits: []*schema.Unit{}, unitsMutex: &sync.Mutex{},
 		unitStatesMutex: &sync.Mutex{}}
 
-	c := &FleetClient{templatePaths: []string{name}, Fleet: &testFleetClient}
+	testConfigBackend := mock.ConfigBackend{Expected: []*model.ConfigNode{{Key: "/deis/platform/enablePlacementOptions", Value: "true"}}}
+
+	c := &FleetClient{templatePaths: []string{name}, Fleet: &testFleetClient, configBackend: testConfigBackend}
 
 	var errOutput string
 	var wg sync.WaitGroup
--- a/deisctl/backend/fleet/fleet.go
+++ b/deisctl/backend/fleet/fleet.go
@@ -6,13 +6,16 @@ import (
 	"path"
 	"text/tabwriter"
 
+	"github.com/deis/deis/deisctl/config"
+
 	"github.com/coreos/fleet/client"
 	"github.com/coreos/fleet/machine"
 )
 
 // FleetClient used to wrap Fleet API calls
 type FleetClient struct {
-	Fleet client.API
+	Fleet         client.API
+	configBackend config.Backend
 
 	// used to cache MachineStates
 	machineStates map[string]*machine.MachineState
@@ -25,7 +28,7 @@ type FleetClient struct {
 
 // NewClient returns a client used to communicate with Fleet
 // using the Registry API
-func NewClient() (*FleetClient, error) {
+func NewClient(cb config.Backend) (*FleetClient, error) {
 	client, err := getRegistryClient()
 	if err != nil {
 		return nil, err
@@ -41,6 +44,6 @@ func NewClient() (*FleetClient, error) {
 	out := new(tabwriter.Writer)
 	out.Init(os.Stdout, 0, 8, 1, '\t', 0)
 
-	return &FleetClient{Fleet: client, templatePaths: templatePaths, runner: sshCommandRunner{},
+	return &FleetClient{Fleet: client, configBackend: cb, templatePaths: templatePaths, runner: sshCommandRunner{},
 		out: out, errWriter: os.Stderr}, nil
 }
--- a/deisctl/backend/fleet/fleet_test.go
+++ b/deisctl/backend/fleet/fleet_test.go
@@ -6,6 +6,9 @@ import (
 	"sync"
 	"testing"
 
+	"github.com/deis/deis/deisctl/config/model"
+	"github.com/deis/deis/deisctl/test/mock"
+
 	"github.com/coreos/fleet/machine"
 	"github.com/coreos/fleet/schema"
 )
@@ -176,8 +179,10 @@ func TestNewClient(t *testing.T) {
 	// set required flags
 	Flags.Endpoint = "http://127.0.0.1:4001"
 
+	testConfigBackend := mock.ConfigBackend{Expected: []*model.ConfigNode{}}
+
 	// instantiate client
-	_, err := NewClient()
+	_, err := NewClient(testConfigBackend)
 	if err != nil {
 		t.Fatal(err)
 	}
--- a/deisctl/backend/fleet/scale_test.go
+++ b/deisctl/backend/fleet/scale_test.go
@@ -7,6 +7,9 @@ import (
 	"sync"
 	"testing"
 
+	"github.com/deis/deis/deisctl/config/model"
+	"github.com/deis/deis/deisctl/test/mock"
+
 	"github.com/coreos/fleet/schema"
 )
 
@@ -30,7 +33,9 @@ func TestScaleUp(t *testing.T) {
 
 	testFleetClient := stubFleetClient{testUnits: testUnits, testUnitStates: []*schema.UnitState{}, unitsMutex: &sync.Mutex{}, unitStatesMutex: &sync.Mutex{}}
 
-	c := &FleetClient{templatePaths: []string{name}, Fleet: &testFleetClient}
+	testConfigBackend := mock.ConfigBackend{Expected: []*model.ConfigNode{{Key: "/deis/platform/enablePlacementOptions", Value: "true"}}}
+
+	c := &FleetClient{templatePaths: []string{name}, Fleet: &testFleetClient, configBackend: testConfigBackend}
 
 	var errOutput string
 	var wg sync.WaitGroup
--- a/deisctl/backend/fleet/unit.go
+++ b/deisctl/backend/fleet/unit.go
@@ -11,6 +11,21 @@ import (
 	"github.com/coreos/fleet/unit"
 )
 
+// path hierarchy for finding systemd service templates
+var templatePaths = []string{
+	os.Getenv("DEISCTL_UNITS"),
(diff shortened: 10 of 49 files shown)


mboersma:
I think this is a reasonable approach. I'm not sure everyone wants a large cluster-type architecture, but this PR just makes it possible with metadata.

I like that by default a newly provisioned cluster will act as it does now--all nodes host all planes.

My concerns are:
- this would need to be a non-breaking change somehow, so Deis has an upgrade path
- implementing a CoreOS Production + Central Services setup would require additional scripting and multiple versions of metadata in user-data...or maybe we still leave that up to the user?

krancour:
@mboersma, my goal was to make it _possible_ to control placement via metadata, but not necessarily to provide a means of doing that straight out of the box.  As long as the possibility exists and there is supporting documentation, I was thinking it was a win.

I agree it would be best if we could somehow make this a non-breaking change.  I'm not sure how that can be accomplished.  I am open to ideas.

aledbf:
@krancour @mboersma if this is merged https://github.com/coreos/fleet/pull/1077 it could be possible to avoid the manual modification of the metadata and apply a "default" in an upgrade

krancour:
@aledbf can you elaborate on what the upgrade process might then look like?

mboersma:
Basically we can't add this metadata to existing nodes, unless I'm mistaken. (And if I am, that's the fix then.) And `MachineMetadata` is a hard requirement that won't be met by old clusters, so nothing can be scheduled. So somehow we'd have to make the `MachineMetadata` sections optional or conditional.

krancour:
@mboersma I considered the possibility of a PR to Fleet to allow for more sophisticated `MachineMetadata` expressions.  I'd have liked to do something like this in the units here:

```
[X-Fleet]
MachineMetadata="controlPlane=true || (controlPlane=undef && dataPlane=undef && routerMesh=undef)"
```

Of course, that seems it would be a major departure from how `MachineMetadata` is parsed and evaluated currently.  A breaking change to Fleet seems a poor way to avoid a breaking change in Deis.  :-/  Back to the drawing board...

coreos/fleet#1077 helps, obviously, but as you noted, it may be quite some time before that hits the stable channel.

What would you think about some means of allowing operators to (easily) "decorate" units with the appropriate `MachineMetaData` for each unit before/during platform installation?

I'm trying to think through the least ugly way of doing that.

iancoffey:
FWIW @krancour, your last suggestion was the approach I was planning to take.

krancour:
@iancoffey, any thoughts on how that's best implemented.  How far did you get down that road?

iancoffey:
Not far, really. Now that the graceful stop is merged I can start doing so, if its helpful. Its a tricky can of worms. I was thinking about whether it makes sense for deisctl to accept options like controlplane_metadata / dataplane_metadata and if supplied, somehow parse the value into new unit templates, so they can match whatever is defined in the cloud config metadata.

Then you could setup whatever metadata you want, and just tell deisctl how to match them to the control/data plane. Maybe?

krancour:
@iancoffey

> Then you could setup whatever metadata you want, and just tell deisctl how to match them to the control/data plane. Maybe?

Is that a step too far, maybe, in terms of flexibility/complexity?  I was rolling with your previous idea of `controlPlane=true/false`, `dataPlane=true/false`, etc... that would make the decoration that needs to be applied to any unit pretty simple.  The flexibility in terms of what gets scheduled to where would have less to do with what an operator wants to do with the units.  Instead, it would ride entirely on judicious application of those "well-known" tags to the machines of a cluster at the time of provisioning (or dynamically later on if coreos/fleet#1077 is merged).

The biggest question in my mind is what is the most unobtrusive way to let an operator opt in to these "decorated" units.

What if operators building new clusters (with metadata already assigned appropriately to each machine) can opt in via `deisctl config` prior to installation?  Units would also have to somehow be refreshed and decorated at that point.  Making all of this seamless seams to me to be the difficult part.

iancoffey:
Yeah agree, allowing arbitrary metadata values is probably a step too far for a first pass. In that case, allowing the units the be refreshed, optionally with the control/data plane data, seems pretty reasonable to me as a place to start.

krancour:
Any thoughts about what the workflow looks like for someone opting into this?  Requiring someone who opts into this to set a config value _and_ refresh units (which would "regenerate" also with decoration?) seems asking for trouble in that it's a two-part process.  One step or the other might be too easily skipped or mangled.  Would it be totally crazy to add a single, separate command to `deisctl` that enables/disables this then refreshes/regenerates (decorates or undecorates) the units all in one fell swoop?

Or I hate to say it, but this almost brings me back to another rabbit hole I'd gone down once before-- the notion that the community would benefit from one or both of 1. a config / install wizard 2. a framework / go library for advanced operators to managed their own advanced deployment options.  The latter could actually be useful not only for Deis users, but for anyone using Fleet to orchestrate the launch of a complex, distributed technology stack.  It may lead to a community-driven ecosystem of tools that stand up Deis and/or other things in specific, useful configurations.

It might be crazy.  Not looking to go off on a tangent here, but it's certainly one way to solve this problem.

krancour:
I'm hoping someone has a simple idea that I'm overlooking.

iancoffey:
Hmmm - Im not seeing a reason why it would need to be a two part process - just regenerating the units with the intended metadata would suffice - thats what I do now.

Im personally :-1: on a wizard, and would probably implement what Ive described above anyway, even if one existed.

krancour:
@iancoffey, the "two part process," as I saw it was 1. config to opt into this advanced placement options feature 2. regenerate (decorate/undecorate) units based on no. 1.  So isn't that asking the operator to do two steps to accomplish one thing?

aledbf:
@krancour why not just add one additional parameter to `deisctl` to use a file with the metadata definition (could be a json file like in AWSTemplateFormatVersion.json)

This file must contain the metadata for each node (with this it could be possible to check the minimum requirement, like at least 3 nodes with `controlPlane=true` or `dataPlane=true` for instance) and with this generate a file in each node (`/run/systemd/system/fleet.service.d/20-cloudinit.conf`) just with the metadata:

```
Environment="FLEET_METADATA=controlPlane=true,dataPlane=true"
```

With this it could be possible to upgrade an existing deis cluster or create a new one.

krancour:
@aledbf, it's not clear to me how your last suggestion works.  `deisctl` doesn't actually play any role in assigning metadata to each machine.  Could you expand on the thought some more to give me some clarity?

aledbf:
> deisctl doesn't actually play any role in assigning metadata to each machine

That's correct. What I'm suggesting is that just using `deisctl` we can "customize" the CoreOS cluster (fleet metadata in this case) as part of the installation (without any manual file modification in any node).
`deisctl` just need to create and upload the `/run/systemd/system/fleet.service.d/20-cloudinit.conf` file

krancour:
What would happen if a machine reboots and cloud-init reruns with whatever cloud-config it already has and clobbers that metadata?

aledbf:
> What would happen if a machine reboots and cloud-init reruns with whatever cloud-config it already has and clobbers that metadata

well that's not possible :)
the files inside `/run/systemd/system/`are appended after the cloud-conf file generation and before any command execution. You will see something like this:

```
core@deis-01 ~ $ systemctl cat fleet
# /etc/systemd/system/fleet.service
[Unit]
Description=fleet
Requires=etcd.service
After=etcd.service

[Service]
ExecStart=/usr/bin/fleetd

# /run/systemd/system/fleet.service.d/20-cloudinit.conf
[Service]
Environment="FLEET_METADATA=controlPlane=true,dataPlane=true"
```

krancour:
Ok... so then what this really boils down to then is an existing "backdoor" means of doing what coreos/fleet#1077 is meant to address?  Yes?

I still have a bad feeling about this.  It requires operators to manage machine-specific configuration _separate_ from the templates (e.g. CloudFormation) that they actually use to provision their clusters.  How would this scale to account for use cases where an operator uses an autoscaling group to manage their data plane?

All this being said... if, like I said, this just amounts to a backdoor means of doing what coreos/fleet#1077 is meant to address (and we can leave the existing provisioning process alone), then maybe we could bake a little "update machine metadata" utility into deisctl, thereby providing an upgrade path for existing clusters.  Is that the idea?

And in such a case, then we could forget all the discussion about conditionally "decorating" units, because if any existing cluster can be updated with appropriate metadata (and this would be covered in upgrade notes), then all the units can reference those tags without issue, yes?

If I am following correctly, I like this very much.  I'd want to play with it a little in a live cluster first to convince myself it all works.  If I'm not following correctly... I apologize.

aledbf:
> Ok... so then what this really boils down to then is an existing "backdoor" means of doing what coreos/fleet#1077 is meant to address? Yes?

This is how azure customize the `etcd` timeout. It feels "wrong" but is just `systemd` :)

> I still have a bad feeling about this. It requires operators to manage machine-specific configuration separate from the templates

Well, what's the alternative? I mean, this should be provider agnostic. I don't think is a good idea to just do this kind of thing for AWS.

> And in such a case, then we could forget all the discussion about conditionally "decorating" units, because if any existing cluster can be updated with appropriate metadata (and this would be covered in upgrade notes), then all the units can reference those tags without issue, yes?

Yes.

iancoffey:
I think the @aledbf approach seems very promising! :+1:

I wasnt aware that we could manage fleet metadata like that.

krancour:
I wasn't really suggesting only doing it for AWS.  CloudFormation was just an example.  Today, the provisioning stuff for all providers uses some common cloud-config that's in contrib/coreos.  My point was that it would be bad to force users to maintain configuration / metadata _separate_ from that.  It adds a lot of complexity and probably won't work well at all for those auto-scaling their data plane.  (Going forward, how would new nodes in an existing cluster get the right metadata?)

Taking a giant step back here, if we want to enable metadata-driven placement options for each plane, such that each plane can optionally be isolated, there are fundamentally two ways to do that without introducing a breaking change:

1. Find a way to update metadata of existing nodes in an existing cluster as well as metadata for future nodes in an existing cluster (owing to the possibility that the data plane will scale; come to think of it, coreos/fleet#1077 might not fully address that case).  If this could be achieved somehow, this would allow a consistent and _static_ set of units to be used that can reference certain, well-known tags like `controlPlane=true` with the assumption such machines will exist somewhere in the cluster, regardless of whether that cluster is new or pre-existing.  I think there are a lot of gotchas that would need to be carefully worked out here.

2. Find a way to conditionally decorate units with appropriate `MachineMetadata` constraints _only_ for new clusters where an operator has explicitly chosen to isolate their planes and has metadata set appropriately to support that.

Do we agree these are the two options?  If so, we can work through the pros and cons of each, because I think there's a lot of complexity hiding in both options.

krancour:
@carmstrong did we agree yesterday that decorating the units (if an operator opts into isolating their planes) is the way to go?

carmstrong:
> did we agree yesterday that decorating the units (if an operator opts into isolating their planes) is the way to go?

Yep. The units themselves should stay the same, and we should add the metadata scheduling on-the-fly in deisctl.

krancour:
@carmstrong, any thoughts on _when_ to do with `deisctl`?  Just in time before installation of each component (if a certain etcd key is set to opt in)?

I had been thinking about the possibility of a separate sub-command that opts you in / performs the decoration.  That seems the most transparent in terms of how it functions, but it has two distinct disadvantages: needs to be re-run if units are refreshed and if `deisctl` is used from multiple machines, probably only the one that the initial install was done from has units on disk that have the decoration.  Problems could occur if maintenance were performed later from a different machine.

I guess I like the first option better... just looking for feedback if you see any glaring issues with it.

carmstrong:
@krancour I think we could have one command that opts-in, perhaps, or a flag to the platform install which would set the etcd key signifying that the cluster separation is desired. Then, any other users wouldn't run into issues because this etcd key would be set. Does that make sense?

krancour:
Yup.  So essentially the first option I gave, but have a dedicated sub-command or install flag that sets the etcd key instead of making an operator set it themselves?  Seems reasonable.

carmstrong:
> have a dedicated sub-command or install flag that sets the etcd key instead of making an operator set it themselves?

I think that makes sense. Maybe something like `deisctl install platform --split-planes` or something that sounds much better than that :) This would set a config (etcd) key and then run a normal install.

We only start caring about this setting once the platform has been installed, right? That's my motivation for thinking it should be a flag to `install platform`.

krancour:
> We only start caring about this setting once the platform has been installed, right?

Maybe.  I'm trying to think through whether operators may ever want to customize units post-decoration, pre-installation.

carmstrong:
> I'm trying to think through whether operators may ever want to customize units post-decoration, pre-installation.

:+1: Although this is really for people using stock units - for folks customizing their own units, we may just say that it's unsupported to use this option. They could do a lot of things there - reordering sections, already have an `X-Fleet` section (which means we can't blindly append that section to the bottom), etc.

Something to think about.

krancour:
Ok.  I agree with all of that.

> They could do a lot of things there - reordering sections, already have an X-Fleet section (which means we can't blindly append that section to the bottom), etc.

Some of the units in question already have that section anyway.  So that's a reminder for me to be careful about how I do this.  What do you think of a second file named `<unit>-decorator` that corresponds to each existing unit?  Each of those can just have the correct trailing lines in it-- which will or will not include `[X-Fleet]` in it, as well as the correct meta-data for the plane in question.

And now that I think about it... if someone were intent on customizing their units and still using this option, there isn't any reason they can't also customize those `<unit>-decorator` files.

krancour:
Just FYI, this is still on my radar.  Will implement the optional unit decoration as discusses as soon as time permits.

krancour:
This is still a work in progress.  It works, but I still need to update `deisctl refresh-units` to account for the decorators (which are separate files in a subdirectory) and still need to update documentation.

mboersma:
@krancour I added `[WIP]` to the subject to help reviewers know they can skip this for now.

krancour:
Note to myself: `controller/scheduler/fleet.py` requires further updates so that isolation of application units to the data plane only occurs if the etcd key `/deis/platform/enablePlacementOptions` is set to `true`.

krancour:
Don't worry about commit message format just yet-- I'll be doing some rebasing and squashing before done.

krancour:
@carmstrong I updated this PR this morning such that `deisctl refresh-units` also refreshes the decorators that we have discussed.  One interesting question this raises for me is how to deal with the possibility that someone runs `deisctl refresh-units` with the `-t` / `--tag` option to call out a specific tag or commit.  It's probably remote, but there's a possibility that someone can call out an older tag or a commit that does _not_ contain the decorators, resulting in 404s.

I wanted to get your take on how to deal with that, because I think this question had to have come up before-- anytime a _new_ component has been added, you may have run into the same dilemma, although I don't see any specific logic in `deisctl` meant to handle this possibility.

The way I'm currently choosing to deal with this is to gracefully handle and 404s that occur when refreshing the decorators.  Does that sound ok?  Or, alternatively, am I overthinking this and should I just not worry about this?  I tend to imagine that `-t` / `--tag` option is used rather infrequently and only by power users at that.

carmstrong:
> Does that sound ok? Or, alternatively, am I overthinking this and should I just not worry about this? I tend to imagine that -t / --tag option is used rather infrequently and only by power users at that.

Yes to all of those. I wouldn't worry too much about this - if someone is using a specific tag other than master, they should also aware that they'll need to make changes as appropriate.

/cc @technosophos

krancour:
@carmstrong sounds good.  Thanks for the sanity check!

mboersma:
Needs rebase, no rush @krancour. :ring:

krancour:
Tests are failing because some of the changes to `deisctl` try to look up a value in etcd prior to creating any new service unit...

(The reason for this is to make a decision to "decorate" or not decorate each component's service unit with additional metadata if and only if the operator has opted into isolating their planes.)

... of course, etcd isn't available during unit tests.  I could mock it out in the manner that was done for the tests in the `deisctl/config` package, but since the code is more functional than OO, it's looking tricky to "inject" that mocked out etcd client into the fleet backend without seriously polluting some interfaces.

I'm open to any ideas anyone has on how to fix these tests without a major refactor.

(Long term, I do wonder if `deisctl` needs some love, but I'll open a separate proposal for that.)

krancour:
Took a bit of effort to grok some new go idioms, but I think I managed to get the testing all sorted out.

Note to self that the controller still needs to be updated to _not_ apply the isolation / placement metadata when the operator has not opted into this feature.

Docs still need updating.

krancour:
Controller should be all set now.  Only docs left to do.

krancour:
This will need another rebase after #3975 is merged... need to add the "decorators" for the new components that k8s introduces.

krancour:
Added decorators for k8s components.  This may need some more manual testing however, because things seem not to be working when I isolate the control and data plane components from each other while also using k8s.

krancour:
Isolating the planes works well with fleet and mesos.  It's failing with k8s and swarm.  Looking into it some more.

krancour:
Swarm issues seem to be on account of #4159

krancour:
Works fine with swarm after #4168

krancour:
Proudly works with all four schedulers now. :fireworks:

carmstrong:
Code LGTM. I like this implementation. Nice work, @krancour. :+1:

krancour:
@carmstrong thanks!  Think the docs are adequate?

carmstrong:
> Think the docs are adequate?

They look good to me!

laurrentt:
Just in time for 1.9 ? :pray:

krancour:
@laurrentt sorry... it's not going to be included in 1.9.0.

mboersma:
Code LGTM.

PR #3720, opened 2015-05-19
docs(config-application): add a config:push / pull example

### LINKED ISSUE #3697: Add `config:push` and `config:pull` examples to docs
http://docs.deis.io/en/latest/using_deis/config-application/ should have examples of how to use `deis config:push` and `deis config:pull`. So sayeth the BDFL. :godmode:

mboersma:
The docs already give the `deis help config` usage string, which has a tidy summary of available options, so this is just a usage example rather than a complete reference.

Closes #3697.

### DIFF
--- a/docs/using_deis/config-application.rst
+++ b/docs/using_deis/config-application.rst
@@ -20,10 +20,31 @@ Use ``deis config`` to modify environment variables for a deployed application.
     config:set         set environment variables for an app
     config:unset       unset environment variables for an app
     config:pull        extract environment variables to .env
+    config:push        set environment variables from .env
 
-    Use `deis help [command]` to learn more
+    Use `deis help [command]` to learn more.
 
 When config is changed, a new release is created and deployed automatically.
+
+You can set multiple environment variables with one ``deis config:set`` command,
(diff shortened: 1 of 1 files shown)


bacongobbler:
:ship:

carmstrong:
Code LGTM.

kalbasit:
I think you should also mention the `-p` option. So in my project I have an encrypted `deis/apps/app-name.conf` file which is the settings of the app in deis and I have a make target that does `deis config:push -p deis/apps/app-name.conf -a app-name` which I find pretty amazing at keeping the config in version control.

bacongobbler:
@kalbasit I think the demo looks fine to me. Those flags are still exposed in `deis help config:push` and this is just supposed to demonstrate `config:push` with default arguments.

mboersma:
> I think you should also mention the -p option.

Trying to strike a balance between brief and informative here... I think putting `-p` in is a bit too detailed, and if the user has been reading the doc up to here they should be willing to try `deis help config:push`.

kalbasit:
I see, then LGTM in that case :)

PR #3729, opened 2015-05-20
[WIP] ref(publisher): tech debt removal

bacongobbler:
A few changes in this PR which are fairly minor but are QoL improvements for maintainers as well as operators:

 - switch to logrus for logging, allowing users to set --log-level to more than just 2 levels
 - updated flag docstrings for more clarity on their impact and usage in publisher
 - log output when publisher is booting (before, publisher would not log anything on stdout)
 - gracefully kill the server on SIGINT/SIGTERM, same as logger
 - log errors on ERROR rather than INFO

I'd be happy to pull out certain pieces for easier reviewing to get it merged/rejected; just say the word! The commits stand alone for the most part, so I just kept it as one PR to prevent having to rebase 3-4 separate PRs, but if it makes life easier then please feel free. Nitpick away!

No milestone has been attached to this as it's doesn't fix anything other than cleanup so no rush on this one

### DIFF
--- a/publisher/main.go
+++ b/publisher/main.go
@@ -48,15 +48,15 @@ func main() {
 		}
 	}()
 
-	server, err := server.New(dockerAddr, etcdAddr, host)
+	server, err := server.New(dockerAddr, etcdAddr, host, publishTTL)
 	if err != nil {
 		log.Fatalf("failed to start (%s)", err)
 	}
 
-	go server.Listen(publishTTL)
+	go server.Listen()
 
 	for {
-		go server.Poll(publishTTL)
+		go server.Poll()
 		time.Sleep(interval)
 	}
 }
--- a/publisher/main.go
+++ b/publisher/main.go
@@ -25,13 +25,13 @@ var (
 )
 
 func init() {
-	flag.StringVar(&bindAddr, "bind-addr", "localhost:6060", "The IP address and port to bind to")
-	flag.DurationVar(&refreshDuration, "refresh-duration", 10 * time.Second, "The time to wait between etcd refreshes")
-	flag.DurationVar(&etcdTTL, "etcd-ttl", refreshDuration * 2, "The TTL for all of the keys in etcd")
-	flag.StringVar(&host, "host", "127.0.0.1", "The host where the publisher is running")
-	flag.StringVar(&dockerAddr, "docker-addr", "unix:///var/run/docker.sock", "The bind address to the docker API")
-	flag.StringVar(&etcdAddr, "etcd-addr", "http://127.0.0.1:4001", "The bind address to the etcd host")
-	flag.StringVar(&logLevel, "log-level", "error", "Acceptable values: error, debug")
+	flag.StringVar(&bindAddr, "bind-addr", "localhost:6060", "address to listen for incoming HTTP requests")
+	flag.DurationVar(&refreshDuration, "interval", 10 * time.Second, "backend polling interval")
+	flag.DurationVar(&etcdTTL, "publish-ttl", refreshDuration * 2, "backend TTL when publishing keys")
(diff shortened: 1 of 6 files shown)


publisher/rootfs/Dockerfile
```diff
@@ -4,7 +4,8 @@ FROM alpine:3.1
 # profiling information without any additional package installation.
 RUN apk add --update-cache curl && rm -rf /var/cache/apk/*
 
-ADD bin/publisher /usr/local/bin/publisher
-ENTRYPOINT ["/usr/local/bin/publisher"]
+COPY . /
```
mboersma: Yes! `COPY` is usually preferable to `ADD`, we should be making this change everywhere as we revisit Dockerfiles.

mboersma:
This looks good at first glance.

> gracefully kill the server on SIGINT/SIGTERM, same as logger

Should this also make the systemd unit start/stop changes in #3703? Or is that going to be a separate, pervasive PR? (I like those changes.)

bacongobbler:
that PR is going to be totally separate. This is just a copy-over from https://github.com/deis/deis/blob/master/logger/main.go#L71-L73 which is executed when those [signals are captured](https://github.com/deis/deis/blob/master/logger/main.go#L54)

carmstrong:
Code LGTM.

carmstrong:
I'll kick off another test.

carmstrong:
This failed in the same place the past two runs (404 on app request), so I'm thinking there is an issue with publisher failing to publish the app.

publisher/main.go
```diff
@@ -2,56 +2,86 @@ package main
...
+		}
 	}()
 
+	server, err := server.New(dockerAddr, etcdAddr, host, publishTTL)
+	if err != nil {
+		log.Fatalf("failed to start (%s)", err)
+	}
+
+	// run Publish() once at boot to publish existing containers
+	go server.Publish()
```
kalbasit: I think you should call `server.Publish()` through this goroutine, otherwise this goroutine might be called at the same time as the goroutine started on line 72 by the ticker channel. Remember in Go, goroutines are not called until the current one is blocked by something (a channel, a resource etc..)

publisher/main.go
```diff
@@ -2,56 +2,86 @@ package main
...
+	publishTTL time.Duration
+	host       string
+	dockerAddr string
+	etcdAddr   string
+	log        = logrus.New()
+	logLevel   string
 )
 
+func init() {
+	flag.StringVar(&bindAddr, "bind-addr", "localhost:6060", "address to listen for incoming HTTP requests")
```
kalbasit: I don't really like the `init()` here, it just adds more burden on the binary boot. Why not simple have them all vars? like

```
var (
  bindAddr = flag.String("bind-addr", "localhost:6060", "address to listen for incoming HTTP requests")
)
```

publisher/server/publisher.go
```diff
@@ -22,9 +22,8 @@ const (
 type Server struct {
 	DockerClient *docker.Client
```
kalbasit: Perhaps you should also switch this to an interface. So we can make a mock docker client in tests..

publisher/main.go
bacongobbler: that looks better. :+1:

publisher/server/publisher.go
bacongobbler: good point, but that's unrelated to the PR. Definitely agree that we should make or use some form of interface here.

kalbasit: :+1: it can be in a separate PR. I think we should make a PR that touches all component to have a common etcd/docker interface.

bacongobbler:
there are tons of changes in here that are hard to review. I'm gonna split this out into separate PRs to make it easier on everyone's eyes. :eyes:

PR #3730, opened 2015-05-20
fix(deisctl): hide random "deis-" services from "deisctl list"

### LINKED ISSUE #3232: bug(deisctl): list should not show apps with the pattern name ^deis-*
If exists an app with the name `deis-demo` it will appear in the output of `deisctl list`

mboersma:
Closes #3232.

### DIFF
--- a/deisctl/backend/fleet/list_units.go
+++ b/deisctl/backend/fleet/list_units.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/coreos/fleet/machine"
 	"github.com/coreos/fleet/schema"
+	"github.com/deis/deis/deisctl/units"
 )
 
 // initialize tabwriter on stdout
@@ -82,8 +83,11 @@ func (c *FleetClient) ListUnits() (err error) {
 	}
 
 	for _, us := range unitStates {
-		if strings.HasPrefix(us.Name, "deis-") {
-			states = append(states, us)
+		for _, prefix := range units.Names {
+			if strings.HasPrefix(us.Name, prefix) {
+				states = append(states, us)
+				break
+			}
 		}
 	}
 	printUnits(states)
--- a/deisctl/Makefile
+++ b/deisctl/Makefile
@@ -4,7 +4,7 @@ include ../includes.mk
 repo_path = github.com/deis/deis/deisctl
 
 GO_FILES = deisctl.go deisctl_test.go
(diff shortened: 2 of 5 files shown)


deisctl/backend/fleet/list_units.go
```diff
@@ -82,8 +104,11 @@ func (c *FleetClient) ListUnits() (err error) {
 	}
 
 	for _, us := range unitStates {
-		if strings.HasPrefix(us.Name, "deis-") {
-			states = append(states, us)
+		for _, prefix := range unitPrefixes {
```
bacongobbler: sad that there's no `in` clause for Go. :(

bacongobbler:
LGTM

deisctl/backend/fleet/list_units.go
mboersma: I thought about making the data structure a Go `map` with empty values just so I wouldn't have to add another loop, but that sounded wrong. I'm assuming this code is more or less optimal after the compiler is done with it. :)

bacongobbler: The only way I could think we could optimize it would be to use `for i := range unitPrefixes { fmt.Println(unitPrefixes[i]) }` which would only allocate the iterator, but then we'd be using it to access the value anyways... meh. The Go std lib uses both ways as well so it's tomato vs tomatoe. I like this approach better, anyways :+1:

deisctl/backend/fleet/list_units.go
```diff
@@ -72,6 +72,28 @@ var (
 	}
 )
 
+// unitPrefixes are compared to unit names, so "deisctl list" doesn't show "deis-demo.service",
+// for example. Update this list when adding a new Deis unit file.
+var unitPrefixes = []string{
```
carmstrong: Could we pull this into a global variable? It'd be nice to have one master list we could use here and in refreshUnits.

mboersma: I did look at using a single data structure for exactly that purpose, but the Go packages that would consume it don't have any common imports. I'll take another look at it tomorrow--it would be lame to maintain essentially the same data in both places.

mboersma:
I updated this to move the unit file names to a common package. I put `units.go` into the units/ dir, which holds the actual unit files but wasn't a go package before.

carmstrong:
Love the change to use the new units! Code LGTM.

PR #3733, opened 2015-05-21
docs(configure-dns): drop references to local.deisapp.com

mboersma:
1-node clusters are not supported, but the "Configure DNS" docs still referred to them, and several places in the codebase still referred to "local.deisapp.com" instead of "local3".


carmstrong:
Good finds! Code LGTM.

bacongobbler:
very nice

LGTM

PR #3734, opened 2015-05-21
fix(controller): merge structure with new structure

### LINKED ISSUE #3494: Do not automatically scale up to 1 after a deploy
If a user has 0 containers of an app running and publishes a new release, we should confirm that Deis doesn't automatically scale up to 1. 

**Example:**
I have an app that defines two proctypes, **web** and **foo**. I keep my **web** scaled to 0, and only want **foo** to be running most of the time. If I deploy my app again, will Deis scale **web** back to 1?

bacongobbler:
If the old structure contained values like {'web': 0}, the container
count would be zero, so the resulting vals structure would be an empty
dictionary. By taking the "requested" structure and updating it with
the actual container queryset, values like {'web': 0} will stay the
same. From that result, the structure will not be considered in an
["initial" state][1], which means that a new deploy will not cause the
containers to be accidentally scaled back up to one.

[1]: https://github.com/deis/deis/blob/ebc3df479a672acb4a81525ba1edd0fb4d70a6a9/controller/api/models.py#L605

closes #3494

### DIFF
--- a/controller/api/models.py
+++ b/controller/api/models.py
@@ -275,7 +275,9 @@ class App(UuidAuditedModel):
         # save new structure to the database
         vals = self.container_set.exclude(type='run').values(
             'type').annotate(Count('pk')).order_by()
-        self.structure = {v['type']: v['pk__count'] for v in vals}
+        new_structure = structure.copy()
+        new_structure.update({v['type']: v['pk__count'] for v in vals})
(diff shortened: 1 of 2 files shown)


mboersma:
Well done. Code LGTM.

carmstrong:
I love tests. Code LGTM.

PR #3736, opened 2015-05-21
ref(controller): remove initial from .deploy()

bacongobbler:
initial is only equal to the time when the structure has no values, so
it makes no sense to check for initial in any other case.

### DIFF
--- a/controller/api/models.py
+++ b/controller/api/models.py
@@ -327,7 +327,7 @@ class App(UuidAuditedModel):
             log_event(self, err, logging.ERROR)
             raise RuntimeError(err)
 
-    def deploy(self, user, release, initial=False):
+    def deploy(self, user, release):
         """Deploy a new release to this application"""
         existing = self.container_set.exclude(type='run')
         new = []
@@ -343,7 +343,7 @@ class App(UuidAuditedModel):
             self._destroy_containers(existing)
 
         # perform default scaling if necessary
-        if initial:
+        if self.structure == {} and release.build is not None:
             self._default_scale(user, release)
 
     def _default_scale(self, user, release):
@@ -602,9 +602,8 @@ class Build(UuidAuditedModel):
                                          build=self,
                                          config=latest_release.config,
                                          source_version=source_version)
-        initial = True if self.app.structure == {} else False
         try:
-            self.app.deploy(user, new_release, initial=initial)
+            self.app.deploy(user, new_release)
             return new_release
         except RuntimeError:
             new_release.delete()


carmstrong:
I like the cleanup, but worried there may be a hidden regression here. Do we have adequate test coverage for these changes?

bacongobbler:
I found that there was a regression with releases:rollback, but the tests caught that and line 333 was born. I guess I could double-check all calls to `app.deploy()` and see if we should be catching that error in other places, but otherwise we should be good.

bacongobbler:
On second thought, it might be a logical error in my code. Let me check that.

bacongobbler:
much better.

carmstrong:
> much better.

What was the bug? Did tests cover it?

bacongobbler:
The tests covered it, but basically what was occurring was the following:

```
><> venv/bin/python manage.py test api.ConfigTest.test_config_set_unicode
WARNING Cannot synchronize with etcd cluster
Creating test database for alias 'default'...
E
======================================================================
ERROR: test_config_set_unicode (api.tests.test_config.ConfigTest)
----------------------------------------------------------------------
  File "/home/bacongobbler/go/src/github.com/deis/deis/controller/venv/local/lib/python2.7/site-packages/mock.py", line 1201, in patched
    return func(*args, **keywargs)
  File "/home/bacongobbler/go/src/github.com/deis/deis/controller/api/tests/test_config.py", line 172, in test_config_set_unicode
    HTTP_AUTHORIZATION='token {}'.format(self.token))
  File "/home/bacongobbler/go/src/github.com/deis/deis/controller/venv/local/lib/python2.7/site-packages/django/test/client.py", line 483, in post
    response = super(Client, self).post(path, data=data, content_type=content_type, **extra)
  File "/home/bacongobbler/go/src/github.com/deis/deis/controller/venv/local/lib/python2.7/site-packages/django/test/client.py", line 302, in post
    return self.request(**r)
  File "/home/bacongobbler/go/src/github.com/deis/deis/controller/venv/local/lib/python2.7/site-packages/django/test/client.py", line 444, in request
    six.reraise(*exc_info)
  File "/home/bacongobbler/go/src/github.com/deis/deis/controller/venv/local/lib/python2.7/site-packages/django/core/handlers/base.py", line 112, in get_response
  File "/home/bacongobbler/go/src/github.com/deis/deis/controller/venv/local/lib/python2.7/site-packages/django/views/decorators/csrf.py", line 57, in wrapped_view
  File "/home/bacongobbler/go/src/github.com/deis/deis/controller/venv/local/lib/python2.7/site-packages/rest_framework/viewsets.py", line 85, in view
  File "/home/bacongobbler/go/src/github.com/deis/deis/controller/venv/local/lib/python2.7/site-packages/rest_framework/views.py", line 407, in dispatch
  File "/home/bacongobbler/go/src/github.com/deis/deis/controller/venv/local/lib/python2.7/site-packages/rest_framework/views.py", line 404, in dispatch
  File "/home/bacongobbler/go/src/github.com/deis/deis/controller/api/views.py", line 84, in create
    return super(AppResourceViewSet, self).create(request, **kwargs)
  File "/home/bacongobbler/go/src/github.com/deis/deis/controller/api/views.py", line 61, in create
    return super(BaseDeisViewSet, self).create(request, *args, **kwargs)
  File "/home/bacongobbler/go/src/github.com/deis/deis/controller/venv/local/lib/python2.7/site-packages/rest_framework/mixins.py", line 21, in create
    self.perform_create(serializer)
  File "/home/bacongobbler/go/src/github.com/deis/deis/controller/api/viewsets.py", line 21, in perform_create
    self.post_save(obj)
  File "/home/bacongobbler/go/src/github.com/deis/deis/controller/api/views.py", line 193, in post_save
    config.app.deploy(self.request.user, self.release)
  File "/home/bacongobbler/go/src/github.com/deis/deis/controller/api/models.py", line 347, in deploy
    self._default_scale(user, release)
  File "/home/bacongobbler/go/src/github.com/deis/deis/controller/api/models.py", line 352, in _default_scale
    if not release.build.sha:
AttributeError: 'NoneType' object has no attribute 'sha'

----------------------------------------------------------------------
Ran 1 test in 0.079s

FAILED (errors=1)
Destroying test database for alias 'default'...
```

By default, initial was set to `False`. Since we deploy [when configuration changes](https://github.com/deis/deis/blob/master/controller/api/views.py#L193), we'd go through the steps to deploy. However, at this point in the tests there's no Build present for the release (i.e. `deis create` into `deis config:set`). `_default_scale()` would be called because no structure existed for this app yet (no scaling operation was done at this time) which would then fail because we're calling `.sha` on a NoneType build object. Checking for both a non-null structure and a non-null build is equivalent to the behaviour it was expecting.

TL;DR our API test coverage is pretty extensive (and awesome).

controller/api/models.py
```diff
@@ -343,7 +343,7 @@ def deploy(self, user, release, initial=False):
             self._destroy_containers(existing)
 
         # perform default scaling if necessary
-        if initial:
+        if self.structure == {} and release.build is not None:
```
carmstrong: Looks like `and release.build is not None` is additional logic that we didn't have before - is this intentional?

bacongobbler: This is intentional so that users who do a `deis config:set` before they do an intial deploy won't blow up on them. If you remove that call, you'll end up with the errors listed above. This is because `_default_scale` requires that the release has a non-null build object.

carmstrong:
Code LGTM with @mboersma's eyes as well.

mboersma:
Code LGTM.

bacongobbler:
We talked about this over last week that the call where the config view calling app.deploy was weird and something we'd want to address by moving it back to the Release object. Do you guys want me to address that here in this PR, or is a followup ok?

mboersma:
A separate PR is totes ok.

PR #3740, opened 2015-05-22
feat(deisctl): add config rm

### LINKED ISSUE #3515: feat(deisctl): deisctl config <component> rm
Some users have been disabling features. The only workaround today is to SSH into one of the CoreOS hosts and run `etcdctl rm` on the specific key. We should map a `deisctl config <component> rm` command to do the equivalent.

Joshua-Anderson:
I went with `rm` over `unset` because `etcdctl` uses `rm`. I didn't add any test for this, as there doesn't seem to be any tests for this part of deisctl. If I overlooked where they were let me know. The tests in `config_test` seemed mostly to test the special behavior of keys and not the etcd integration.

Fixes #3515

### DIFF
--- a/deisctl/config/etcd.go
+++ b/deisctl/config/etcd.go
@@ -42,6 +42,11 @@ func (c *etcdClient) Get(key string) (string, error) {
 	return resp.Node.Value, nil
 }
 
+func (c *etcdClient) Delete(key string) error {
+	_, err := c.etcd.Delete(key, false)
+	return err
+}
+
 func (c *etcdClient) Set(key string, value string) (string, error) {
 	resp, err := c.etcd.Set(key, value, 0) // don't use TTLs
 	if err != nil {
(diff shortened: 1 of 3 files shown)


carmstrong:
> I went with rm over unset because etcdctl uses rm.

:+1: 

This is awesome. I was working with a customer the other day and had to jump onto the host and use `etcdctl` to remove a key. Awesome work, @Joshua-Anderson!

carmstrong:
Code LGTM.

deisctl/config/etcd.go
```diff
@@ -42,6 +42,14 @@ func (c *etcdClient) Get(key string) (string, error) {
 	return resp.Node.Value, nil
 }
 
+func (c *etcdClient) Delete(key string) error {
+	_, err := c.etcd.Delete(key, false)
+	if err != nil {
```
bacongobbler: we should remove all this `if` logic and just return err, since we return it in both cases (nil and non-nil)

bacongobbler:
code LGTM barring my one comment

Joshua-Anderson:
@bacongobbler Done!

carmstrong:
Thanks @Joshua-Anderson!

PR #3745, opened 2015-05-26
feat(client): add --username flag to auth:passwd

### LINKED ISSUE #3292: Request: `deis auth:passwd` takes --user argument for admins to change user passwords
Changing your own password with `deis auth:passwd` is easy.
Changing the password for a user who has forgot their password is not.

http://stackoverflow.com/questions/1873806/changing-password-in-django

```
$ docker exec -ti deis-controller ./manage.py shell
WARNING Cannot synchronize with etcd cluster
Python 2.7.6 (default, Mar 22 2014, 22:59:56)
[GCC 4.8.2] on linux2
Type "help", "copyright", "credits" or "license" for more information.
(InteractiveConsole)
>>> from django.contrib.auth.models import User
>>> u = User.objects.get(username__exact='theirusername')
>>> u.set_password('iforgotmypassword')
>>> u.save()
```

There really should be an easier way.

bacongobbler:
closes #3292

### DIFF
--- a/client/deis.py
+++ b/client/deis.py
@@ -895,9 +895,11 @@ Make sure that the Controller URI is correct and the server is running.
 
         Options:
           --password=<password>
-            provide the current password for the account.
+            the current password for the account.
           --new-password=<new-password>
-            provide a new password for the account.
+            the new password for the account.
+          --username=<username>
+            the account's username.
         """
         if not self._settings.get('token'):
             raise EnvironmentError(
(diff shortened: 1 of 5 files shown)


client/deis.py
```diff
@@ -912,7 +914,13 @@ def auth_passwd(self, args):
             if new_password != confirm:
                 self._logger.error('Password mismatch, not changing.')
                 sys.exit(1)
-        payload = {'password': password, 'new_password': new_password}
+        payload = {
+            'password': password,
+            'new_password': new_password,
+            'username': self._settings['username'],
```
mboersma: You could save two lines of code with:
```python
'username': args.get('--username', self._settings['username']),
```

mboersma:
Code LGTM.

client/deis.py
bacongobbler: that looks much better!

carmstrong:
Code LGTM.

PR #3746, opened 2015-05-26
ref(client): clarify usage of ps:restart

### LINKED ISSUE #3639: clarify usage of ps:restart
`deis help ps` reads as "restart process types (e.g. web, worker)" which implies that you can only restart a specific process type. The documentation at `deis help ps:restart` also shows "Restarts an application's processes by type" which is untrue. It allows you to restart all containers, a specific process type, or an individual container:

```
deis ps:restart # restarts all processes
deis ps:restart web # restarts all web processes
deis ps:restart web.1 # restarts web.1
```

We should clarify this in the client help.

bacongobbler:
closes #3639

### DIFF
--- a/client/deis.py
+++ b/client/deis.py
@@ -1711,7 +1711,7 @@ Make sure that the Controller URI is correct and the server is running.
         Valid commands for processes:
 
         ps:list        list application processes
-        ps:restart     restart process types (e.g. web, worker)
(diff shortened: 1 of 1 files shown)


mboersma:
LGTM.

carmstrong:
LGTM.

PR #3747, opened 2015-05-26
fix(client): require the tag for builds:create

### LINKED ISSUE #2830: Docker image pull hangs if latest tag does not exist
Steps to reproduce:
 - deis create foo
 - deis pull giorgos/masterfirefoxos

waits forever while it should return saying that no 'latest' tag exists and that user should define a tag to pull.

bacongobbler:
closes #2830

### DIFF
--- a/client/deis.py
+++ b/client/deis.py
@@ -296,6 +296,22 @@ def encode(obj):
         return obj
 
 
+def parse_repository_tag(repo):
+    """Parses a given docker image and splits the tag from the repository.
+
+    See github.com/docker/docker-py#be73aaf, lines 188-197
+    """
+    column_index = repo.rfind(':')
+    if column_index < 0:
+        return repo, None
+    tag = repo[column_index + 1:]
+    slash_index = tag.find('/')
+    if slash_index < 0:
+        return repo[:column_index], tag
+
+    return repo, None
+
+
 def readable_datetime(datetime_str):
     """
     Return a human-readable datetime string from an ECMA-262 (JavaScript)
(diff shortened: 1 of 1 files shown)


carmstrong:
Should we also enforce this at the backend/API level, or is not worth bothering?

bacongobbler:
I recall us having this discussion last week and we decided that it was fine to just patch it in the client for now

carmstrong:
> I recall us having this discussion last week and we decided that it was fine to just patch it in the client for now

That's fine, that's why I was asking :)

carmstrong:
Code LGTM.

mboersma:
Code LGTM.

bacongobbler:
One thing I haven't tested was `deis pull ip-addr:5000/deis/example-go`, which would pass. I think it's high time we add client unit tests. Would that be acceptable?

carmstrong:
> One thing I haven't tested was deis pull ip-addr:5000/deis/example-go, which would pass.

Good point.

> think it's high time we add client unit tests. Would that be acceptable?

Rewriting the client in Go was one thing @Joshua-Anderson mentioned he'd be interested in taking a stab at once he starts :) I think that'd be a good time to write unit tests, as well.

bacongobbler:
okay, I've just imported the `parse_repository_tag` method from docker-py so we don't have to install another library. The logic is simple enough, anyhoo. I'll punt on client tests :sheep:

carmstrong:
Amended code LGTM.

mboersma:
LGTM++.

PR #3749, opened 2015-05-26
feat(router): allow customizing controller subdomain

### LINKED ISSUE #3488: Feature request: customize controller host name
Is there a config option to set the deis controller url? (deis.example.com)
if not there should be. I am runnig in a corporate environment with strict DNS rules

bacongobbler:
closes #3488

### DIFF
--- a/controller/api/models.py
+++ b/controller/api/models.py
@@ -92,7 +92,7 @@ def validate_app_structure(value):
 
 def validate_reserved_names(value):
     """A value cannot use some reserved names."""
-    if value in ['deis']:
+    if value in settings.DEIS_RESERVED_NAMES:
         raise ValidationError('{} is a reserved name.'.format(value))
 
 
--- a/controller/api/tests/test_app.py
+++ b/controller/api/tests/test_app.py
@@ -156,10 +156,6 @@ class AppTest(TestCase):
                                     HTTP_AUTHORIZATION='token {}'.format(self.token))
         self.assertContains(response, 'App IDs can only contain [a-z0-9-]', status_code=400)
         url = '/v1/apps'
-        body = {'id': 'deis'}
-        response = self.client.post(url, json.dumps(body), content_type='application/json',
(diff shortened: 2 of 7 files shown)


carmstrong:
Is `validate_reserved_names` covered by unit tests for various names?

carmstrong:
We need to add this setting to "Customizing Controller".

carmstrong:
There are also numerous places where we refer to "deis.<domain>". While I think updating all of these to include a disclaimer for a custom subdomain, we should evaluate all uses of it.

bacongobbler:
> Is validate_reserved_names covered by unit tests for various names?

Yes: https://github.com/deis/deis/blob/master/controller/api/tests/test_app.py#L158-L162

> we should evaluate all uses of it.

I'm not quite sure on what "it" is supposed to represent. Do you mean that we add disclaimers where we say `deis.example.com`?

carmstrong:
> I can add tests for this.

:+1:

> I'm not quite sure on what "it" is supposed to represent.

Sorry - looks like GitHub stripped out my `deis.<domain>`. I just want to look at places where we say `deis.<domain>` in the docs and see if we should mention the fact that the host is customizable. 

I just did a quick pass, and if we call this out anywhere, it looks like http://docs.deis.io/en/latest/using_deis/register-user/#register-with-a-controller is the place. Maybe add a note there saying that while deis. is the default, this is a customizable setting, and link to Customizing Controller?

bacongobbler:
Awesome. Thanks for clarifying. I'll update the docs and check for anywhere else :)

carmstrong:
> I'll update the docs and check for anywhere else :)

Cool. We probably don't need it anywhere we're just using it as an example command, but if we're talking about the controller and how to talk to it, we should probably mention the fact that it's customizable. "Register User" is the only place I found, but it's worth another pass. :+1:

bacongobbler:
done!

carmstrong:
Code LGTM.

mboersma:
Code LGTM.

PR #3752, opened 2015-05-27
Clean up old git repos and images on destroy

### LINKED ISSUE #3249: Deis apps:destroy does not remove the git repo with the same name
if a new application is created with the same name, everything is already up to date on git push

mboersma:
This fixes two problems in Deis master:
- `deis destroy` leaves a git repo dropping in builder at `/home/git/appname.git`
- `deis scale cmd=0` deletes the `/deis/services/appname` dir after 20 seconds, but that causes the (fixed) `check-repos` script to purge its git repo prematurely

My assumption is that the templating for `check-repos` worked in earlier versions of `confd` that were more specific to `etcd` behavior. This PR switches to `etcd watch` in a loop rather than rely on `confd` templating to launch the script. It also reverts #3348 so that `/deis/services/appname` dirs don't have a TTL, since `deis destroy` deletes them reliably in my testing.

Closes #3249, refs #3348.

### DIFF
--- a/builder/rootfs/bin/boot
+++ b/builder/rootfs/bin/boot
@@ -77,10 +77,14 @@ docker build -t deis/slugrunner /usr/local/src/slugrunner/
 /usr/sbin/sshd -D -e &
 SSHD_PID=$!
 
+# start a cleanup script to remote old repositories and images
+/bin/cleanup &
+CLEANUP_PID=$!
+
 # smart shutdown on SIGINT and SIGTERM
 function on_exit() {
-	kill -TERM $DOCKER_PID $SSHD_PID
-	wait $DOCKER_PID $SSHD_PID 2>/dev/null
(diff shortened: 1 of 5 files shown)


bacongobbler:
nice cleanup to run on an actual change in /deis/services by utilizing etcd watch rather than a confd reload. Code LGTM!

mboersma:
> by utilizing etcd watch rather than a confd reload

It adds a bit of complexity, but what it gains apparently is notification when a directory itself under `/deis/services/` is added or removed. Using just `confd` polling or watching, it only cares about keys.

I also tried `etcd exec-watch` which would seem to be perfect for this case, but it was unreliable. It would exec the command once or twice and then just stall, so I went with this approach.

carmstrong:
Remember that we did see performance issues when we switched everything over to using watch, but I think one judicious use of it is fine. And this is an appropriate place to use it.

Code LGTM.

Issue #2906, opened 2015-01-15
Component status of "running" doesn't mean the component is functioning properly


carmstrong:
This commonly trips people up, especially when investigating a component and seeing that its journal is filled with `ERROR Config check failed: exit status 1` even thought the service is technically "running".

#2701 may be an opportunity to fix things up - perhaps we should have the container exit whenever a config check fails, so it's apparent that it's not functioning properly?

carmstrong:
@bacongobbler Think we can tackle this as part of the Go-init refactor?

bacongobbler:
Hopefully, yes. With `/bin/boot` gone, PID 1 will be reserved for the actual process, which means that the process should die when these errors are coughed up.

bacongobbler:
Alternatively, if we just wanted to fix the `ERROR Config check failed: exit status 1` error we can just tell confd to write to a PIDfile and capture/trap that as most other users do.

bacongobbler:
This looks more like a limitation of a delayed boot with systemd. Refactoring to Go won't solve this problem. For example, given that the database has been running for a while then is rebooted and scheduled on a new host, bin/boot will still have to replay backups from ceph before it is actually booted. This will act the same as a boot script or as a Go static binary. :(

I'm looking into alternative approaches to resolve this issue, particularly around tweaking our systemd units.

carmstrong:
> I'm looking into alternative approaches to resolve this issue, particularly around tweaking our systemd units.

Yeah, wondering if recovery could be in an ExecStartPost or something. Hmm....

bacongobbler:
closing as irrelevant for [v2](https://deis.com/blog/2016/workflow-stable/). kubernetes reports the proper status of the container.

PR #3757, opened 2015-05-27
[WIP] fix(*): fail fast if confd fails to template

bacongobbler:
99 times out of 100, `confd -onetime` fails for a component because a
component is still booting and has not yet published its keys to etcd.
In these cases, we should fail fast and fail hard, which will cause the
systemd units to trigger an auto-restart.

refs #2906. This should help with the majority of "issues" users are running into. The `config check failed` issue isn't resolved by this though which is why this doesn't close the issue. This does help point out to the user which components are failing, however.


bacongobbler:
closing as this will make #3797 rear its head **much** more often than we'd like. We can address this once we fix up that problem child.

jchauncey:
This would be a really nice feature to have within deis for people wanting to alert on individual components going down. At the current time it is possible that a component could be in a bad state because confd cannot read a key from etcd. This causes the container to just spin in the background instead of exiting. 

By having the container exit we can take advantage of fleet restarting the service automatically. However, that of course could present its own problem (fleet keeps restarting the service even though its not healthy and should be dead for a reason).

PR #3762, opened 2015-05-28
docs(managing_deis): Add notes on persisting CoreOS upgrades

### LINKED ISSUE #3149: docs: Document Digital Ocean's behavior to re-image the host on reboot
This explains issues like #2737.

### LINKED ISSUE #3155: Error in upgrading-deis.rst
In the docs this is the procedure to do a manual update:

> $ ssh core@<server ip>
> $ sudo su
> $ echo GROUP=stable > /etc/coreos/update.conf
> $ systemctl unmask update-engine.service
> $ systemctl start update-engine.service
> $ update_engine_client -update
> $ systemctl stop update-engine.service
> $ systemctl mask update-engine.service
> $ reboot

this is wrong (at least now).
Why? https://github.com/coreos/bugs/issues/262

the boot partition is updated after the reboot so if the update-engine service is disabled it will not run and update the boot flags.
Only after this output from `journalctl -u update-engine.service`
`[0226/232543:INFO:update_attempter.cc(674)] Updating boot flags...`
is safe to disable the update-engine service. (after the next update it will use the updated version)

This means that https://github.com/deis/deis/blob/master/contrib/coreos/user-data.example#L25 must be remove so after the reboot with the update update-engine.service will start automatically

reference https://github.com/coreos/bugs/issues/262

carmstrong:
closes #3149
closes #3155

### DIFF
--- a/docs/managing_deis/upgrading-deis.rst
+++ b/docs/managing_deis/upgrading-deis.rst
@@ -188,7 +188,26 @@ While typically not recommended, it is possible to trigger an update of a CoreOS
 Deis releases may recommend a CoreOS upgrade - in these cases, the release notes for a Deis release
 will point to this documentation.
 
-To update CoreOS, run the following commands:
+Checking the CoreOS version
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+You can check the CoreOS version by running the following command on the CoreOS machine:
+
+.. code-block:: console
+
+    $ cat /etc/os-release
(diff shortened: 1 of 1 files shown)


carmstrong:
This passed the docs tests in Jenkins and is ready for review.

bacongobbler:
lgtm! These are good changes.

mboersma:
LGTM.

PR #3763, opened 2015-05-28
fix(*): docker.service depends on its mount

### LINKED ISSUE #3677: docker.service needs to depend on the mount.
Ran into some edge cases where the mount would not complete before docker started. This would cause docker run to never work until the docker daemon was restarted. We fixed this on our end by putting an After and Require statement in the docker.service.

carmstrong:
Just specifying that the mount needs to be started before Docker is
insufficient - since Docker is socket activated, it has a chance
to start before the mount. This commit adds a new drop-in unit
which configures Docker to only start after its mount on providers
where we use a separate mount for Docker (AWS, Azure, and GCE).

/cc @jwaldrip

closes #3677

### DIFF
--- a/contrib/ec2/gen-json.py
+++ b/contrib/ec2/gen-json.py
@@ -28,6 +28,11 @@ MOUNT_DOCKER_VOLUME = '''
   Where=/var/lib/docker
   Type=ext4
 '''
+DOCKER_DROPIN = '''
+  [Unit]
+  Requires=var-lib-docker.mount
+  After=var-lib-docker.mount
+'''
 FORMAT_ETCD_VOLUME = '''
   [Unit]
   Description=Formats the etcd device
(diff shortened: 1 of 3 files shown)


contrib/ec2/gen-json.py
```diff
@@ -60,13 +65,20 @@
   RemainAfterExit=yes
   ExecStart=/usr/bin/chown -R etcd:etcd /media/etcd
 '''
+ETCD_DROPIN = '''
```
bacongobbler: is etcd socket-activated on EC2 as well? Just curious why this is necessary since the PR didn't say anything about etcd

carmstrong: No, but since we need etcd to start after the mount, it's better to be explicit than rely on ordering and *hope* things start up in the order we want. Now etcd cannot start until the mount is ready.

bacongobbler:
code LGTM

mboersma:
Code LGTM.

PR #3767, opened 2015-05-28
fix(builder): generate key on boot

### LINKED ISSUE #3132: question(builder): the published image uses the same ssh keys?
This https://github.com/deis/deis/blob/master/builder/image/Dockerfile#L28 should not run in the boot script instead of the the docker build?

bacongobbler:
closes #3132

### DIFF
--- a/builder/rootfs/bin/boot
+++ b/builder/rootfs/bin/boot
@@ -39,6 +39,21 @@ function etcd_safe_mkdir {
 	set -e
 }
 
+function etcd_set_default_stdin {
+	set +e
+	ERROR=$(etcdctl --no-sync -C $ETCD mk $ETCD_PATH/$1 2>&1 >/dev/null)
+	if [[ $? -ne 0 && $(echo $ERROR | grep -ive "key already exists") ]]; then
+		echo "etcd_set_default_stdin: an etcd error occurred ($ERROR)"
+        echo "aborting..."
+		exit 1
+	fi
+	set -e
+}
+
+function etcd_get {
+	etcdctl --no-sync -C $ETCD get $ETCD_PATH/$1
+}
+
 etcd_safe_mkdir $ETCD_PATH/users
 
 # wait for confd to run once and install initial templates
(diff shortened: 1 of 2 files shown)


carmstrong:
On second thought, this means that we're going to get an unknown key warning every time builder is restarted, right? Hm.... 

We may want to generate keys on first boot and then store them in etcd or something to be templated later.

bacongobbler:
I could try introducing a function that pulls the key from etcd or auto-generates a new one. `/deis/builder/sshPrivateKey` seem like a good endpoint to store the key?

carmstrong:
> Does /deis/builder/sshPrivateKey seem like a good endpoint to store the key?

SGTM

bacongobbler:
done according to @carmstrong's suggestion. the ssh keys, once generated, are now saved in etcd to persist for future reboots.

builder/rootfs/bin/boot
```diff
@@ -74,6 +88,27 @@ done
 docker build -t deis/slugbuilder /usr/local/src/slugbuilder/
 docker build -t deis/slugrunner /usr/local/src/slugrunner/
 
+function gen_host_keys {
+	# generate the keys, then set them up in etcd if not already set
+	/usr/bin/ssh-keygen -A
```
carmstrong: Does it make sense to run `ssh-keygen` before checking if the keys are already there? Seems unnecessary if we're just going to pull from etcd anyhow.

builder/rootfs/bin/boot
```diff
@@ -39,6 +39,20 @@ function etcd_safe_mkdir {
 	set -e
 }
 
+function etcd_set_default_stdin {
+	set +e
+	etcdctl --no-sync -C $ETCD mk $ETCD_PATH/$1 >/dev/null 2>&1
+	if [[ $? -ne 0 && $? -ne 4 ]]; then
```
mboersma: This should be updated as per #3791:
```shell
  ERROR="$(etcdctl --no-sync -C $ETCD mk $ETCD_PATH/$1 $2 2>&1 >/dev/null)"
  if [[ $? -ne 0 && $(echo $ERROR | grep -ive "key already exists") ]]; then
```

builder/rootfs/bin/boot
bacongobbler: it's a little unnecessary, though this was the most efficient way I could think of to generate the keys without introducing too much branch complexity. If you'd prefer to save a few milliseconds, I can wrap this and check if one of the keys have already been saved in etcd. Would that be ok?

carmstrong: It's not really about saving milliseconds, but I think it'd make the code a little more correct and easy to follow. Remember before that we've seen race conditions where an etcd key is set and then is attempted to be read immediately following (setting and using the slugbuilder/slugrunner keys). Let's do a sanity check before we generate new keys, and if they exist in etcd, just use them.

carmstrong: Additionally, the logic seems incorrect to me - it looks like we are generating new keys on *every boot*, then saving them to etcd and re-templating them? This doesn't seem like at all what we want,  unless I'm missing something.

The templating after seems completely unnecessary.

bacongobbler: The logic follows like so:

 - generate new keys on every boot
 - set keys in etcd *if not already set*
 - read keys from etcd, write to the local filesystem

In essence, the first time the keys are generated they are pushed to etcd. After a reboot, the keys will attempt to write but hit the "Key not found" error, making it a no-op. The keys that were previously set in etcd are now read and written to the local filesystem.

carmstrong: I see. But without comments to that effect or a close read of the code to notice that `etcdctl mk` is used, the code looks incorrect. Let's refactor it to make it more readable. I think branching logic is fine here to pay for clarify. :+1:

bacongobbler: done!

carmstrong:
Much easier to follow - thanks @bacongobbler! Code LGTM.

mboersma:
Code LGTM.

PR #3768, opened 2015-05-28
fix(router): avoid consistently hashing on empty string

### LINKED ISSUE #3508: Affinity arg is not working correctly

wenzowski:
Sorry this was so slow.

closes #3508

### DIFF
--- a/router/image/templates/nginx.conf
+++ b/router/image/templates/nginx.conf
@@ -158,6 +158,11 @@ http {
     {{ $useSSL := or (getv "/deis/router/sslCert") "false" }}
     {{ $domains := ls "/deis/domains" }}
     {{ $certs := ls "/deis/certs" }}
+    ## workaround for nginx hashing empty string bug http://trac.nginx.org/nginx/ticket/765
+    {{ if exists "/deis/router/affinityArg" }}
+    set_random $prng 0 99;
+    set_if_empty $arg_{{ getv "/deis/router/affinityArg" }} $prng;
+    {{ end }}
     ## start service definitions for each application
     {{ range $app := lsdir "/deis/services" }}
     {{ $upstreams := printf "/deis/services/%s/*" $app}}


wenzowski:
Eventually, this needs to be fixed upstream.
see http://trac.nginx.org/nginx/ticket/765#ticket

carmstrong:
Code LGTM.

carmstrong:
Kicked off a new Jenkins build: https://ci.deis.io/job/test-ec2/1128/console

router/image/templates/nginx.conf
```diff
@@ -158,6 +158,10 @@ http {
     {{ $useSSL := or (getv "/deis/router/sslCert") "false" }}
     {{ $domains := ls "/deis/domains" }}
     {{ $certs := ls "/deis/certs" }}
+    {{ if exists "/deis/router/affinityArg" }}
+    set_random $prng 0 99;
+    set_if_empty $arg_{{ $affinityArg }} $prng;
```
mboersma: I'm missing where `$affinityArg` gets initialized. Should this be `getv "/deis/router/affinityArg"` instead?

bacongobbler:
could you please reference the bug in a comment so we know to remove this logic once the nginx bug is resolved and we bump nginx?

wenzowski:
Nginx runs trac not gh :(

router/image/templates/nginx.conf
wenzowski: By default it isn't set, which is why I'm using `if exists`. It gets initialized manually with a `deisctl config router set affinityArg=a` for example.

Since it's the name of the `GET` variable containing the affinity token it can be an arbitrary string variable name. So if it were set to empty string or a string that nginx's conf file parser thinks is invalid things would break.

wenzowski: oh, I see what you mean. yes that's wrong. fixing

wenzowski: amended and force-pushed

bacongobbler:
I'm talking about a one-liner comment in the source code that links to http://trac.nginx.org/nginx/ticket/765 :)

wenzowski:
ah. yes indeed

wenzowski:
updated & rebased

mboersma:
@helgi pointed out that the [nginx issue](http://trac.nginx.org/nginx/ticket/765#ticket) was closed as `wontfix`. So this PR becomes more important.

wenzowski:
@mboersma comment updated

mboersma:
Code LGTM.

dshafik:
Brought this up to @linuxjedi at Nginx and his comment was:

> I'm not surprised at the "wontfix". The function needs to be compatible with Memcached which will hash on empty.
>  
> I believe the map Maxim was suggesting would hash on the remote_addr instead if the arg was empty. Although without knowing the remote application it is unknown if this gives you what you need

bacongobbler:
thanks for taking a look into this, @dshafik :)

LinuxJedi:
If you hit Nginx issues again and need help getting answers please feel free to contact me or anyone on the developer relations team.
(sorry for hijacking your report to say that)

carmstrong:
> (sorry for hijacking your report to say that)

Not at all! Se certainly appreciate the offer, @LinuxJedi!

wenzowski:
Thanks for the offer @LinuxJedi Yes, I understood Maxim was probably suggesting a hash of the remote_addr as a way to avoid the emptystring problem, which wasn't what I was trying to request. In the case where affinity is only necessary if a GET variable is set, like this one, the roundtrip to memcached could be avoided entirely when an emptystring is encountered if affinity is not needed for these requests by falling back to round robin for these requests. Having ngninx/memcached hash a pseudo-random number seems like unnecessary overhead for the requests that do not require affinity. I don't know the nginx internals very well though, so I understand the wontfix is probably for a good reason.

fabiob:
The router's nginx does not have the required module for `set_random` to work.

See #3964 for more info.

PR #3771, opened 2015-05-29
fix(database): work around wal-e error by pinning pbr lib

mboersma:
Deis CI is currently stalled on wal-e/wal-e#181, so this is an attempt to work around the issue by pinning `pbr` to a last-known-good version.


carmstrong:
Code LGTM.

aledbf:
Code LGTM.

PR #3776, opened 2015-05-29
fix(logspout): use $IMAGE_PREFIX like other components

mboersma:
The `Makefile` for logspout was still hard-coded to create a Docker image with a deis/ prefix. This brings it in line with other components.


bacongobbler:
LGTM

carmstrong:
Code LGTM.

mboersma:
I'm merging this because it is polluting the production logspout with CI tags. :sweat:

PR #3777, opened 2015-06-01
fix(controller): Add EnvironmentFile to the serivce files created by fleet.py

egbertp:
The service file that is created by the controller after a succesfull
build of a new container, didn't have access to the /etc/environment file
in CoreOS.

People that run etcd on the non-standard 127.0.0.1:4001 address, need to
make the etcd address availalbe to fleetctl in the .service file.
If you don't this, the etcdctl commands used in the ExecStartPre won't
be able to communicate with the correct etcd service.

### DIFF
--- a/controller/scheduler/fleet.py
+++ b/controller/scheduler/fleet.py
@@ -377,6 +377,7 @@ SchedulerClient = FleetHTTPClient
 
 CONTAINER_TEMPLATE = [
     {"section": "Unit", "name": "Description", "value": "{name}"},
+    {"section": "Service", "name": "EnvironmentFile", "value": "/etc/environment"},
     {"section": "Service", "name": "ExecStartPre", "value": '''/bin/sh -c "IMAGE=$(etcdctl get /deis/registry/host 2>&1):$(etcdctl get /deis/registry/port 2>&1)/{image}; docker pull $IMAGE"'''},  # noqa
     {"section": "Service", "name": "ExecStartPre", "value": '''/bin/sh -c "docker inspect {name} >/dev/null 2>&1 && docker rm -f {name} || true"'''},  # noqa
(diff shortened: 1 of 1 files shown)


deis-admin:

bacongobbler:
This is intentional. Applications shouldn't know anything about the host, as that allows attackers to use methods like [webserver fingerprinting](https://www.owasp.org/index.php/Fingerprint_Web_Server_%28OTG-INFO-002%29) to determine known vulnerabilities, as well as granting them internal knowledge of the cluster, which they should never have access to. I can see the reasoning behind this, but it introduces security flaws.

bacongobbler:
related: https://github.com/deis/deis/issues/1078#issuecomment-74879266

carmstrong:
> People that run etcd on the non-standard 127.0.0.1:4001 address, need to
> make the etcd address availalbe to fleetctl in the .service file.

Perhaps your use case is served by the work @krancour is doing in #3679?

krancour:
@carmstrong, @egbertp, yes, this same change is also in #3679.

@bacongobbler I'd put a lot of thought into the security ramifications of this (because @carmstrong had raised them to me once before).   Unless I am missing something, I don't believe they're quite as dire as they appear on the surface...

This would grant additional environmental context to Fleet / systemd when they _start_ the unit.  Unless the unit explicitly passes any of that context to the Docker container with `-e` flags (which it does not), I don't believe that any application gains any new awareness of the host environment.

I have actually put this to the test as well... I am running a custom build of Deis currently that already includes this change (from #3679).  I can `nse` to an application container and verify that the variables I normally poke into `/etc/environment` are inaccessible.

krancour:
@egbertp, this change should probably be made to the `RUN_TEMPLATE` as well as the `CONTAINER_TEMPLATE`.  Could you add that?

carmstrong:
> Unless the unit explicitly passes any of that context to the Docker container with -e flags (which it does not), I don't believe that any application gains any new awareness of the host environment.

That's actually a good point. In my head, I was thinking that the context would be passed to the application context as environment variables, but it doesn't sound like that's the case here.

bacongobbler:
Okay, that makes sense. In that case I'd be more in favour of closing this for #3679 since that addresses this problem across the platform rather than in one specific area.

mboersma:
Closing this in favor of #3679.

PR #3786, opened 2015-06-03
fix(database): avoid pbr runtime error by pinning oslo-config

mboersma:
This is a slightly better fix for the wal-e runtime error that was hampering CI testing last week.

Refs #3771. See wal-e/wal-e#183.


bacongobbler:
Ha, I thought the PR was talking about PBR.

Pabst Blue Ribbon

LGTM!

mboersma:
I think the PBR double-entendre is intentional. :beers:

carmstrong:
Code LGTM.

PR #3791, opened 2015-06-04
fix(*): check for key already exists errors only

### LINKED ISSUE #3790: etcdctl fails to set keys during provisioning
Testing showed that often on vagrant, and occasionally on EC2, the `boot` scripts of controller, builder, or database can somehow get to the `confd --onetime` section without actually having set keys or created etcd dirs with `etcd_safe_mkdir` or `etcd_set_default`. The user-visible symptom is that builder loops forever, waiting for `confd` templating, and eventually goes into `activating/auto-restart` status.

bacongobbler:
refs #3790

This change is necessary as etcdctl globs all server errors as exit code 4, including the infamous `all the given peers are not reachable` as well as `Key already exists`. Instead of checking the error code, we can check what's on stderr so we can safely ignore `Key already exists` "errors".

### DIFF
--- a/controller/bin/boot
+++ b/controller/bin/boot
@@ -26,9 +26,10 @@ sleep $(($ETCD_TTL+1))
 
 function etcd_set_default {
 	set +e
-	etcdctl --no-sync -C $ETCD mk $ETCD_PATH/$1 $2 >/dev/null 2>&1
-	if [[ $? -ne 0 && $? -ne 4 ]]; then
-		echo "etcd_set_default: an etcd error occurred. aborting..."
+	ERROR="$(etcdctl --no-sync -C $ETCD mk $ETCD_PATH/$1 $2 2>&1 >/dev/null)"
+	if [[ $? -ne 0 && $(echo $ERROR | grep -ive "key already exists") ]]; then
+		echo "etcd_set_default: an etcd error occurred ($ERROR)"
+		echo "aborting..."
 		exit 1
 	fi
 	set -e
@@ -36,9 +37,10 @@ function etcd_set_default {
 
 function etcd_safe_mkdir {
 	set +e
-	etcdctl --no-sync -C $ETCD mkdir $1 >/dev/null 2>&1
-	if [[ $? -ne 0 && $? -ne 4 ]]; then
-		echo "etcd_safe_mkdir: an etcd error occurred. aborting..."
+	ERROR="$(etcdctl --no-sync -C $ETCD mkdir $1 2>&1 >/dev/null)"
+	if [[ $? -ne 0 && $(echo $ERROR | grep -ive "key already exists") ]]; then
+		echo "etcd_safe_mkdir: an etcd error occurred ($ERROR)"
+		echo "aborting..."
 		exit 1
 	fi
 	set -e
--- a/builder/rootfs/bin/boot
+++ b/builder/rootfs/bin/boot
@@ -30,9 +30,10 @@ sleep $(($ETCD_TTL+1))
 
 function etcd_safe_mkdir {
 	set +e
-	etcdctl --no-sync -C $ETCD mkdir $1 >/dev/null 2>&1
-	if [[ $? -ne 0 && $? -ne 4 ]]; then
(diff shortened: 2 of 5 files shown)


controller/bin/boot
```diff
@@ -26,9 +26,10 @@ sleep $(($ETCD_TTL+1))
 
 function etcd_set_default {
 	set +e
-	etcdctl --no-sync -C $ETCD mk $ETCD_PATH/$1 $2 >/dev/null 2>&1
-	if [[ $? -ne 0 && $? -ne 4 ]]; then
-		echo "etcd_set_default: an etcd error occurred. aborting..."
+	ERROR="$(etcdctl --no-sync -C $ETCD mk $ETCD_PATH/$1 $2 2>&1 >/dev/null)"
```
carmstrong: I'm not really sure how this works to capture when we're redirecting output to `/dev/null` - it reads to me like this output will always be an empty string.

bacongobbler: The code was taken from http://stackoverflow.com/a/962268. It's weird but stderr is redirected to stdout, stdout is redirected to /dev/null, and then stderr (which is now in stdout) is read into ERROR.

kalbasit: so you only want stderr in ERROR? Why not `ERROR="$(etcdctl --no-sync -C $ETCD mk $ETCD_PATH/$1 $2 >/dev/null)"`, this should send stdout to `/dev/null`

bacongobbler: we need stderr on stdout in order to capture it, hence the redirections.

bacongobbler:
fixed up the logic to only check for the string. I manually tested this logic via 

```
core@deis-01 ~ $ ERROR="$(etcdctl --no-sync mk /test foo 2>&1 >/dev/null)"
core@deis-01 ~ $ ERROR="$(etcdctl --no-sync mk /test foo 2>&1 >/dev/null)"
core@deis-01 ~ $ if [[ $? -ne 0 && $(echo $ERROR | grep -ive "key already exists") ]]; then echo $ERROR; fi
```

mboersma:
I've run this many times and I'm confident it filters out only the "key already exists" errors now. Code LGTM.
```console
Jun 05 16:06:40 deis-01 sh[2828]: d9163538dd6c: Download complete
Jun 05 16:06:40 deis-01 sh[2828]: d9163538dd6c: Download complete
Jun 05 16:06:40 deis-01 sh[2828]: Status: Downloaded newer image for 192.168.59.103:5000/deisci/controller:git-b2a77e3
Jun 05 16:06:40 deis-01 systemd[1]: Started deis-controller.
Jun 05 16:07:07 deis-01 sh[3147]: etcd_set_default: an etcd error occurred (Error:  501: All the given peers are not reachable (Tried to connect to each peer twice and failed) [0])
Jun 05 16:07:07 deis-01 sh[3147]: aborting...
Jun 05 16:07:07 deis-01 systemd[1]: deis-controller.service: main process exited, code=exited, status=1/FAILURE
Jun 05 16:07:07 deis-01 docker[3696]: Error response from daemon: No such container: deis-controller
Jun 05 16:07:07 deis-01 docker[3696]: time="2015-06-05T16:07:07Z" level="fatal" msg="Error: failed to remove one or more containers"
Jun 05 16:07:07 deis-01 systemd[1]: Unit deis-controller.service entered failed state.
Jun 05 16:07:07 deis-01 systemd[1]: deis-controller.service failed.
Jun 05 16:07:12 deis-01 systemd[1]: deis-controller.service holdoff time over, scheduling restart.
Jun 05 16:07:12 deis-01 systemd[1]: Starting deis-controller...
Jun 05 16:07:12 deis-01 systemd[1]: Started deis-controller.
```

mboersma:
Could we make a similar change to `etcd_safe_mkdir` in the builder `boot` script? I have also seen it fail with the same etcd 501 error.

bacongobbler:
done

mboersma:
Code LGTM.

carmstrong:
Code LGTM. Come on Jenkins...

PR #3792, opened 2015-06-04
fix(client): Fix attribute errors on windows

nathansamson:
Fixes a regression where starting up the client in windows would result in a crash, becuase windows does not have support for pipes, and so no SIGPIPE as well.

I didn't test if things still worked in Unix, but I assume they do...

### DIFF
--- a/client/deis.py
+++ b/client/deis.py
@@ -76,8 +76,9 @@ import urllib3.contrib.pyopenssl
 
 # Don't throw IOError when a pipe closes
 # https://github.com/deis/deis/issues/3764
-from signal import signal, SIGPIPE, SIG_DFL
-signal(SIGPIPE, SIG_DFL)
+import signal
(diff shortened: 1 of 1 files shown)


deis-admin:

mboersma:
Jenkins, add to whitelist.

nathansamson:
good we have a CI suite...

nathansamson:
sorry, my git skills seem to be degrading, I accidentetely added bunch of unrelated changes...

carmstrong:
Code LGTM. @nathansamson could you try rebasing off master? We merged some fixes that should help this pass.

mboersma:
Code LGTM. Thanks @nathansamson!

nathansamson:
I see this has been added to milestone 1.8, while I think as this is a regression from 1.6, should be part of 1.7

bacongobbler:
code LGTM!

PR #3795, opened 2015-06-04
fix(logger): poll etcd rather than use etcd watch

bacongobbler:
refs #3790 

Part 1 of 2. We've seen watch thrash the etcd cluster which could explain why we're seeing a lot of `501: All the given peers are not reachable` during stress tests.

### DIFF
--- a/logger/main.go
+++ b/logger/main.go
@@ -46,7 +46,6 @@ func main() {
 	client := etcd.NewClient([]string{"http://" + publishHost + ":" + publishPort})
 
 	signalChan := make(chan os.Signal, 1)
-	drainRespChan := make(chan *etcd.Response)
 	drainChan := make(chan string)
 	stopChan := make(chan bool)
 	exitChan := make(chan bool)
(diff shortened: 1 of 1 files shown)


carmstrong:
@kalbasit Does this seem sound to you?

mboersma:
Code LGTM. Tested multiple times after rebasing this off master.

carmstrong:
Code LGTM. Let's ship this.

logger/main.go
```diff
@@ -58,18 +57,31 @@ func main() {
...
+		for {
+			resp, err := client.Get(publishPath+"/drain", false, false)
+			if err != nil {
+				log.Printf("warning: could not retrieve drain URI from etcd: %v\n", err)
+				continue
+			}
+			if resp != nil && resp.Node != nil {
+				drainChan <- resp.Node.Value
+			}
+			time.Sleep(time.Duration(publishInterval))
```
kalbasit: Apologies for missing to review this. In general, it's better to use a ticker than to sleep. That's because, when exiting, you really should not leave goroutines running. So I would re-write this block as:

```
// HACK (bacongobbler): poll etcd for changes in the log drain value
// etcd's .Watch() implementation is broken when you use TTLs
//
// https://github.com/coreos/etcd/issues/2679
go func() {
  t := time.NewTicker(time.Duration(publishInterval))
  for {
    select {
      case <-exitChan:
        return
      case <-t.C:
        resp, err := client.Get(publishPath+"/drain", false, false)
        if err != nil {
          log.Printf("warning: could not retrieve drain URI from etcd: %v\n", err)
          continue
        }
        if resp != nil && resp.Node != nil {
          drainChan <- resp.Node.Value
        }
    }
  }
}()
```

kalbasit: I will submit a PR to fix it.

PR #3798, opened 2015-06-05
feat(client): Various improvements to config operations

nathansamson:
This includes 2 improvements to config operations.

Also seems to include some unrelated commits, as I branched of tag v1.6.1 to work with my test cluster...

PR #3799, opened 2015-06-05
feat(client): Various improvements to config operations

nathansamson:
This includes 2 improvements to config operations.

Replaces #3798 

Note that this version is not manually tested by me, as it is not compatible with my cluster. As the changes are in code that seemingly didn't change since 1.6.1 I don't expect any troubles, but you never know...

### DIFF
--- a/client/deis.py
+++ b/client/deis.py
@@ -1207,19 +1207,8 @@ Make sure that the Controller URI is correct and the server is running.
             if len(items) == 0:
                 self._logger.info('No configuration')
                 return
-            keys = sorted(values)
 
-            if not oneline:
-                width = max(map(len, keys)) + 5
-                for k in keys:
-                    k, v = encode(k), encode(values[k])
-                    self._logger.info(("{k:<" + str(width) + "} {v}").format(**locals()))
-            else:
-                output = []
-                for k in keys:
-                    k, v = encode(k), encode(values[k])
-                    output.append("{k}={v}".format(**locals()))
-                self._logger.info(' '.join(output))
+            self._display_configuration(values, oneline=oneline)
         else:
             raise ResponseError(response)
 
@@ -1282,8 +1271,7 @@ Make sure that the Controller URI is correct and the server is running.
             if len(items) == 0:
                 self._logger.info('No configuration')
                 return
-            for k, v in values.items():
-                self._logger.info("{}: {}".format(encode(k), encode(v)))
+            self._display_configuration(values, oneline=False)
         else:
             raise ResponseError(response)
 
@@ -1328,8 +1316,7 @@ Make sure that the Controller URI is correct and the server is running.
             if len(items) == 0:
                 self._logger.info('No configuration')
                 return
-            for k, v in values.items():
-                self._logger.info("{k}: {v}".format(**locals()))
+            self._display_configuration(values, oneline=False)
         else:
             raise ResponseError(response)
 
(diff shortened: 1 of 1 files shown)


client/deis.py
```diff
@@ -1348,6 +1335,24 @@ def _read_config_from_path(self, path):
 
         return env_dict
 
+    def _display_configuration(self, configuration, oneline=False, highlight=[]):
```
bacongobbler: let's make the `oneline` and `highlight` options as [keyword arguments](http://stackoverflow.com/a/1419160) to make this a little easier to maintain.

carmstrong:
@nathansamson Are you able to make the change @bacongobbler requested? Thanks for taking the time to submit this!

nathansamson:
@carmstrong done. Somehow I had missed the comment

client/deis.py
```diff
@@ -1328,8 +1316,7 @@ def config_unset(self, args):
             if len(items) == 0:
                 self._logger.info('No configuration')
                 return
-            for k, v in values.items():
-                self._logger.info("{k}: {v}".format(**locals()))
+            self._display_configuration(values, oneline=False)
```
bacongobbler: `oneline` is set to False by default so we can remove the invocation from here.

client/deis.py
```diff
@@ -1348,6 +1335,27 @@ def _read_config_from_path(self, path):
...
+        highlight = kwargs.get('highlight', [])
+
+        keys = sorted(configuration)
+        if not oneline:
+            width = max(map(len, keys)) + 5
+            for k in keys:
+                k, v = encode(k), encode(configuration[k])
+                key = "{k:<" + str(width) + "}"
+                if k in highlight:
+                    self._logger.info(("\033[1m" + key + "\033[0m {v}").format(**locals()))
```
bacongobbler: let's use the `colored` library as we do here for better portability: https://github.com/deis/deis/blob/master/client/deis.py#L682-L691

mboersma:
Highlighting the config that actually changed is nice. (I added asterisks to indicate where bold green ANSI highlighting was applied.)
```
$ deis config:set FOO=bar
Creating config... done, v2

=== outlaw-aqualung
*FOO*    bar
$ deis config:set BOO=far
Creating config... done, v3

=== outlaw-aqualung
*BOO*    far
FOO      bar
$ deis config:set FOO=baz
Creating config... done, v4

=== outlaw-aqualung
BOO      far
*FOO*    baz
$ deis config:list
=== outlaw-aqualung Config
BOO      far
FOO      baz
$ deis config:set FOO=bax BAX=foo
Creating config... done, v5

=== outlaw-aqualung
*BAX*    foo
BOO      far
*FOO*    bax
```

If this PR could use the termcolor library as suggested I think it's ready to go. Something like:
```python
self._logger.info("{} {}".format(colored(key, 'green'), v))
```

bacongobbler:
s/green/magenta. Deis needs more pink! (kidding)

mboersma:
The second commit should use a lower-case verb as per our [code standards](http://docs.deis.io/en/latest/contributing/standards/#subject-line):
```
feat(client): highlight changed config
```

nathansamson:
Hi,

I will close this pull request in the hope that @Joshua-Anderson will include something similar in #4015 

Otherwise I might try a third attempt in the new codebase

Regards,
Nathan

PR #3800, opened 2015-06-05
chore(*): update Docker base image to alpine:3.2

mboersma:
See http://alpinelinux.org/posts/Alpine-3.2.0-released.html


carmstrong:
Code LGTM.

mboersma:
There's more to do here; for some reason the registry functional test always fails with a socket error. I'll close this and re-open it when it's actually ready for prime time.

PR #3801, opened 2015-06-05
feat(router): use new reuseport option

aledbf:
http://nginx.com/blog/socket-sharding-nginx-release-1-9-1/

### DIFF
--- a/router/build.sh
+++ b/router/build.sh
@@ -10,7 +10,7 @@ if [[ -z $DOCKER_BUILD ]]; then
   exit 1
 fi
 
-export VERSION_NGINX=nginx-1.9.0
+export VERSION_NGINX=nginx-1.9.1
 export VERSION_NAXSI=0d53a64ed856e694fcb4038748c8cf6d5551a603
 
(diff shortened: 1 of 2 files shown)


carmstrong:
@aledbf What does this change mean? Is there anything we should look out for, or potential issues from this?

aledbf:
This should be a transparent update. 
We should see less latency per request in when the cluster is under high loads.

kalbasit:
The benchmark looks quite amazing. :+1:

carmstrong:
@aledbf Could you rebase this? @Joshua-Anderson's commit to retry pushes to DockerHub should hopefully help.

aledbf:
@carmstrong done. The problem was that the directive `reuseport` must be specified only once and in the default server.

carmstrong:
We should probably split out the nginx bump commit so it's not hidden.

mboersma:
Code LGTM.

carmstrong:
Code LGTM. Thanks @aledbf!

Issue #3722, opened 2015-05-19
Feature Request: Maintenance Mode


jwaldrip:
Allow for a maintenance mode config option that will stop serving web traffic and send it to a static maintenance page.

ex:

```sh
$ deis maintenance:set_page http://s3.aws.com/mybucket/mypage.html
$ deis maintenance:on
$ curl http://mysite.com
<html>
We are under maintenance!!!
</html>
```

Joshua-Anderson:
Heroku's CLI docs for maintennance mode are [here](https://devcenter.heroku.com/articles/maintenance-mode), for reference. I like this because when deploying you often have to migrate your app, which requires stopping the web server. Giving a nice error page is much better than the a 404 error.

gabrtv:
Big +1 for this.  I don't see this being a big effort either.

kalbasit:
:+1: 

FYI for a client, I solved this by creating a `maintenance` application which is an Nginx serving the maintenance page for any request and then did:

```
deis domains:remove www.example.com -a example && deis domains:add www.example.com -a maintenace
```

PR #3805, opened 2015-06-06
WIP: feat(*): Allow applications to be in maintenance mode

nathansamson:
Closes: #3722

Note this is very much a non working wip.

TODO:
[ ] Test the basics, I just wrote the code without even testing if it made syntactical (or semantical) sense. I will try  this later on a computer with more RAM in vagrant
[ ] Add unit tests
[ ] Add documentation for controller API
[ ] Add user documentation
[ ] Change client
[ ] Add custom error/maintenance pages

If someone from the core dev team could tell me this is more or less the right way to implement this feature, let me now

### DIFF
--- a/controller/api/models.py
+++ b/controller/api/models.py
@@ -410,6 +410,19 @@ class App(UuidAuditedModel):
         escaped_command = command.replace("'", "'\\''")
         return c.run(escaped_command)
 
+    def maintenance_on(self, user):
+        msg = "{} enables miantenance mode".format(user.username)
+        log_event(self, msg)
+        _etcd_client.write('/deis/maintenance/{}'.format(id), True)
+
+    def maintenance_on(self, user):
+        msg = "{} enables miantenance mode".format(user.username)
+        log_event(self, msg)
+        try:
+            _etcd_client.delete('/deis/maintenance/{}'.format(id))
+        except KeyError:
+            pass
+
 
 @python_2_unicode_compatible
 class Container(UuidAuditedModel):
--- a/controller/api/urls.py
+++ b/controller/api/urls.py
@@ -56,6 +56,10 @@ urlpatterns = patterns(
         views.AppViewSet.as_view({'get': 'logs'})),
     url(r"^apps/(?P<id>{})/run/?".format(settings.APP_URL_REGEX),
         views.AppViewSet.as_view({'post': 'run'})),
+    url(r"^apps/(?P<id>{})/maintenance/on?".format(settings.APP_URL_REGEX),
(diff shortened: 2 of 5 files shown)


controller/api/models.py
```diff
@@ -410,6 +410,19 @@ def _has_hostname(image):
         escaped_command = command.replace("'", "'\\''")
         return c.run(escaped_command)
 
+    def maintenance_on(self, user):
+        msg = "{} enables miantenance mode".format(user.username)
+        log_event(self, msg)
+        _etcd_client.write('/deis/maintenance/{}'.format(id), True)
+
+    def maintenance_on(self, user):
```
kalbasit: I believe you meant `maintenance_off` here.

nathansamson: yeah I noticed as well. As said, thiis was done in 20m or so.

I was more intrested in general remarks of setting an etc key, and then using that in the router is a good way forward.
Also I wanted people to know I was onto this, to not do double work

mboersma:
> I was more intrested in general remarks

@nathansamson at first glance, I think this is valid as a general approach: set a per-app `etcd` key that deis-router uses to tailor the HTTP response based on whether maintenance mode is on or not.

controller/api/models.py
```diff
@@ -410,6 +410,19 @@ def _has_hostname(image):
         escaped_command = command.replace("'", "'\\''")
         return c.run(escaped_command)
 
+    def maintenance_on(self, user):
+        msg = "{} enables miantenance mode".format(user.username)
+        log_event(self, msg)
+        _etcd_client.write('/deis/maintenance/{}'.format(id), True)
```
mboersma: This should probably be `.format(self.id), True`.

nathansamson:
Closing, as I tried (but failed) to get a dev environment up & running on my computer (8GB ram).

Even when not running a UI, and do everything from a shell (effectively using almost all ram for nodes), builder never worked properly in my test VM.

I am planning to buy a new computer in the short-term with 16GB (or more) RAM, but for now I'll close the issue...

I might work on it in a few weeks time again, but for now I have no viable option to do so...

bacongobbler:
@nathansamson most of us deploy Deis on one of the providers. DO's a pretty cheap option for testing changes. You should be able to runthe unit tests on your computer without an issue, but running the entire platform is certainly a resource hog.

nathansamson:
Note: I've reworked on this in https://github.com/nathansamson/deis/tree/maintenance2

That version is now working (after fixing a lot of stupid bugs).

After I add documentation + some form of tests I'll reopen

PR #3816, opened 2015-06-08
fix(client): Bump pyinstaller to fix pyinstaller/pyinstaller#152

Joshua-Anderson:
I bumped Pyinstaller to the latest commit ([diff](https://github.com/pyinstaller/pyinstaller/compare/7413317...32bbb954b355937ccfe377afbe56979db79a7b30)), to fix pyinstaller/pyinstaller#152, which prevents compilation if the python headers are in certain locations (in my case, where homebrew installed python was subject to this).

I'm not sure what your policy is on bumping Pyinstaller, but I tested the client and it seemed to work after the bump.

This is a major bump, with over 200 commits since the last pyinstaller update, so this definitely needs an careful test.

### DIFF
--- a/client/Makefile
+++ b/client/Makefile
@@ -1,6 +1,6 @@
 
 build: setup-venv
-	venv/bin/pip install -r requirements.txt git+https://github.com/pyinstaller/pyinstaller@7413317
+	venv/bin/pip install -r requirements.txt git+https://github.com/pyinstaller/pyinstaller@32bbb954b355937ccfe377afbe56979db79a7b30
 	venv/bin/pyinstaller deis.spec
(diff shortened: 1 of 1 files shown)


mboersma:
Generally speaking that's probably safe. We just need to be at or past the previous commit in order to get `requests` packaged properly. See #2721. (I wish PyInstaller would make a new official release.)

mboersma:
I never saw the original build error, but I tested `deis` CLI building and basic operation with this PR. LGTM.

carmstrong:
> but I tested deis CLI building and basic operation with this PR.

Good enough for me. Code LGTM.

PR #3822, opened 2015-06-08
fix(publisher): ensure creating the directory if none exists

### LINKED ISSUE #3819: After upgrade to 1.7, sites are 404'ing and cannot push
I am using the ruby buildpack, it pushes, builds the image, and then i get the following:

```
/home/git/builder: line 175: cd: /home/git/site.git: No such file or directory
```

bacongobbler:
See comment for more information.

Closes #3819

### DIFF
--- a/publisher/server/publisher.go
+++ b/publisher/server/publisher.go
@@ -242,9 +242,10 @@ func (s *Server) removeEtcd(key string, recursive bool) {
 	}
 }
 
-// updateDir updates the given directory for a given ttl. It succeeds
-// only if the given directory already exists.
+// updateDir updates or creates the given directory for a given ttl.
 func (s *Server) updateDir(directory string, ttl uint64) {
+	// Create directory with reckless abandon. If anything occurs, UpdateDir will catch it!
+	s.EtcdClient.CreateDir(directory, ttl)
 	if _, err := s.EtcdClient.UpdateDir(directory, ttl); err != nil {
 		log.Println(err)
 	}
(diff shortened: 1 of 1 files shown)


mboersma:
Code LGTM, testing upgrade scenario now.

bacongobbler:
for anyone else who wants to test this scenario:

 - deploy v1.6.1
 - deploy an app, any app (I used [example-go](https://github.com/deis/example-go))
 - upgrade to v1.7.0, notice that `deis open` shows a 404
 - upgrade publisher with `deisctl config publisher set image=deisci/publisher:jenkins-test-ec2-1242`

Your app should be working, and /deis/services/<appname> should be properly populated again.

mboersma:
I was not able to recreate step 3, "notice that `deis open` shows a 404", at least on vagrant, despite waiting for a few minutes after `deisctl uninstall platform`. I tried to recreate the conditions with `deis scale web=0` and removing the key and dir with `etcdctl`, then upgrading, and a 404 persisted after ugrading to v1.7, but `deis scale web=1` repaired it immediately.

tl:dr, couldn't reproduce, so couldn't verify, at least on vagrant. Trying EC2 now...

bacongobbler:
We must be getting inconsistent results then. Here's a raw deploy for me: https://gist.github.com/bacongobbler/4bb73ec9f1889594f1f5

EDIT: for some reason, rebooting publisher on v1.7 no longer causes this issue. I guess that means that my theory has been thrown out of the question :(

carmstrong:
@bacongobbler Should we close this until we find the root issue with publisher, or is this a useful bugfix to have anyway? It doesn't seem to be causing the root issue.

bacongobbler:
yes, this doesn't seem to be our smoking gun. Back to the drawing board. :pencil2:

bacongobbler:
I'll pull out the logging in a separate PR as those are still useful. It'll probably get slogged in after I deal with #3729 since that touches logging as well.

PR #3826, opened 2015-06-09
fix(router): fix markdown table rendering

linki:
small change so the markdown table is correctly rendered by github

### DIFF
--- a/router/README.md
+++ b/router/README.md
@@ -34,8 +34,8 @@ The rules included are from this project [doxi-rules](https://bitbucket.org/lazy
 
 Only these modules are enabled:
 
-|--|--|
-|File| |
+|File|Purpose|
+|----|-------|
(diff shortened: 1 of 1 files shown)


deis-admin:

mboersma:
Nice catch--I didn't notice this markdown table was broken. But this still doesn't render right for me--I think you have to extend the first row of hyphens to have the same length as the header row for it to render properly as a [GitHub table](https://help.github.com/articles/github-flavored-markdown/#tables):

```
|File|Purpose|
|----|-------|
```
That worked for me when previewing.

linki:
thanks, my bad. should have checked it with github before creating the PR :hushed: should work fine now.

mboersma:
LGTM. Thanks @linki!

carmstrong:
LGTM.

PR #3828, opened 2015-06-09
fix(router): set_random belongs in a `location` block

wenzowski:
Corrects placement of patch from #3768 since `set_random` is only defined within a `location` block.


mboersma:
According to nginx docs, [set_random](http://wiki.nginx.org/HttpSetMiscModule#set_random) should indeed be inside a `location` (or `location if`) block. Too bad `nginx` didn't enforce that syntax.

Code LGTM.

carmstrong:
Code LGTM.

PR #3833, opened 2015-06-09
fix(tests): fix assertions on default beverage

krancour:
This has burned me two times too many:

https://github.com/deis/deis/blob/master/tests/ps_test.go#L41

This fixes an issue where integration tests that would otherwise
succeed, locally, fail close to the end because the user executing
the tests has set the DEIS_DRINK_OF_CHOICE environment variable,
which interferes with the assertion that the default beverage to
be enjoyed whilst scaling is coffee.  This corrects that problem
by unsetting the DEIS_DRINK_OF_CHOICE environment variable prior
to testing.


mboersma:
Great--now I remember I had to remove `export DEIS_DRINK_OF_CHOICE=Fanta` from my .bashrc because the tests expect `coffee`. Thanks for fixing this. Code LGTM.

carmstrong:
Code LGTM. :+1:

PR #3837, opened 2015-06-09
chore(*): bump CoreOS to 681.0.0

mboersma:
See https://coreos.com/releases/#681.0.0

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -56,7 +56,7 @@ Vagrant.configure("2") do |config|
   config.ssh.insert_key = false
 
   config.vm.box = "coreos-%s" % $update_channel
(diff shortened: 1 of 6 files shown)


carmstrong:
We may as well bump Docker's builder to 1.6.2 as part of this, right? Or is that blocked by registry v2?

mboersma:
> Or is that blocked by registry v2?

Yep. They API changed in an incompatible way, so Docker 1.6.x won't work with the existing registry, but I'm working on that in my [new-dev-registry](https://github.com/mboersma/deis/tree/new-docker-registry) branch.

carmstrong:
Code LGTM.

aledbf:
Code LGTM.

PR #3842, opened 2015-06-10
docs(installing_deis): etcd doesn't require an odd # of nodes

mboersma:
ping @gabrtv.

### DIFF
--- a/docs/installing_deis/aws.rst
+++ b/docs/installing_deis/aws.rst
@@ -70,10 +70,7 @@ By default, the script will provision 3 servers. You can override this by settin
 
     $ export DEIS_NUM_INSTANCES=5
 
-Note that for scheduling to work properly, clusters must consist of at least 3 nodes and always
-have an odd number of members. For more information, see `etcd disaster recovery`_.
-
-Deis clusters of less than 3 nodes are unsupported.
+A Deis cluster must have 3 or more nodes. See :ref:`cluster-size` for more details.
 
 
 Generate a New Discovery URL
--- a/docs/installing_deis/azure.rst
+++ b/docs/installing_deis/azure.rst
@@ -76,12 +76,11 @@ With that, let's run the azure-coreos-cluster script which will create the CoreO
          --data-disk
          --custom-data azure-user-data
 
-This script will by default provision a 3 node cluster but you can increase this with the
+This script will by default provision a 3-node cluster, but you can increase this with the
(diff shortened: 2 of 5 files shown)


gabrtv:
Thanks for the quick turnaround. LGTM!

carmstrong:
We say that because of https://coreos.com/docs/cluster-management/scaling/etcd-optimal-cluster-size/

"The other important cluster optimization is to always have an odd active cluster size (i.e. activeSize). Adding an odd node to the number of peers doesn't change the size of the majority and therefore doesn't increase the total latency of the majority as described above. But, you gain a higher tolerance for peer failure by adding the extra machine. You can see this in practice when comparing two even and odd sized clusters:"

Maybe require is too strong a word, but I've always thought it was a best practice for failure tolerance.

gabrtv:
While it's true a standalone etcd cluster needs an odd number of nodes, this has not been true of Deis clusters since introduction of standby mode in etcd 0.4: https://coreos.com/blog/etcd-0.4.0/

mboersma:
Thanks @carmstrong--I knew that was in there somewhere but couldn't find it. Sounds like an odd-sized cluster is somewhere between an optimization and best practice, but not strictly required. We could change the wording here to "recommended"...

carmstrong:
@gabrtv I guess I still don't understand why standby mode changes this, since it has to do with failure tolerance. I thought standby mode was for larger clusters so we're not waiting on writes to persist throughout the whole cluster?

gabrtv:
After some reading it seems standby mode has been deprecated for proxy mode in etcd2: https://github.com/coreos/etcd/blob/556713739c01fa1ebafdf53c8d1f87d6d33367bf/Documentation/backward_compatibility.md#standby

Either way we should merge this PR as we don't want to scare folks away from even-sized clusters (either planned or accidental) because they work just fine per the table on: https://coreos.com/docs/cluster-management/scaling/etcd-optimal-cluster-size/#odd-active-cluster-size

mboersma:
We should ensure that we have links to specific etcd docs justifying this and talking about tuning.

mboersma:
Updated with a consistent reference to our Cluster Size section, and added a link to CoreOS' "optimal cluster size" doc.

carmstrong:
LGTM. I see the issue with our "clusters must have an odd number of nodes" language.

gabrtv:
LGTM, thanks everyone!

PR #3846, opened 2015-06-10
test(deisctl): add race detection to deisctl

Joshua-Anderson:
The deisctl client didn't have race detection enabled on the tests while the rest of deis did. It can't hurt to have it and it could catch some nasty bugs.


mboersma:
The [race detector](https://golang.org/doc/articles/race_detector.html) should definitely be on for `go test`, but it looks like Deis defines `GOTEST = $(GO) test --cover --race -v` in includes.mk and then only uses it in logger tests. So the `-race` flag isn't actually on for most of our tests.

This PR LGTM as-is, but either we should add to it or open another one to try to get all Deis tests using the `-race` flag.

bacongobbler:
indeed. Standardization for tests across components is key. :key:

Joshua-Anderson:
Didn't look carefully enough at the other unit tests. Sorry :frowning:. I'll open a new PR that better fixes the problem in a sec.

mboersma:
No apologies necessary--it's very helpful that you noticed this in the first place and pounced on it.

PR #3854, opened 2015-06-11
fix(logger): adjust publish and ttl times to seconds

### LINKED ISSUE #3853: deis-logger is eating whole cpu
3-node cluster, fresh 1.7.0 with 1.6.1 publisher (due to issue with 404's).

One node where I've deis-logger, logger and etcd process together are eating whole cpu. If I stop deis-logger.service, etcd cpu usage is dropping. journalctl -f all together on all nodes combined is showing on average maybe <100 lines per minute.

journalctl -f -u for etcd or deis-logger doesn't show any warnings/errors.

Strace on etcd process is showing lots of GET requests to '/v2/keys/deis/logs/drain?con...', but if I remove log drain it doesn't change much in case of cpu usage.

Any advice where I should look?

mboersma:
Closes #3853.

### DIFF
--- a/logger/main.go
+++ b/logger/main.go
@@ -59,7 +59,7 @@ func main() {
 
 	go syslogd.Listen(exitChan, cleanupChan, drainChan, fmt.Sprintf("%s:%d", logAddr, logPort))
 	if enablePublish {
-		go publishService(exitChan, client, publishHost, publishPath, strconv.Itoa(logPort), uint64(time.Duration(publishTTL).Seconds()))
+		go publishService(exitChan, client, publishHost, publishPath, strconv.Itoa(logPort), uint64(time.Duration(publishTTL)*time.Second))
 	}
 
 	// HACK (bacongobbler): poll etcd for changes in the log drain value
@@ -76,7 +76,7 @@ func main() {
 			if resp != nil && resp.Node != nil {
 				drainChan <- resp.Node.Value
 			}
-			time.Sleep(time.Duration(publishInterval))
+			time.Sleep(time.Duration(publishInterval) * time.Second)
 		}
 	}()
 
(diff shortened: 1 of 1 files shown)


logger/main.go
```diff
@@ -76,7 +76,7 @@ func main() {
 			if resp != nil && resp.Node != nil {
 				drainChan <- resp.Node.Value
 			}
-			time.Sleep(time.Duration(publishInterval))
+			time.Sleep(time.Duration(publishInterval) * time.Seconds)
```
aledbf: time.Sleep(publishInterval * time.Seconds) is enough, no need to use time.Duration

mackwic:
Very slightly related but I don't know if relevant: from the [Go `flag` reference](http://golang.org/pkg/flag/#pkg-overview), the correct way to pass command-line argument is:  `/bin/logger -flag=value`.

But, the [Dockerfile use the `--flag` syntax](https://github.com/deis/deis/blob/master/logger/image/Dockerfile#L4). Is it an error ?

bacongobbler:
@mackwic both are fine. From the same page:

> One or two minus signs may be used; they are equivalent.

bacongobbler:
@aledbf's comment looks like a good change to implement, otherwise LGTM

carmstrong:
LGTM.

mboersma:
On a new `make dev-cluster` vagrant cluster from master, `top` shows the logger process to consume about 35% of CPU shares, even without any apps deployed. After this change, logger consumes a negligible (<1%) share.

logger/main.go
mboersma: I couldn't get the `go` compiler to pass that version. [`time.Sleep`](http://golang.org/pkg/time/#example_Sleep) expects a `Duration`, and this is the [canonical way](http://golang.org/pkg/time/#Duration) to convert an integer to seconds.

PR #3858, opened 2015-06-11
ref(*): disable pip version check and caching

mboersma:
The constant reminders that pip has a new version involve a network transaction and are not helpful to our build process, and some python libraries were being inadvertently cached at /root/.cache/ inside Docker images.


carmstrong:
> and some python libraries were being inadvertently cached

That's not good. Code LGTM.

bacongobbler:
I think it's fine, it's just adding more files to the final docker image that were unnecessary, unless I'm mistaken.

Code LGTM

PR #3859, opened 2015-06-12
test(deisctl): add tests to config package

Joshua-Anderson:


deisctl/config/config_test.go
```diff
@@ -2,11 +2,124 @@ package config
 
 import (
 	"encoding/base64"
+	"fmt"
 	"io/ioutil"
 	"os"
 	"testing"
 )
 
+type store struct {
```
bacongobbler: hmm, `store` is a little misleading here, since it's just representing a single key/value pair. let's use `KVPair` instead, with a store being a slice of key/value pairs. For example:

```
type KVPair struct {
    Key string
    Value string
}
type MockStore []KVPair
```

That way it's more clear to the reader :)

deisctl/config/config_test.go
```diff
@@ -2,11 +2,124 @@ package config
...
+	for _, expect := range m.expected {
+		if expect.key == key {
+			return nil
+		}
+	}
+	return fmt.Errorf("%s does not exist", m.expected)
+}
+
+type mockWriter struct {
+	output *string
```
bacongobbler: Instead of using a string here, would it be better to use a [`bytes.Buffer`](http://golang.org/pkg/bytes/#Buffer) instead? That way we don't have to do ugly type casting (which causes a [full copy of the buffer into the string](http://stackoverflow.com/a/9649061)) and instead just use [`io.Copy`](http://golang.org/pkg/io/#Copy).

bacongobbler: also since it's just a buffer, we don't need to create a struct for it. Just calling it `type mockWriter bytes.Buffer` should do the trick :)

Joshua-Anderson:
@bacongobbler Done!

Joshua-Anderson:
Wow. Docker Hub failed 3 times in a row on this PR :clap:. Could somebody please restart the CI build?

bacongobbler:
https://ci.deis.io/job/test-ec2/1318/

mboersma:
LGTM. Excellent stuff @Joshua-Anderson! I ran the tests locally and put the `deisctl` binary through its paces.

bacongobbler:
code LGTM

PR #3860, opened 2015-06-12
test(deisctl): add tests for CMD package

Joshua-Anderson:


deisctl/cmd/cmd.go
```diff
@@ -416,7 +416,11 @@ func RefreshUnits(dir string, tag string) error {
 	for _, unit := range units.Names {
 		src := rootURL + tag + "/deisctl/units/" + unit + ".service"
 		dest := filepath.Join(dir, unit+".service")
+<<<<<<< HEAD
```
carmstrong: Looks like merge conflicts...

Joshua-Anderson:
I fixed the git conflict. On my local machine on the same commit it wasn't there, as I fixed it a while ago. :confused:

deisctl/cmd/cmd.go
```diff
@@ -405,7 +405,7 @@ func Config(target string, action string, key []string) error {
 // RefreshUnits overwrites local unit files with those requested.
 // Downloading from the Deis project GitHub URL by tag or SHA is the only mechanism
 // currently supported.
-func RefreshUnits(dir string, tag string) error {
+func RefreshUnits(dir string, tag string, get func(string) (*http.Response, error)) error {
```
bacongobbler: Instead of making a custom `get` function, you could simply run a [httptest](http://golang.org/pkg/net/http/httptest/) server that hosts the units and refactor out the hardcoded rootURL. I think that would be a cleaner approach. What do you think about that?

bacongobbler: i.e. make url a parameter, and when we call RefreshUnits, we send in that URL. In the default case, that'd be the github URL and for the httptest server that'd be `srv.URL`.

Joshua-Anderson:
@bacongobbler Switched to `httptest` server.

bacongobbler:
There's some cleanup that we can probably do in a future refactor (specifically looking at the checkUnits control flow), but from a testing perspective this looks good to me!

bacongobbler:
issue seems unrelated to your PR. I've kicked off another job for you: https://ci.deis.io/job/test-ec2/1317/

carmstrong:
Code LGTM.

mboersma:
LGTM. I also ran the new tests and did some manual testing of the `deisctl` binary since this involved code refactoring.

Thank you for adding tests! :green_heart:

PR #3866, opened 2015-06-12
feat(router): enable HSTS when enforceHTTPS is set

wenzowski:
When browsers see the HSTS header on an HTTPS request then they rewrite
all links for the current domain that point at HTTP resources to point
to HTTPS resources. When /deis/router/enforceHTTPS is set, using HSTS
avoids the extranneous 301 redirect to the HTTPS resource and prevents
[some threats][1]. The HTTPS Strict Transport Security header mechanism
is defined in [RFC-6797][2]

[1]: https://www.owasp.org/index.php/HTTP_Strict_Transport_Security
[2]: https://tools.ietf.org/html/rfc6797

### DIFF
--- a/docs/customizing_deis/router_settings.rst
+++ b/docs/customizing_deis/router_settings.rst
@@ -60,7 +60,7 @@ setting                                      description
 /deis/router/gzipDisable                     nginx gzipDisable setting (default: "msie6")
 /deis/router/gzipTypes                       nginx gzipTypes setting (default: "application/x-javascript application/xhtml+xml application/xml application/xml+rss application/json text/css text/javascript text/plain text/xml")
 /deis/router/hsts/enabled                    enable HTTP Strict Transport Security headers for HTTPS requests (default: false)
-/deis/router/hsts/maxAge                     maximum number of seconds user agents should observe HSTS rewrites (default: 2628000)
+/deis/router/hsts/maxAge                     maximum number of seconds user agents should observe HSTS rewrites (default: 10886400)
 /deis/router/hsts/includeSubDomains          enforce HSTS for requests on all subdomains (default: false)
 /deis/router/hsts/preload                    allow the domain to be included in the HSTS preload list (default: false)
 /deis/router/maxWorkerConnections            maximum number of simultaneous connections that can be opened by a worker process (default: 768)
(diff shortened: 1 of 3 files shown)


wenzowski:
The [`add_header`](http://nginx.org/en/docs/http/ngx_http_headers_module.html) mechanism would set `Strict-Transport-Security` blindly for both HTTP and HTTPS requests. Since the header is only valid on HTTPS requests a map is used in this PR to ensure it is only set on HTTPS requests.

The `always` keyword ensures the header field will be added regardless of the HTTP response code.

bacongobbler:
Hmm. I'm wondering if this should be hidden behind a feature flag or not, considering that this is an opt-in security enhancement. \cc @deis/core-maintainers

wenzowski:
I chose this to be triggered by `/deis/router/enforceHTTPS` since it really is an extension of that: hsts forces the browser to use https on the domain for the specified number of seconds unless a subsequent request overwrites the maxAge. There should probably be a way to opt out of the hsts header but still enforce 301 redirects to https, even if the default is to trigger this with the same key.

croemmich:
This is great, I was just thinking about doing this yesterday! I'm with @bacongobbler, this should probably be behind a feature flag. Some user's may not want it and others may not understand it and break things if they decide to disable SSL.

Maybe use these keys instead?
/deis/router/hsts/enabled (default: false)
/deis/router/hsts/maxAge (default: 2628000)
/deis/router/hsts/includeSubDomains (default: false)

wenzowski:
Great suggestions @croemmich, updated.

I've updated the default `maxAge` so the `preload` key will produce a header that is compatible with the Chrome [preload list](https://hstspreload.appspot.com/)

wenzowski:
I also don't see a use case for disabling `enforceHTTPS` but enabling HSTS, so setting `/deis/router/hsts/enabled` to `true` also enables `enforceHTTPS` but `/deis/router/enforceHTTPS` no longer enables HSTS so this is now completely backwards compatible.

mboersma:
> so setting /deis/router/hsts/enabled to true also enables enforceHTTPS but /deis/router/enforceHTTPS no longer enables HSTS so this is now completely backwards compatible.

That sounds like the way to go. Code LGTM. This needs some manual testing IMHO to ensure everything is working as we think it is on a real cluster with SSL.

carmstrong:
Code LGTM.

mboersma:
Tested, LGTM.

Issue #3812, opened 2015-06-07
Deis networking V2


carmstrong:
There are fundamental but breaking improvements to the Deis platform and the way components communicate that will likely necessitate a Deis v2.0 release. These changes include:

- Overlay network
- ~~Internal service discovery (#3072)~~
- Migration strategy to networking v2

carmstrong:
We decided to drop internal service discovery from this requirement, and instead consider networking v2 the introduction of the overlay network in #3868.

mboersma:
Deis will have flannel in v1.9 thanks to #4103.

PR #3868, opened 2015-06-13
Proposal feat(*): use flannel as overlay network provider

aledbf:
**Network:** 10.244.0.0/16

**Subnets:**
* *Netmask:* 255.255.255.0 (/24)
* *Number of subnets:* 256
* *Number of IP addresses per subnet:* 256

This allows a cluster up to 256 subnets of 256 different IP addresses, i.e. the maximum size of the cluster is 256 nodes.


contrib/coreos/user-data.example
```diff
@@ -33,6 +33,41 @@ coreos:
...
+      EnvironmentFile=/etc/environment
+      Environment="TMPDIR=/var/tmp/"
+      Environment="DOCKER_HOST=unix:///var/run/early-docker.sock"
+      Environment="FLANNEL_VER=0.4.1"
+      LimitNOFILE=40000
+      LimitNPROC=1048576
+      ExecStartPre=/sbin/modprobe ip_tables
+      ExecStartPre=/usr/bin/mkdir -p /run/flannel
+      ExecStartPre=/usr/bin/touch /run/flannel/options.env
+      ExecStartPre=-/usr/bin/etcdctl mk /coreos.com/network/config '{"Network":"172.17.0.0/16", "SubnetLen": 20, "SubnetMin":"172.17.0.0","SubnetMax":"172.17.208.0", "Backend": {"Type": "vxlan"}}' || true
```
gabrtv: I believe we need to change the network range to avoid an overlap with our Vagrant address scheme: `172.17.8.0/24`

carmstrong:
@aledbf Is this safe to drop-in now? Does this break upgrades? Should we instead consider this for networking v2 (#3812) which will likely necessitate Deis v2.0?

aledbf:
@carmstrong I wish but unfortunately the deis-publisher and deis-controller must be changed:
- deis-publisher:
  - https://github.com/deis/deis/blob/master/publisher/server/publisher.go#L116

    change the ip to use the assigned in the overlay network and the port to use
```
        apiContainer, err := s.DockerClient.InspectContainer(container.ID)
        hostAndPort := apiContainer.NetworkSettings.IPAddress + ":" + port
```
```
        docker inspect  -f "{{ .NetworkSettings.Ports }}"
```
- deis-controller:
  - https://github.com/deis/deis/blob/master/controller/scheduler/fleet.py#L382

    remove -P from the template: it causes problems with the overlay network because it uses 0.0.0.0 for mapping

carmstrong:
We discussed today that this should land soon after v1.8 ships so that we can put it through its paces in master and ship in v1.9.

aledbf:
Maybe it could be a good idea to use the new `aws-vpc` [Amazon VPC route table](http://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/VPC_Route_Tables.html) option in v0.5.0 for when the cluster is running using Amazon as provider.

carmstrong:
@aledbf We'll want to rebase this and update the targets to `etcd2.service` once #4005 lands. We'll want to ship this immediately after so it can bake in master for 1.9.

krancour:
@aledbf why the 16 subnet / node limit?  That seems a bit restrictive.

aledbf:
@krancour because of the selected netmask (/20). I can change that to /21 so each subnet can hold 2048 IPs. Then the limit of nodes can be 32.

krancour:
I should have phrased the question differently.  Was there a specific reason for the /20 to begin with?

aledbf:
@krancour the only "reason" is because /19 is not big enough.

krancour:
@aledbf after our discussion yesterday, are we mutually comfortable using the defaults documented here instead?  (Overlay network that is /16 with subnets that are /24)  

https://coreos.com/flannel/docs/latest/flannel-config.html

That would allow us up to 256 nodes with up to 256 containers each-- and that seems pretty reasonable to me.  (Granted there are other things like etcd that may tip over at that scale, but I think this would prevent the overlay network itself from being the things that stops us from scaling.)

krancour:
Publishing seems not to be working.  I'm looking into that.

krancour:
This line is the culprit:

https://github.com/aledbf/deis/blob/overlay_network/publisher/server/publisher.go#L115

`PublicPort` seems to reference the host end of the port mapping.  Those port mappings are not longer included, so `PublicPort` resolves to 0.  `PrivatePort`, however, seems to correctly return whatever port the container exposes.

```
port := strconv.Itoa(int(p.PrivatePort))
```

With that change, this all seems to work flawlessly.

krancour:
> Does this break upgrades? Should we instead consider this for networking v2 (#3812) which will likely necessitate Deis v2.0?

Yes and yes.

publisher/server/publisher.go
```diff
@@ -112,8 +112,9 @@ func (s *Server) publishContainer(container *docker.APIContainers, ttl time.Dura
 		for _, p := range container.Ports {
 			// lowest port wins (docker sorts the ports)
 			// TODO (bacongobbler): support multiple exposed ports
-			port := strconv.Itoa(int(p.PublicPort))
-			hostAndPort := s.host + ":" + port
+			port := strconv.Itoa(int(p.PrivatePort))
+			apiContainer, _ := s.DockerClient.InspectContainer(container.ID)
+			hostAndPort := apiContainer.NetworkSettings.IPAddress + ":" + port
 			if s.IsPublishableApp(containerName) && s.IsPortOpen(hostAndPort) {
```
smothiki: in fleet.py container template still has -P so this change is not required

carmstrong:
> Does this break upgrades? Should we instead consider this for networking v2 (#3812) which will likely necessitate Deis v2.0?

I was under the impression that this PR was a prerequisite for the Kubernetes PR - is that not true? /cc @aledbf @kmala @smothiki

kmala:
@carmstrong Yes,this PR is a mandatory requirement for the kubernetes PR.

publisher/server/publisher.go
krancour: Let me look into that.  It definitely does not work without this change, but the -P that remains in fleet.py might be unnecessary now.

krancour:
@aledbf can you fill me in on why the port exposed by the builder changed from 22 to 2223 and builder now has host networking?

I _think_ I understand why it was done.  Correct me if I am wrong, but in the absence of the port mapping of 2223 on the host to 22 in the container (as it used to be before the overlay network) you wish to ensure that the builder is _still_ accessible via the host's port 2223.  Am I on the right track?

IMO, host networking is heavy-handed and should be avoided whenever possible.  Since the builder does get its own IP on the overlay network, just like every other container, do you think maybe it could make more sense to just modify how the builder advertises it's IP:port and how the router discovers it?

publisher/server/publisher.go
krancour: @smothiki the -P in fleet.py actually doesn't belong.

krancour:
> Since the builder does get its own IP on the overlay network, just like every other container, do you think maybe it could make more sense to just modify how the builder advertises it's IP:port... ?

^ Seems the same could be said for many other components as well.  Using `deis-controller` or `deis-database` as examples-- they are still advertising themselves as follows:

```
core@deis-03 ~ $ etcdctl get /deis/controller/host
172.17.8.101
core@deis-03 ~ $ etcdctl get /deis/controller/port
8000
```

Any other component connecting to the controller does so using that IP:port-- which also utilizes a port mapping of 8000 --> 80.  This is unnecessary if we take full advantage of the overlay network.  For instance, we could drop the port mapping and have controller advertise something like 10.244.something.something:80 instead.

aledbf:
> can you fill me in on why the port exposed by the builder changed from 22 to 2223 and builder now has host networking?

@krancour deis-builder exposes the port 22 so without a mapping it produces a conflict with the ssh daemon in the node. The only reason for this is to use the same IP from the node where is running. 
I don't see any problem if the builder uses one IP from the overlay network but if for this the boot script requires a change to get the assigned IP (now uses the value of the env variable HOST)

krancour:
> Does this break upgrades? Should we instead consider this for networking v2 (#3812) which will likely necessitate Deis v2.0?
> Yes and yes.

Defending that statement... upgrade from 1.8.0 didn't go smoothly.  Some of the changes to the components assume the existence of the overlay network.

I'm experimenting right now to see if there's a more minimal set of changes that could be adopted such that a fresh cluster can run flannel to satisfy the needs of the kubernetes preview, but also such that an existing cluster that doesn't pick up flannel from the updated cloud-config during upgrade in place can still continue to function with the 1.9 components.

krancour:
> exposes the port 22 so without a mapping it produces a conflict with the ssh daemon in the node

But only because we dropped to `--net=host`, right?  If we can avoid doing that there's no conflict.

krancour:
Consensus seems to be that we need to figure out how to make this not break upgrades.

contrib/coreos/user-data.example
```diff
@@ -113,9 +148,13 @@ write_files:
...
-        EnvironmentFile=/etc/environment_proxy
-        Environment="DOCKER_OPTS=--insecure-registry 10.0.0.0/8 --insecure-registry 172.16.0.0/12 --insecure-registry 192.168.0.0/16 --insecure-registry 100.64.0.0/10"
+      [Unit]
+      Requires=flanneld.service
+      After=flanneld.service
+
+      [Service]
+      EnvironmentFile=/etc/environment_proxy
+      Environment="DOCKER_OPTS=--insecure-registry 10.0.0.0/8 --insecure-registry 172.16.0.0/12 --insecure-registry 192.168.0.0/16 --insecure-registry 100.64.0.0/10 --iptables=false"
   - path: /run/deis/bin/get_image
```
smothiki: @aledbf should the docker opts contain -ip-masq=false as we already started flannel with --ip-masq=true

carmstrong:
Replaced by #4103

PR #3872, opened 2015-06-15
feat(router): add Diffie-Hellman parameter for DHE ciphersuites

wenzowski:
The openssl defaults for Ephemeral Diffie-Hellman exchange are considered by some to be inadequate.

Generate a 2048-bit key and upload to the cluster with:
```
openssl dhparam -out ./dhparam.pem 2048
deisctl config router set sslDhparam=./dhparam.pem
```

### DIFF
--- a/deisctl/config/config.go
+++ b/deisctl/config/config.go
@@ -13,7 +13,8 @@ import (
 var fileKeys = []string{
 	"/deis/platform/sshPrivateKey",
 	"/deis/router/sslCert",
-	"/deis/router/sslKey"}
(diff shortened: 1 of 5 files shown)


carmstrong:
Code LGTM.

technosophos:
So this will simply allow admins to define parameters for their Ephemeral DH generation? AFAIK, numbits is the only thing you can set for this, right? I'm good with that.

LGTM

wenzowski:
@technosophos yes numbits is only [required param](https://www.openssl.org/docs/apps/dhparam.html)

Issue #3723, opened 2015-05-19
Feature Request: Configurable Upchecks


jwaldrip:
Allow the configuration of "up checks" that will be called to confirm if a container is up during a deploy. The upcheck should be in the form of an http action and url and optional response substring.

`deis upcheck:set get /heartbeat "lubdub"`
`deis upcheck:unset`

carmstrong:
This sounds related to #1829 and #3674. @jwaldrip what do you think of the discussions in those issues?

jwaldrip:
1829 is canary deploys, I am not talking about doing that here. This is just true zero downtime deploys. Simple and straight forward.

carmstrong:
So to define this in terms of code/scope/work: you're proposing a per-app healthcheck that the controller would use to confirm a container is running beyond checking the controller state, correct?

PierreKircher:
@carmstrong the thing we see at brandfolder is an old version gets destroyed before a new one is up and running .. 

so publisher needs a check at least for status 200 ( and possible a regex body matcher configureable via etcd ofc ) before mapping that stuff to etcd ,, that prevents the issues seen in #3674

the other problem is that the controller tells the scheduler to delete the old version as soon as the new version is there even if it did not boot up sucessfull .. we could write a state with publisher to etcd .. and clean them up with a job in the controller instead of destroying them directly .. 

that way we might have 2 versions online for a few seconds but archice a true 0 downtime .. 

id like to hear thougths on that

carmstrong:
This is being tracked in #3813, which is a solution to address a number of issues for application health checks during deploy and publishing.

PR #3874, opened 2015-06-15
feat(publisher): check app config for HTTP health check

### LINKED ISSUE #3813: Application healthcheck endpoint for checks during deploy and publishing
During an application release/deploy, the container just checks to make sure that the fleet unit reports that it is running: https://github.com/deis/deis/blob/c2cffb969a1edaaa456033a05b7a6958eea02bad/controller/scheduler/fleet.py#L193

Just because a unit is running doesn't mean that it's ready/healthy.

Additionally, there is actually a bug in the way that publisher checks for applications to be listening before it publishes them - @gabrtv discovered that the Docker proxy responsible for routing requests from the host into containers will actually accept on the external port immediately (even if no daemon in the container is bound to that port). 

The proposal is to:
- Add a new `healthcheck` endpoint and CLI command to configure a healthcheck endpoint for the application
- Modify the controller so that during a deploy, controller will wait for each deployed container to report healthy if a healthcheck endpoint is configured for the application
- Modify the publisher to check for an appropriate (HTTP 200?) response from the healthcheck endpoint before publishing an app container if a healthcheck endpoint is configured for the application

This solution addresses several issues and feature requests surrounding application deploys and health checks in the router, including #3478, #3723, and #3674.

bacongobbler:
closes #3813 

Note that this is considered a breaking change as all apps are enforced to have a health check. By default this checks for a 200 OK response at the root URL, but users can configure this by setting HEALTHCHECK_URL and HEALTHCHECK_STATUS_CODE.

### DIFF
--- a/controller/api/models.py
+++ b/controller/api/models.py
@@ -1004,6 +1004,34 @@ def _etcd_purge_app(**kwargs):
         pass
 
 
+def _etcd_publish_config(**kwargs):
+    config = kwargs['instance']
+    # we purge all existing config when adding the newest instance. This is because
+    # deis config:unset would remove an existing value, but not delete the
+    # old config object
+    try:
+        _etcd_client.delete('/deis/services/{}/config'.format(config.app),
+                            prevExist=True, dir=True, recursive=True)
+    except KeyError:
+        pass
+    if kwargs['created']:
+        for k, v in config.values.iteritems():
+            _etcd_client.write(
+                '/deis/services/{}/config/{}'.format(
+                    config.app,
+                    unicode(k).encode('utf-8').lower()),
+                unicode(v).encode('utf-8'))
+
+
+def _etcd_purge_config(**kwargs):
+    config = kwargs['instance']
+    try:
+        _etcd_client.delete('/deis/services/{}/config'.format(config.app),
+                            prevExist=True, dir=True, recursive=True)
+    except KeyError:
+        pass
+
+
 def _etcd_publish_cert(**kwargs):
     cert = kwargs['instance']
     if kwargs['created']:
@@ -1069,3 +1097,5 @@ if _etcd_client:
     post_delete.connect(_etcd_purge_app, sender=App, dispatch_uid='api.models')
     post_save.connect(_etcd_publish_cert, sender=Certificate, dispatch_uid='api.models')
     post_delete.connect(_etcd_purge_cert, sender=Certificate, dispatch_uid='api.models')
+    post_save.connect(_etcd_publish_config, sender=Config, dispatch_uid='api.models')
+    post_delete.connect(_etcd_purge_config, sender=Config, dispatch_uid='api.models')
--- a/controller/api/models.py
+++ b/controller/api/models.py
@@ -299,6 +299,45 @@ class App(UuidAuditedModel):
         if set([c.state for c in to_add]) != set(['up']):
             err = 'warning, some containers failed to start'
             log_event(self, err, logging.WARNING)
+        # if the user specified a health check, try checking to see if it's running
+        try:
+            config = self.config_set.latest()
+            if 'HEALTHCHECK_URL' in config.values.keys():
+                # check the app to see if it's healthy
+                try:
+                    self._do_healthcheck(config)
+                except Exception as e:
+                    log_event(self, str(e), logging.ERROR)
+                    self._destroy_containers(to_add)
+                    raise
+        except Config.DoesNotExist:
+            pass
+
+    def _do_healthcheck(self, config):
+        timeout = time.time() + 60  # 1 minute from now
+        while True:
+            if time.time() > timeout:
+                raise RuntimeError(
+                    'app failed to respond to health check within 60 seconds of launch')
+            try:
+                response = requests.get(
+                    'http://{}{}'.format(
+                        self.url,
+                        config.values['HEALTHCHECK_URL']),
+                    timeout=5)
+                expected_status_code = config.values['HEALTHCHECK_STATUS_CODE'] if \
+                    'HEALTHCHECK_STATUS_CODE' in config.values.keys() else \
+                    settings.DEIS_HEALTHCHECK_STATUS_CODE
+                if str(response.status_code) != str(expected_status_code):
+                    err = "aborting, app failed health check (got '{}', expected: '{}')".format(
+                        response.status_code,
+                        expected_status_code)
+                    raise RuntimeError(err)
+                break
+            except requests.exceptions.ConnectionError:
+                continue
+            except requests.exceptions.Timeout:
+                continue
 
     def _restart_containers(self, to_restart):
         """Restarts containers via the scheduler"""
--- a/controller/deis/settings.py
+++ b/controller/deis/settings.py
@@ -290,6 +290,9 @@ DEIS_DATETIME_FORMAT = '%Y-%m-%dT%H:%M:%S%Z'
 # names which apps cannot reserve for routing
 DEIS_RESERVED_NAMES = ['deis']
 
+# the default expected status code for healthchecks
+DEIS_HEALTHCHECK_STATUS_CODE = '200'
+
 # default scheduler settings
 SCHEDULER_MODULE = 'scheduler.mock'
 SCHEDULER_TARGET = ''  # path to scheduler endpoint (e.g. /var/run/fleet.sock)
--- a/controller/api/tests/test_config.py
+++ b/controller/api/tests/test_config.py
@@ -15,14 +15,36 @@ from django.contrib.auth.models import User
 from django.test import TransactionTestCase
 from rest_framework.authtoken.models import Token
 
-from api.models import Config
+from api.models import App, Config
 
 
-def mock_import_repository_task(*args, **kwargs):
+def mock_status_ok(*args, **kwargs):
     resp = requests.Response()
     resp.status_code = 200
     resp._content_consumed = True
     return resp
+
+
+def mock_status_not_found(*args, **kwargs):
+    resp = requests.Response()
+    resp.status_code = 404
+    resp._content_consumed = True
+    return resp
+
+
+def mock_request_timed_out(*args, **kwargs):
+    raise requests.exceptions.Timeout()
+
+
+def mock_request_connection_error(*args, **kwargs):
+    raise requests.exceptions.ConnectionError()
+
+
+def mock_time(*args, **kwargs):
+    if not hasattr(mock_time, "counter"):
+        mock_time.counter = 0  # it doesn't exist yet, so initialize it
+    mock_time.counter += 1
+    return mock_time.counter
 
 
 class ConfigTest(TransactionTestCase):
@@ -35,7 +57,7 @@ class ConfigTest(TransactionTestCase):
         self.user = User.objects.get(username='autotest')
         self.token = Token.objects.get(user=self.user).key
 
-    @mock.patch('requests.post', mock_import_repository_task)
+    @mock.patch('requests.post', mock_status_ok)
     def test_config(self):
         """
         Test that config is auto-created for a new app and that
@@ -107,7 +129,7 @@ class ConfigTest(TransactionTestCase):
         self.assertEqual(response.status_code, 405)
         return config5
 
-    @mock.patch('requests.post', mock_import_repository_task)
+    @mock.patch('requests.post', mock_status_ok)
     def test_response_data(self):
         """Test that the serialized response contains only relevant data."""
         body = {'id': 'test'}
@@ -132,7 +154,7 @@ class ConfigTest(TransactionTestCase):
         }
         self.assertDictContainsSubset(expected, response.data)
 
-    @mock.patch('requests.post', mock_import_repository_task)
+    @mock.patch('requests.post', mock_status_ok)
     def test_config_set_same_key(self):
         """
         Test that config sets on the same key function properly
@@ -156,7 +178,7 @@ class ConfigTest(TransactionTestCase):
         self.assertIn('PORT', response.data['values'])
         self.assertEqual(response.data['values']['PORT'], '5001')
 
-    @mock.patch('requests.post', mock_import_repository_task)
+    @mock.patch('requests.post', mock_status_ok)
     def test_config_set_unicode(self):
         """
         Test that config sets with unicode values are accepted.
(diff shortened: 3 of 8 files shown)


bacongobbler:
Still needs functional tests, but wanted to make sure that this is the right design decision.

carmstrong:
> Note that this is considered a breaking change as all apps are enforced to have a health check.

I'm pretty sure we wanted this to be backwards-compatible. The check should look like this:

- If an app has a healthcheck configured:
  - Check the healthcheck

We should also check the healthcheck endpoint in the controller during a deploy so that this serves @jwaldrip's use case in #3723 - if the healthcheck fails, the new deploy should be considered a failure and we should roll back. It's not enough just to check that the containers are running.

Please re-read #3813 and ensure we're addressing everything in there. If we're missing something, we should update that issue as appropriate so that this PR truly addresses all concerns in that issue.

/cc @gabrtv @technosophos

bacongobbler:
OK I'll see what I can do about that

technosophos:
So in this case, the healthcheck is only performed if the upcheckUrl is set for the given app? This makes sense to me. And it looks like the current version (00a29d90fed004277d7e6a7a9aa4f2f4a9055e41) no longer breaks backward compatibility, right?

bacongobbler:
@technosophos correct. The idea is that the health check will only occur if the user opts into the feature, which then we can health check to our heart's content.

bacongobbler:
Just as a personal reminder, if #3692 is merged before this one then I'll need to add a call to `._destroy_containers()` if the health check fails on the container's end.

bacongobbler:
Well that's a funny error:

```
config:set FOO=讲台 --app=configsample
Creating config...  ... 500 INTERNAL SERVER ERROR
<h1>Server Error (500)</h1>
```

Logs following:

```
ERROR Internal Server Error: /v1/apps/configsample/config
  File "/usr/lib/python2.7/site-packages/django/core/handlers/base.py", line 112, in get_response
  File "/usr/lib/python2.7/site-packages/django/views/decorators/csrf.py", line 57, in wrapped_view
  File "/usr/lib/python2.7/site-packages/rest_framework/viewsets.py", line 85, in view
  File "/usr/lib/python2.7/site-packages/rest_framework/views.py", line 407, in dispatch
  File "/usr/lib/python2.7/site-packages/rest_framework/views.py", line 404, in dispatch
  File "/app/api/views.py", line 124, in create
    return super(AppResourceViewSet, self).create(request, **kwargs)
  File "/app/api/views.py", line 101, in create
    return super(BaseDeisViewSet, self).create(request, *args, **kwargs)
  File "/usr/lib/python2.7/site-packages/rest_framework/mixins.py", line 21, in create
    self.perform_create(serializer)
  File "/app/api/viewsets.py", line 20, in perform_create
    obj = serializer.save(owner=self.request.user)
  File "/usr/lib/python2.7/site-packages/rest_framework/serializers.py", line 164, in save
    self.instance = self.create(validated_data)
  File "/usr/lib/python2.7/site-packages/rest_framework/serializers.py", line 768, in create
    instance = ModelClass.objects.create(**validated_data)
  File "/usr/lib/python2.7/site-packages/django/db/models/manager.py", line 157, in create
    return self.get_queryset().create(**kwargs)
  File "/usr/lib/python2.7/site-packages/django/db/models/query.py", line 322, in create
    obj.save(force_insert=True, using=self.db)
  File "/app/api/models.py", line 692, in save
    return super(Config, self).save(**kwargs)
  File "/usr/lib/python2.7/site-packages/django/db/models/base.py", line 545, in save
    force_update=force_update, update_fields=update_fields)
  File "/usr/lib/python2.7/site-packages/django/db/models/base.py", line 582, in save_base
    update_fields=update_fields, raw=raw, using=using)
  File "/usr/lib/python2.7/site-packages/django/dispatch/dispatcher.py", line 185, in send
    response = receiver(signal=self, sender=sender, **named)
  File "/app/api/models.py", line 1028, in _etcd_publish_config
    _etcd_client.write('/deis/services/{}/config/{}'.format(config.app, k.lower()), v)
  File "/usr/lib/python2.7/site-packages/etcd/client.py", line 269, in write
    response = self.api_execute(path, method, params=params)
  File "/usr/lib/python2.7/site-packages/etcd/client.py", line 570, in api_execute
    redirect=self.allow_redirect)
  File "/usr/lib/python2.7/site-packages/urllib3/request.py", line 133, in request_encode_body
    body, content_type = urlencode(fields), 'application/x-www-form-urlencoded'
  File "/usr/lib/python2.7/urllib.py", line 1338, in urlencode
    v = quote_plus(str(v))
UnicodeEncodeError: 'ascii' codec can't encode characters in position 0-1: ordinal not in range(128)
```

I guess that this means I need to do something like https://github.com/deis/deis/pull/2813 for the new _etcd_publish_config() function.

controller/api/models.py
```diff
@@ -298,6 +298,24 @@ def _start_containers(self, to_add):
         if set([c.state for c in to_add]) != set(['up']):
             err = 'warning, some containers failed to start'
             log_event(self, err, logging.WARNING)
+        # if the user specified a health check, try checking to see if it's running
+        try:
```
carmstrong: Should we add some sort of sleep/retry logic here? I could see apps that take a while to start up result in failure each time because the healthchecks don't respond quickly enough.

bacongobbler: I think setting a [request timeout](http://requests.readthedocs.org/en/latest/user/quickstart/#timeouts) of 60 seconds [a la Heroku](https://devcenter.heroku.com/articles/error-codes#r10-boot-timeout) would make sense. I don't believe we should add some retry logic though as the first response should be enough to determine if the app is responding as expected.

carmstrong: Are we sure about that? I think one of the issues enumerated in #3813 is that the Docker proxy accepts connections before the container is even listening. We should test this to make sure that the first connect will block until we actually get a response, either good or bad.

mboersma:
>  I'll need to add a call to ._destroy_containers()

ping.

controller/api/models.py
bacongobbler: Good call. Turns out that the proxy will respond and result in an aborted request, so I guess this would have to be re-written to keep trying for 60 seconds until the connection is not reset.

```
docker@boot2docker:~$ docker run -dp 8080:8080 deis/example-go bash -c 'sleep 60 && /go/bin/example-go'
6a24ec14c20663828c155a1e08d591b3e12d7e5d072782b78f31c7d0a8365d2e
docker@boot2docker:~$ docker run -v $PWD:/app -e HOST=192.168.59.104 python:2.7 bash -c 'pip install requests && python /app/test.py'
Collecting requests
  Downloading requests-2.7.0-py2.py3-none-any.whl (470kB)
Installing collected packages: requests
Successfully installed requests-2.7.0
  File "/app/test.py", line 4, in <module>
    response = requests.get('http://{}:8080'.format(os.environ['HOST']), timeout=60)
  File "/usr/local/lib/python2.7/site-packages/requests/api.py", line 69, in get
    return request('get', url, params=params, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/requests/api.py", line 50, in request
    response = session.request(method=method, url=url, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/requests/sessions.py", line 465, in request
    resp = self.send(prep, **send_kwargs)
  File "/usr/local/lib/python2.7/site-packages/requests/sessions.py", line 573, in send
    r = adapter.send(request, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/requests/adapters.py", line 415, in send
    raise ConnectionError(err, request=request)
requests.exceptions.ConnectionError: ('Connection aborted.', error(104, 'Connection reset by peer'))
```

Script to test was as follows:

```
import os
import requests

response = requests.get('http://{}:8080'.format(os.environ['HOST']), timeout=60)

print response
```

bacongobbler:
I've addressed all the comments, but it seems like I've hit a (related-yet-unrelated) router issue: https://gist.github.com/bacongobbler/ff38c24b08c438803544

I'll patch that up as it blocks this PR.

bacongobbler:
fixed! It was an error with the publisher health checks... :sheep:

controller/deis/settings.py
```diff
@@ -290,6 +290,9 @@
 # names which apps cannot reserve for routing
 DEIS_RESERVED_NAMES = ['deis']
 
+# the default expected status code for healthchecks
+DEIS_HEALTHCHECK_STATUS_CODE = '200'
```
carmstrong: Should we also put this in the confd template for the settings so it can be overridden by setting an etcd key (like `/deis/controller/healthcheckDefaultStatusCode`)?

bacongobbler: I made a mini comment about this in the commit:

> By default, the expected status code is 200 unless the user specifies
otherwise. This variable is non-configurable on the platform side as
both the controller and the publisher have to rely on this value, and I
could not find a happy medium for setting up a shared config variable.

Since it affects two components, I couldn't figure out a happy medium, so I left that part out :)

carmstrong: Ok. I think that's fine - if someone really wants this, they can use a custom component or figure out how to customize the setting in the general case.

carmstrong:
@bacongobbler Is this still a WIP or is it ready for review? I kicked off another Jenkins run.

bacongobbler:
There's still some WIP on this because I can't seem to get the darned unicode parsing to work.

mboersma:
I can recreate this failure in isolation after I do `from __future__ import unicode_literals` as models.py does. But removing that import, and/or not casting to unicode, doesn't help. The real problem seems to be an assumption in python's `urllib` that it's dealing with an ASCII string:
```
Jun 25 17:51:54 deis-03 sh[19730]: File "/usr/lib/python2.7/site-packages/urllib3/request.py", line 133, in request_encode_body
Jun 25 17:51:54 deis-03 sh[19730]: body, content_type = urlencode(fields), 'application/x-www-form-urlencoded'
Jun 25 17:51:54 deis-03 sh[19730]: File "/usr/lib/python2.7/urllib.py", line 1338, in urlencode
Jun 25 17:51:54 deis-03 sh[19730]: v = quote_plus(str(v))
Jun 25 17:51:54 deis-03 sh[19730]: UnicodeEncodeError: 'ascii' codec can't encode characters in position 0-1: ordinal not in range(128)
```

bacongobbler:
we eventually found the issue. It turns out that urllib doesn't accept unicode strings as input. Encoding the string from unicode to utf-8 resolves it. Once jenkins goes green this is ready for review

controller/api/models.py
```diff
@@ -299,6 +299,45 @@ def _start_containers(self, to_add):
...
+            config = self.config_set.latest()
+            if 'HEALTHCHECK_URL' in config.values.keys():
+                # the app has 60 seconds to connect
+                num_retries = 60
+                for _ in xrange(num_retries):
```
mboersma: This could potentially loop 60 times with a 60s timeout. Should `num_retries` be used for both values here? An hour seems excessive, even if it's unlikely.

bacongobbler: Potentially, yes. In reality the request won't time out because the docker userland proxy will return a ConnectionError. In this case, I'm exploiting the timeout in the unit tests by raising a Timeout exception so we won't call `time.sleep(1)`, making the unit tests block for 60 seconds. However, I could always just stub out `time.sleep()` with the mock library as I'm doing right now, so that would remove this issue. Does that seem fair?

mboersma: > stub out time.sleep() with the mock library

If that's local to the test and does the job, it sounds preferable.

carmstrong:
I'd like us to bang on this pretty hard when testing the upgrade scenario to 1.8 -- anything that affects users' deployed applications means careful testing is warranted.

I'd like to have a handful of apps on a cluster scaled to a number of containers, then upgrade to 1.8 and ensure they're still publishing and happy. Then, configure healthchecks on them and deploy a few new applications both with and without healthchecks.

controller/api/models.py
carmstrong: @bacongobbler Did we make this change yet?

controller/api/models.py
```diff
@@ -1069,3 +1134,5 @@ def create_auth_token(sender, instance=None, created=False, **kwargs):
     post_delete.connect(_etcd_purge_app, sender=App, dispatch_uid='api.models')
     post_save.connect(_etcd_publish_cert, sender=Certificate, dispatch_uid='api.models')
     post_delete.connect(_etcd_purge_cert, sender=Certificate, dispatch_uid='api.models')
+    post_save.connect(_etcd_publish_config, sender=Config, dispatch_uid='api.models')
```
carmstrong: @bacongobbler We should also update http://docs.deis.io/en/latest/managing_deis/upgrading-deis/#triggering-an-upgrade as part of this PR since we'll now need to republish Config objects after an upgrade.

bacongobbler: good catch!

carmstrong: Sorry, wrong link: http://docs.deis.io/en/latest/managing_deis/backing_up_data/#finishing-up

controller/api/models.py
bacongobbler: just now, yes :)

controller/api/models.py
bacongobbler: done

bacongobbler:
This is ready for another round of reviews

carmstrong:
Code LGTM. Weird that we don't see Jenkins triggering another build, though...

mboersma:
Code LGTM.

bacongobbler:
should this be manually tested before merging or shall we exercise this during the release testing? I've tested this manually and the base case seems to work but I'm not too sure about the Config publish/purge stuff in the controller, since that's untested.

carmstrong:
> should this be manually tested before merging or shall we exercise this during the release testing?

All PRs should be fully tested before merging. I don't want to sacrifice the stability of master for the sake of shipping a release.

> I've tested this manually and the base case seems to work but I'm not too sure about the Config publish/purge stuff in the controller, since that's untested.

We should definitely be testing base and edge/extreme cases for all PRs as they're authored.

bacongobbler:
ok, I'll test that and make sure nothing pops up.

bacongobbler:
I found that the config is not removed from etcd after a config:unset, so I'll add some integration tests and a fix for that.

bacongobbler:
ping @carmstrong @mboersma for another round of reviews

mboersma:
Code LGTM.

carmstrong:
Code LGTM. @bacongobbler have you tested this in various edge cases? Are you confident in it? If so, we can merge it. I don't want basic bugs cropping up and restarting our release testing tomorrow, although almost certainly we'll run into a minor issue somewhere.

bacongobbler:
Yes, I've tested it in various edge cases, up to and including

 - config:set HTTP_HEALTHCHECK_URL twice, overwriting the previous value
 - config:unset HTTP_HEALTHCHECK_URL
 - deis destroy removing config values

That should cover the most various edge cases, so I'm confident in this. Merging!

laurrentt:
I wrote something similar in the controller to push configs named WHITELIST to etcd. I'm glad to see I won't be needing this anymore. Do you think anyone would be interested by a per-app IP whitelisting feature in the router ?

PR #3881, opened 2015-06-16
fix(builder): add "z" flag for compressed tar archives

### LINKED ISSUE #3871: Deploy of a new version fails with: "tar: invalid tar magic " after Deis update 1.5 to 1.7
Today we updated our development platform form deis 1.5 to 1.7.2.
After the update the push of play  application fails  after image is built with:

Successfully built 8aa9c3adcbbe
-----> Pushing image to private registry

tar: invalid tar magic

Adding a Procfile does not mitigate the problem.

Play Version 2.4 , Jvm 1.8, deis deployed on Ec2.

mboersma:
When a Procfile isn't present in a `git push` operation, builder tries to extract or fabricate one, but the tar commands would fail with "invalid tar magic" since the compression flag "z" was missing.

I tested this by removing the Procfile from [example-play]() and committing it before `git push`. It failed before this change, but afterward the Procfile was correctly filled in (here's with debug output):
```console
Successfully built 39111ec4e108
+ puts-step 'Pushing image to private registry'
+ echo '-----> Pushing image to private registry'
-----> Pushing image to private registry
+ docker push 172.17.8.100:5000/ultima-underdog:git-f67f5ca1
+ echo

+ '[' -f Procfile ']'
+ '[' -f /home/git/ultima-underdog.git/build/tmp.B9dJa9cCgo/slug.tgz ']'
+ tar -tzf /home/git/ultima-underdog.git/build/tmp.B9dJa9cCgo/slug.tgz ./Procfile
++ extract-types
++ tar --to-stdout -xzf /home/git/ultima-underdog.git/build/tmp.B9dJa9cCgo/slug.tgz ./.release
+ PROCFILE='{"web":"play run --http.port=$PORT $PLAY_OPTS"}'
+ puts-step 'Launching... '
+ echo '-----> Launching... '
-----> Launching... 
```

Closes #3871.

### DIFF
--- a/builder/rootfs/etc/confd/templates/builder
+++ b/builder/rootfs/etc/confd/templates/builder
@@ -140,11 +140,11 @@ if [ -f Procfile ]; then
 elif [ -f $TMP_DIR/slug.tgz ]; then
     # Sometimes, the buildpack will generate a Procfile instead of populating /bin/release
     # /bin/release was unofficially deprecated for declaring default process types
-    if tar -tf $TMP_DIR/slug.tgz ./Procfile &> /dev/null;
+    if tar -tzf $TMP_DIR/slug.tgz ./Procfile &> /dev/null;
     then
-        PROCFILE="$(tar --to-stdout -xf $TMP_DIR/slug.tgz ./Procfile | yaml2json-procfile)"
+        PROCFILE="$(tar --to-stdout -xzf $TMP_DIR/slug.tgz ./Procfile | yaml2json-procfile)"
     else
-        PROCFILE=$(tar --to-stdout -xf $TMP_DIR/slug.tgz ./.release | extract-types)
+        PROCFILE=$(tar --to-stdout -xzf $TMP_DIR/slug.tgz ./.release | extract-types)
     fi
 else
     PROCFILE="{}"


bacongobbler:
Ah! I knew there was a reason behind this. A long time ago there we removed the `pigz` package from slugbuilder, causing the "slug.tgz" to be a tar file. Looks like https://github.com/deis/deis/commit/835774225cd27d64425f23727075833d8cbc2b44 still relied on the release being a tar file rather than a tarball, which I'm surprised this error never came up before.

Code LGTM!

mboersma:
> Looks like 8357742 still relied on the release being a tar file rather than a tarball

Ah, that actually makes sense now. Glad to have an explanation and hopefully a fix.

carmstrong:
Code LGTM. Are there any issues with upgrades/backwards compatibility?

mboersma:
> Are there any issues with upgrades/backwards compatibility?

#2218 and progrium/cedarish#9 and #2915 make it appear that this is a change Deis should have made a while ago, so I don't think this creates new backward compatibility issues. @bacongobbler?

bacongobbler:
^^

carmstrong:
Code LGTM.

carmstrong:
> is it not critic enough to bump to 1.7.4 ?

I don't think users run into this unless they don't have a Procfile, right? Maybe I'm misunderstanding how users can run into this.

mboersma:
> I don't think users run into this unless they don't have a Procfile, right?

That's correct, and the workaround is just to commit a Procfile and `git push` again. So I know it's a PITA @lorieri but unfortunately this bug has been in Deis for a couple releases, and probably doesn't quite rise to level of a patch release (unless the other maintainers disagree).

bacongobbler:
I agree. There's a workaround and it'll get fixed in v1.8 so I don't think a patch release is **completely** necessary.

lorieri:
Ok :)

PR #3885, opened 2015-06-17
Fleet: remove containers on exit

jwaldrip:


deis-admin:

controller/scheduler/fleet.py
```diff
@@ -379,7 +379,7 @@ def attach(self, name):
     {"section": "Unit", "name": "Description", "value": "{name}"},
     {"section": "Service", "name": "ExecStartPre", "value": '''/bin/sh -c "IMAGE=$(etcdctl get /deis/registry/host 2>&1):$(etcdctl get /deis/registry/port 2>&1)/{image}; docker pull $IMAGE"'''},  # noqa
     {"section": "Service", "name": "ExecStartPre", "value": '''/bin/sh -c "docker inspect {name} >/dev/null 2>&1 && docker rm -f {name} || true"'''},  # noqa
-    {"section": "Service", "name": "ExecStart", "value": '''/bin/sh -c "IMAGE=$(etcdctl get /deis/registry/host 2>&1):$(etcdctl get /deis/registry/port 2>&1)/{image}; docker run --name {name} {memory} {cpu} {hostname} -P $IMAGE {command}"'''},  # noqa
+    {"section": "Service", "name": "ExecStart", "value": '''/bin/sh -c "IMAGE=$(etcdctl get /deis/registry/host 2>&1):$(etcdctl get /deis/registry/port 2>&1)/{image}; docker run --rm --name {name} {memory} {cpu} {hostname} -P $IMAGE {command}"'''},  # noqa
```
bacongobbler: I can see that this is necessary for the run template, but in this one we have an ExecStop operation that calls `/usr/bin/docker rm -f {name}` on line 384. Doesn't that accomplish the same thing here?

carmstrong:
For some background, we were discussing this in IRC last night: https://botbot.me/freenode/deis/2015-06-17/?msg=42064022&page=1

Apparently, our `ExecStop` isn't really doing anything, and there are a lot of orphaned containers laying about. This could certainly impact the "out of disk space" issues folks are seeing.

We should investigate by doing some scaling operations and seeing what the state of stale containers is, then testing this change everywhere (including for Deis components themselves) and sees if that keeps things cleaner.

bacongobbler:
I believe the reason why we keep run containers around is due to the `deis run` scheduler logic for fleet. After the container is done, it SSH's into the box and inspects the container's output and return code. We should probably do the cleanup there instead of after the container is finished running.

bacongobbler:
> Apparently, our ExecStop isn't really doing anything, and there are a lot of orphaned containers laying about. This could certainly impact the "out of disk space" issues folks are seeing.

In that case, I guess we should remove line 384, add in the `--rm` flag to the container template, and shift over the operation to remove the container after [the run operation is done with the container](https://github.com/deis/deis/blob/428ecc5990acff7741372026c74a3e8bdd93c35f/controller/scheduler/fleet.py#L332-L334). @jwaldrip would you be up for that?

carmstrong:
I reimplemented this as #3918. Thanks @jwaldrip!

PR #3887, opened 2015-06-17
tests(deisctl): Add tests to backend package

Joshua-Anderson:


bacongobbler:
I'm gonna need a :coffee: to review this, but I'm all for adding tests :fireworks:

deisctl/backend/fleet/unit.go
```diff
@@ -63,8 +56,13 @@ func (c *FleetClient) lastUnit(component string) (num int, err error) {
 
 // NewUnit takes a component type and returns a Fleet unit
 // that includes the relevant systemd service template
-func NewUnit(component string) (uf *unit.UnitFile, err error) {
-	template, err := readTemplate(component)
+func NewUnit(c *FleetClient, component string) (uf *unit.UnitFile, err error) {
+	uf, err = doNewUnit(component, c.templatePaths)
```
bacongobbler: If NewUnit requires the client's templatePaths field, I feel like we should be making this a FleetClient function rather than a package function. That way we don't have to do this roundabout `doNewUnit` function. Any thoughts on this?

deisctl/backend/fleet/journal_test.go
```diff
@@ -0,0 +1,75 @@
...
+			MachineID:    "test-1",
+		},
+	}
+
+	c := &FleetClient{Fleet: &stubFleetClient{testUnits: testUnits, testMachineStates: testMachines,
+		unitsMutex: &sync.Mutex{}}}
+
+	testWriter := bytes.Buffer{}
+
+	err := doJournal(c, "router", &testWriter, mockJournalCommandRunner{validUnits: []string{"deis-router@1.service", "deis-router@2.service"}})
```
bacongobbler: I feel like we should be testing the public functions like `c.Journal()` rather than `doJournal` for increased coverage. The tests should cover how a user would typically use this library to consume fleet, IMO :)

bacongobbler: (in this case the consumer would be `deisctl` itself)

Joshua-Anderson: I created the `do*` methods because I needed dependency injection. Functions like journal called things that wrote to `STDOUT` or tried to `SSH` into a machine. These actions make sense, but you can't test `STDOUT` or deisctl actually trying to ssh into a machine. So I created `doJournal` which allows me to replace `STDOUT` and `ssh` with stubs that can actually be tested. I then had `Journal` call `doJournal` with `STDOUT` and `ssh`, keeping the public api the same.

bacongobbler: Would it make more sense to make the writer and the  as struct fields of FleetClient instead of part of the method signature? I can't help but feel that we could simplify the logic by introducing a `Runner` field which is of type `commandRunner` as well as an `Output` field of type `io.Writer`. That way your stubFleetClient could just specify the `Runner` as type `mockCommandRunner` and call `.Journal()` directly.

Joshua-Anderson: Just about to suggest that :smile:. I'll talk to you about this over chat.

mboersma:
If you're updating this PR, please amend the [commit message](http://docs.deis.io/en/latest/contributing/standards/#commit-style) to use the category `test` and start with a lower-case noun, like:
```
test(deisctl): add unit tests to backend package
```
:japanese_ogre:

Joshua-Anderson:
@bacongobbler Done!

mboersma:
Code LGTM. I like the refactoring to allow mocks. I ran the tests locally and tested the `deisctl` binary. Nice work! :green_heart:

carmstrong:
Deferring to @technosophos or @bacongobbler for another code review on this one.

bacongobbler:
phew... took me a while, but code LGTM

carmstrong:
Good enough for me. Great work, @Joshua-Anderson!

PR #3890, opened 2015-06-17
ref(Makefiles): use wildcard for GO_FILES

mboersma:
Many components have a Makefile with `GO_FILES` hard-coded to the list of .go files in the current directory. Using the `wildcard` func will stop the need for this maintenance.

I noticed we weren't linting `domains_test.go` and `users_test.go`, then decided to fix it in general rather than just add two more files to tests/Makefile.


bacongobbler:
LGTM :)

carmstrong:
Code LGTM.

PR #3895, opened 2015-06-18
fix(deisctl): Fix panic when deisctl is called with no arguments

Joshua-Anderson:
Note, `deisctl` now returns `0` when called with no arguments rather than `1` as it previously had.


Joshua-Anderson:
Closing in favor of a more comprehensive PR

PR #3898, opened 2015-06-18
ref(deisctl): clean up some parsing and error behavior

Joshua-Anderson:
Note: `./deisctl` will trigger a docopt error because it wasn't given any command line arguments. However, this error is empty, so go just prints an empty line. I think it looks better, but it is a slight change in behavior.  The advantage is that if docopt has an actual error, instead of throwing it away like we currently do we print it.

Before:

``` shell
$ deisctl
Usage: deisctl [options] <command> [<args>...]
$
```

After:

``` shell
$ deisctl
Usage: deisctl [options] <command> [<args>...]

$
```

### DIFF
--- a/deisctl/deisctl.go
+++ b/deisctl/deisctl.go
@@ -60,15 +60,16 @@ Options:
 	argv, helpFlag := parseArgs(argv)
 	// give docopt an optional final false arg so it doesn't call os.Exit()
 	args, err := docopt.Parse(usage, argv, false, version.Version, true, false)
-	if err != nil || len(args) == 0 {
-		if helpFlag {
-			fmt.Print(usage)
-			return 0
-		} else if argv[0] == "--version" {
-			return 0
-		}
+
+	if err != nil && err.Error() != "" {
+		fmt.Println(err)
 		return 1
 	}
+
+	if len(args) == 0 {
+		return 0
+	}
+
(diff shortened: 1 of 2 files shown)


bacongobbler:
Nice cleanup, though for now I wouldn't print the nil error. If we bump docopt and it prints an error message, we can update the code but for now let's keep it neat and pretty :)

The rest is covered by unit tests (including the `--help` flag stuff), so I'm confident that this won't break backwards compatibility. I'm assuming that docopt is already doing this for us?

Joshua-Anderson:
@bacongobbler The problem is that `docopt` can return errors with text. It's just that the most common error (if no args are passed to deisctl `docopt`) docop throws an error of `""`.

I see these options:

1. Don't print docopt errors, which might mask some legitimate errors.
2. Print errors, but this prints the `""` error of no arguments.
3. We check if the error is a null string and if it is don't print it.

bacongobbler:
> We check if the error is a null string and if it is don't print it.

I like that idea. If docopt isn't giving us any information then there's nothing to report!

Joshua-Anderson:
@bacongobbler Done!

mboersma:
Code LGTM.

bacongobbler:
code LGTM

PR #3900, opened 2015-06-18
feat(router): enable OCSP stapling by default

wenzowski:
This allows nginx to cache the [OCSP] response from the Certificate Authority
and return a signed copy of it to the User Agent. This reduces the
number of roundtrips necessary to verify the certificate and therefore
results in both a performance improvement and reduced load on the OCSP
server.

If an invalid response is stapled and sent to the client it is ignored
so disabling verification by default does no harm.

This currently requires `/deis/router/ocsp/responder` to be manually set to the
IP Address of the certificate's OCSP server. This will not be necessary if
a dns resolver is configured so nginx can look up the domain name specified
by the certificate.

[OCSP]: https://tools.ietf.org/html/rfc2560

### DIFF
--- a/router/image/templates/nginx.conf
+++ b/router/image/templates/nginx.conf
@@ -39,6 +39,10 @@ http {
     gzip_vary {{ or (getv "/deis/router/gzipVary") "on" }};
     {{ end }}
 
+    # DNS servers to query for address resolution. Needed to resolve OCSP responder.
+    resolver 8.8.8.8 8.8.4.4 valid=300s;
+    resolver_timeout 5s;
+
     {{ $useFirewall := or (getv "/deis/router/firewall/enabled") "false" }}{{ if eq $useFirewall "true" }}# include naxsi rules
     include     /opt/nginx/firewall/naxsi_core.rules;
     include     /opt/nginx/firewall/web_apps.rules;
--- a/router/image/templates/nginx.conf
+++ b/router/image/templates/nginx.conf
@@ -202,6 +202,11 @@ http {
         ssl_ciphers '{{ getv "/deis/router/sslCiphers" }}';
         ssl_prefer_server_ciphers on;
         {{ end }}
+        ssl_stapling {{ or (getv "/deis/router/ocsp/stapling") "on" }};
+        ssl_stapling_verify {{ or (getv "/deis/router/ocsp/verify") "off" }};
+        {{ if exists "/deis/router/ocsp/responder" }}
+        ssl_stapling_responder {{ getv "/deis/router/ocsp/responder" }};
+        {{ end }}
         {{/* if there's no app SSL cert but we have a router SSL cert, enable that instead */}}
         {{/* TODO (bacongobbler): wait for https://github.com/kelseyhightower/confd/issues/270 */}}
         {{/* so we can apply this config to just subdomains of the platform domain. */}}
--- a/router/boot.go
+++ b/router/boot.go
@@ -60,6 +60,7 @@ func main() {
 	mkdirEtcd(client, "/deis/certs")
 	mkdirEtcd(client, "/deis/router/hosts")
 	mkdirEtcd(client, "/deis/router/hsts")
+	mkdirEtcd(client, "/deis/router/ocsp")
 
 	setDefaultEtcd(client, etcdPath+"/gzip", "on")
 
--- a/router/image/templates/deis.conf
+++ b/router/image/templates/deis.conf
@@ -14,4 +14,9 @@ ssl_protocols TLSv1 TLSv1.1 TLSv1.2;
 ssl_ciphers '{{ getv "/deis/router/sslCiphers" }}';
 ssl_prefer_server_ciphers on;
 {{ end }}
+ssl_stapling {{ or (getv "/deis/router/ocsp/stapling") "on" }};
+ssl_stapling_verify {{ or (getv "/deis/router/ocsp/verify") "off" }};
+{{ if exists "/deis/router/ocsp/responder" }}
+ssl_stapling_responder {{ getv "/deis/router/ocsp/responder" }};
+{{ end }}
 {{ end }}


wenzowski:
2nd commit adds [Google Public DNS](https://developers.google.com/speed/public-dns/docs/using) as the default resolver. Will that work for now?

aledbf:
@wenzowski why not just use the `ssl_stapling_file` and avoid all the dns queries?

wenzowski:
@aledbf because the stapled ocsp response generally needs to be refreshed at regular intervals.

It's not at all uncommon for a certificate authority to return an ocsp response that is valid for less than a week. The shorter the validity period the more likely a revoked certificate will be caught, so shorter is better. By allowing nginx to fetch the stapling file itself no manual intervention is necessary for the stapled response to be kept up to date.

aledbf:
@wenzowski you are right but using digicert certificates for instance the ocsp is validation mess, most of the time nginx ends with a timeout and then you need to change the dns query times in nginx.

wenzowski:
I think I accounted for that with an explicit dns resolver timeout of 5s. Unless you mean the http request to the digicert ocsp server times out? If their ocsp responder is over capacity then ocsp stapling seems critical. Validation is off by default in this PR since `ssl_trusted_certificate` for each keypair would be necessary for most users.

It sounds like there's a use case for an etcd key for the `ssl_stapling_file`.

aledbf:
> Unless you mean the http request to the digicert ocsp server times out?

Yes, and that introduces latency processing the request

wenzowski:
If it delays the nginx response that's a bug. It's supposed to be non-blocking. So the first request each router instance responds to gets returned without a stapled ocsp even though the flag is turned on. Unless there is a bug, nginx continues to respond without a stapled ocsp until it receives one from the certificate authority. And when the stapled ocsp is missing or invalid the browser takes care of querying the endpoint specified by the cert itself. And depending on the browser _it_ can block until it receives a valid response but nginx shouldn't.

router/image/templates/nginx.conf
```diff
@@ -39,6 +39,10 @@ http {
     gzip_vary {{ or (getv "/deis/router/gzipVary") "on" }};
     {{ end }}
 
+    # DNS servers to query for address resolution. Needed to resolve OCSP responder.
+    resolver 8.8.8.8 8.8.4.4 valid=300s;
```
carmstrong: I'm concerned with hard-coding this.. I could imagine scenarios where users want to use an internal DNS server or use clusters that are completely cut-off from the internet, and this would break.

aledbf:
@wenzowski this is what I use to generate the stapling file with a crontab so I can avoid any external request:

```
./ocsp.sh ./DigiCertCASHA2.pem <path to the certificate>
<certificate>: good
	This Update: Jun 18 21:15:00 2015 GMT
	Next Update: Jun 25 21:30:00 2015 GMT

------
#!/bin/sh
ISSUER_CER=$1
SERVER_CER=$2

URL=$(openssl x509 -in $SERVER_CER -text | grep "OCSP – URI:" | cut -d: -f2,3)
openssl ocsp -noverify -no_nonce -respout ocsp.resp -issuer $ISSUER_CER -cert $SERVER_CER -url $URL
```

router/image/templates/nginx.conf
mboersma: If `resolver` has to be set for OCSP, maybe its value should be (yet another) etcd key.

wenzowski: Could parse `/etc/resolve.conf` and grab the [authoritative nameserver configuration](http://wiki.alpinelinux.org/wiki/Configure_Networking#Configuring_DNS) then dump that into an etcd key so it's accessible from the template?

wenzowski:
Unless I misunderstand @aledbf that snippet is still doing a request to the `OCSP - URI:` you've cut from the parsed certificate. The difference appears to be that openssl is aware of the system's dns configuration while nginx is not.

There's already an `/etc/resolv.conf` [parser](https://golang.org/src/net/dnsconfig_unix.go) in `net`, so my preference would be to pull the dns servers info out of there and drop it into the nginx template assuming the bug you mentioned is resolved as of 1.9.2.

Unfortunately that's `dnsReadConfig` which isn't exported, so I would need to get some go-fu to grab it.

Regarding the cron approach, what happens when nginx tries to access a missing or empty local stapling file when that directive is set? If cron is the way to go, there are [minor considerations](http://wiki.alpinelinux.org/wiki/Alpine_Linux:FAQ#My_cron_jobs_don.27t_run.3F) for Alpine.

technosophos:
I have founds some suggestions that if `resolver` is not explicitly set, it will use the system's resolver:
https://www.digicert.com/ssl-support/nginx-enable-ocsp-stapling-on-server.htm

Have we tested that? It seems to me that that would be the most desirable.

wenzowski:
> For a resolution of the OCSP responder hostname, the resolver directive should also be specified.
> <cite>[Nginx docs](http://nginx.org/en/docs/http/ngx_http_ssl_module.html#ssl_stapling)</cite>

It did not work for me without explicitly setting the resolver.

technosophos:
I think we're at the point where we might as well just go with `8.8.8.8 8.8.4.4`. Maybe we could do these as template variables just in case the need arises to change it? Thoughts, @wenzowski ?

aledbf:
Today I found that if the dns server in `resolver` is not running nginx just returns error 500 for any url  without any explanation in the logs.

wenzowski:
Well that's a significant reason to not merge this. On-disk stapling file it is.

technosophos:
What's the best practice for maintaining the on-disk stapling file? @aledbf should we use your method and add staple file generation as a cron job?

carmstrong:
ping @wenzowski and @aledbf - what's the next step to shipping this?

carmstrong:
Closing this for now, as it hasn't been updated in a few weeks and will need to be rebased. We'd love to review once it's ready!

PR #3901, opened 2015-06-18
ref(deisctl): fix error to string conversion in tests

Joshua-Anderson:
The original code was very hackish. I discovered the `Error()` method and updated the code to use that. Sorry for the original horrible code :frowning:


mboersma:
Code LGTM.

carmstrong:
Code LGTM. Thanks @Joshua-Anderson!

Issue #3913, opened 2015-06-23
JSON load failure on "deis apps:run 世界"


mboersma:
Many test runs are now failing with a JSON error trying to parse the results of `docker inspect` over SSH:
```python
INFO appssample: test runs 'echo Hello, 世界'
ERROR appssample: appssample_v2.run.1 (run): Expecting object: line 40 column 28 (char 1077)
ERROR Internal Server Error: /v1/apps/appssample/run
Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/django/core/handlers/base.py", line 112, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/lib/python2.7/site-packages/django/views/decorators/csrf.py", line 57, in wrapped_view
    return view_func(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/rest_framework/viewsets.py", line 85, in view
    return self.dispatch(request, *args, **kwargs)
  File "/usr/lib/python2.7/site-packages/rest_framework/views.py", line 407, in dispatch
    response = self.handle_exception(exc)
  File "/usr/lib/python2.7/site-packages/rest_framework/views.py", line 404, in dispatch
    response = handler(request, *args, **kwargs)
  File "/app/api/views.py", line 205, in run
    output_and_rc = app.run(self.request.user, request.data['command'])
  File "/app/api/models.py", line 412, in run
    return c.run(escaped_command)
  File "/app/api/models.py", line 546, in run
    rc, output = self._scheduler.run(job_id, image, entrypoint, command)
  File "/app/scheduler/fleet.py", line 328, in run
    container = json.loads(_output)
  File "/usr/lib/python2.7/json/__init__.py", line 338, in loads
    return _default_decoder.decode(s)
  File "/usr/lib/python2.7/json/decoder.py", line 366, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/lib/python2.7/json/decoder.py", line 382, in raw_decode
    obj, end = self.scan_once(s, idx)
ValueError: Expecting object: line 40 column 28 (char 1077)
10.21.1.168 "POST /v1/apps/appssample/run HTTP/1.0" 500 27 "python-requests/2.5.1 CPython/2.7.6 Linux/3.16.0-41-generic"
```
[Here is one example](https://ci.deis.io/job/test-ec2/1461/consoleFull#682192135502066c60b2236c09cc9bf8).

[...]

PR #3914, opened 2015-06-23
[WIP] ref(controller): log raw output of "docker inspect" on error

mboersma:
This is to investigate the raw output that is causing #3913 failures, it doesn't fix anything (yet). 

Apologies for submitting an exploratory PR, but the network bandwidth here at DockerCon 2015 doesn't permit me to test this locally.


carmstrong:
Code LGTM.

bacongobbler:
this is blocking most PRs from going green (#3874 just hit this as well). From the logs it doesn't look like the error we're looking for isn't showing itself. I have a cluster up so I'll see what I can dig up :)

mboersma:
> I'll see what I can dig up

That would be great. If we can just trigger the error and catch the raw text I think it will illuminate what's going wrong, but this has been quite the Heisenbug.

mboersma:
Closing in favor of #3930.

PR #3915, opened 2015-06-24
feat(router): enable ssl caching and expose ssl_protocols for configuration

croemmich:
This pull will enable ssl session caching by default and reduce the ssl_buffer size to 4k to slightly increase the time-to-first-byte. Additionally it makes ssl_protocols configurable.

I split out the SSL configuration to a separate config file. As a side effect, the ssl_dhparam option is now affecting domain certificates, as it was likely intended to.

With the current SSL settings and these, it's now possible to score an A+ on https://www.ssllabs.com/ssltest.

### DIFF
--- a/router/image/templates/ssl.conf
+++ b/router/image/templates/ssl.conf
@@ -1,0 +1,25 @@
+# set the allowed protocols
+ssl_protocols {{ or (getv "/deis/router/sslProtocols") "TLSv1 TLSv1.1 TLSv1.2" }};
+
+# turn on session caching to drastically improve performance
+{{ if exists "/deis/router/sslSessionCache" }}
+ssl_session_cache {{ getv "/deis/router/sslSessionCache" }};
+ssl_session_timeout {{ or (getv "/deis/router/sslSessionTimeout") "10m" }};
+{{ end }}
+
+# allow configuring ssl session tickets
+ssl_session_tickets {{ or (getv "/deis/router/sslSessionTickets") "on" }};
+
+# slightly reduce the time-to-first-byte
+ssl_buffer_size {{ or (getv "/deis/router/sslBufferSize") "4k" }};
+
+# allow configuring custom ssl ciphers
+{{ if exists "/deis/router/sslCiphers" }}
+ssl_ciphers '{{ getv "/deis/router/sslCiphers" }}';
+ssl_prefer_server_ciphers on;
+{{ end }}
+
+# allow custom DH parameters
+{{ if exists "/deis/router/sslDhparam" }}
+ssl_dhparam /etc/ssl/dhparam.pem;
+{{ end }}
--- a/docs/customizing_deis/router_settings.rst
+++ b/docs/customizing_deis/router_settings.rst
@@ -70,6 +70,11 @@ setting                                      description
 /deis/router/sslCiphers                      cluster-wide enabled SSL ciphers
 /deis/router/sslKey                          cluster-wide SSL private key
 /deis/router/sslDhparam                      cluster-wide SSL dhparam
+/deis/router/sslProtocols                    nginx ssl_protocols setting (default: TLSv1 TLSv1.1 TLSv1.2)
+/deis/router/sslSessionCache                 nginx ssl_session_cache setting (default: not set)
+/deis/router/sslSessionTickets               nginx ssl_session_tickets setting (default: on)
+/deis/router/sslSessionTimeout               nginx ssl_session_timeout setting (default: 10m)
+/deis/router/sslBufferSize                   nginx ssl_buffer_size setting (default: 4k)
 /deis/router/workerProcesses                 nginx number of worker processes to start (default: auto i.e. available CPU cores)
 /deis/router/proxyProtocol                   nginx PROXY protocol enabled
 /deis/router/proxyRealIpCidr                 nginx IP with CIDR used by the load balancer in front of deis-router (default: 10.0.0.0/8)
--- a/router/image/conf.d/ssl.conf.toml
+++ b/router/image/conf.d/ssl.conf.toml
@@ -1,0 +1,10 @@
+[template]
+src   = "ssl.conf"
+dest  = "/opt/nginx/conf/ssl.conf"
+uid = 0
+gid = 0
+mode  = "0644"
+keys = [
+  "/deis/router",
+]
+reload_cmd = "/opt/nginx/sbin/nginx -s reload"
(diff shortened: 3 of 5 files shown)


carmstrong:
I love the splitting out of the SSL config to its own file. Code LGTM!

carmstrong:
This seems to have passed Jenkins, and is good to merge pending another review.

router/image/templates/ssl.conf
```diff
@@ -0,0 +1,23 @@
+# set the allowed protocols
+ssl_protocols {{ or (getv "/deis/router/sslProtocols") "TLSv1 TLSv1.1 TLSv1.2" }};
+
+# turn on session caching to drastically improve performance
+ssl_session_cache {{ or (getv "/deis/router/sslSessionCache") "shared:SSL:32m" }};
```
bacongobbler: this session is only shared amongst all workers across a single router. Doesn't that defeat the purpose when the next incoming request is redirected to another router? The session would be lost.

aledbf: @bacongobbler you are right. This is how cloudflare handle the sharing part https://blog.cloudflare.com/tls-session-resumption-full-speed-and-secure/ and this is what could provide this to nginx using lua https://github.com/openresty/lua-nginx-module/issues/331 (but is not ready, yet)

croemmich:
@bacongobbler You're correct. However, many load balancers have sticky sessions which will always direct a user's requests to the same router.

bacongobbler:
We might want to hide that under a feature flag if that's the case. many != all. I agree that this would be handy but some users may be confused when they upgrade and suddenly have their apps change because the routers enabled session caching. :)

croemmich:
@bacongobbler I can certainly do that, however having it on by default shouldn't cause any change in functionality since session ids were already being assigned by Nginx. It would be a strong preference of mine to turn caching on by default and instruct users to enable sticky session in their load balancers, but it's your call.

As an aside, I've always found the default of 'none' to be slightly odd. It causes Nginx to assign an id that client are supposed to send back. However, when the value of 'none' is used the server completely ignores the id from the client resulting in, though small, unnecessary overhead. A better default would be 'off' in which the server doesn't even assign the ids.

technosophos:
I think we'd like a feature flag on it, at least while we get some evidence about what its impact will be. Sound reasonable, @croemmich

croemmich:
@technosophos Sounds good, I'll push an update soon!

croemmich:
The feature flag is in place, thanks!

technosophos:
Code LGTM.

We should manually verify.

router/image/templates/ssl.conf
```diff
@@ -0,0 +1,25 @@
...
+
+# allow configuring custom ssl ciphers
+{{ if exists "/deis/router/sslCiphers" }}
+ssl_ciphers '{{ getv "/deis/router/sslCiphers" }}';
+ssl_prefer_server_ciphers on;
+{{ end }}
+
+# allow custom DH parameters
+{{ if exists "/deis/router/sslDhparam" }}
+ssl_dhparam /etc/ssl/dhparam.pem;
```
aledbf: where this file is generated?

croemmich: It's generated by Confd. Here is the configuration: https://github.com/deis/deis/tree/master/router/image/conf.d

mboersma:
Code LGTM, but this still needs some manual testing, and it would be great if @aledbf also signed off on it.

carmstrong:
@wenzowski also typically likes to see these sorts of changes

router/image/templates/ssl.conf
```diff
@@ -0,0 +1,25 @@
...
+ssl_protocols {{ or (getv "/deis/router/sslProtocols") "TLSv1 TLSv1.1 TLSv1.2" }};
+
+# turn on session caching to drastically improve performance
+{{ if exists "/deis/router/sslSessionCache" }}
+ssl_session_cache {{ getv "/deis/router/sslSessionCache" }};
+ssl_session_timeout {{ or (getv "/deis/router/sslSessionTimeout") "10m" }};
+{{ end }}
+
+# turn off tickets, as they can cause issues in HA environments
+ssl_session_tickets off;
```
aledbf: I think this should be optional. The cost of not using session resumption is too high
https://blog.cloudflare.com/tls-session-resumption-full-speed-and-secure/

mboersma:
@croemmich is it reasonable to have the default be `ssl_session_tickets on;` instead? I'm not that knowledgeable in this area, but [TLS session resumption](http://tools.ietf.org/html/rfc5077) seems like a win, and I trust @aledbf's experience with this.

croemmich:
@mboersma Yes, but probably only if we make sure all of the nodes have the same `ssl_session_ticket_key`. I'll include that as well as frequent rotation of the keys.

carmstrong:
> I'll include that as well as frequent rotation of the keys.

How are you thinking of implementing the key rotation?

croemmich:
@carmstrong: Here's my plan, let me know if you can think of a better way to do it.

The router will have an additional cron job to generate and rotate the keys. The router will have 2 keys configured at any time; the current key and the previous key, which will ensure old tickets aren't dropped when the keys are rotated. I'll probably set the rotation period to 12 hours. I'm not sure how easy it will be to make configurable. I'll use an etcd key with a ttl slightly longer than the rotation period to ensure only one router is rotating keys.

carmstrong:
@croemmich Would key rotation need to be on by default? That sounds like some additional complexity, but given the security-sensitive nature of the feature, I think it makes sense. I'm just thinking about all the ways it could break. The routers have to be the most stable part of the cluster so that apps don't go down.

/cc @wenzowski @technosophos

technosophos:
@croemmich A two-key rotation like that is pretty standard, so I am good with that. So the way you rotate the keys is by doing something like this?

```
ssl_session_ticket_file b.crt   # <-- current
ssl_session_ticket_file c.crt  # <-- old
```

Then the key rotation hits, and then then you rewrite the file as:

```
ssl_session_ticket_file a.crt   # <-- brand new
ssl_session_ticket_file b.crt  # <-- not current anymore
# And c is now rotated out.
```

Is that right?

Then it sounds like the session timeout is directly correlated with the key rotation schedule, right? So we'd need to confd the rotation script as well, and use the same time?

croemmich:
@technosophos That's correct. I believe session timeouts can be shorter than the ticket expiry, but not longer.

croemmich:
@carmstrong The feature should likely be enabled by default.

croemmich:
I'm going to roll the key rotation in to #4182 so we can hopefully get this in to 1.9. Please review. Thanks!

/cc @mboersma @technosophos

PR #3918, opened 2015-06-25
fix(*): ensure container removal with --rm

carmstrong:
@jwaldrip reported in #3885 that we seemingly leave behind orphaned containers that are not cleaned up after they are removed. This PR ensures all unit files use `--rm` to automatically clean up the container (`database` was lacking it before). This also removes the unnecessary `ExecStopPost` to clean up the container, as this is seemingly unnecessary (and may not even work). A side benefit of this change is that we no longer see "container does not exist" in the Docker logs whenever a service is stopped (as the container is already cleaned up by the time the `ExecStopPost` hits).

Note that we don't change the run template in the fleet scheduler to use `--rm` since as @bacongobbler mentioned, we have to log in and perform a `docker inspect` to check on the success of the container.

replaces #3885

### DIFF
--- a/controller/scheduler/fleet.py
+++ b/controller/scheduler/fleet.py
@@ -379,9 +379,8 @@ CONTAINER_TEMPLATE = [
     {"section": "Unit", "name": "Description", "value": "{name}"},
     {"section": "Service", "name": "ExecStartPre", "value": '''/bin/sh -c "IMAGE=$(etcdctl get /deis/registry/host 2>&1):$(etcdctl get /deis/registry/port 2>&1)/{image}; docker pull $IMAGE"'''},  # noqa
     {"section": "Service", "name": "ExecStartPre", "value": '''/bin/sh -c "docker inspect {name} >/dev/null 2>&1 && docker rm -f {name} || true"'''},  # noqa
-    {"section": "Service", "name": "ExecStart", "value": '''/bin/sh -c "IMAGE=$(etcdctl get /deis/registry/host 2>&1):$(etcdctl get /deis/registry/port 2>&1)/{image}; docker run --name {name} {memory} {cpu} {hostname} -P $IMAGE {command}"'''},  # noqa
+    {"section": "Service", "name": "ExecStart", "value": '''/bin/sh -c "IMAGE=$(etcdctl get /deis/registry/host 2>&1):$(etcdctl get /deis/registry/port 2>&1)/{image}; docker run --name {name} --rm {memory} {cpu} {hostname} -P $IMAGE {command}"'''},  # noqa
     {"section": "Service", "name": "ExecStop", "value": '''/usr/bin/docker stop {name}'''},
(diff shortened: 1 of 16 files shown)


carmstrong:
Jenkins error seems totally unrelated:
```
INFO appssample: test runs 'echo Hello, 世界'
ERROR appssample: appssample_v2.run.1 (run): Expecting object: line 58 column 25 (char 1548)
ERROR Internal Server Error: /v1/apps/appssample/run
  File "/usr/lib/python2.7/site-packages/django/core/handlers/base.py", line 112, in get_response
  File "/usr/lib/python2.7/site-packages/django/views/decorators/csrf.py", line 57, in wrapped_view
  File "/usr/lib/python2.7/site-packages/rest_framework/viewsets.py", line 85, in view
  File "/usr/lib/python2.7/site-packages/rest_framework/views.py", line 407, in dispatch
  File "/usr/lib/python2.7/site-packages/rest_framework/views.py", line 404, in dispatch
  File "/app/api/views.py", line 205, in run
    output_and_rc = app.run(self.request.user, request.data['command'])
  File "/app/api/models.py", line 412, in run
    return c.run(escaped_command)
  File "/app/api/models.py", line 546, in run
    rc, output = self._scheduler.run(job_id, image, entrypoint, command)
  File "/app/scheduler/fleet.py", line 328, in run
    container = json.loads(_output)
  File "/usr/lib/python2.7/json/__init__.py", line 338, in loads
    return _default_decoder.decode(s)
  File "/usr/lib/python2.7/json/decoder.py", line 366, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/lib/python2.7/json/decoder.py", line 382, in raw_decode
    obj, end = self.scan_once(s, idx)
ValueError: Expecting object: line 58 column 25 (char 1548)
172.17.42.1 "POST /v1/apps/appssample/run HTTP/1.0" 500 27 "python-requests/2.5.1 CPython/2.7.6 Linux/3.16.0-41-generic"
sh: locale: not found
```

bacongobbler:
That sounds like #3913

bacongobbler:
While we don't need to address it in this PR, we can remove the container in `.run()` after we've inspected its output/status code. Would you be willing to PR that?

carmstrong:
> While we don't need to address it in this PR, we can remove the container in .run() after we've inspected its output/status code.

We're already doing that in a `finally` block after inspecting the status code. Does that not get run in all cases? From what I can tell, most of the `.run` is wrapped in a `try/finally` which does clean up the container.

bacongobbler:
Yes, that should be run in all cases. I just never bothered to check :)

code LGTM!

carmstrong:
> I just never bothered to check :)

Probably a good thing to do before suggesting a PR author goes and does it ;)

technosophos:
LGTM

PR #3921, opened 2015-06-25
docs(installing_deis): update DO supported regions

bacongobbler:
DO now has support for SFO1 and FRA1, so this updates the docs accordingly.


mboersma:
I see the same from `docl`. LGTM.

carmstrong:
LGTM.

PR #3927, opened 2015-06-26
update doc links to deis.com

bacongobbler:

PR #3928, opened 2015-06-26
Update Engine Yard links to Deis.com

gpxl:
Requested by DMC


gpxl:
Dupe of https://github.com/deis/deis/pull/3927

PR #3929, opened 2015-06-26
fix(logger): reduce severity of log failure

bacongobbler:
Sending a log to an external server should not kill the logger on an error. For example, if the host is down or unavailable, the logger should just log that the error message has been lost, but carry on.


technosophos:
Agree. LGTM.

carmstrong:
Code LGTM.

mboersma:
Code LGTM.

PR #3930, opened 2015-06-27
[DO NOT MERGE] debug deis run issue

bacongobbler:
The other PR seems to hit the same issue, but isn't showing the raw issue. I've been testing for a couple days with no luck... maybe Jenkins can light the way.


controller/scheduler/fleet.py
```diff
@@ -325,7 +331,10 @@ def _do_ssh(cmd):
             _rc, _output = _do_ssh('docker inspect {name}'.format(**locals()))
             if _rc != 0:
                 raise RuntimeError('could not determine exit code')
-            container = json.loads(_output)
+            try:
+                _container = json.loads(_output)
```
mboersma: Should be `container` here.

carmstrong:
@bacongobbler Is this good to close?

bacongobbler:
Yes :)

PR #3931, opened 2015-06-27
fix(controller): allow non superusers to manage their keys

rochacon:
Users were unable to remove their SSH keys. This PR fixes the Key management view configuration that was blocking the API access. The wrong view configuration was being inherited from `api.views.BaseDeisViewSet`. 

Error using API directly:

```
$ http DELETE deis.platform/v1/keys/rochacon@machine.local "Authorization:token <rochacon-token>"
HTTP/1.1 403 FORBIDDEN
Allow: GET, DELETE, HEAD, OPTIONS
Connection: keep-alive
Content-Encoding: gzip
Content-Type: application/json
DEIS_API_VERSION: 1.4
DEIS_PLATFORM_VERSION: 1.7.3
Date: Sat, 27 Jun 2015 00:58:41 GMT
Server: nginx/1.9.0
Transfer-Encoding: chunked
Vary: Accept-Encoding
X_DEIS_API_VERSION: 1.4
X_DEIS_PLATFORM_VERSION: 1.7.3

{
    "detail": "You do not have permission to perform this action."
}
```

Error using the CLI:

```
$ deis keys
=== rochacon Keys
rochacon@machine.local ssh-rsa AAAAB3Nz...2eSfWk4irQ

$ deis keys:remove rochacon@machine.local
Removing rochacon@machine.local SSH Key... 403 FORBIDDEN
Detail:
You do not have permission to perform this action.
```

### DIFF
--- a/controller/api/tests/test_key.py
+++ b/controller/api/tests/test_key.py
@@ -152,3 +152,13 @@ class KeyTest(TestCase):
     def test_rsa_key_fingerprint(self):
         fp = fingerprint(RSA_PUBKEY)
         self.assertEquals(fp, '54:6d:da:1f:91:b5:2b:6f:a2:83:90:c4:f9:73:76:f5')
+
+    def test_key_api_with_non_superuser_rsa(self):
+        self.user = User.objects.get(username='autotest2')
+        self.token = self.user.auth_token.key
+        self._check_key(RSA_PUBKEY)
+
+    def test_key_api_with_non_superuser_ecdsa(self):
+        self.user = User.objects.get(username='autotest2')
+        self.token = self.user.auth_token.key
+        self._check_key(ECDSA_PUBKEY)
--- a/controller/api/views.py
+++ b/controller/api/views.py
@@ -287,6 +287,7 @@ class CertificateViewSet(BaseDeisViewSet):
 class KeyViewSet(BaseDeisViewSet):
     """A viewset for interacting with Key objects."""
     model = models.Key
+    permission_classes = [IsAuthenticated, permissions.IsOwner]
     serializer_class = serializers.KeySerializer
 
 


deis-admin:

rochacon:
Rebased branch to fix commit subject length and use latest master.

bacongobbler:
thanks for this! We're currently debugging issues with the integration tests right now (more than half are failing inconspicuously). We'll jump on this as soon as that's done. Thank you for your patience! :)

bacongobbler:
Aha, so it seems that the BaseDeisViewSet has the permission set of `IsAppUser`, which of course there's no app tied to a Key, in which case the key will fail to be updated.

Code LGTM! We just need to wait on our end to fix Jenkins.

carmstrong:
Can admins still remove any user's key? I think admins should be able to remove a key, and a user should be able to remove his/her own key.

bacongobbler:
There is a `IsOwnerOrAdmin` permission set. Considering that admins had global access before, it only makes sense to keep that same behaviour, since there's no argument against it. @rochacon would you mind making that change?

rochacon:
@carmstrong @bacongobbler Agreed, admins must be able to remove users keys. I'll add a test for that and change the proper permission list.

mboersma:
Jenkins, add to whitelist.

rochacon:
Seems like admins were never able to delete users keys. I've made the following test and it breaks with the old permissions, the new ones and with `IsOwnerOrAdmin`.

I believe it is breaking due to [this line](https://github.com/deis/deis/blob/b174faeeb72f1c7cfd42eaf2545474573b07f104/controller/api/viewsets.py#L17) in `api.viewsets.OwnerViewSet`

Test:

```python
def test_admin_should_be_able_to_remove_users_keys(self):
    # Create Key with non superuser
    user = User.objects.get(username='autotest2')
    user_token = user.auth_token.key
    body = {'id': 'mykey@box.local', 'public': RSA_PUBKEY}
    response = self.client.post('/v1/keys', json.dumps(body), content_type='application/json',
                                HTTP_AUTHORIZATION='token {}'.format(user_token))
    self.assertEqual(response.status_code, 201)
    # Use superuser to remove key
    admin = User.objects.get(username='autotest')
    admin_token = admin.auth_token.key
    response = self.client.delete('/v1/keys/{}'.format(body['id']),
                                  HTTP_AUTHORIZATION='token {}'.format(admin_token))
    self.assertEqual(response.status_code, 204)

```

bacongobbler:
In that case you may need something like [the following](https://github.com/deis/deis/blob/4e2204ee303f569dda81cfc3d1e0f5e2a9d964cd/controller/api/views.py#L152-L153) in the KeyViewSet so it overrides the parent's behaviour. Good catch

rochacon:
@bacongobbler I wasn't sure if I should add this, since it looks like a new feature. Will override `get_queryset` and update the PR.

rochacon:
@bacongobbler @carmstrong, currently admins are not allowed to manage users keys. So, I'll not include this feature here.

I did started to implement this feature but there were some side-effects that deserves more UX design, such as admins listing all registered keys when using `deis keys`.

I'll open a new issue so we can discuss this.

carmstrong:
> I'll open a new issue so we can discuss this.

Sounds good to me. I created an issue (#3976) to ensure admins have proper permissions. Users and keys are two models of potentially many that they don't have proper access to.

carmstrong:
@rochacon Could you try rebasing this? We merged some test fixes that will hopefully allow this to pass. Thanks!

rochacon:
@carmstrong Sure do, I'll rebase this and update the PR.

I have a draft of the issue for proper admin access, but got really busy and didn't opened it. I'll add some comments to #3976 with the initial work I did.

rochacon:
Rebased

bacongobbler:
@rochacon would you mind rebasing this once more? We had to fix up some of the stuff in master, specifically https://github.com/deis/deis/pull/3957 s what's blocking this PR from passing.

rochacon:
No problem @bacongobbler, rebased it

I see you guys released v1.8.0, I was hoping to have this fix in that version. Will the next release be the v1.9.0 milestone or is there a bugfix release planned?

bacongobbler:
since this isn't technically a bugfix it'll get into v1.9. Sorry about that!

mboersma:
Code LGTM. Note that this doesn't try to fix everything in this area: see #3976.

bacongobbler:
code LGTM. Thanks @rochacon :)

rochacon:
I'm back from vacations, nice to see this merged in! \o/

PR #3933, opened 2015-06-29
fix(builder): only build slugbuilder/slugrunner when necessary

carmstrong:
On start, the slugbuilder and slugrunner images are built from
the builder's source directory. This is unnecessary on restarts,
and prevents optimizations where the slugbuilder and slugrunner images
are already pre-cached (for example, by bind-mounting a host volume
into builder with the images).

This commit simply wraps these `docker build` instructions in
`docker history` checks to see if the images exist before
attempting to build them.

/cc @iancoffey @technosophos

### DIFF
--- a/builder/rootfs/bin/boot
+++ b/builder/rootfs/bin/boot
@@ -85,9 +85,14 @@ while [[ ! -e /var/run/docker.sock ]]; do
 	sleep 1
 done
 
-# build required images
-docker build -t deis/slugbuilder /usr/local/src/slugbuilder/
-docker build -t deis/slugrunner /usr/local/src/slugrunner/
+# build required images if they don't already exist
+if ! docker history deis/slugbuilder >/dev/null 2>&1; then
+	docker build -t deis/slugbuilder /usr/local/src/slugbuilder/
+fi
+
+if ! docker history deis/slugrunner >/dev/null 2>&1; then
+	docker build -t deis/slugrunner /usr/local/src/slugrunner/
+fi
 
 function gen_host_keys {
 	if ! etcd_get sshHostKey; then


carmstrong:
This also enables users to override our slugbuilder and slugrunner if their heart so desires.

builder/rootfs/bin/boot
```diff
@@ -85,9 +85,14 @@ while [[ ! -e /var/run/docker.sock ]]; do
 	sleep 1
 done
 
-# build required images
-docker build -t deis/slugbuilder /usr/local/src/slugbuilder/
-docker build -t deis/slugrunner /usr/local/src/slugrunner/
+# build required images if they don't already exist
+if ! docker history deis/slugbuilder >/dev/null 2>&1; do
```
iancoffey: is do/done valid in a bash if statement?

carmstrong: Well, that's embarrassing :/

technosophos:
Agreed and will roll into the new builder, too. But are we running the risk of not getting updates to slug* when we reboot?

bacongobbler:
Yes, that's an issue that I see with this PR. If someone were to hack on the builder and modify slug*, then run a `make deploy` their changes won't occur.

carmstrong:
> If someone were to hack on the builder and modify slug*, then run a make deploy their changes won't occur.

If the builder restarts, its Docker graph is wiped, no? So restarting builder would indeed cause it to start up and build from the source tree as intended. This change is mainly for use cases where we want to pre-cache slugbuilder and slugrunner as an optimization.

I also think a slight inconvenience to the dev workflow is warranted for a potentially substantial improvement on the deployment side.

technosophos:
I agree about the optimization. LGTM, then.

bacongobbler:
> If the builder restarts, its Docker graph is wiped, no?

We still have a volume for /var/lib/docker, so the data is cached unless we remove that :)

https://github.com/deis/deis/blob/master/deisctl/units/deis-builder.service#L7

carmstrong:
> We still have a volume for /var/lib/docker

Ah, I forgot about that. Maybe the solution is to force a build of slugrunner and slugbuilder as part of the makefile `make deploy` workflow for builder. Either way, I think this change should carry on as it is so we can make use of this optimization in Deis 1.8. The rewrite of builder will undoubtedly change some of how this is handled, anyway. And we so rarely make changes to slugbuilder and slugrunner that I'm okay with requiring a manual rebuild of them for now.

carmstrong:
@bacongobbler I still prefer to implement this as-is rather than remove the volume from builder. Faster builder startup and building of applications is much preferred, and I don't want to make significant changes so late in bash-builder's life.

mboersma:
> I'm okay with requiring a manual rebuild of them for now.

I am too. Code LGTM.

bacongobbler:
I'm on board. Code LGTM.

PR #3937, opened 2015-06-29
feat(deisctl): specify # of routers to install

laurrentt:
When installing the platform with deisctl, 3 routers are started by
default. This causes an annoying freeze when you use custom component
placement and you have less than 3 instances flagged as routerMesh.
Setting the $DEIS_NUM_ROUTERS variable prior calling "deisctl install
platform" will allow you to choose how many routers are to be installed
in your cluster.

### DIFF
--- a/deisctl/client/client.go
+++ b/deisctl/client/client.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"strconv"
 
 	"github.com/deis/deis/deisctl/backend"
 	"github.com/deis/deis/deisctl/backend/fleet"
@@ -109,7 +110,7 @@ Examples:
 // Install loads the definitions of components from local unit files.
 // After Install, the components will be available to Start.
 func (c *Client) Install(argv []string) error {
-	usage := `Loads the definitions of components from local unit files.
+	usage := fmt.Sprintf(`Loads the definitions of components from local unit files.
 
 After install, the components will be available to start.
 
@@ -120,12 +121,23 @@ After install, the components will be available to start.
 
 Usage:
   deisctl install [<target>...] [options]
-`
+
+Options:
+  --router-mesh-size=<num>  Number of routers to be loaded when installing the platform [default: %d].
+`, cmd.DefaultRouterMeshSize)
 	// parse command-line arguments
 	args, err := docopt.Parse(usage, argv, true, "", false)
 	if err != nil {
 		return err
 	}
+
+	meshSizeArg, _ := args["--router-mesh-size"].(string)
+	parsedValue, err := strconv.ParseUint(meshSizeArg, 0, 8)
+	if err != nil || parsedValue < 1 {
+		fmt.Print("Error: argument --router-mesh-size: invalid value, make sure the value is an integer between 1 and 255.\n")
+		return err
+	}
+	cmd.RouterMeshSize = uint8(parsedValue)
 
 	return cmd.Install(args["<target>"].([]string), c.Backend, cmd.CheckRequiredKeys)
 }
--- a/deisctl/cmd/cmd.go
+++ b/deisctl/cmd/cmd.go
@@ -27,6 +27,8 @@ const (
 	StatelessPlatformCommand string = "stateless-platform"
 	swarm                    string = "swarm"
 	mesos                    string = "mesos"
+	// DefaultRouterMeshSize defines the default number of routers to be loaded when installing the platform.
+	DefaultRouterMeshSize uint8 = 3
 )
 
 // ListUnits prints a list of installed units.
@@ -44,6 +46,9 @@ var Stdout io.Writer = os.Stderr
 
 // Location to write standard error information. By default, this is the os.Stderr.
 var Stderr io.Writer = os.Stdout
+
+// Number of routers to be installed. By default, it's DefaultRouterMeshSize.
+var RouterMeshSize = DefaultRouterMeshSize
 
 // Scale grows or shrinks the number of running components.
 // Currently "router", "registry" and "store-gateway" are the only types that can be scaled.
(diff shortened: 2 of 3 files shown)


deis-admin:

deisctl/utils/utils_test.go
```diff
@@ -18,3 +18,24 @@ func TestResolvePath(t *testing.T) {
...
+
+func TestGetEnvUint(t *testing.T) {
+	defaultValue := uint64(3)
+	goodVars := []string{"1", "2", "4"}
+	expectedGoodUints := []uint64{1, 2, 4}
+	badVars := []string{"-1", "abc"}
+	for index, goodVar := range goodVars {
+		os.Setenv("DEIS_NUM_ROUTERS", goodVar)
+		result := GetEnvUint("DEIS_NUM_ROUTERS", defaultValue)
+		if expectedGoodUints[index] !=  result {
```
mboersma: The extra space here will fail `make -C deisctl test-style`, since we expect all .go code to have `gofmt` run on it.

laurrentt:
Sorry I'm just not used to `gofmt`. I ran `make -C deisctl test-style` and everything looks fine.

mboersma:
Jenkins, add to whitelist.

technosophos:
Thinking out loud, here, is it better to set this sort of information as a flag, rather than as an environment variable?

It looks to me like this value is only used once, and that there is no compelling reason to store the data in the environment for more than the installation. So maybe it would be better to pass in something like `--num-routers=5` instead?

laurrentt:
@technosophos I agree but for consistency I used the same technique as http://docs.deis.io/en/latest/installing_deis/aws/#choose-number-of-instances. I'll let the maintainers decide and change it if needed.

bacongobbler:
@technosophos is one of the maintainers now :)

I agree that it should be hidden under a feature flag. Beforehand this was all done with bash but it makes more sense now that we have an official binary (deisctl).

laurrentt:
@bacongobbler \*facepalm\* ...I should have checked his profile first :). I'll make the change when I have the time.

technosophos:
No worries. You are right that the existing code does tend to prefer environment variables over flags, which is why I left the question open.

mboersma:
> I agree that it should be hidden under a feature flag.

+1. Although Deis does have `$DEIS_NUM_INSTANCES` it's more or less deprecated, and having an explicit command-line way to do this is preferable.

technosophos:
Any update on this? I think we could get it on the 1.9 release.

laurrentt:
@technosophos I'll try to have it done by today

laurrentt:
@mboersma Don't know if you received a notification but I made the change and added an command-line argument to set the router mesh size instead of using an env var. Let me know if I need to change something.

deisctl/client/client.go
```diff
@@ -120,13 +121,25 @@ After install, the components will be available to start.
 
 Usage:
   deisctl install [<target>...] [options]
-`
+
+Options:
+  --router-mesh-size=<num>  Number of routers to be loaded when installing the platform [default: %d].
```
bacongobbler: `default: %d`? Shouldn't this be a 3, or is this getting parsed somewhere? :)

bacongobbler: Ah, I see the call to `fmt.Sprintf()` now. Disregard!

deisctl/client/client.go
```diff
@@ -120,13 +121,25 @@ After install, the components will be available to start.
...
+`, cmd.DefaultRouterMeshSize)
 	// parse command-line arguments
 	args, err := docopt.Parse(usage, argv, true, "", false)
 	if err != nil {
 		return err
 	}
 
+	meshSizeArg, ok := args["--router-mesh-size"].(string)
+	if ok {
```
bacongobbler: I know that this will likely never happen, but I still think we should have a catch case in case `--router-mesh-size` comes up blank, or else we should just remove this `if ok` branch logic and just go about our business.

technosophos: If `!ok` then this would set `RouterMeshSize` to `0`, which then goes to DefaultRouterMeshSize, right? Or are you talking about the case where if someone does "--router-mesh-size=", we may end up with an error.

As a nice touch, I like the fact that we can specify mesh sizes in octal!

laurrentt:
@bacongobbler @technosophos I've removed the `if ok` and handled both blank mesh size and out of range values.

mboersma:
I tested this on a 7-node vagrant cluster with various arguments to `--router-mesh-size=`. Made sure that `deisctl uninstall platform` cleans everything up too. 

The only discrepency I found is that we install the default 3 routers if `--router-mesh-size=0`, but I think that should either do what it says (not install any router units), or be disallowed:
```console
$ deisctl install platform --router-mesh-size=1.5
Error: Unable to parse --router-mesh-size option, make sure the value is an integer between 1 and 255.
Error: strconv.ParseUint: parsing "1.5": invalid syntax
$ deisctl install platform --router-mesh-size=-1
Error: Unable to parse --router-mesh-size option, make sure the value is an integer between 1 and 255.
Error: strconv.ParseUint: parsing "-1": invalid syntax
$ deisctl install platform --router-mesh-size=lalala
Error: Unable to parse --router-mesh-size option, make sure the value is an integer between 1 and 255.
Error: strconv.ParseUint: parsing "lalala": invalid syntax
$ deisctl install platform --router-mesh-size=0
● ▴ ■
■ ● ▴ Installing Deis...
▴ ■ ●
Storage subsystem...
deis-store-daemon.service: loaded
deis-store-monitor.service: loaded
deis-store-volume.service: loaded
deis-store-metadata.service: loaded
deis-store-gateway@1.service: loaded
Logging subsystem...
deis-logspout.service: loaded
deis-logger.service: loaded
Control plane...
deis-controller.service: loaded
deis-database.service: loaded
deis-builder.service: loaded
deis-registry@1.service: loaded
Data plane...
deis-publisher.service: loaded
Routing mesh...
deis-router@2.service: loaded
deis-router@3.service: loaded
deis-router@1.service: loaded
Done.

Please run `deisctl start platform` to boot up Deis.
```

mboersma:
Also, as per our [http://docs.deis.io/en/latest/contributing/standards/#single-issue](code standards), this should be squashed into a single git commit with a subject like `feat(deisctl): specify # of routers to install`.

laurrentt:
@mboersma done ! Let me know if there's anything else.

mboersma:
LGTM, tested on a vagrant 7-node cluster.

carmstrong:
Code LGTM. Thanks @laurrentt!

PR #3939, opened 2015-06-30
ref(router): update nginx to 1.9.2

aledbf:


bacongobbler:
It looks like docker had choked on an earlier push, preventing builds to succeed on the master node. I've rebooted the docker daemon on the master node so it should be good to go, though CI may hit some other issues we're trying to debug at the moment.

mboersma:
From http://nginx.org/en/CHANGES:

```
Changes with nginx 1.9.2                                         16 Jun 2015

    *) Feature: the "backlog" parameter of the "listen" directives of the
       mail proxy and stream modules.
    *) Feature: the "allow" and "deny" directives in the stream module.
    *) Feature: the "proxy_bind" directive in the stream module.
    *) Feature: the "proxy_protocol" directive in the stream module.
    *) Feature: the -T switch.
    *) Feature: the REQUEST_SCHEME parameter added to the fastcgi.conf,
       fastcgi_params, scgi_params, and uwsgi_params standard configuration
       files.
    *) Bugfix: the "reuseport" parameter of the "listen" directive of the
       stream module did not work.
    *) Bugfix: OCSP stapling might return an expired OCSP response in some
       cases.
```

LGTM.

Refs #3900.

bacongobbler:
we never used the reuseport parameter in the stream module, but good to know! LGTM

carmstrong:
Code LGTM.

aledbf:
@bacongobbler specifying `reuseport` in the default server it applies to all including the stream.
(to be strict with the definition, you can only specify `reuseport`once)

Issue #3814, opened 2015-06-07
Upgrade registry to docker-registry v2 (Docker Distribution)


carmstrong:
There are several issues relating to registry performance (#3580, #3479).

We should upgrade to the Go-based rewrite of the [official Docker registry](https://docs.docker.com/registry/), which is now called [Docker distribution](https://github.com/docker/distribution). We'll also need to patch in our changes to support the `deis pull` workflow.

Additionally, there is also a [native driver](https://github.com/docker/distribution/blob/master/registry/storage/driver/rados/rados.go) to consume Ceph RADOS which is likely going to be significantly more performant than our current S3-on-Ceph implementation. We should investigate this, although it is likely to complicate our upgrade path.

Upgrading to registry v2 also allows us to support Docker 1.6-built images, resolving issues like #3550.

mboersma:
Using registry:2.0.1 requires a rewrite of the code in controller/registry/private.py to use the much different Docker Registry v2 API. I'm not sure that our current approach for appending a "last-mile" layer to the image will work with the new Docker image format: there is no "upload layer with JSON metadata" endpoint, only a way to upload blobs already created by Docker.

aledbf:
@mboersma this is the PR in the new registry that could help for "the last-mile" https://github.com/docker/distribution/pull/610 but is not ready (yet)

fabiob:
@mboersma, where can I find more about the "last-mile" thing deis need?

Can't we just make `deis-builder` append a few lines on the Dockerfile, just before building?

Machool:
@mboersma , do you know when registry v2 will be available in deis? is it in 1.10.0?

SomeoneWeird:
Is there any update on this?

pomke:
considering dockerhub etc have now moved on from the older version, and it's becoming difficult to setup the older docker registry (docs are vanishing, etc), this unresolved issue makes deis unfortunately a non-viable-solution for it's primary use case (running pre-built docker images).   I realise that existing users are happy with their old/working deployments but for new people deis is a non-operational death-star. :/

lorieri:
@mboersma https://github.com/docker/distribution/issues/422

mboersma:
@lorieri exciting! This is actually implemented now in 2.1.1, right? I may not be able to hook up delete right away, but at least the potential is there to keep deis-registry storage from growing forever.

mboersma:
The two difficulties currently are:
[...]

Issue #3479, opened 2015-04-14
Cannot deploy, registry timeout


glogiotatidis:
On a 1.5.0 cluster sometimes deploy (using `deis pull`) either takes a lot of time or fails. Looking deeper in the logs it seems that docker has issues connecting to the internal registry. Sometimes after a couple of tries it fixes itself but once I had to `fleetctl stop` and `fleetctl start deis-registry@1` to make it work.

```
Apr 10 10:02:58 ip-10-152-74-240 sh[20008]: Pulling repository 10.152.74.8:5000/masterfirefoxos-prod
Apr 10 10:03:58 ip-10-152-74-240 sh[20008]: time="2015-04-10T10:03:58Z" level="fatal" msg="Get http://10.152.74.8:5000/v1/repositories/masterfirefoxos-prod/images: read tcp 10.152.74.8:5000: i/o timeout"
Apr 10 10:03:58 ip-10-152-74-240 systemd[1]: masterfirefoxos-prod_v258.cmd.2.service: control process exited, code=exited status=1
Apr 10 10:03:58 ip-10-152-74-240 systemd[1]: Failed to start masterfirefoxos-prod_v258.cmd.2.
Apr 10 10:03:58 ip-10-152-74-240 systemd[1]: Unit masterfirefoxos-prod_v258.cmd.2.service entered failed state.
Apr 10 10:03:58 ip-10-152-74-240 systemd[1]: masterfirefoxos-prod_v258.cmd.2.service failed.
Apr 10 10:04:03 ip-10-152-74-240 systemd[1]: masterfirefoxos-prod_v258.cmd.2.service holdoff time over, scheduling restart.
Apr 10 10:04:03 ip-10-152-74-240 systemd[1]: Stopping masterfirefoxos-prod_v258.cmd.2...
Apr 10 10:04:03 ip-10-152-74-240 systemd[1]: Starting masterfirefoxos-prod_v258.cmd.2...
```

mboersma:
We have also occasionally seen the stock `registry:0.9.1` die when using it for local testing. They might be the same issue. You can set `loglevel` to "debug" in [docker-registry "classic"](https://github.com/docker/docker-registry#general-options) to see more.

glogiotatidis:
Is there a plan to fix this? We're randomly getting these errors.

glogiotatidis:
A bit more background info: this started on deis 1.5.0 but it's really bad on 1.5.2. If you need those debug messages I can provide, just let me know.

[...]

PR #3942, opened 2015-06-30
chore(registry): use updated registry V1 fork

mboersma:
See https://github.com/deis/docker-registry/compare/new-repository-import-v091...deis:new-repository-import-master. It was suggested to me at DockerCon that some post-v0.9.1 fixes in classic registry could improve stability, and that `gevent` and/or `gunicorn` had been buggy, so this rebases our fork and updates those packages. If #3814 isn't ready for the next release, this is worth testing to see if it fixes classic registry stability issues.

Refs #3479.


bacongobbler:
code LGTM. Doesn't look like master has anything crazy.

carmstrong:
Code LGTM.

PR #3944, opened 2015-06-30
fix(controller): handle unicode when logging run error

mboersma:
This should fix the specific error in this CI build: https://ci.deis.io/job/test-ec2/1554/consoleFull#682192135502066c60b2236c09cc9bf8


bacongobbler:
This shouldn't normally occur, since 1554 originated from #3930 which should be a raw dump of `docker inspect` (or so we hope). I'm all for future-proofing the platform if this makes sense, though.

mboersma:
This is just one of several similar constructs, so a general fix would have to be applied elsewhere, and this didn't illuminate anything when run through CI. I'll re-open it if this looks like the correct fix while we're testing offline.

PR #3945, opened 2015-06-30
docs(managing_deis): add docs to remove Ceph dependency

carmstrong:
Note that I haven't tested this myself - this is just a translation of the documentation @Blystad provided in #2812. I think we should still discuss what changes in Deis itself would be reasonable to implement and would make this easier for users.

I also didn't include the upgrading instructions, since a modified but traditional upgrade should work just fine. Docs could use some review and ideally some testing.

refs #2812

/cc @Blystad @krancour @lazabogdan


carmstrong:
Replaced by #3947

PR #3946, opened 2015-07-01
fix(contrib/rackspace): add double quotes around POSIX vars

kzap:
put double quotes around paths incase running from osx or paths with spaces

### DIFF
--- a/contrib/rackspace/provision-rackspace-cluster.sh
+++ b/contrib/rackspace/provision-rackspace-cluster.sh
@@ -44,13 +44,13 @@ if [ -z "$DEIS_NUM_INSTANCES" ]; then
 fi
 
 # check that the CoreOS user-data file is valid
-$CONTRIB_DIR/util/check-user-data.sh
+"$CONTRIB_DIR/util/check-user-data.sh"
 
 i=1 ; while [[ $i -le $DEIS_NUM_INSTANCES ]] ; do \
     echo_yellow "Provisioning deis-$i..."
     # This image is CoreOS 633.1.0 in the stable channel
     # TODO: Update to 681.2.0 when it's available in the stable channel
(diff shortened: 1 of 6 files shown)


deis-admin:

contrib/rackspace/provision-rackspace-cluster.sh
```diff
@@ -5,10 +5,10 @@
 
 set -e
 
-THIS_DIR=$(cd $(dirname $0); pwd) # absolute path
-CONTRIB_DIR=$(dirname $THIS_DIR)
+THIS_DIR="$(cd "$(dirname $0)"; pwd)" # absolute path
```
mboersma: Two sets of quotes should be enough here:
```bash
THIS_DIR=$(cd "$(dirname $0)"; pwd) # absolute path
CONTRIB_DIR=$(dirname "$THIS_DIR")
```
Then quote `"$CONTRIB_DIR" as you did below. I tested this on a Mac in a folder with spaces in the name.

mboersma:
There are similar statements with `THIS_DIR` and `CONTRIB_DIR` in other `contrib/` scripts that should also be quoted, and as I mentioned above, I think the PR adds more quotes than necessary.

@kzap if you are willing to make similar changes wherever `THIS_DIR` and `CONTRIB_DIR` shows up, that would be very helpful for anyone who uses Deis code in a folder containing spaces. Or if that's more than you bargained for, just let us know and we can finish that up. Thanks!

kzap:
@mboersma ok updated the other scripts but not sure if they work as i didnt test on those providers

mboersma:
Refs #4044.

carmstrong:
@kzap Are you able to rebase this against master? We've had a lot of changes lately. Thanks!

bacongobbler:
@carmstrong I think this is superseded by #4044 which will catch this in the future :)

mboersma:
@kzap my apologies; we really appreciate all the work you did here, but I'm going to close this in favor of #4044 which takes a more comprehensive approach to quoting shell variables. If you have time to review #4044 and double-check that it covers the same bases, we would value any feedback you have. Thanks again!

kzap:
not sure on shellcheck, but if it will catch the errors then ok :)

PR #3947, opened 2015-07-01
feat(deisctl): support stateless control plane without Ceph

### LINKED ISSUE #2812: Make store component optional by allowing external database, registry, and log collector

carmstrong:
Adds support to `deisctl` so users can specify `stateless-platform` instead of `platform` for `deisctl install`, `deisctl start`, `deisctl stop`, and `deisctl uninstall`. This new command will skip `store-*`, `database`, and `logger` components.

Also adds documentation from @Blystad detailing how to run a stateless cluster, although the instructions have been adapted to make use of the new `stateless-platform` command in `deisctl`.

I think this is enough support to consider #2812 closed. We can probably refine this in the future, such as removing the dummy `deis-store-volume` component, but I think it's a good start.

replaces #3945
closes #2812 

/cc @Blystad @krancour @lazabogdan

### DIFF
--- a/deisctl/cmd/cmd.go
+++ b/deisctl/cmd/cmd.go
@@ -22,7 +22,9 @@ import (
 const (
 	// PlatformCommand is shorthand for "all the Deis components."
 	PlatformCommand string = "platform"
-	swarm           string = "swarm"
+	// StatelessPlatformCommand is shorthand for the components except store-*, database, and logger.
+	StatelessPlatformCommand string = "stateless-platform"
+	swarm                    string = "swarm"
 )
 
 // ListUnits prints a list of installed units.
@@ -67,9 +69,10 @@ func Start(targets []string, b backend.Backend) error {
 	// if target is platform, install all services
 	if len(targets) == 1 {
 		if targets[0] == PlatformCommand {
-			return StartPlatform(b)
-		}
-		if targets[0] == swarm {
+			return StartPlatform(b, false)
+		} else if targets[0] == StatelessPlatformCommand {
+			return StartPlatform(b, true)
+		} else if targets[0] == swarm {
 			return StartSwarm(b)
 		}
 	}
@@ -102,7 +105,7 @@ deisctl config platform set sshPrivateKey=<path-to-key>
 	return nil
 }
 
-func startDefaultServices(b backend.Backend, wg *sync.WaitGroup, outchan chan string, errchan chan error) {
+func startDefaultServices(b backend.Backend, stateless bool, wg *sync.WaitGroup, outchan chan string, errchan chan error) {
 
 	// create separate channels for background tasks
 	_outchan := make(chan string)
@@ -110,34 +113,49 @@ func startDefaultServices(b backend.Backend, wg *sync.WaitGroup, outchan chan st
 	var _wg sync.WaitGroup
 
 	// wait for groups to come up
-	outchan <- fmt.Sprintf("Storage subsystem...")
-	b.Start([]string{"store-monitor"}, wg, outchan, errchan)
-	wg.Wait()
-	b.Start([]string{"store-daemon"}, wg, outchan, errchan)
-	wg.Wait()
-	b.Start([]string{"store-metadata"}, wg, outchan, errchan)
-	wg.Wait()
+	if !stateless {
+		outchan <- fmt.Sprintf("Storage subsystem...")
+		b.Start([]string{"store-monitor"}, wg, outchan, errchan)
+		wg.Wait()
+		b.Start([]string{"store-daemon"}, wg, outchan, errchan)
+		wg.Wait()
+		b.Start([]string{"store-metadata"}, wg, outchan, errchan)
+		wg.Wait()
 
-	// we start gateway first to give metadata time to come up for volume
-	b.Start([]string{"store-gateway@*"}, wg, outchan, errchan)
-	wg.Wait()
-	b.Start([]string{"store-volume"}, wg, outchan, errchan)
-	wg.Wait()
+		// we start gateway first to give metadata time to come up for volume
+		b.Start([]string{"store-gateway@*"}, wg, outchan, errchan)
+		wg.Wait()
+		b.Start([]string{"store-volume"}, wg, outchan, errchan)
+		wg.Wait()
+	}
 
 	// start logging subsystem first to collect logs from other components
 	outchan <- fmt.Sprintf("Logging subsystem...")
-	b.Start([]string{"logger"}, wg, outchan, errchan)
-	wg.Wait()
+	if !stateless {
+		b.Start([]string{"logger"}, wg, outchan, errchan)
+		wg.Wait()
+	}
 	b.Start([]string{"logspout"}, wg, outchan, errchan)
 	wg.Wait()
 
-	b.Start([]string{
-		"database", "registry@*", "controller", "builder",
-		"publisher", "router@*"},
-		&_wg, _outchan, _errchan)
+	if stateless {
+		b.Start([]string{
+			"registry@*", "controller", "builder",
+			"publisher", "router@*"},
+			&_wg, _outchan, _errchan)
+	} else {
+		b.Start([]string{
+			"database", "registry@*", "controller", "builder",
+			"publisher", "router@*"},
+			&_wg, _outchan, _errchan)
+	}
 
 	outchan <- fmt.Sprintf("Control plane...")
-	b.Start([]string{"database", "registry@*", "controller"}, wg, outchan, errchan)
+	if stateless {
+		b.Start([]string{"registry@*", "controller"}, wg, outchan, errchan)
+	} else {
+		b.Start([]string{"database", "registry@*", "controller"}, wg, outchan, errchan)
+	}
 	wg.Wait()
 	b.Start([]string{"builder"}, wg, outchan, errchan)
 	wg.Wait()
@@ -157,9 +175,10 @@ func Stop(targets []string, b backend.Backend) error {
 	// if target is platform, stop all services
 	if len(targets) == 1 {
 		if targets[0] == PlatformCommand {
-			return StopPlatform(b)
-		}
-		if targets[0] == swarm {
+			return StopPlatform(b, false)
+		} else if targets[0] == StatelessPlatformCommand {
+			return StopPlatform(b, true)
+		} else if targets[0] == swarm {
 			return StopSwarm(b)
 		}
 	}
@@ -178,7 +197,7 @@ func Stop(targets []string, b backend.Backend) error {
 	return nil
 }
 
-func stopDefaultServices(b backend.Backend, wg *sync.WaitGroup, outchan chan string, errchan chan error) {
+func stopDefaultServices(b backend.Backend, stateless bool, wg *sync.WaitGroup, outchan chan string, errchan chan error) {
 
 	outchan <- fmt.Sprintf("Routing mesh...")
 	b.Stop([]string{"router@*"}, wg, outchan, errchan)
@@ -189,22 +208,33 @@ func stopDefaultServices(b backend.Backend, wg *sync.WaitGroup, outchan chan str
 	wg.Wait()
 
 	outchan <- fmt.Sprintf("Control plane...")
-	b.Stop([]string{"controller", "builder", "database", "registry@*"}, wg, outchan, errchan)
+	if stateless {
+		b.Stop([]string{"controller", "builder", "registry@*"}, wg, outchan, errchan)
+	} else {
+		b.Stop([]string{"controller", "builder", "database", "registry@*"}, wg, outchan, errchan)
+	}
 	wg.Wait()
 
 	outchan <- fmt.Sprintf("Logging subsystem...")
-	b.Stop([]string{"logger", "logspout"}, wg, outchan, errchan)
+	if stateless {
+		b.Stop([]string{"logspout"}, wg, outchan, errchan)
+	} else {
+		b.Stop([]string{"logger", "logspout"}, wg, outchan, errchan)
+	}
 	wg.Wait()
 
-	outchan <- fmt.Sprintf("Storage subsystem...")
-	b.Stop([]string{"store-volume", "store-gateway@*"}, wg, outchan, errchan)
-	wg.Wait()
-	b.Stop([]string{"store-metadata"}, wg, outchan, errchan)
-	wg.Wait()
-	b.Stop([]string{"store-daemon"}, wg, outchan, errchan)
-	wg.Wait()
-	b.Stop([]string{"store-monitor"}, wg, outchan, errchan)
-	wg.Wait()
+	if !stateless {
+		outchan <- fmt.Sprintf("Storage subsystem...")
+		b.Stop([]string{"store-volume", "store-gateway@*"}, wg, outchan, errchan)
+		wg.Wait()
+		b.Stop([]string{"store-metadata"}, wg, outchan, errchan)
+		wg.Wait()
+		b.Stop([]string{"store-daemon"}, wg, outchan, errchan)
+		wg.Wait()
+		b.Stop([]string{"store-monitor"}, wg, outchan, errchan)
+		wg.Wait()
+	}
+
 }
 
 // Restart stops and then starts the specified components.
@@ -247,9 +277,10 @@ func Install(targets []string, b backend.Backend, checkKeys func() error) error 
 	// if target is platform, install all services
 	if len(targets) == 1 {
 		if targets[0] == PlatformCommand {
-			return InstallPlatform(b, checkKeys)
-		}
-		if targets[0] == swarm {
+			return InstallPlatform(b, checkKeys, false)
+		} else if targets[0] == StatelessPlatformCommand {
+			return InstallPlatform(b, checkKeys, true)
+		} else if targets[0] == swarm {
 			return InstallSwarm(b)
 		}
 	}
@@ -268,18 +299,28 @@ func Install(targets []string, b backend.Backend, checkKeys func() error) error 
 	return nil
 }
 
-func installDefaultServices(b backend.Backend, wg *sync.WaitGroup, outchan chan string, errchan chan error) {
+func installDefaultServices(b backend.Backend, stateless bool, wg *sync.WaitGroup, outchan chan string, errchan chan error) {
 
-	outchan <- fmt.Sprintf("Storage subsystem...")
-	b.Create([]string{"store-daemon", "store-monitor", "store-metadata", "store-volume", "store-gateway@1"}, wg, outchan, errchan)
-	wg.Wait()
+	if !stateless {
+		outchan <- fmt.Sprintf("Storage subsystem...")
+		b.Create([]string{"store-daemon", "store-monitor", "store-metadata", "store-volume", "store-gateway@1"}, wg, outchan, errchan)
+		wg.Wait()
+	}
 
 	outchan <- fmt.Sprintf("Logging subsystem...")
-	b.Create([]string{"logger", "logspout"}, wg, outchan, errchan)
+	if stateless {
+		b.Create([]string{"logspout"}, wg, outchan, errchan)
+	} else {
+		b.Create([]string{"logger", "logspout"}, wg, outchan, errchan)
+	}
 	wg.Wait()
 
 	outchan <- fmt.Sprintf("Control plane...")
-	b.Create([]string{"database", "registry@1", "controller", "builder"}, wg, outchan, errchan)
+	if stateless {
+		b.Create([]string{"registry@1", "controller", "builder"}, wg, outchan, errchan)
+	} else {
+		b.Create([]string{"database", "registry@1", "controller", "builder"}, wg, outchan, errchan)
+	}
 	wg.Wait()
 
 	outchan <- fmt.Sprintf("Data plane...")
@@ -297,9 +338,10 @@ func installDefaultServices(b backend.Backend, wg *sync.WaitGroup, outchan chan 
 func Uninstall(targets []string, b backend.Backend) error {
 	if len(targets) == 1 {
 		if targets[0] == PlatformCommand {
-			return UninstallPlatform(b)
-		}
-		if targets[0] == swarm {
+			return UninstallPlatform(b, false)
+		} else if targets[0] == StatelessPlatformCommand {
+			return UninstallPlatform(b, true)
+		} else if targets[0] == swarm {
 			return UnInstallSwarm(b)
 		}
 	}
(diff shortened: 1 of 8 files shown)


docs/managing_deis/running-stateless-cluster.rst
```diff
@@ -0,0 +1,214 @@
...
+The :ref:`builder` component is responsible for building applications deployed
+to Deis via the ``git push`` workflow. It pushes to registry to store releases,
+so it will require no changes.
+
+Store
+^^^^^
+
+The :ref:`store` components implement Ceph itself. In a Ceph-less cluster, we
+will skip the installation and starting of these components. We will also
+mock a fake ``deis-store-volume`` component so that the controller and logger
```
Blystad: Previously we write that we will not be running logger, yet here we say that the mock deis-store-volume is required for the logger. I suggest removing the logger part of the line :)

docs/managing_deis/running-stateless-cluster.rst
```diff
@@ -0,0 +1,214 @@
...
+Fake deis-store-volume
+^^^^^^^^^^^^^^^^^^^^^^
+
+The ``deis-store-volume`` component mounts the shared Ceph filesystem used by
+both the :ref:`logger` and :ref:`controller`. For the controller to run, it needs
+to think that the log directory exists, so we create a fake ``deis-store-volume``
+component.
+
+This code snippet is intended to be run on a CoreOS host because it uses
+``fleetctl`` directly.
```
Blystad: Would it be possible to make deisctl install this for us? To remove a manual step.

docs/managing_deis/running-stateless-cluster.rst
```diff
@@ -0,0 +1,214 @@
...
+Confirm installation
+^^^^^^^^^^^^^^^^^^^^
+
+That's it! Deis is now running without Ceph. Issue a ``deisctl list`` to confirm
+that the services are started, and see :ref:`using_deis` to start using the cluster.
+
+Upgrading Deis
+--------------
+
+When following the :ref:`upgrading-deis` documentation, be sure to use
```
Blystad: I'm a bit sceptical to wether the normal upgrading deis documentation will work or not.

For inplace upgrade it might. For the new-cluster guide I think the instructions in my PDF is more correct or better then the current instructions in the documentation. Reason being that we store all configuration in a external database, all logs in a external endpoint, and as such upgrading is easier. In addition we don't have to recreate every app, just restart them.

deisctl/cmd/platform.go
```diff
@@ -11,12 +11,18 @@ import (
...
+func InstallPlatform(b backend.Backend, checkKeys func() error, stateless bool) error {
 
 	if err := checkKeys(); err != nil {
 		return err
 	}
 
+	if stateless {
+		fmt.Println("Warning: With a stateless control plane, `deis logs` will be unavailable.")
+		fmt.Println("Additionally, components will need to be configured to use external persistent stores.")
+		fmt.Println("See the official Deis documentation for details on running a stateless control plane.")
```
bacongobbler: A link to the documentation would be great here so the user doesn't have to go fishing for it :)

deisctl/cmd/cmd.go
```diff
@@ -268,18 +299,28 @@ func Install(targets []string, b backend.Backend, checkKeys func() error) error
...
+		b.Create([]string{"store-daemon", "store-monitor", "store-metadata", "store-volume", "store-gateway@1"}, wg, outchan, errchan)
+		wg.Wait()
+	}
 
 	outchan <- fmt.Sprintf("Logging subsystem...")
-	b.Create([]string{"logger", "logspout"}, wg, outchan, errchan)
+	if stateless {
+		b.Create([]string{"logspout"}, wg, outchan, errchan)
+	} else {
+		b.Create([]string{"logger", "logspout"}, wg, outchan, errchan)
```
bacongobbler: there seems to be a lot of if/else with duplicate code, but different component lists. What if we converted the list of strings into actual component structures that contained the component name and a boolean flag representing if the component is "stateless"? That way we can programatically remove certain components from the list based on whether it's stateless or not and reduce LoC. Would that seem fair to undertake with this PR or would a followup be better (considering that we're close to PR freeze day)?

docs/managing_deis/running-stateless-cluster.rst
carmstrong: I'm not happy with the fake unit being the long-term solution here, so I didn't want to add it to `deisctl`. I'm thinking if controller ensures that the log directory exists when it starts up, that may be the better fix, but I didn't want to go down that route just yet. I think we ship this as-is and then refine the process later.

docs/managing_deis/running-stateless-cluster.rst
carmstrong: Yep, I'll make that change.

docs/managing_deis/running-stateless-cluster.rst
carmstrong: > For inplace upgrade it might.

That's the recommended upgrade path, so we should recommend that for stateless upgrades as well. If there's an issue with an in-place upgrade for stateless clusters, we should see if we can fix that.

I'm not sure when someone would prefer to ever deploy a new cluster and cut over as your instructions suggest when they can instead just upgrade in-place, unless I'm missing something?

deisctl/cmd/platform.go
carmstrong: I thought about that, but didn't want to link to the "latest" docs because that may not agree with the user's provisioned version of Deis.

deisctl/cmd/cmd.go
carmstrong: > Would that seem fair to undertake with this PR or would a followup be better (considering that we're close to PR freeze day)?

I thin we can probably make another pass at deisctl in general to clean up some repeated code. In this case, I preferred the verbosity of enumerating what exactly we're starting each time. It may be less code to do it your way, but I think it's more readable this way. Someone reading the code can see what exactly the `stateless` flag skips.

deisctl/cmd/platform.go
bacongobbler: With other versions of `deisctl`, we modify `refresh-units` to pull from a specific tag if it's an official release. Perhaps we could do the same here? (i.e. v1.8.0 tagged release of deis docs for deisctl v1.8.0, otherwise latest)

This is a minor thing since stateless-platform is a very specific edge case, but just some food for thought :)

carmstrong: Yeah, I think that's a bit too transparent, though. We have no way of knowing which unit they're using. And I want @krancour to think about a better way to handle this other than mocking the unit (again, I think having controller ensure the log directory exists on startup is the way to go here), so I don't want to spend too much time worrying about the mock unit.

bacongobbler:
the approach looks good to me; now it's just time to test this sucker.

docs/managing_deis/running-stateless-cluster.rst
```diff
@@ -0,0 +1,213 @@
...
+* Log drain service with syslog log format compatibility (like `Papertrail`_)
+
+Understanding component changes
+-------------------------------
+
+Either directly or indirectly, all components in the :ref:`control-plane`
+require Ceph (:ref:`store`). Some components require changes to accommodate
+the removal of Ceph. The necessary changes are described below.
+
+Logger
```
bacongobbler: A lot of this information is very useful; we should possibly take some of this info and dump it into the [component docs](http://docs.deis.io/en/latest/understanding_deis/components/) once we merge this :+1:

docs/managing_deis/running-stateless-cluster.rst
bacongobbler: It doesn't need to be baked into deisctl! Since it's a glorified fleetctl wrapper we can use our own tooling for this!

For example, using the [ping example service from fleet](https://raw.githubusercontent.com/coreos/fleet/master/examples/ping.service):

```
><> cd $DEISCTL_UNITS
><> wget [ping.service]
><> mv ping.service deis-ping.service
><> deisctl install ping
deis-ping.service: loaded                                 
><> deisctl start ping
deis-ping.service: active/running                                 
><> deisctl journal ping
-- Logs begin at Wed 2015-07-01 09:14:37 . --
Jul 01 16:49:27 deis-03 systemd[1]: Started PING.
Jul 01 16:49:27 deis-03 bash[1675]: ping
Jul 01 16:49:27 deis-03 systemd[1]: Starting PING...
Jul 01 16:49:28 deis-03 bash[1675]: ping
Jul 01 16:49:29 deis-03 bash[1675]: ping
Jul 01 16:49:30 deis-03 bash[1675]: ping
Jul 01 16:49:31 deis-03 bash[1675]: ping
Jul 01 16:49:32 deis-03 bash[1675]: ping
Jul 01 16:49:33 deis-03 bash[1675]: ping
```

Another solution would be to change the last few calls to fleetctl to the following:

```
$ DEISCTL_UNITS=/tmp deisctl install store-volume
$ DEISCTL_UNITS=/tmp deisctl start store-volume
$ rm /tmp/deis-store-volume.service
```

I agree with @Blystad that using existing tooling would make for a happier installation experience, but the controller fix would also be a better solution so we don't have to hack this in. This is an officially supported workflow so we should make this as smooth as possible for users, in my opinion. :)

deisctl/cmd/platform.go
bacongobbler: :+1:

deisctl/cmd/cmd.go
bacongobbler: :+1:

deisctl/cmd/platform.go
krancour: @carmstrong 

> again, I think having controller ensure the log directory exists on startup is the way to go here

As far as I can tell (and I've double-checked this a few times) it already works that way.  Sanity check for me... have you seen it behave to the contrary?

carmstrong: > it already works that way

I know you had mentioned changing the controller to just `Wants` instead of `Requires` `deis-store-volume`, but I think @Blystad said it'll come up just fine either way. Does this mean we can just skip the mock `deis-store-volume` entirely?

carmstrong: I confirmed that controller refuses to start without `deis-store-volume`:
```
Jul 01 22:27:33 deis-02 fleetd[1050]: ERROR manager.go:129: Failed to trigger systemd unit deis-controller.service start: Unit deis-store-volume.service failed to load: No such file or directory.
```

Going to see if making controller just `Wants` it resolves the issue.

carmstrong: It does. Made that change, and now we don't even need the mock `deis-store-volume`.

carmstrong:
I changed deis-controller to only `Wants=deis-store-volume.service` instead of requiring it, removing the need to mock the unit. This is looking pretty clean, and is ready for final testing and review.

carmstrong:
This is failing on the known `deis run` issue with the test infrastructure, so at least it passed docs and provisioning...

deisctl/cmd/platform.go
krancour: :+1: I've been running my clusters just like that for a while, but it's awesome to have the result I was having independently verified.  imo, this does a lot to smooth out the difficulties of configuring a ceph-less cluster.

szymonpk:
Today I've played around a little bit with cephless deis, it's working rather fine. However few things:

1. In documentation you don't mention need of exporting DEISCTL_TUNNEL, setting key and domain via deisctl. You don't link anywhere to installation manual. I know it's for power users, but still.
2. `s3secure=false` is completely [ignored](https://github.com/deis/deis/blob/master/registry/templates/create_bucket#L12). In case of local s3 transport it may be fine, but for amazon I would like use https if possible.
3. boto's [get_all_buckets](https://github.com/deis/deis/blob/master/registry/templates/create_bucket#L16) method requires special s3 permission, which I don't wan't to give deis s3 user, more about this in 5. Maybe we can use `lookup`?
4. boto's [create_bucket](https://github.com/deis/deis/blob/master/registry/templates/create_bucket#L17) with single name param will always create bucket in US region, it may cause higher bill if you have bucket in different region and for sure it will degrade performance.
5. I've prepared [IAM policy](https://gist.github.com/szymonpk/07f3a0f18d78e6821983) which may be included into doc. It will limit permissions to just necessary ones, however I don't like `ListAllMyBuckets` described in 3.

bacongobbler:
I followed the documentation to a tee, but I see this error every time in the controller during boot:

```
Jul 02 18:12:20 deis-02 sh[12158]: Traceback (most recent call last):
Jul 02 18:12:20 deis-02 sh[12158]: File "./manage.py", line 13, in <module>
Jul 02 18:12:20 deis-02 sh[12158]: execute_from_command_line(sys.argv)
Jul 02 18:12:20 deis-02 sh[12158]: File "/usr/lib/python2.7/site-packages/django/core/management/__init__.py", line 399, in execute_from_command_line
Jul 02 18:12:20 deis-02 sh[12158]: utility.execute()
Jul 02 18:12:20 deis-02 sh[12158]: File "/usr/lib/python2.7/site-packages/django/core/management/__init__.py", line 392, in execute
Jul 02 18:12:20 deis-02 sh[12158]: self.fetch_command(subcommand).run_from_argv(self.argv)
Jul 02 18:12:20 deis-02 sh[12158]: File "/usr/lib/python2.7/site-packages/django/core/management/base.py", line 242, in run_from_argv
Jul 02 18:12:20 deis-02 sh[12158]: self.execute(*args, **options.__dict__)
Jul 02 18:12:20 deis-02 sh[12158]: File "/usr/lib/python2.7/site-packages/django/core/management/base.py", line 284, in execute
Jul 02 18:12:20 deis-02 sh[12158]: self.validate()
Jul 02 18:12:20 deis-02 sh[12158]: File "/usr/lib/python2.7/site-packages/django/core/management/base.py", line 310, in validate
Jul 02 18:12:20 deis-02 sh[12158]: num_errors = get_validation_errors(s, app)
Jul 02 18:12:20 deis-02 sh[12158]: File "/usr/lib/python2.7/site-packages/django/core/management/validation.py", line 34, in get_validation_errors
Jul 02 18:12:20 deis-02 sh[12158]: for (app_name, error) in get_app_errors().items():
Jul 02 18:12:20 deis-02 sh[12158]: File "/usr/lib/python2.7/site-packages/django/db/models/loading.py", line 196, in get_app_errors
Jul 02 18:12:20 deis-02 sh[12158]: self._populate()
Jul 02 18:12:20 deis-02 sh[12158]: File "/usr/lib/python2.7/site-packages/django/db/models/loading.py", line 75, in _populate
Jul 02 18:12:20 deis-02 sh[12158]: self.load_app(app_name, True)
Jul 02 18:12:20 deis-02 sh[12158]: File "/usr/lib/python2.7/site-packages/django/db/models/loading.py", line 99, in load_app
Jul 02 18:12:20 deis-02 sh[12158]: models = import_module('%s.models' % app_name)
Jul 02 18:12:20 deis-02 sh[12158]: File "/usr/lib/python2.7/site-packages/django/utils/importlib.py", line 40, in import_module
Jul 02 18:12:20 deis-02 sh[12158]: __import__(name)
Jul 02 18:12:20 deis-02 sh[12158]: File "/usr/lib/python2.7/site-packages/guardian/models.py", line 93, in <module>
Jul 02 18:12:20 deis-02 sh[12158]: monkey_patch_user()
Jul 02 18:12:20 deis-02 sh[12158]: File "/usr/lib/python2.7/site-packages/guardian/__init__.py", line 24, in monkey_patch_user
Jul 02 18:12:20 deis-02 sh[12158]: User = get_user_model()
Jul 02 18:12:20 deis-02 sh[12158]: File "/usr/lib/python2.7/site-packages/django/contrib/auth/__init__.py", line 127, in get_user_model
Jul 02 18:12:20 deis-02 sh[12158]: user_model = get_model(app_label, model_name)
Jul 02 18:12:20 deis-02 sh[12158]: File "/usr/lib/python2.7/site-packages/django/db/models/loading.py", line 271, in get_model
Jul 02 18:12:20 deis-02 sh[12158]: self._populate()
Jul 02 18:12:20 deis-02 sh[12158]: File "/usr/lib/python2.7/site-packages/django/db/models/loading.py", line 75, in _populate
Jul 02 18:12:20 deis-02 sh[12158]: self.load_app(app_name, True)
Jul 02 18:12:20 deis-02 sh[12158]: File "/usr/lib/python2.7/site-packages/django/db/models/loading.py", line 99, in load_app
Jul 02 18:12:20 deis-02 sh[12158]: models = import_module('%s.models' % app_name)
Jul 02 18:12:20 deis-02 sh[12158]: File "/usr/lib/python2.7/site-packages/django/utils/importlib.py", line 40, in import_module
Jul 02 18:12:20 deis-02 sh[12158]: __import__(name)
Jul 02 18:12:20 deis-02 sh[12158]: File "/app/api/models.py", line 1057, in <module>
Jul 02 18:12:20 deis-02 sh[12158]: _etcd_client.get('/deis')
Jul 02 18:12:20 deis-02 sh[12158]: File "/usr/lib/python2.7/site-packages/etcd/client.py", line 453, in get
Jul 02 18:12:20 deis-02 sh[12158]: return self.read(key)
Jul 02 18:12:20 deis-02 sh[12158]: File "/usr/lib/python2.7/site-packages/etcd/client.py", line 344, in read
Jul 02 18:12:20 deis-02 sh[12158]: self.key_endpoint + key, self._MGET, params=params, timeout=timeout)
Jul 02 18:12:20 deis-02 sh[12158]: File "/usr/lib/python2.7/site-packages/etcd/client.py", line 582, in api_execute
Jul 02 18:12:20 deis-02 sh[12158]: return self._handle_server_response(response)
Jul 02 18:12:20 deis-02 sh[12158]: File "/usr/lib/python2.7/site-packages/etcd/client.py", line 598, in _handle_server_response
Jul 02 18:12:20 deis-02 sh[12158]: etcd.EtcdError.handle(**r)
Jul 02 18:12:20 deis-02 sh[12158]: File "/usr/lib/python2.7/site-packages/etcd/__init__.py", line 142, in handle
Jul 02 18:12:20 deis-02 sh[12158]: raise exc(msg)
Jul 02 18:12:20 deis-02 sh[12158]: KeyError: 'Key not found : /deis'
Jul 02 18:12:20 deis-02 systemd[1]: deis-controller.service: main process exited, code=exited, status=1/FAILURE
```

...Which is odd because I can see that the keyspace was created already. :confused:

bacongobbler:
Got it to work following the documentation provided. I created a trial Papertrail account (this will create a default route which will accept incoming UDP connections), spawned a free tier postgres server on RDS and created an S3 bucket, configured the cluster to use these credentials et voila, it works!

Code LGTM!

bacongobbler:
@szymonpk your points provide great feedback, however I think points 2-5 should be addressed in a separate PR so we can review that separately. Would you be up for a PR to address your comments?

carmstrong:
> I think points 2-5 should be addressed in a separate PR

:+1: especially if that could potentially change the default behavior for "regular" clusters. But great feedback, @szymonpk! I agree that supporting TLS would be much preferred.

szymonpk:
@bacongobbler should I create PR? I'm not very familiar with deis components hacking yet but I can try.

bacongobbler:
One issue with this workflow is that API logs for the app are lost and are not sent to papertrail. This is because the controller writes directly to the mounted log file, so `deis logs` shows some output:

```
><> deis logs
2015-07-02T18:34:53UTC deis[api]: bacongobbler created initial release
2015-07-02T18:46:08UTC deis[api]: bacongobbler deployed 14fb753
2015-07-02T18:46:10UTC deis[api]: bacongobbler scaled containers web=1
```

I know that this is meant to be improved upon, but just showing what I see :)

carmstrong:
> This is because the controller writes directly to the mounted log file, so deis logs shows some output:

Yeah, I think the warning "`deis logs` will not work" is good enough :) If the controller restarts and moves hosts, you wouldn't see those things in there. We should file a bug for that separately though, @bacongobbler - ideally when using an external log drain, we're shipping all logs there.

bacongobbler:
@szymonpk a PR or filing a separate issue so we can track your issue once this is merged should do.

mboersma:
Code LGTM.

carmstrong:
Merging - @bacongobbler could you please own creating an issue for your discovery and for @szymonpk's if he doesn't get to it?

szymonpk:
@carmstrong sure, first thing in the morning.

krancour:
I agree all logs should go to one place regardless of whether that is logger or somewhere external like papertrail when running ceph-less.  imo, we could stop having controller write directly to the log files and have it log to stdout like all other components and let logspout worry about forwarding those (to wherever) as in the usual case for apps and other components.  Any reason not to do that?

carmstrong:
> Any reason not to do that?

None that I can think of immediately, which means it's worth opening an issue and discussing! :+1:

bacongobbler:
The logger is the one that filters messages by tag. If we ship them to the logger directly by sending them to stdout from the controller, they'll all be filtered to `deis.log` since the log format for those api messages are `deis[api]`. The log format is important to discriminate which logs came from the controller and which logs came from the app itself. You'd probably need to figure out how to change thhat in the logger somehow if you want to approach the problem that way.

krancour:
@bacongobbler I'm confused by your previous comment.  Can you elaborate / clarify for my benefit?

I think we're talking about having `deis-controller` logs written not (necessarily) to `deis-logger`, but to wherever (`deis-logger` [possibly] _or_ external log service) _via_ `deis-logspout` in the same way as all applications do already.  Much as every log message from an application is tagged with the application's name, wouldn't log messages from `deis-controller` also be clearly tagged with their source?  Isn't `deis-controller`, from a logging standpoint, "just another application?"

If you're saying otherwise, that's where my confusion is.  Can you set me straight?

PR #3955, opened 2015-07-02
bug(deisctl): uninstall deis-store-admin as well

slack:
Remove deis-store-admin when using `deisctl uninstall platform`.

Before (after running `deisctl uninstall platform`):
```
core@ip-10-86-135-213 ~ $ fleetctl list-units
UNIT				MACHINE				ACTIVE	SUB
deis-store-admin.service	ae23b83e.../10.235.130.28	active	running
deis-store-admin.service	b5920086.../10.86.135.213	active	running
deis-store-admin.service	bf327080.../10.17.142.183	active	running
deis-store-admin.service	e678cbc5.../10.35.8.20		active	running
deis-store-admin.service	eb826a49.../10.87.0.137		active	running
```

After (fixed `deisctl`): 
```
core@ip-10-86-135-213 ~ $ fleetctl list-units
UNIT	MACHINE	ACTIVE	SUB
```


mboersma:
`deisctl install platform` doesn't install deis-store-admin though: it's optional like the swarm- units. So I don't think `deisctl uninstall platform` should try to remove deis-store-admin by default.

carmstrong:
> So I don't think deisctl uninstall platform should try to remove deis-store-admin by default.

Agreed - I'm also concerned this will emit errors for most users.

PR #3611, opened 2015-05-01
Updated copyright (it's 2015!)

SamuelMarks:

PR #3956, opened 2015-07-02
[WIP] ref(client): rewrite client in go.

Joshua-Anderson:
##### Endpoints:
- [x] auth
- [x] apps
- [x] keys
- [x] users

This PR duplicates the behavior of the python client as closely as possible.

I'd love any feedback anybody has on this PR, especially design.


.gitignore
```diff
@@ -26,6 +26,7 @@ builder/rootfs/usr/bin
 cache/image/bin/
 client/dist/
 client/makeself/
+client-go/deis
```
bacongobbler: I'd make it replace the `client/` dir entirely. We can archive the python client and move it elsewhere for historical purposes.

client-go/Makefile
```diff
@@ -0,0 +1,51 @@
...
+COMPONENT = $(notdir $(repo_path))
+IMAGE =  $(IMAGE_PREFIX)/$(COMPONENT):$(BUILD_TAG)
+
+build:
+	CGO_ENABLED=0 godep go build -a -installsuffix cgo -ldflags '-s' -o deis .
+
+installer:
+	rm -rf dist && mkdir -p dist
+	CGO_ENABLED=0 godep go build -a -installsuffix cgo -ldflags '-s' -o dist/deis .
+	@if [ ! -d makeself ]; then git clone -b single-binary https://github.com/deis/makeself.git; fi
```
bacongobbler: this part isn't necessary :)

client-go/Makefile
```diff
@@ -0,0 +1,51 @@
...
+		&& echo 'Please move deis to a directory in your search PATH.' \
+		&& echo \
+		&& echo 'See http://docs.deis.io/ for documentation.' \
+		&& echo"
+
+install:
+	godep go install -v .
+
+setup-root-gotools:
+	sudo GOPATH=/tmp/tmpGOPATH go get -u -v golang.org/x/tools/cmd/cover
```
bacongobbler: what's with the GOPATH change, followed by an immediate `rm -rf`?

.gitignore
carmstrong: My original thought was to have it written in parallel like @Joshua-Anderson is doing here so that we could begin asking users to test it before we make it the official client, but @mboersma made the point that our CI runs will fully exercise the client and validate it pretty effectively. I'd like to see what he thinks about swapping them out in one fell swoop.

client-go/README.md
```diff
@@ -0,0 +1,215 @@
+# Deis Control Utility
+
+`deisctl` is a command-line utility used to provision and operate a Deis cluster.
```
bacongobbler: oopsie!

.gitignore
mboersma: In the end it should replace the `client/` dir IMHO, but it would be safest to be able to make a python client build up until that point. I think going back in history with `git` is sufficient to allow that, or else the old client code could move to client-old or somesuch. tl:dr, however you want to do it, but when we find an unpexected issue with the go CLI we will be glad we still have the python CLI working as a fallback.

client-go/cmd/auth.go
```diff
@@ -0,0 +1,212 @@
...
+	"net/url"
+	"os"
+	"strings"
+
+	"github.com/deis/deis/client-go/controller/client"
+	"golang.org/x/crypto/ssh/terminal"
+)
+
+// Register a user with the controller
+func Register(controller string, username string, password string, email string,
```
bacongobbler: this is slightly confusing. In terms of Go best practices, the `cmd` package should be a "dumb" implementation of each command, only being given an argument list which then it can pass along to the underlying client libraries. For example, in golang.org/src/cmd/go/build.go, they use a [Command structure](http://golang.org/src/cmd/go/build.go#L30), which then calls [rubBuild](http://golang.org/src/cmd/go/build.go#L261) that takes an argument list. This is also replicated amongst other projects in the community. I think it'd be better to standardize and follow the same best practices as other projects. That way each command can specify its own flagset that is independent of the root binary's flagset, among other benefits (standardization, ability for users to call cmd.Register with an argument list without it breaking across versions etc.). What do you think about taking this approach?

tests/utils/itutils.go
```diff
@@ -20,7 +20,7 @@ import (
 )
 
 // Deis points to the CLI used to run tests.
-var Deis = "deis "
+var Deis = os.Getenv("DEIS_BINARY")
```
bacongobbler: this will break local environments that do not have $DEIS_BINARY set (I don't run test-integration.sh locally, just boot up my own cluster then run `make -C tests test-full`). Would you mind changing this to:

```
var Deis = os.Getenv("DEIS_BINARY")
if Deis == "" {
    Deis = "deis "
}
```

While we're at it, we should remove the space requirement in $DEIS_BINARY and just update the one reference to this variable :)

client-go/Makefile
Joshua-Anderson: I'm not sure, I duplicated it from deisctl and didn't look carefully at it.

tests/utils/itutils.go
Joshua-Anderson: Done!

client-go/cmd/auth.go
bacongobbler: we talked about this offline. I'm fine with the approach.

carmstrong:
@Joshua-Anderson I think we should develop this in parallel without hooking it up to Jenkins. Then, we can PR and merge each sub-command of the client as you write it. I think that's preferred to reviewing one monster PR.

Joshua-Anderson:
Added `app` endpoint.

Joshua-Anderson:
@carmstrong I removed Jenkins hook and pushed all current code. This can be merged if we are interested in merging the WIP code.

carmstrong:
> This can be merged if we are interested in merging the WIP code.

We are indeed - thanks! We'll get to this after we ship 1.8.

Joshua-Anderson:
merged in `keys` endpoint

carmstrong:
> merged in keys endpoint

I thought we were going to keep this PR the base, which we can review/merge, and then PR the sub-sections one at a time so they are easier to review? This PR is quickly turning into a beast to review.

Joshua-Anderson:
`ps` endpoint is complete. It is available [here](https://github.com/Joshua-Anderson/deis/commit/8cf539e28a74e901fb8990682c552f74511b903c) and I will propose a PR with it after this PR is merged.

Joshua-Anderson:
`config` endpoint is complete. It is available here Joshua-Anderson/deis@1be4380f434a6df5d2bad2a30118ae61170f1d3b and I will propose a PR with it after this PR is merged.

Joshua-Anderson:
`domains` endpoint is complete. It is available at Joshua-Anderson/deis@8a0f6a390debde23ab674cfc5c10d1708e31bd7c and I will propose a PR with it after this PR is merged.

bacongobbler:
I'd also add plugins to the list of features required for this new client: http://docs.deis.io/en/latest/customizing_deis/cli-plugins/

Joshua-Anderson:
Yeah. I missed those at first glance (not well documented in the code). I''ll open a PR after this one is merged in.

client-go/controller/client/client.go
```diff
@@ -0,0 +1,96 @@
...
+
+	return &Client{HTTPClient: CreateHTTPClient(settings.SslVerify), SSLVerify: settings.SslVerify,
+		URL: *URL, Token: settings.Token, Username: settings.Username}, nil
+}
+
+// Save settings to a file
+func (c Client) Save() error {
+	settings := settingsFile{Username: c.Username,
+		SslVerify:  c.SSLVerify,
+		Controller: c.URL.String(), Token: c.Token}
```
bacongobbler: The controller value gets saved accidentally as `http://deis.domain.com/v1/auth/register` after a `deis login`. We should strip out the URL path here

client-go/cmd/apps.go
```diff
@@ -0,0 +1,245 @@
...
+	}
+
+	app, err := apps.Get(*c, appID)
+
+	if err != nil {
+		return err
+	}
+
+	fmt.Printf("=== %s Application\n", app.ID)
+	fmt.Println("updated: ", app.Updated)
```
bacongobbler: The old client would return the raw JSON object string, though I like this format better. However, if we're deviating from Heroku's look-and-feel here, I think it'd look better if this was tab-ified, much like Heroku's client:

```
$ heroku info -a polar-shelf-1435
=== polar-shelf-1435
Dynos:         0
Git URL:       https://git.heroku.com/polar-shelf-1435.git
Owner Email:   me@bacongobbler.com
Region:        us
Stack:         cedar-14
Web URL:       https://polar-shelf-1435.herokuapp.com/
Workers:       0
```

bacongobbler:
- [x] apps
 - [x] auth
 - [x] keys
 - [x] users

Only my comments are barring an LGTM. Looks great, and the performance is noticeably faster!

carmstrong:
@Joshua-Anderson This will also need to be rebased off master to pass CI.

Joshua-Anderson:
@bacongobbler Done!

carmstrong:
I'd also like either @mboersma or @technosophos (ideally both) to review. Deferring my LGTM to their review.

client-go/LICENSE
```diff
@@ -0,0 +1,13 @@
+Copyright 2014 Engine Yard, Inc.
```
mboersma: For a new copyright notice we should use the year the code was introduced: 2015.

client-go/Makefile
mboersma: This is a voodoo incantation necessary to get `go vet` installed on some platforms, like my Mac. We tried to remove it once and found yes, it still causes errors to install `go vet` as other than root.

client-go/LICENSE
sgoings: Why don't we keep the copyright year current everywhere?

client-go/cmd/apps.go
```diff
@@ -0,0 +1,245 @@
...
+	"os"
+	"strings"
+
+	"github.com/deis/deis/client-go/controller/client"
+	"github.com/deis/deis/client-go/controller/models/apps"
+	"github.com/deis/deis/client-go/controller/models/config"
+)
+
+// AppCreate creates an app with the application
+func AppCreate(id string, buildpack string, remote string, noRemote bool) error {
```
mboersma: The first two parameter types can be omitted here, if you like this style:
```go
func AppCreate(id, buildpack, remote string, noRemote bool) error
```

Joshua-Anderson: I don't care either way. Is there a standard throughout the deis project. If the function was only one type, i.e. `func Test(first, second string)` I generally went with but if there were more than one distinct type I added a type to everything to avoid confusion. I'm surprised that `gofmt

client-go/LICENSE
mboersma: I'm not a lawyer, but my arguments for leaving it alone generally are:
1. It's a pain
2. Copyright can't be reasserted unless meaningful changes were actually made
2a. Lawyers themselves disagree about the previous point
3. Updating the copyright year extends the legal horizon more than a century from now, which seems unlikely to be consequential to Deis

client-go/cmd/apps.go
mboersma: There isn't a standard for this that I'm aware of in our codebase; up to you. Your argument makes sense to me.

bacongobbler: Indeed. The tooling dictates the format, generally speaking, so if `gofmt` isn't complaining it's not a huge deal.

client-go/LICENSE
bacongobbler: related: #3611

sgoings: Thanks. Just have been a part of projects in the past that are adamant about keeping that copyright year up to date.

client-go/controller/client/client.go
Joshua-Anderson: Fixed!

client-go/cmd/apps.go
Joshua-Anderson: Done!

bacongobbler:
code LGTM!

client-go/Makefile
```diff
@@ -0,0 +1,39 @@
...
+GO_FILES = $(wildcard *.go)
+GO_PACKAGES = parser cmd controller/api controller/client $(wildcard controller/models/*)
+GO_PACKAGES_REPO_PATH = $(addprefix $(repo_path)/,$(GO_PACKAGES))
+
+COMPONENT = $(notdir $(repo_path))
+IMAGE =  $(IMAGE_PREFIX)/$(COMPONENT):$(BUILD_TAG)
+
+build:
+	CGO_ENABLED=0 godep go build -a -installsuffix cgo -ldflags '-s' -o deis .
+
```
mboersma: We should add the check to ensure this really got built as a static binary:
```
$(call check-static-binary,deis)
```

client-go/cmd/apps.go
```diff
@@ -0,0 +1,245 @@
...
+		return err
+	}
+
+	fmt.Printf("done, created %s\n", app.ID)
+
+	if buildpack != "" {
+		configValues := map[string]string{
+			"BUILDPACK_URL": buildpack,
+		}
+		err = config.Set(*c, app.ID, configValues)
```
mboersma: I think Deis go code prefers the one-line form for error checks if you don't need to refer to the return values later. (We should have this in our coding standards doc if there is consensus on this.) Doesn't have to be changed; just sayin'.
```go
if err = config.Set(*c, app.ID, configValues); err != nil {
        return err
}
```

client-go/cmd/apps.go
```diff
@@ -0,0 +1,245 @@
...
+		return err
+	}
+
+	if appID == "" {
+		appID, err = c.DetectApp()
+
+		if err != nil {
+			return err
+		}
+	}
```
mboersma: This stanza starting with `c, err := client.Load()` is repeated in several functions. Can we make it a helper function to keep the code DRY?

client-go/controller/client/auth.go
```diff
@@ -0,0 +1,243 @@
...
+	headers.Add("Content-Type", "application/json")
+	addUserAgent(&headers)
+
+	res, err := rawRequest(client, "POST", URL.String(), bytes.NewBuffer(body), headers, 201)
+
+	if err != nil {
+		return err
+	}
+
+	res.Body.Close()
```
mboersma: I think it would be more defensive coding to move this above the error check and call it as `defer res.Body.Close()`.

client-go/controller/client/auth.go
```diff
@@ -0,0 +1,243 @@
...
+	if err != nil {
+		return err
+	}
+
+	resBody, err := ioutil.ReadAll(res.Body)
+
+	token := api.AuthLoginResponse{}
+
+	err = json.Unmarshal([]byte(resBody), &token)
+	res.Body.Close()
```
mboersma: Same as above--I think it would be safer coding to move this right after `rawRequest` and use `defer`.

client-go/controller/client/auth.go
```diff
@@ -0,0 +1,243 @@
...
+		return err
+	}
+
+	controllerClient := Client{Username: username, SSLVerify: sslVerify,
+		URL: controllerURL, Token: token.Token}
+
+	err = controllerClient.Save()
+
+	if err != nil {
+		return nil
```
mboersma: Should be `return err` I assume (and I would make it a one-line check with `controllerClient.Save()` above).

client-go/controller/client/auth.go
```diff
@@ -0,0 +1,243 @@
...
+	client, err := Load()
+
+	if err != nil {
+		return err
+	}
+
+	body, status, err := client.BasicRequest("DELETE", "/v1/auth/cancel/", []byte{})
+
+	if status != 204 {
+		return fmt.Errorf("Cancelation failed: %s", body)
```
mboersma: Should be "Cancellation" with two Ls, even if the python client also spelled it wrong. :smile:

client-go/cmd/apps.go
technosophos: The exception being if the `if` line would then be too long?

mboersma: > if the `if` line would then be too long

Of course, although our [coding standards](http://docs.deis.io/en/latest/contributing/standards/#code-standards) allow lines up to 99 chars, so that shouldn't happen often.

client-go/controller/client/client.go
```diff
@@ -0,0 +1,96 @@
...
+	"net/url"
+	"os"
+	"path"
+)
+
+// Client oversees the interaction between the client and controller
+type Client struct {
+	// HTTP client used to communicate with the API.
+	HTTPClient *http.Client
+	SSLVerify  bool
```
mboersma: `SSLVerify` needs a comment as well so it doesn't feel left out.

client-go/controller/client/http.go
```diff
@@ -0,0 +1,138 @@
...
+	req.Header = headers
+
+	res, err := client.Do(req)
+
+	if err != nil {
+		return nil, err
+	}
+
+	if res.StatusCode != expectedStatusCode {
+		defer res.Body.Close()
```
mboersma: Should this go on line 32 instead, right after `client.Do`?

bacongobbler: it should go right after the error has been confirmed non-nil or else it will raise a panic() trying to call close on a nil response body.

mboersma: Right you are: line 36 then.

Joshua-Anderson: I don't want to close the body because I want to return the response. The reason why I am closing the body here is a that this block runs when the webserver returned an error, so this function is returning with an error too and needs to close the body.

technosophos: @Joshua-Anderson is right to do it this way. The body only gets closed on the error condition. In all other cases, it's the responsibility of the caller to close the body.

mboersma: Yep, I missed the return. Thanks!

Joshua-Anderson:
Thanks for the feedback @mboersma @technosophos I think I implemented all of it.

Joshua-Anderson:
@technosophos I still need to switch over to using the color library in my code, will do that tomorrow.

client-go/Makefile
mboersma: You can remove these sections of the Makefile once we merge #4050.

Joshua-Anderson:
Now the code is using the prettyprint package to print colors and is ready to go.

mboersma:
Code LGTM.

technosophos:
I think we should merge this so that @Joshua-Anderson can get on with the rest of the tasks.

Code LGTM.

PR #3968, opened 2015-07-06
fix(tests): check https://$DEV_REGISTRY

bacongobbler:
In the case of DockerHub which just closed down http:// access this morning,
we should also be checking https://, which is still up.


carmstrong:
Code LGTM.

bacongobbler:
@mboersma also found another issue in CI, which appears to be due to the fact that we're using the FQDN for DockerHub, so I've added that here as well.

https://ci.deis.io/job/test-ec2/1675/consoleFull#682192135502066c60b2236c09cc9bf8

mboersma:
LGTM. I tested this manually to ensure a local registry still works. Note that this can turn a bunch of the `docker tag` commands into no-ops, which is apparently fine:
```console
$ docker tag -f deisci/swarm:jenkins-test-ec2-1677 deisci/swarm:jenkins-test-ec2-1677
```

gabrtv:
This passed the relevant tests.  BDFL-merging despite test failure as this is blocking CI.

PR #3972, opened 2015-07-06
[wip] feat(*): label all of the deis platform components

slack:
Requires Docker 1.6, so blocked until that becomes available. Making a PR now for thoughts and discussion!

Some thoughts on a `io.deis` label namespace would be great:
* `io.deis.application=name`
* `io.deis.environment=<production|staging>`

Also mixing in any fleet machine metadata or customer-provided metadata could certainly be useful, but that is largely a thought experiment at the moment.

### DIFF
--- a/deisctl/units/deis-builder.service
+++ b/deisctl/units/deis-builder.service
@@ -8,7 +8,7 @@ ExecStartPre=/bin/sh -c "docker inspect deis-builder-data >/dev/null 2>&1 || doc
 ExecStartPre=/bin/sh -c "IMAGE=`/run/deis/bin/get_image /deis/builder` && docker history $IMAGE >/dev/null 2>&1 || docker pull $IMAGE"
 ExecStartPre=/bin/sh -c "docker inspect deis-builder >/dev/null 2>&1 && docker rm -f deis-builder || true"
 ExecStartPre=-/bin/sh -c "/sbin/losetup -f"
-ExecStart=/bin/sh -c "IMAGE=`/run/deis/bin/get_image /deis/builder` && docker run --name deis-builder --rm -p 2223:2223 --volumes-from=deis-builder-data -c 800 -e EXTERNAL_PORT=2223 -e HOST=$COREOS_PRIVATE_IPV4 --privileged -v /etc/environment_proxy:/etc/environment_proxy $IMAGE"
+ExecStart=/bin/sh -c "IMAGE=`/run/deis/bin/get_image /deis/builder` && docker run --label io.deis.control=builder --name deis-builder --rm -p 2223:2223 --volumes-from=deis-builder-data -c 800 -e EXTERNAL_PORT=2223 -e HOST=$COREOS_PRIVATE_IPV4 --privileged -v /etc/environment_proxy:/etc/environment_proxy $IMAGE"
 ExecStartPost=/bin/sh -c "echo 'Waiting for builder on 2223/tcp...' && until ncat $COREOS_PRIVATE_IPV4 2223 --exec '/usr/bin/echo dummy-value' >/dev/null 2>&1; do sleep 1; done"
 ExecStop=-/usr/bin/docker stop deis-builder
 Restart=on-failure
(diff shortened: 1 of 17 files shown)


bacongobbler:
just to gather your thoughts, what's would this feature help you accomplish? Just looking for a way to group components by their label for monitoring purposes? :)

carmstrong:
> Just looking for a way to group components by their label for monitoring purposes? :)

That's the motivation as I understand it. More visibility for metrics is a :+1: from me.

slack:
Yup, want to be able to use upstream tools that use docker labels to group container metrics both for graphing and alerting.

For example:
*  `if CPU > 80% and container.label matches "io.deis.component=store*" then warn`
* show me a graph of page faults where `container.label matches "io.deis.application=foozy-biz"`

Once Docker 1.6 lands, I'll deploy these branches and get some real-world examples that make more sense. :dancers:

bacongobbler:
It might make sense to [add labels](https://github.com/deis/deis/blob/master/controller/scheduler/fleet.py#L126) to the [apps as well](https://github.com/deis/deis/blob/master/controller/scheduler/fleet.py#L382), if you're up for it :)

slack:
Will do, stopped as soon as I realized we weren't on Docker 1.6.

carmstrong:
> stopped as soon as I realized we weren't on Docker 1.6.

I'm pretty sure we're at Docker 1.6.2 on the hosts - the Docker-in-Docker within builder is only Docker 1.5.0 so we can't yet add tooling to individual apps, but this PR as-is could potentially be merged.

slack:
I was deploying units against a v1.7.3 cluster which is CoreOS 647.2.0 and thus Docker 1.5.0. Explains that!

mboersma:
Deis will be using Docker 1.5.0 for at least the next release: see #3992. There is a simultaneous `docker pull` bug present in Docker 1.6.2 and 1.7.0 that needs to be fixed before we can upgrade. :no_mouth:

carmstrong:
Docker 1.7.1 is now in the alpha channel: https://coreos.com/releases/#745.1.0

I'm tracking this PR for Deis v1.10, since I'm hoping it will have landed in stable by then.

carmstrong:
@marineam said it may be a while :(

```
1:49:54 PM <marineam> carmstrong: it will be a while, 1.7.1 still has a bug with cgroups that prevents it from working with rkt on the same host so it isn't even going to beta yet :( we are working on sorting out a fix.
```

slack:
Will pick up after deis/deis#4438 is merged.

slack:
With the docker bump merged, I'd like to re-kindle the labels conversation.

How do we want to manage the label/tag namespace?

This PR is using `io.deis.component` for platform components and will add `io.deis.application=appname`

Should it be sliced and diced differently?

bacongobbler:
sounds fine to me!

carmstrong:
> io.

I assume the `io.` prefix is standard? If so, the proposed namespace sounds good to me.

bacongobbler:
Yeah it's known as [reverse domain name notation](https://en.wikipedia.org/wiki/Reverse_domain_name_notation). It's fairly common with packages to avoid namespace collisions.

jchauncey:
Yeah that's how Java does packages

carmstrong:
> Yeah it's known as reverse domain name notation.

Familiar with the scheme, but didn't recognize it here - used to seeing `com.`

Code LGTM.

kalbasit:
Accidentally closed the PR.

mboersma:
Is there a way to make this backward-compatible? Users who have upgraded from Deis v1.10.0 clusters will still have Docker 1.5 on their hosts.

slack:
@mboersma hmm, probably not. We'd need to make `deisctl` pick a set of units and/or have the controller generate app unit files based on the cluster's docker version.

Supporting BC seems like it may add too much complexity for an arguable amount of benefit. Does the project have a history of dealing with stuff like this?

jchauncey:
Seems like a great v2 feature where we might be able to break backwards
compatibility

carmstrong:
We could theoretically wrap the `docker run` in a shell script which would check the version of Docker on the host and optionally include the `--label` if it's supported. But not sure that is worth trading clarity and simplicity for.

I'd let the code change wait for v2. In the interim, we could add documentation so users can add this to their own units if they so desire.

slack:
With k8s in the picture for v2 this doesn't really make sense at all.

PR #3975, opened 2015-07-07
feat(k8s) : add k8s scheduler support to deis

kmala:


contrib/coreos/user-data.example
```diff
@@ -20,19 +20,59 @@ coreos:
...
-      ListenStream=2375
-      BindIPv6Only=both
-      Service=docker.service
+      [Service]
+      Type=notify
+      Restart=always
+      RestartSec=5
+      Environment="TMPDIR=/var/tmp/"
+      Environment="DOCKER_HOST=unix:///var/run/early-docker.sock"
+      Environment="FLANNEL_VER=0.4.1"
```
aledbf: @kmala maybe this could be [v0.5.0](https://github.com/coreos/flannel/releases/tag/v0.5.0)? 
(It contains a fix for vxlan)

aledbf:
@kmala the current wget used in the to install the binaries run every time the service is started/restarted. Maybe it could be a good idea to check if the binary exists so is downloaded just once? (in total is ~120MB every time)

router/image/templates/nginx.conf
```diff
@@ -340,6 +340,157 @@ http {
...
     }
+    #start k8s apps
+    {{ range $kapp := ls "/registry/services/specs/default" }}
+    {{ $upstreams := printf "/registry/services/specs/default/%s" $kapp}}
+    upstream {{ $kapp }} {
+        {{ if exists "/deis/router/affinityArg" }}
+        hash $arg_{{ getv "/deis/router/affinityArg" }} consistent;
+        {{ end }}
+        {{$appdir := printf "/registry/services/specs/default/%s" $kapp}}{{range gets $appdir}}
+        server {{$data := json .Value}}{{$data.spec.clusterIP}}:80;
```
aledbf: in some cases (like the kubernetes service that uses the port 443) this will not work.

deisctl/units/deis-kube-apiserver.service
```diff
@@ -0,0 +1,28 @@
...
+ExecStartPre=/usr/bin/chmod +x /opt/bin/kube-apiserver
+ExecStart=/opt/bin/kube-apiserver \
+  --admission_control=NamespaceLifecycle,NamespaceAutoProvision,LimitRanger,SecurityContextDeny,ServiceAccount,ResourceQuota \
+  --allow_privileged=true \
+  --insecure_bind_address=0.0.0.0 \
+  --insecure_port=8080 \
+  --kubelet_https=true \
+  --secure_port=6443 \
+  --service-cluster-ip-range=10.100.0.0/16 \
+  --etcd_servers=http://127.0.0.1:4001 \
```
aledbf: Why not use something like
```
--etcd-servers=http://$(curl 127.0.0.1:4001/v2/machines | sed 's/ //g') \
```

to provide more than one serve in case of an error in etcd?

aledbf:
@kmala why etcd and not etcd2? (all the k8s guides use etcd2)

deisctl/units/deis-kube-kubelet.service
```diff
@@ -0,0 +1,16 @@
+[Unit]
+Description=Kubernetes Kubelet
+Documentation=https://github.com/GoogleCloudPlatform/kubernetes
+[Service]
+EnvironmentFile=/etc/environment
+ExecStartPre=/usr/bin/wget -P /opt/bin https://storage.googleapis.com/kubernetes-release/release/v0.19.3/bin/linux/amd64/kubelet
+ExecStartPre=/usr/bin/chmod +x /opt/bin/kubelet
+ExecStartPre=/usr/bin/mkdir -p /opt/kubernetes/manifests/
+ExecStart=/bin/bash -c '/opt/bin/kubelet --address=0.0.0.0 --port=10250 --hostname_override=$COREOS_PRIVATE_IPV4 --api_servers=`/usr/bin/etcdctl get /deis/scheduler/k8s/master`:8080 --allow_privileged=true --cluster_dns=10.100.0.10 --cluster_domain=cluster.local --logtostderr=true --config=/opt/kubernetes/manifests/ --healthz_bind_address=0.0.0.0 --healthz_port=10248'
```
bacongobbler: I thought we were "containerizing" these services so it's easier to run these locally without meddling with the host filesystem. Does it not make sense to run these agents inside a container?

aledbf: > I thought we were "containerizing" these services so it's easier to run these locally without meddling with the host filesystem, as well as better packaging, etc

@bacongobbler yes it makes sense but is not ready for that yet.

https://github.com/GoogleCloudPlatform/kubernetes/issues/4869
https://github.com/kelseyhightower/kubeos

kmala:
I had made change to the  flannel version, skip the download if the binary exists,and pass all the nodes to the etcd_servers parameter.

deisctl/units/deis-kube-apiserver.service
```diff
@@ -0,0 +1,28 @@
...
+ExecStartPre=/usr/bin/chmod +x /opt/bin/kube-apiserver
+ExecStart=/bin/bash -c "/opt/bin/kube-apiserver \
+  --admission_control=NamespaceLifecycle,NamespaceAutoProvision,LimitRanger,SecurityContextDeny,ServiceAccount,ResourceQuota \
+  --allow_privileged=true \
+  --insecure_bind_address=0.0.0.0 \
+  --insecure_port=8080 \
+  --kubelet_https=true \
+  --secure_port=6443 \
+  --service-cluster-ip-range=10.100.0.0/16 \
+  --etcd_servers=http://$(curl 127.0.0.1:4001/v2/machines | sed 's/ //g') \
```
aledbf: @kmala I suggested this change but this is the correct one:
```
--etcd_servers=$(curl -s 127.0.0.1:4001/v2/machines | sed 's/ //g') \
```

aledbf:
@kmala I did something similar (but different) to download and check the k8s binary version:
https://github.com/aledbf/deis/blob/kubernetes/deisctl/units/deis-kube-apiserver.service#L11
https://github.com/aledbf/deis/blob/kubernetes/contrib/coreos/user-data.example#L147
https://github.com/aledbf/deis/blob/kubernetes/contrib/coreos/user-data.example#L306

So just changing the value of `K8S_VERSION` and restarting the service it will update the binary

aledbf:
@kmala the memory limits are being ignored. The [format](https://github.com/GoogleCloudPlatform/kubernetes/blob/master/docs/resource_quota_admin.md) is not the same used in docker run

controller/scheduler/k8s.py
```diff
@@ -0,0 +1,556 @@
...
+              name,status, reason, data)
+          raise RuntimeError(errmsg)
+      parsed_json =  json.loads(data)
+      return parsed_json
+
+    def deploy(self, name, image, command, **kwargs):
+        app_name = kwargs.get('aname', {})
+        old_rc = self._get_old_rc(app_name)
+        new_rc = self._create_rc(name, image, command, **kwargs)
+        desired = int(old_rc["spec"]["replicas"])
```
aledbf: If for some reason is not possible to get this it throws an error
```
Jul 08 21:25:39 nodo-2 sh[1111]: desired = int(old_rc["spec"]["replicas"])
Jul 08 21:25:39 nodo-2 sh[1111]: TypeError: 'int' object has no attribute '__getitem__'
```

kmala:
I had took changes from you to download if a new binary is present.Currently there is already a file /etc/environment and hence i had named it to a different one.
I had also made change for the memory limits to work.

aledbf:
@kmala yes, the limits works and the scale too (without any error about the replicationController).

aledbf:
@kmala testing `config:set` and `config:unset` I see this error:
`cannot be updated: the object has been modified; please apply your changes to the latest version and try again`

aledbf:
> Currently there is already a file /etc/environment

That was my intention to not add another configuration file.
My real objective is to have just one place where is required to change the kubernetes binaries and only with a restart it will check if it exists, the version and download the file just when necessary.

kmala:
I had fixed the issue with the config.It was occurring because the RC is taking sometime to get created in few cases.

carmstrong:
@kmala You'll need to remove a lot of the etcd2 changes that bump the CoreOS release or move from the stable channel - those are no longer necessary. See #4005

kmala:
@carmstrong Yeah sure...I will change them.

mboersma:
I know this is a WIP, but just to note: this is [failing in CI](https://ci.deis.io/job/test-ec2/1828) because of python style errors. Run `make -C controller test-style` to see them.

aledbf:
@kmala you need to add 
```
mkdirEtcd(client, "/registry/services/specs/default")
```
after https://github.com/kmala/deis/blob/k8s/router/boot.go#L62 to pass the router test

docs/customizing_deis/choosing-a-scheduler.rst
```diff
@@ -120,19 +125,55 @@ The Swarm Scheduler is now active. Commands such as ``deis destroy`` or
...
+.. note::
+
+    **Known Issues**
+
+    - Kubernetes ApiServer is not HA. If ApiServer is rescheduled it won't effect the existing Apps
+      but affects creation of new Apps.
+    - Installing flannel is not backwards compatible as it changes Docker bridge which requires restarting
+      Docker Engine.
+    - Docker Engine is started with arguments ``--iptables=false --ip-masq=false`` for flannel ip_tables entries to take
+      precedence over docker ip_tables.
```
aledbf: This is not correct. There is no docker iptables rules at all

kmala:
requires #4024

bacongobbler:
> /home/jenkins/workspace/test-ec2/src/github.com/deis/deis/docs/customizing_deis/choosing-a-scheduler.rst:14: ERROR: Unknown target name: "mesos with marathon framework".

Please make sure `make test-unit` runs smoothly locally. Once it goes green and everything seems stable I'll give it a whirl. Thanks! :)

mboersma:
This also needs a rebase before you test and resubmit it. :recycle:

controller/scheduler/k8s.py
```diff
@@ -0,0 +1,636 @@
...
+            create = True
+            rc = self._get_rc_(name)
+            if ("observedGeneration" in rc["status"]
+                    and rc["metadata"]["generation"] == rc["status"]["observedGeneration"]):
+                break
+            time.sleep(1)
+        return json.loads(data)
+
+    def create(self, name, image, command, **kwargs):
+        self._create_rc(name, image, command, **kwargs)
```
aledbf: @kmala you need to change the creation order. First the service and then the rc because with the current order the spreading of pods in the cluster is not possible.

https://github.com/GoogleCloudPlatform/kubernetes/issues/7257
https://github.com/GoogleCloudPlatform/kubernetes/issues/11144

bacongobbler: +1. Makes sense according to the linked issues.

aledbf:
@kmala I found a way to improve the spreading using a custom policy.
The difference with the default is that the `ServiceSpreadingPriority` weight is 2

```
{
    "kind": "Policy",
    "apiVersion": "v1",
    "predicates": [
        {
            "name": "PodFitsPorts"
        },
        {
            "name": "PodFitsResources"
        },
        {
            "name": "NoDiskConflict"
        },
        {
            "name": "MatchNodeSelector"
        },
        {
            "name": "HostName"
        }
    ],
    "priorities": [
        {
            "name": "LeastRequestedPriority",
            "weight": 1
        },
        {
            "name": "BalancedResourceAllocation",
            "weight": 1
        },
        {
            "name": "ServiceSpreadingPriority",
            "weight": 2
        },
        {
            "name": "EqualPriority",
            "weight": 1
        }
    ]
}
```

mboersma:
Needs rebase against master, sorry for the hassle.

controller/scheduler/k8s.py
```diff
@@ -0,0 +1,647 @@
...
+        loc.update(re.match(MATCH, container_fullname).groupdict())
+        mem = kwargs.get('memory', {}).get(loc['c_type'])
+        cpu = kwargs.get('cpu', {}).get(loc['c_type'])
+        if mem or cpu:
+            containers[0]["resources"] = {"limits": {}}
+        if mem:
+            if mem[-2:-1].isalpha() and mem[-1].isalpha():
+                mem = mem[:-1]
+            mem = mem+"i"
+            containers[0]["resources"]["limits"]["memory"] = mem
```
aledbf: This creates a resource reservation (ram) not a limit in the usage. So in a node with 8GB, a limit of 1G means only 8 pods
```
Combined requested resources <total ram in bytes of reserver resources> from existing pods exceeds capacity <total ram in bytes in the node> on node
````

File: controller/scheduler/k8s.py [hunk 10]
```diff
@@ -0,0 +1,647 @@
...
+        conn_serv = httplib.HTTPConnection(self.target+":"+self.port)
+        conn_serv.request('POST', '/api/'+self.apiversion+'/namespaces/default/services',
+                          headers=headers, body=copy.deepcopy(template))
+        resp = conn_serv.getresponse()
+        data = resp.read()
+        reason = resp.reason
+        status = resp.status
+        conn_serv.close()
+        if not 200 <= status <= 299:
+            errmsg = "Failed to create Service:{} {} {} - {}".format(
```
aledbf: Is not possible to check the error and in case the error is `AlreadyExists` just continue?

kmala: we can check that but i think we shouldn't ignore it as we won't get the AlreadyExists error in any correct/valid scenario.

controller/scheduler/k8s.py
kmala: Currently this is how kubernetes scheduler works as far as i know.There is open issue(https://github.com/GoogleCloudPlatform/kubernetes/issues/168) where they want to take the resource request into consideration for scheduling than the limits.

aledbf: @kmala ok, but then this difference should be mentioned in the docs.

Right now (using fleet) the ram limits means that the memory usage in the container cannot grow only until the limit. After that the container is restarted.

controller/scheduler/k8s.py
aledbf: @kmala  doing deis pull I always receive the error `"message": "service \"app-20830\" already exists",`. Only after removing the service is possible to create a new version

controller/scheduler/k8s.py
kmala: I updated the doc with the difference.

controller/scheduler/k8s.py
kmala: This must have happened because of a issue in the deletion of the App.I had corrected the code.

docs/customizing_deis/choosing-a-scheduler.rst
```diff
@@ -120,19 +125,54 @@ The Swarm Scheduler is now active. Commands such as ``deis destroy`` or
...
+    $deisctl install k8s && deisctl start k8s
+
+Then set the controller's ``schedulerModule`` to "k8s":
+
+.. code-block:: console
+
+    $ deisctl config controller set schedulerModule=k8s
+
+The Kubernetes scheduler is now active. Commands such as ``deis destroy`` or
+``deis scale web=9`` will use `kubernetes ApiServer`_ to manage app PODS.
```
mboersma: The trailing underscore `_` tells ReStructuredText to look for a URL reference with this name, which it didn't find. I'm not sure what was intended here. Please run `make -C docs/ test` to work this out, and please run the test suite in general on this PR before submitting changes.

controller/scheduler/k8s.py
aledbf: @kmala it "feels" weird not be able to reuse the service that already exists. 
Not reusing the service means that the assigned IP is going to be different so if some user is using the env variables to get the host and port (following the [environment-guide](https://github.com/GoogleCloudPlatform/kubernetes/tree/master/docs/user-guide/environment-guide) example) will stop working because of the change.

mboersma:
There are flannel-related changes in #3868 that aren't in this PR: [one](https://github.com/deis/deis/pull/3868/files#diff-43358db3bfa05551b90bcf90c48dfe2cR382) [two](https://github.com/deis/deis/pull/3868/files#diff-ee0f424af16faeb11858134c924da6a9R117). @aledbf should those changes be here as well, or should we work on merging #3868 before this?

aledbf:
@mboersma  #3868 contains additional modifications required to allow deis-publisher keep working with the overlay network using fleet or mesos as scheduler

mboersma:
`deisctl` doesn't build in this PR because it needs to catch up with the changes in #4024 (use io.Writer).

mboersma:
This is failing because `fleet` isn't accessible outside the cluster or because `etcd2` is unhappy apparently:
```
Using ELB deis-test-DeisWebE-8MKTXTRT6Q8M at deis-test-DeisWebE-8MKTXTRT6Q8M-914251698.us-west-2.elb.amazonaws.com
[0;32mYour Deis cluster has been successfully deployed to AWS CloudFormation and is started.[0m
[0;32mPlease continue to follow the instructions in the documentation.[0m
[0;32mEnabling proxy protocol[0m
Error: 501: All the given peers are not reachable (Tried to connect to each peer twice and failed) [0]
[0;31m#[0m
[0;31m# Enabling proxy protocol failed, please enable proxy protocol [0m
[0;31m# manually after finishing your deis cluster installation.[0m
[0;31m#[0m
[0;31m# deisctl config router set proxyProtocol=1[0m
[0;31m#[0m

>>> Waiting for etcd/fleet at ec2-54-68-178-239.us-west-2.compute.amazonaws.com <<<


>>> Timeout waiting for etcd/fleet <<<

INFO client.go:291: Failed getting response from http://127.0.0.1:4001/: ssh: rejected: connect failed (Connection refused)
ERROR client.go:213: Unable to get result for {Get /_coreos.com/fleet/states}, retrying in 100ms
INFO client.go:291: Failed getting response from http://127.0.0.1:4001/: ssh: rejected: connect failed (Connection refused)
ERROR client.go:213: Unable to get result for {Get /_coreos.com/fleet/states}, retrying in 200ms
INFO client.go:291: Failed getting response from http://127.0.0.1:4001/: ssh: rejected: connect failed (Connection refused)
ERROR client.go:213: Unable to get result for {Get /_coreos.com/fleet/states}, retrying in 400ms
Error: timeout reached

>>> Error detected, dumping logs <<<
```

carmstrong:
@kmala Is this passing for you locally?

kmala:
No..Its failing locally also because the etcd2 will fail to start. I am just waiting for the etcd2 PR to get merged or i have to change my code to use etcd.

mboersma:
> I am just waiting for the etcd2 PR to get merged

Soon! :smile:

controller/scheduler/k8s.py
carmstrong: @kmala Did we make this change? The code still looks like the original ordering.

kmala: No we din't make this change as the creation of service requires port of the containers which is available only if we create RC/POD and inspect the image. Instead we are adding a custom policy to the kubernetes scheduler service for the spreading to work properly.

mboersma:
It looks like the current problem is that user-data has grown to just over 16K with this PR, which prevents provisioning from succeeding on AWS. I am trying to remove extraneous bytes from user-data so this can fit, but this PR should also discard anything that's not absolutely required.

kmala:
I had made changes to my PR to make sure that the size is less than 16k.So we don't require any changes to the existing user data file at least for now.

deisctl/cmd/cmd.go
```diff
@@ -85,6 +86,9 @@ func Start(targets []string, b backend.Backend) error {
 		case swarm:
 			return StartSwarm(b)
 		}
+		if targets[0] == k8s {
```
mboersma: This and the next three similar clauses should be rewritten as:
```go
case k8s:
```

docs/customizing_deis/choosing-a-scheduler.rst
```diff
@@ -122,16 +122,49 @@ The Swarm Scheduler is now active. Commands such as ``deis destroy`` or
...
+    The Kubernetes Scheduler is a technology preview and is not recommended for
+    production use.
+
+`Kubernetes`_ orchestration system for docker containers:
+
+    Kubernetes provides APIs to manage, deploy and scale docker containers. Kubernetes deploys docker containers as `PODS`_,
+    unique entity across a cluster. Containers inside a POD share the same namespaces. More information about `PODS`_.
+
+To run kubernetes on CoreOS requires `flannel`_ which provides subnet to each host for use with containers runtimes.
+Deis repository provides a new user-data file to bootstrap CoreOs machines on the cluster with flannel.
```
mboersma: I would reword this a bit:
```
Kubernetes on CoreOS requires `flannel`_, which provides a subnet for each host to use with containers.
Deis' user-data file activates flannel when provisioning a new cluster.
```

docs/customizing_deis/choosing-a-scheduler.rst
```diff
@@ -122,16 +122,49 @@ The Swarm Scheduler is now active. Commands such as ``deis destroy`` or
...
-    2015-04-30T17:31 02a570: (from 172.17.8.100:5000/hungry-variable:v5 node:deis-01) start
-    2015-04-30T17:31 61e59c: (from 172.17.8.100:5000/hungry-variable:v5 node:deis-02) create
-    2015-04-30T17:31 61e59c: (from 172.17.8.100:5000/hungry-variable:v5 node:deis-02) start
+    $ deisctl config controller set schedulerModule=k8s
+
+The Kubernetes scheduler is now active. Commands such as ``deis destroy`` or
+``deis scale web=9`` will use kubernetes ApiServer to manage app PODS.
+
+For each App deis creates a replication controller which manages application PODS and a kubernetes service which acts as a proxy and routes traffic to the PODS associated with the App. A new release uses rolling deploy mechanism. If you are using
+kubernetes scheduler for Deis you can avoid publisher component.
```
mboersma: I would reword this:
```
For each app, Deis creates a replication controller to manage application PODs and a Kubernetes service
which proxies traffic to those PODs. A new app release will use a "rolling deploy" mechanism.
```
Maybe explain specifically how "rolling deploy" differs from what we do with other schedulers? Also I wouldn't bother mentioning that deis-publisher isn't used in this situation. That's an implementation detail, and we don't want to encourage users to rip publisher out IMHO.

docs/customizing_deis/choosing-a-scheduler.rst
```diff
@@ -122,16 +122,49 @@ The Swarm Scheduler is now active. Commands such as ``deis destroy`` or
...
-    2015-04-30T17:31 172.17.8.100:5000/hungry-variable:v5: (from  node:deis-01) pull
-    2015-04-30T17:31 172.17.8.100:5000/hungry-variable:v5: (from  node:deis-02) pull
-    2015-04-30T17:31 02a570: (from 172.17.8.100:5000/hungry-variable:v5 node:deis-01) create
-    2015-04-30T17:31 02a570: (from 172.17.8.100:5000/hungry-variable:v5 node:deis-01) start
-    2015-04-30T17:31 61e59c: (from 172.17.8.100:5000/hungry-variable:v5 node:deis-02) create
-    2015-04-30T17:31 61e59c: (from 172.17.8.100:5000/hungry-variable:v5 node:deis-02) start
+    $ deisctl config controller set schedulerModule=k8s
+
+The Kubernetes scheduler is now active. Commands such as ``deis destroy`` or
+``deis scale web=9`` will use kubernetes ApiServer to manage app PODS.
```
mboersma: Should probably be:
> ...will use Kubernetes' ApiServer to manage app PODs.

bacongobbler: Shouldn't POD be a proper noun? It doesn't stand for anything, AFAIK. A pod is a pod.

mboersma: It's not even a proper noun, just a noun, so I guess "pod" is what we should use everywhere.

mboersma:
My previous comments applied to a specific commit, but the changes I referred to were overwritten by a later commit. That kind of change shouldn't end up in a final PR--that we tried something in one commit and reverted it in a later one is "inside baseball" stuff.

carmstrong:
We'll also need to add the kubernetes units to the `deisctl refresh-units` command: https://github.com/deis/deis/blob/bb9e90a310c4ad5ad9c56a2a6ba15d1ded90accf/deisctl/units/units.go#L4-L26

rimusz:
awesome :+1:

mboersma:
Code LGTM.

router/image/conf.d/nginx.conf.toml
```diff
@@ -12,6 +12,7 @@ keys = [
   "/deis/builder",
   "/deis/store/gateway",
   "/deis/certs",
+  "/registry/services/",
```
carmstrong: Why are we using this path here? Are these keys all related to the `deis-registry` component? If not, maybe we should consider changing this.

carmstrong: Ah, I just noticed this was `/registry` and not `/deis/registry`. Disregard - this is reasonable.

carmstrong:
Should we note in the documentation that the Kubernetes scheduler is only possible for newly-provisioned clusters? The changes in cloud-config will not exist for upgraded clusters, and I imagine k8s probably won't function without those changes.

/cc @kmala @smothiki @technosophos

docs/customizing_deis/choosing-a-scheduler.rst
```diff
@@ -122,16 +122,50 @@ The Swarm Scheduler is now active. Commands such as ``deis destroy`` or
...
+
+.. note::
+
+    **Known Issues**
+
+    - Kubernetes ApiServer is not HA. If ApiServer is rescheduled it will reschedule all the kubernetes units.
+    - Kubernetes does resource based scheduling. Specifying limits will create reservation of the specified resource on the node.
+    - Installing flannel is not backwards compatible as it changes Docker bridge which requires restarting
+      Docker Engine.
+    - Docker Engine is started with arguments ``--iptables=false --ip-masq=false`` for the docker not to create IP masquerading rule and use the one created by flannel.
```
smothiki: this should be changed . We are not doing this anymore

docs/customizing_deis/choosing-a-scheduler.rst
```diff
@@ -122,16 +122,50 @@ The Swarm Scheduler is now active. Commands such as ``deis destroy`` or
...
+A new release uses rolling deploy mechanism where in we create and replace the pod with the new release one by one until all the pods are replaced properly or rollback to the older release
+if any error occurs which differs from the existing schedulers where we create all the containers at a time and replace them with the old ones.
+
+.. note::
+
+    **Known Issues**
+
+    - Kubernetes ApiServer is not HA. If ApiServer is rescheduled it will reschedule all the kubernetes units.
+    - Kubernetes does resource based scheduling. Specifying limits will create reservation of the specified resource on the node.
+    - Installing flannel is not backwards compatible as it changes Docker bridge which requires restarting
```
smothiki: @carmstrong  we are mentioning installing flannel is not back wards compatible. This implies that coreOS with flannel requires new cluster. Let me know if you want this to be rephrased or add a note saying kubernetes installation requires new coreOS cluster

docs/customizing_deis/choosing-a-scheduler.rst
```diff
@@ -122,16 +122,50 @@ The Swarm Scheduler is now active. Commands such as ``deis destroy`` or
 To monitor Swarm Scheduler operations, watch the logs of the swarm-manager
 component, or spy on Docker events directly on the swarm-manager machine:
 
+Kubernetes Scheduler
+--------------------
+
+.. important::
+
+    The Kubernetes Scheduler is a technology preview and is not recommended for
+    production use.
```
technosophos: Here would be a good place to say "The Kubernetes preview can only be used on new clusters. There is no migration path from an existing cluster to the Kubernetes technology preview."

mboersma:
Updated the "choosing a scheduler" and roadmap docs after comments. Please re-review.

bacongobbler:
I'm concerned about the idea of writing out a script to download the kubernetes binaries in user data. If someone does an in-place upgrade, they'll have to bump the kubernetes binaries on each host. I guess since this is a tech preview it's fine but I can't help but feel that we can install them from the units themselves. Same goes for the scheduler policy file.

I guess my concerns are always about config mgmt/upgrade path but this will do for a tech preview.

controller/api/models.py
```diff
@@ -267,11 +292,15 @@ def scale(self, user, structure):  # noqa
...
                 container_num += 1
                 diff -= 1
+
         if changed:
-            if to_add:
-                self._start_containers(to_add)
-            if to_remove:
-                self._destroy_containers(to_remove)
+            if "scale" in dir(self._scheduler):
+                self._scale_containers(scale_types, to_remove)
```
bacongobbler: why do we need a separate `_scale_containers` call here just for kubernetes?

controller/api/models.py
```diff
@@ -179,6 +179,28 @@ def __str__(self):
...
         return self.id + '.' + settings.DEIS_DOMAIN
 
+    def _get_job_id(self, container_type):
+        app = self.id
+        release = self.release_set.latest()
+        version = "v{}".format(release.version)
+        job_id = "{app}_{version}.{container_type}".format(**locals())
+        return job_id
+
+    def _get_command(self, container_type):
```
bacongobbler: this code is identical to the Container class' implementation. If we're already knowledgeable about the container's process type, why aren't we calling the container's function instead?

controller/api/models.py
kmala: Because we need to just update the RC with with the number of replicas and it will create or delete the extra pods for us which is different from the existing implementations.

bacongobbler:
It works, which is great, but there are a few maintenance concerns here:

 - no unit tests in the controller, specifically around the new `scale` and `deploy` functions for the kubernetes scheduler. We should be mocking these calls, at the very least
 - a lot of concepts like the Container class are completely irrelevant with kubernetes, so I feel like we should either refactor some of the code back into the App class, specifically the calls `_get_job_id` and `_get_command`.

I still feel this should be a fork rather than merged into master, considering that there are a **lot** of changes that take the current Fleet-centric world turned upside-down and convolutes the controller and router's codebases significantly, but that's just my opinion.

controller/api/models.py
kmala: I am not having the container object at the place where i am using it.I can use the container's function by passing the container's list from the scale and deploy to functions where i am using it and finding the unique types and count from the lists.

mboersma:
@bacongobbler I understand your concerns: it is a lot of controller code that feels like it could still stand some refactoring for simplicity...

@carmstrong @krancour @technosophos could you review this please so we can settle on a plan of action WRT v1.9? I've contributed to it now so I can't really give an impartial review any longer.

carmstrong:
Code LGTM. While I agree that there could be some minor cleanup, I'd argue that the changes are pretty minor. We could do some refactoring in a follow-up PR if necessary.

bacongobbler:
this doesn't seem to break existing clusters, so code LGTM.

smothiki:
qvjuafhj

rimusz:
@smothiki ha ha nice one

deisctl/units/deis-kube-apiserver.service
```diff
@@ -0,0 +1,30 @@
+[Unit]
+Description=Kubernetes API Server
+Documentation=https://github.com/GoogleCloudPlatform/kubernetes
+Requires=etcd2.service fleet.service docker.service flanneld.service
```
krancour: We are actually running etcd2 _as_ "etcd.service" still.  This should probably change.

carmstrong: :+1: Good catch. We're masking `etcd2.service` as well.

kmala: I think we can remove the etcd2.service as flannel requires etcd and apiserver requires flannel

krancour: > I think we can remove the etcd2.service as flannel requires etcd and apiserver requires flannel

That should be fine.

kmala: #4151

tilgovi:
Awesome!!

Issue #3774, opened 2015-05-29
app routes return 404 after all instances are down


bacongobbler:
Instead of a 503, `deis open` is returning a 404 for an existing app with all web apps scaled down to zero.

```
><> cd example-ruby-sinatra/
><> deis ps
=== ruby Processes

><> curl http://ruby.local3.deisapp.com
<html>
<head><title>404 Not Found</title></head>
<body bgcolor="white">
<center><h1>404 Not Found</h1></center>
<hr><center>nginx/1.9.0</center>
</body>
</html>
```

I think this occurs because we're using `lsdir` instead of `ls` in [the router template](https://github.com/deis/deis/blob/master/router/image/templates/nginx.conf#L162), which only iterates through all subkeys which also have subkeys. When all processes are scaled down, all subkeys are expired, therefore the router template removes the entry.

bacongobbler:
I have [a branch](https://github.com/bacongobbler/deis/tree/3774-use-ls) but it looks like there's a bug with `ls` in confd. It doesn't detect directories, only keys.

bacongobbler:
confd has been removed in the v2 router and we won't be able to get to this for the LTS release (#4776) so closing this.

PR #3982, opened 2015-07-07
fix(controller): sleep until publisher's refresh interval

bacongobbler:
Deis' publisher component has a maximum latency of 10 seconds before
it refreshes its config from etcd. For a true zero-downtime
deployment, we want to make sure that the new containers are
templated in the router before performing the health check and/or
recycling the old containers.

The FIXME here would be to utilize confd --watch in the router and in
publisher, but that is blocked due to an [issue upstream with etcd](1).

[1]: coreos/etcd#2679

### DIFF
--- a/controller/api/models.py
+++ b/controller/api/models.py
@@ -1058,14 +1058,14 @@ def _etcd_publish_config(**kwargs):
     # deis config:unset would remove an existing value, but not delete the
     # old config object
     try:
-        _etcd_client.delete('/deis/services/{}/config'.format(config.app),
+        _etcd_client.delete('/deis/config/{}'.format(config.app),
                             prevExist=True, dir=True, recursive=True)
     except KeyError:
         pass
     if kwargs['created']:
         for k, v in config.values.iteritems():
             _etcd_client.write(
-                '/deis/services/{}/config/{}'.format(
+                '/deis/config/{}/{}'.format(
                     config.app,
                     unicode(k).encode('utf-8').lower()),
                 unicode(v).encode('utf-8'))
@@ -1074,7 +1074,7 @@ def _etcd_publish_config(**kwargs):
 def _etcd_purge_config(**kwargs):
     config = kwargs['instance']
     try:
-        _etcd_client.delete('/deis/services/{}/config'.format(config.app),
+        _etcd_client.delete('/deis/config/{}'.format(config.app),
                             prevExist=True, dir=True, recursive=True)
     except KeyError:
         pass
--- a/controller/api/models.py
+++ b/controller/api/models.py
@@ -314,6 +314,15 @@ class App(UuidAuditedModel):
             pass
 
     def _do_healthcheck(self, config):
+        # HACK (bacongobbler): sleep until publisher has hit the refresh interval. Publisher may
+        # take up to 20 seconds before it refreshes the keyspace containers, which we want to make
+        # sure that they're responding before performing the health check and/or recycle the old
+        # containers.
+        #
+        # The FIXME here would be to utilize confd --watch in both publisher and the router, but
+        # that is blocked dueto an issue upstream with etcd:
+        # https://github.com/coreos/etcd/issues/2679
+        time.sleep(20)
         timeout = time.time() + 60  # 1 minute from now
         while True:
             if time.time() > timeout:
--- a/controller/api/tests/test_build.py
+++ b/controller/api/tests/test_build.py
@@ -24,6 +24,7 @@ def mock_import_repository_task(*args, **kwargs):
     return resp
 
 
+@mock.patch('time.sleep', lambda func: func)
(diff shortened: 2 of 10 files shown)


bacongobbler:
tested manually with example-scala. Both with and without HEALTHCHECK_URL set no 502 errors occur, which confirms my suspicion.

carmstrong:
This seems to make sense and seems like a reasonable solution, although I can't shake off the feeling that it's a bit hacky. Code LGTM. Would like @gabrtv's review as well.

mboersma:
Code looks good, but I'm testing manually on EC2 now.

mboersma:
This does appear to fix the transient 502 errors during `deis config:set` that I saw yesterday. I tested with Deis master on EC2 and saw the 502s again, then rebuilt and deployed controller from this PR, and no 502s.

But `deis scale web=0` now leads to a permanent `HTTP/1.1 502 Bad Gateway` response when completed--I thought that was supposed to be a 404 in that case. Also scaling back up from zero containers made the healthcheck fail and refuse to deploy (it worked the second time):
```console
$ deis scale web=0
Scaling processes... but first, Slurm!
done in 5s
=== ironic-lakeside Processes

$ deis scale web=1
Scaling processes... but first, Slurm!
503 SERVICE UNAVAILABLE
Detail:
aborting, app failed health check (got '502', expected: '200')
```
That corresponds to this `curl` output every 1/4 second:
```console
HTTP/1.1 502 Bad Gateway
HTTP/1.1 502 Bad Gateway
HTTP/1.1 502 Bad Gateway
HTTP/1.1 200 OK
HTTP/1.1 502 Bad Gateway
HTTP/1.1 200 OK
HTTP/1.1 502 Bad Gateway
HTTP/1.1 200 OK
HTTP/1.1 200 OK
HTTP/1.1 200 OK
HTTP/1.1 200 OK
HTTP/1.1 200 OK
HTTP/1.1 200 OK
HTTP/1.1 200 OK
HTTP/1.1 502 Bad Gateway
HTTP/1.1 502 Bad Gateway
HTTP/1.1 502 Bad Gateway
HTTP/1.1 502 Bad Gateway
HTTP/1.1 502 Bad Gateway
HTTP/1.1 502 Bad Gateway
```

bacongobbler:
> But deis scale web=0 now leads to a permanent HTTP/1.1 502 Bad Gateway response when completed--I thought that was supposed to be a 404 in that case.

That's partially to be expected. It should be a 503 (all instances are down). I was seeing something similar in #3774.

What does seem odd is the healthcheck failing after scaling back up again. We're waiting for the containers to come back up, then sleeping for 10 seconds for the router to update. After that we perform the healthcheck. Do you see anything in the router or in publisher that would explain this behaviour

mboersma:
Agreed it should be a 503 when there are no instances, but apparently this PR changes that behavior from 404 -> 502, which I'm not sure is desirable.

I've tried several times to reproduce the healthcheck failure on scaling up from 0 and it hasn't happened again. Instead we see `curl` output just like above, but it resolves to a steady 200 response. I've looked through publisher and router logs and don't see anything obviously wrong.

bacongobbler:
Ah, that's because deis-publisher has a [10 second refresh interval](https://github.com/deis/deis/blob/59ff8d2e979e4eae7feee06c63da838e4587f259/publisher/main.go#L17), which essentially means we're hitting the router *right* when the etcd keys are theoretically just getting published. Therefore, we need to wait for 10 seconds (next publish interval), then for another 10 seconds for the router's theoretical interval. It's ugly, but this is something we can eventually fix upstream to remove this.

carmstrong:
@bacongobbler This looks like a real failure:
```
=== healthchecksample
DEIS_APP: healthchecksample
HEALTHCHECK_URL: /
HEALTHCHECK_STATUS_CODE: 404
DEIS_RELEASE: v5
--- FAIL: TestHealthcheck (164.58s)
	itutils.go:283: <nil>
FAIL
exit status 1
FAIL	github.com/deis/deis/tests	164.584s
```

bacongobbler:
yes, testing manually it seems like it's not failing as it should be.

bacongobbler:
> This looks like a real failure:

This looks to be a chicken-and-egg problem with the healthcheck code. When we push an app with HEALTHCHECK_STATUS_CODE=999, publisher will not publish the app because a GET / will return a 200 response, not a 999 response. The router then has no containers published, so it won’t template the app, causing it to return 404s. When the *controller* does the health check, it will get a 404 response instead of a 200 response, which means that the user will see an error identical to `aborting, app failed health check (got '404', expected: '999’)`. For now I've updated the integration tests to follow the typical workflow, but I'll file an issue once this is merged to fix this issue.

carmstrong:
@bacongobbler We can close this until the healthcheck PR is reopened, right?

bacongobbler:
Yes :)

PR #3984, opened 2015-07-07
chore(*): bump CoreOS to 717.1.0

mboersma:
See https://coreos.com/releases/#717.1.0


bacongobbler:
I guess we can close this considering that we're waiting on a bugfix from docker. :)

mboersma:
There is also a CoreOS 717.3.0 now, but it has Docker 1.6.2 as well. Since we need a workaround for a Docker issue and intend to merge #3992, I'll close this.

PR #3992, opened 2015-07-09
chore(*): revert CoreOS to 647.2.0

carmstrong:
Newer CoreOS releases include Docker 1.6, which is fundaementally
broken for simultaneous pulls of the same filesystem layers -
see https://github.com/docker/docker/issues/9997

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -52,16 +52,28 @@ def vm_cpus
 end
 
 Vagrant.configure("2") do |config|
-  # always use Vagrants insecure key
+  # always use Vagrant's insecure key
   config.ssh.insert_key = false
-
   config.vm.box = "coreos-%s" % $update_channel
-  config.vm.box_version = "= 647.2.0"
-  config.vm.box_url = "http://%s.release.core-os.net/amd64-usr/current/coreos_production_vagrant.json" % $update_channel
+
+  # HACK to bypass the limitation of the CoreOS "current" metadata for boxes -
+  # it only specifies the latest release, so we cannot use it for older releases.
+  # TODO(carmstrong) we can remove this once we're back to using "specific release or more recent"
+  if $update_channel == "stable"
+    config.vm.box_version = "= 647.2.0"
+    config.vm.box_url = "http://stable.release.core-os.net/amd64-usr/647.2.0/coreos_production_vagrant.json"
+  else
+    config.vm.box_version = ">= 681.2.0"
+    config.vm.box_url = "http://%s.release.core-os.net/amd64-usr/current/coreos_production_vagrant.json" % $update_channel
+  end
 
   ["vmware_fusion", "vmware_workstation"].each do |vmware|
     config.vm.provider vmware do |v, override|
-      override.vm.box_url = "http://%s.release.core-os.net/amd64-usr/current/coreos_production_vagrant_vmware_fusion.json" % $update_channel
+      if $update_channel == "stable"
+        override.vm.box_url = "http://stable.release.core-os.net/amd64-usr/647.2.0/coreos_production_vagrant_vmware_fusion.json" % $update_channel
+      else
+        override.vm.box_url = "http://%s.release.core-os.net/amd64-usr/current/coreos_production_vagrant_vmware_fusion.json" % $update_channel
+      end
     end
   end
 
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -56,7 +56,7 @@ Vagrant.configure("2") do |config|
   config.ssh.insert_key = false
 
   config.vm.box = "coreos-%s" % $update_channel
-  config.vm.box_version = ">= 681.2.0"
+  config.vm.box_version = "= 647.2.0"
   config.vm.box_url = "http://%s.release.core-os.net/amd64-usr/current/coreos_production_vagrant.json" % $update_channel
 
   ["vmware_fusion", "vmware_workstation"].each do |vmware|
--- a/contrib/azure/azure-coreos-cluster
+++ b/contrib/azure/azure-coreos-cluster
@@ -35,8 +35,8 @@ parser.add_argument('--affinity-group', default='',
                    help='optional, overrides location if specified')
 parser.add_argument('--ssh', default=22001, type=int,
                    help='optional, starts with 22001 and +1 for each machine in cluster')
-parser.add_argument('--coreos-image', default='2b171e93f07c4903bcad35bda10acf22__CoreOS-Stable-681.2.0',
-                   help='optional, [2b171e93f07c4903bcad35bda10acf22__CoreOS-Stable-681.2.0]')
+parser.add_argument('--coreos-image', default='2b171e93f07c4903bcad35bda10acf22__CoreOS-Stable-647.2.0',
+                   help='optional, [2b171e93f07c4903bcad35bda10acf22__CoreOS-Stable-647.2.0]')
 parser.add_argument('--num-nodes', default=3, type=int,
                    help='optional, number of nodes to create (or add), defaults to 3')
 parser.add_argument('--virtual-network-name',
--- a/contrib/ec2/deis.template.json
+++ b/contrib/ec2/deis.template.json
@@ -44,7 +44,7 @@
         "m4.xlarge",
         "m4.2xlarge",
         "m4.4xlarge",
-        "m4.10xlarge",        
+        "m4.10xlarge",
         "m1.medium",
         "m1.large",
         "m1.xlarge",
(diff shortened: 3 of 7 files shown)


carmstrong:
We may need to change the format of the Vagrantfile to specify the release .box and not the .json, as it only publishes the latest version - see https://github.com/coreos/coreos-vagrant/issues/83#issuecomment-119814476

jwaldrip:
I hope this doesn't cause any problems for us. Our servers just auto-updated to 717.1.0

carmstrong:
@jwaldrip We're still debugging the issues with Docker and DOcker Hub. I put this PR together so we'd have it as a last resort.

carmstrong:
We'll be sure to update as we know more!

jwaldrip:
Okay great. Thanks!

bacongobbler:
Checked the AMIs against what was in https://github.com/deis/deis/blob/v1.7.3/contrib/ec2/deis.template.json and they all match up.

Do you think it's possible to revert https://github.com/deis/deis/blob/master/contrib/linode/provision-linode-cluster.py#L416 as well? ping @croemmich

bacongobbler:
not a blocker since it's a community-supported provider, so this has my LGTM :)

jwaldrip:
So... does this mean y'all are going to revert?

bacongobbler:
Yes. There are a few issues affecting us from migrating to a newer version of docker, specifically we're seeing the same issues as shown in docker/docker#13971 on some clusters. This may not be a problem in docker v1.7.1, but that hasn't been released yet.

From internal communication:

> If you read all the comments, they do not have a handle on an actual fix yet, despite having closed the issue. `flock` doesn't look like a reasonable workaround to me--we would have to serialize all the `docker pull`s on each host, which I assume would allow to measure provisioning in geological time.
> Possible solutions:
> 1. Serialize all `docker pull` ops with `flock` in the unit files (see http://linux.die.net/man/1/flock and https://github.com/docker/docker/issues/13971#issuecomment-112568990)
> 2. Revert to CoreOS 647.x.x which has Docker 1.5.0 (same as we shipped for v1.7.3 of Deis)
> 3. Convince CoreOS to revert to Docker 1.5.0 in stable
> 4. Hack our user-data to use Docker 1.5.0 on current CoreOS versions
> 5. Wait until Docker 1.7.1 arrives in CoreOS, assuming it fixes this which is not yet clear
> And a corrolary: if we upgrade to Registry V2 inside Deis, this bug will likely happen internally to Deis just on `deis scale web=10`.

Based on those proposed solutions, reverting to a known working version of Docker looks to be the easiest solution and wait on a fix from Docker. Since v1.7.3 locks to 647.2.0, this isn't a backwards-incompatible change.

mboersma:
Newer versions of CoreOS include Docker 1.6.2, which has a big bug we run into often on provisioning: docker/docker#13971. You'll see one or more Deis components in the `auto-restart` state forever, and logs will show a message like this:
```
FATA[0084] Error pulling image (app-15) from aaron42net/dockertest, Driver overlay failed to create image rootfs e8d7ddc9a499bd28ecdd5c12bf0b38ed3f247b46b10673fc5061f659f73740cc: lstat /var/lib/docker/overe8d7ddc9a499: Error downloading dependent layers
```
Our understanding after talking with Docker engineers is that this happens when two `docker pull` tasks fetch the same layer and collide, they can write a corrupted layer that will persist and prevent all images using that layer from starting. (The other way this can manifest is with a `-bash: curl: command not found` or similar message when starting the image.)

We hope this is fixed in Docker 1.7.1, but for now using Docker 1.5.0 looks like the only safe bet.

LGTM.

bacongobbler:
code LGTM.

Vagrantfile
```diff
@@ -52,16 +52,20 @@ def vm_cpus
 end
 
 Vagrant.configure("2") do |config|
-  # always use Vagrants insecure key
+  # always use Vagrant's insecure key
```
mboersma: With this typo fix, we are no longer bug-compatible with [upstream](https://github.com/coreos/coreos-vagrant/blob/master/Vagrantfile#L48). :smiling_imp:

carmstrong: All our upstream are belong to us

carmstrong: This commit also breaks upstream compatibility in much bigger ways... :innocent:

Vagrantfile
```diff
@@ -52,16 +52,20 @@ def vm_cpus
...
   config.ssh.insert_key = false
-
   config.vm.box = "coreos-%s" % $update_channel
-  config.vm.box_version = ">= 681.2.0"
-  config.vm.box_url = "http://%s.release.core-os.net/amd64-usr/current/coreos_production_vagrant.json" % $update_channel
+  config.vm.box_version = "= 647.2.0"
```
mboersma: This does work for coreos-stable, even if you delete that version.
```console
$ vagrant box remove coreos-stable --box-version 647.2.0 && vagrant up
```
But it won't work for coreos-alpha:
```console
$ COREOS_CHANNEL=alpha vagrant up
Bringing machine 'deis-01' up with 'virtualbox' provider...
Bringing machine 'deis-02' up with 'virtualbox' provider...
Bringing machine 'deis-03' up with 'virtualbox' provider...
==> deis-01: Running triggers before up...
==> deis-01: Box 'coreos-alpha' could not be found. Attempting to find and install...
    deis-01: Box Provider: virtualbox
    deis-01: Box Version: = 647.2.0
You specified a box version constraint with a direct box file
path. Box version constraints only work with boxes from Vagrant
Cloud or a custom box host. Please remove the version constraint
and try again.
```

carmstrong: > This does work for coreos-stable, even if you delete that version.

Funny, it didn't work for me, and *shouldn't* work, as far as I can tell - see https://github.com/coreos/coreos-vagrant/issues/83#issuecomment-119814476

Without these changes, the client tries to fetch the metadata for 647.2.0 from http://stable.release.core-os.net/amd64-usr/current/coreos_production_vagrant.json, but the only version listed there is the latest in the channel. So, it errs out.

As for the issue with different channels, I'm not too concerned about supporting the update channel in the Vagrantfile workflow, as that is a seldom-used special case.

bacongobbler: This box version wouldn't work for the alpha channel anyways:

```
><> curl  http://alpha.release.core-os.net/amd64-usr/647.2.0/coreos_production_vagrant.json
<?xml version='1.0' encoding='UTF-8'?><Error><Code>NoSuchKey</Code><Message>The specified key does not exist.</Message></Error>
```

http://alpha.release.core-os.net/amd64-usr/index.html confirms this as well. Does this block test-coreos-alpha?

mboersma: > it didn't work for me, and shouldn't work

I was surprised it did. Not sure why.

The real issue with breaking the update channel is it kills our nightly "canary" jobs in CI. It is useful to see when breaking changes are introduced into coreos-beta and coreos-alpha so we can remedy things before they get to stable. Without the vagrant cloud update channels working, I don't know how we can build and test against the latest versions.

carmstrong: > It is useful to see when breaking changes are introduced into coreos-beta and coreos-alpha so we can remedy things before they get to stable.

Agreed, but this is intended to be a short-lived bandaid. I'm open to suggestions, but it seems like we need to restrict clusters to 647.2.0 for them to work properly. Or, do we move forward with Vagrant using whatever version it pleases, and living with the caveat that local development workflows may be broken due to Docker 1.6.2?

bacongobbler: Hmm. Since the beta/alpha jobs just want to use the latest version on their respective channels, maybe we can make this a switch flag if the channel == stable. Would that be a reasonable workaround?

mboersma: I think either solution would be ok: allow Vagrant to use current versions of CoreOS (knowing that Docker 1.6.2 may be buggy), or make the stable channel code conditional in the Vagrantfile if possible. Or we can just disable the test-coreos-beta and test-coreos-alpha jobs in CI temporarily, until this overall situation is improved--they are valuable, but not crucial.

bacongobbler:
I'm happy with the workaround for vagrant. Once again, this has my LGTM :)

mboersma:
Code LGTM.

jwaldrip:
I have seen this on my clusters way back in the day. But it was related to docker being activated by socket before the lib/docker mount was ready.

mboersma:
Sorry, hopefully this revert isn't painful for you @jwaldrip. For Deis this is purely a provisioning-time issue as far as we can tell--it won't happen when fetching images from Deis' internal V1 registry during a `deis scale` or similar operation.

So if you've updated past 647.2.0, you *might* run into this issue when upgrading Deis, but if you're able to get the next release installed I don't think you will see operational problems. YMMV of course.

PR #4004, opened 2015-07-09
docs(managing_deis): rename stateless platform docs

carmstrong:
@gabrtv pointed out that calling it a stateless platform isn't quite yet correct. Even though the `deisctl` command is `deisctl install stateless-platform`, clarifying the documentation should be sufficient.


bacongobbler:
lgtm

mboersma:
LGTM.

PR #3748, opened 2015-05-26
[WIP] chore(*): support etcd2

carmstrong:
**Not to be merged as-is!**

This commit updates provision scripts, documentation, and cloud-config
settings to support etcd 2, a significant rewrite which has shown
to be more stable and resilient than etcd 0.x.

This commit will be amended once a CoreOS release with etcd 2 hits
the stable channel, and sufficient testing has been done to have
confidence in etcd 2 for supporting Deis.

PR #4005, opened 2015-07-09
[WIP] chore(*): use etcd2

### LINKED ISSUE #3564: Adapt for etcd2
etcd2 is in the alpha channel now. We'll need to:

- [ ] change some settings in user-data (see https://gist.github.com/carmstrong/c5c4d60c59fbf1558d77)
- [ ] ensure confd supports etcd2
- [ ] support upgrades to etcd 2 from etcd 0.4.x

carmstrong:
This commit configures hosts to use etcd2. Note that it doesn't
yet have upgrade logic between the versions. We also don't run
etcd2 in a container (even though it's recommended by
@kelseyhightower) because of the following reasons:

* An additional network dependency on startup until host is operational
  (pulling the container from Quay)
* Additional complexity customizing the unit to bind-mount different
  etcd data locations depending on the provider
* The version of etcdctl on the host won't necessarily match the
  version of etcd (relatively minor)

replaces #3748
closes #3564

### DIFF
--- a/contrib/azure/azure-coreos-cluster
+++ b/contrib/azure/azure-coreos-cluster
@@ -68,7 +68,7 @@ coreos:
     addr: $private_ipv4:4001
     peer-addr: $private_ipv4:7001
   units:
-    - name: etcd.service
+    - name: etcd2.service
       command: start
     - name: fleet.service
       command: start
--- a/contrib/azure/create-azure-user-data
+++ b/contrib/azure/create-azure-user-data
@@ -62,7 +62,7 @@ def main():
     configuration = combine_dicts(configuration_coreos_template,
                                   configuration_azure_template)
 
-    configuration["coreos"]["etcd"]["discovery"] = url
+    configuration["coreos"]["etcd2"]["discovery"] = url
 
     with azure_user_data as outfile:
         try:
--- a/contrib/coreos/user-data.example
+++ b/contrib/coreos/user-data.example
@@ -1,16 +1,17 @@
 #cloud-config
 ---
 coreos:
-  etcd:
+  etcd2:
     # generate a new token for each unique cluster from https://discovery.etcd.io/new
     # uncomment the following line and replace it with your discovery URL
     # discovery: https://discovery.etcd.io/12345693838asdfasfadf13939923
-    addr: $private_ipv4:4001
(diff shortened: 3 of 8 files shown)


aledbf:
@carmstrong why not start using etcd2 in a container as part of the migration?
(this allows the use of etcd2 in CoreOS 647.0.0)

bacongobbler:
Some components like Flannel uses etcd as a backing service. Since Flannel needs to start before Docker, and Flannel requires etcd, it's another chicken-and-egg problem.

aledbf:
@bacongobbler  not really
```
units:
  - name: etcd2.service
    command: start
    content: |
      [Unit]
      Description=etcd2 container
      After=early-docker.service
      Before=early-docker.target

      [Service]
      ExecStartPre=-/usr/bin/docker kill etcd2
      ExecStartPre=-/usr/bin/docker rm etcd2
      ExecStartPre=/usr/bin/docker pull quay.io/coreos/etcd:v2.0.13
      ExecStart=/usr/bin/docker run -p 4001:4001 -p 2380:2380 -p 2379:2379 --name etcd2 quay.io/coreos/etcd:v2.0.13 \
        -name %m \
        -advertise-client-urls http://${private_ipv4}:2379,http://${private_ipv4}:4001 \
        -listen-client-urls http://localhost:2379,http://localhost:4001,http://0.0.0.0:2379,http://0.0.0.0:4001 \
        -initial-advertise-peer-urls http://${private_ipv4}:2380 \
        -listen-peer-urls http://0.0.0.0:2380 \
        -initial-cluster-token etcd-cluster \
        -initial-cluster ${ETCD_PEERS} \
        -initial-cluster-state new \
        --discovery #https://discovery.etcd.io/12345693838asdfasfadf13939923
      ExecStop=/usr/bin/docker stop etcd2
```

carmstrong:
@aledbf Don't we need to persist the etcd data on the host through the use of volumes? It seems like we'd wipe away all the data on service restart.

aledbf:
@carmstrong ops, I forgot the volume part, but the service is the same

carmstrong:
I prefer the approach in #4024

mboersma:
I concur. Thanks as always for showing us the light @aledbf.

PR #3965, opened 2015-07-05
fix(router): add set-misc module to nginx build

fabiob:
This is the missing piece, so that the PR #3768 can work.

Fixes #3964.

PR #3954, opened 2015-07-02
Add verification of downloaded nginx archive.

bladealslayer:
The archive is downloaded over insecure connection. Added a SHA256 checksum verification. The checkum can be generated by person updating the scipt, after manually verifying their own downloaded nginx archive via provided PGP signature (http://nginx.org/download/$VERSION_NGINX.tar.gz.asc).

PR #4008, opened 2015-07-10
fix(router): add set-misc module to nginx build

### LINKED ISSUE #3964: AffinityArg hangs router on 1.7.3
I'm trying to apply the affinityArg setting on the router…

```
deisctl config router set affinityArg=sid
```

…but I keep getting this error on the logs:

```
[INFO] - 2015/07/04 09:01:12 [emerg] 136#0: unknown directive "set_random" in /opt/nginx/conf/.nginx.conf102133528:140
[INFO] - waiting for confd to write initial templates...
```

If I remove the setting, everything goes back as normal.

Have I done something wrong?

mboersma:
The commit c7c895c fixes the affinityArg hashing issue by using the `set_random` directive on the nginx configuration, but forgets to include the module in which it is declared. This commit fixes this, while also tries to make the build script a little cleaner.

Closes #3964.

Replaces #3965--thanks @fabiob!


carmstrong:
Code LGTM.

router/build.sh
```diff
@@ -38,14 +40,11 @@ apk add --update-cache \
...
-curl -sSL https://github.com/nbs-system/naxsi/archive/$VERSION_NAXSI.tar.gz -o $BUILD_PATH/$VERSION_NAXSI.tar.gz
-
-# expand the source files
-tar xzf $VERSION_NGINX.tar.gz
-tar xzf $VERSION_NAXSI.tar.gz
+# download and extract the source files
+curl -sSL http://nginx.org/download/$VERSION_NGINX.tar.gz | tar xz
+curl -sSL https://github.com/nbs-system/naxsi/archive/$VERSION_NAXSI.tar.gz | tar xz
+curl -sSL https://github.com/simpl/ngx_devel_kit/archive/v$VERSION_NDK.tar.gz | tar xz
+curl -sSL https://github.com/openresty/set-misc-nginx-module/archive/v$VERSION_SETMISC.tar.gz | tar xz
```
sgoings: Why not loop around your sources, then `curl <item> | tar xz`?

mboersma: That could be more DRY, but I'm ok with this minor instance of repetition unless @fabiob is motivated to rewrite it.

fabiob: I plan to rewrite this to include @bladealslayer changes on #3954 – also, to include verification of the plugins as well

fabiob: Actually, I just noticed that the current rebase removes @bladealslayer's changes, so the rewrite is a must.

mboersma: Good catch. Restoring that and adding verification of the other tarballs would be great. :green_heart:

fabiob: See the new PR: #4028

mboersma:
Closing in favor of #4028.

PR #4009, opened 2015-07-10
fix(controller): revert HTTP healthcheck feature

bacongobbler:
this new feature does not give the user a true zero-downtime effect as what was expected from #3813, so in order to ship 1.8 we're pulling it out and going to try again for v1.9. I'll also be closing #3982 and re-introducing a new PR for this.

re-opens #3813.


mboersma:
Code LGTM.

technosophos:
Code LGTM

PR #4024, opened 2015-07-14
chore(*): use etcd2 inside a container

aledbf:
alternative implementation of ref #4005

**Update 1:**
The changes in the python script are required because the `etcd2` section in cloud-unit is not valid until CoreOS >= 653.0.0

**Update 2:**
The reason of the symlink in the download of etcdctl is because the provided tar version in alpine does not support the parameter `--strip-components`

**Update 3:**
The new section in `provision-ec2-cluster.sh` is to wait until `etcd` is up and running before any `deisctl` command. (it can up to 2 minutes to download and start the container the first time)

**Update 4:** change the graceful-deis-shutdown script to use the new member removal command

**Update 5:** change the logic to obtain the etcd leader in `deis-swarm`

**Update 6:** the default port in fleet is 4001. This was changed [here](https://github.com/coreos/fleet/commit/b731b3cb6717282a97e92b7f71e308f4d314d023) but still is not released

**Update 7:** the old ports (4001 and 7001) are required because the provided `etcdctl` and `fleetctl` binaries in CoreOS are using `etcd v0.4.9`

**Update 8:** the new `github.com/ugorji` go dependency is because is used by `go-etcd`


carmstrong:
@aledbf My other concern is that the local `etcdctl` version on the CoreOS hosts won't necessarily match the daemon version. While this *shouldn't* be a big deal, it wouldn't surprise me if there are oddities. We can control the version inside the containers themselves, but not the version on the host. Thoughts?

aledbf:
> the local etcdctl version on the CoreOS hosts won't necessarily match the daemon version

Correct but deis provides a "custom" `etcdctl` binary in each component image, so from that perspective I don't see a problem.

aledbf:
@carmstrong any suggestion for this PR?
About `etcdctl` this is passing without any change in the versions so it's using `etcdctl v0.4.9`

carmstrong:
> any suggestion for this PR?

Let's bump the versions of `etcdctl` in all of the components to match the etcd version - even though it works as-is, I'd like to have bugfixes and backup/restore functionality in the client should it be necessary in the components.

carmstrong:
I also wouldn't mind entirely dropping the old ports for etcd (4001 and 7001) and switching everything over to use the new IANA ports.

aledbf:
@carmstrong done and passing

carmstrong:
@aledbf What if we dropped ports 4001 and 7001 entirely? We reference them in the bin/boot scripts in the components, but as long as we updated the ports there, I think we'd get away with only using the new ones.

We can also remove the `debug-etcd` service, since the new `etcdctl cluster-health` command replaces it. :+1:

aledbf:
@carmstrong done. It passed twice so I think is ready for another review.

For now is not possible to remove the port 4001 and 7001 from the etcd container (Update 7)

swarm/swarm.go
```diff
@@ -51,19 +57,29 @@ func run(cmd string) {
 }
 
 func getleaderHost() string {
```
aledbf: @smothiki cam you verify this change? (Is not "easy" to obtain the leader in etcd2)

smothiki: Sure, I might remove the function as well now swarm doesn't require etcd leader for service discovery.

carmstrong:
I just realized that in my overzealous desire to kick etcd 0.4 to the curb, I made this PR backwards-incompatible -- we can't bump `etcdctl` in the containers because we don't know if users will be running 0.4.x or 2.x. For several releases most likely, users will still be running etcd 0.4 on their hosts.

We'll probably need to stick to the old `etcdctl` and ports for the time being.

tests/fixtures/test-etcd/Dockerfile
```diff
@@ -3,11 +3,10 @@ FROM alpine:3.1
...
 
-ENV ETCD_VERSION v0.4.9
+# install etcdctl
+ENV ETCD_VERSION=2.0.13
+RUN curl -sSL https://github.com/coreos/etcd/releases/download/v${ETCD_VERSION}/etcd-v${ETCD_VERSION}-linux-amd64.tar.gz | \
+    tar zxv -C /usr/local/bin && ln -s /usr/local/bin/etcd-v${ETCD_VERSION}-linux-amd64/etcd* /usr/local/bin
 
-# install etcd and etcdctl
-RUN curl -sSL https://github.com/coreos/etcd/releases/download/$ETCD_VERSION/etcd-$ETCD_VERSION-linux-amd64.tar.gz \
-  | tar -vxz -C /usr/local/bin --strip=1
```
mboersma: You can extract things directly without using symlinks as long as you install alpine's `tar` package, which is more full-featured than the built-in busybox version.

mboersma:
Also, at least against vagrant, this PR does nothing currently. We still get CoreOS 647.2.0, so the `etcd2` section in user-data is ignored and old `etcd` comes up instead. I'd like to get #4032 merged as a way to fix that.

contrib/coreos/user-data.example
```diff
@@ -1,25 +1,57 @@
...
   fleet:
     # We have to set the public_ip here so this works on Vagrant -- otherwise, Vagrant VMs
     # will all publish the same private IP. This is harmless for cloud providers.
     public-ip: $private_ipv4
     # allow etcd to slow down at times
     etcd_request_timeout: 3.0
+    etcd_servers: http://127.0.0.1:2379
   units:
   - name: etcd.service
+    mask: true
```
aledbf: @mboersma the etcd service is masked so the user-data section for etcd is not used

mboersma: I was referring to what you said about the `etcd2` section doing nothing until CoreOS >= 653.0.0. The important point is that running this PR leaves you with a cluster that is *not* running `etcd2` in containers or otherwise.

mboersma: For now, prefixing tests or provisioning with `COREOS_CHANNEL=alpha|beta` allows this to be tested on vagrant, but we may need to change things if it has to work on 647.2.0.

tests/fixtures/test-etcd/Dockerfile
aledbf: @mboersma ok

contrib/coreos/user-data.example
mboersma: I am very wrong, and quite happy about it. This does seem to work just fine on current CoreOS 647.2.0:
```console
core@deis-01 ~ $ ps aux | grep etcd
etcd      1064  0.0  0.6 212924 13216 ?        Ssl  19:22   0:00 /usr/bin/docker run --net=host --rm --volume=/var/lib/etcd2:/var/lib/etcd2 --volume=/usr/share/ca-certificates:/etc/ssl/certs:ro -p 4001:4001 -p 2380:2380 -p 2379:2379 -p 7001:7001 --name etcd2 quay.io/coreos/etcd:v2.0.13 -name 59b8789c69504c34bb93c2c1b18c4765 -data-dir /var/lib/etcd2 -advertise-client-urls http://172.17.8.100:2379,http://172.17.8.100:4001 -listen-client-urls http://0.0.0.0:2379,http://0.0.0.0:4001 -initial-advertise-peer-urls http://172.17.8.100:2380,http://172.17.8.100:7001 -listen-peer-urls http://0.0.0.0:2380,http://0.0.0.0:7001 --heartbeat-interval 400 --election-timeout 2000 --discovery https://discovery.etcd.io/2059ff6f760766eab9a54335272a2e4b
root      1089  0.9  0.6  16112 13964 ?        Ssl  19:22   0:04 /etcd -name 59b8789c69504c34bb93c2c1b18c4765 -data-dir /var/lib/etcd2 -advertise-client-urls http://172.17.8.100:2379,http://172.17.8.100:4001 -listen-client-urls http://0.0.0.0:2379,http://0.0.0.0:4001 -initial-advertise-peer-urls http://172.17.8.100:2380,http://172.17.8.100:7001 -listen-peer-urls http://0.0.0.0:2380,http://0.0.0.0:7001 --heartbeat-interval 400 --election-timeout 2000 --discovery https://discovery.etcd.io/2059ff6f760766eab9a54335272a2e4b
core      1400  0.0  0.0   4288   272 pts/0    R+   19:30   0:00 grep --colour=auto etcd
core@deis-01 ~ $ cat /etc/os-release 
NAME=CoreOS
ID=coreos
VERSION=647.2.0
VERSION_ID=647.2.0
BUILD_ID=
PRETTY_NAME="CoreOS 647.2.0"
ANSI_COLOR="1;32"
HOME_URL="https://coreos.com/"
BUG_REPORT_URL="https://github.com/coreos/bugs/issues"
```

aledbf: @mboersma :) I was starting a new cluster to check this

swarm/swarm.go
smothiki: Right now we are using docker 1.5.0 and swarm required docker version >= 1.6.0. Also etcd2 doesn't support getting machines API.

aledbf: @smothiki the code that I changed in the `getleaderHost` obtains the machines and the leader. I ask for a review of those changes

smothiki: @aledbf my bad. leaderInfo has no attribute name. https://github.com/coreos/etcd/blob/master/Documentation/api.md#self-statistics .  
```curl 172.17.8.101:2379/v2/stats/self
{"name":"526602f0dd3e4ce3aa3d35b065447c24","id":"5531f89a10a42d8e","state":"StateLeader","startTime":"2015-07-21T23:33:31.597378Z","leaderInfo":{"leader":"5531f89a10a42d8e","uptime":"2m2.2133334s","startTime":"2015-07-21T23:33:33.022862067Z"},"recvAppendRequestCnt":0,"sendAppendRequestCnt":1801,"sendPkgRate":14.510472876845256,"sendBandwidthRate":1187.2468907834789}```

smothiki: My bad code looks good. working, I didn't check the LeaderInfo struct before.

aledbf: > LeaderInfo struct before.

@smothiki yeah, they mix name and id in different struct

carmstrong:
Closing in favor of #4084. Thanks as always, @aledbf!

Issue #3880, opened 2015-06-16
deisctl gets to "failed/failed" on a service, and then hangs.


technosophos:
On occasion, if I get a service to exit unexpectedly, then deisctl will hang on any subsequent commands.

Example:

```
$ deisctl stop builder
deis-builder.service: failed/failed
(hangs)
```

With a little bit of debugging code in `fleet/stop.go`, I can see that it's stuck in the for loop there:

```go
		lastSubState = currentState.SystemdSubState
		fmt.Printf("Waiting for state %s, but in state %s\n", desiredState, currentState.SystemdSubState) // Added
		time.Sleep(250 * time.Millisecond)
```

Which prints this out every 250 msec:

```
Waiting for state dead, but in state failed
```

I'm not sure if we're legitimately waiting for something to happen, or if `failed` or `dead` should be treated like a special case.

bacongobbler:
This has been an issue with deisctl since inception. It's probably a mis-use of the fleet client API that's causing this issue. IIRC the idea is that you give the fleet API a DesiredState (in this case, `loaded/dead`) and deisctl is just looping and waiting for that unit to show that it's now in that desired state.

mboersma:
Closed by #4095.

PR #4025, opened 2015-07-14
fix(deisctl): Switch from channels to io.Writers.

technosophos:
Fixes issue #3971 

deisctl used channels to send output and errors to a goroutine which
then wrote them. This patch replaces that with simpler code that simply
writes messages to Stdout and Stderr (configurable for outher writers).

PR #4026, opened 2015-07-14
fix(deisctl): Exit deisctl when start/stop hits state failed.

technosophos:
This fixes issue #3880. In start and stop commands, when the state hits failed (regardless of the expected state), deisctl exits.

### DIFF
--- a/deisctl/backend/fleet/start.go
+++ b/deisctl/backend/fleet/start.go
@@ -86,6 +86,12 @@ func doStart(c *FleetClient, target string, wg *sync.WaitGroup, out, ew io.Write
 		}
 
 		lastSubState = currentState.SystemdSubState
+
+		if lastSubState == "failed" {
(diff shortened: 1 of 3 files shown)


technosophos:
Testing instructions:

- Install this version of `deisctl`
- Bring up a deis cluster `deisctl start platform`
- Start and stop a component a few times `deisctl start controller && deisctl stop controller`

You should not get the deisctl client into the state where it is hung waiting for a failed service to stop/start.

Caveat: There is still one known issue with builder during which it _appears_ to be started, while it hasn't in fact finished. That will interfere with this test. So don't test with builder.

mboersma:
This code looks good, but I think the PR needs to be rebased to excise the first two commits now that we merged #4025.

technosophos:
Closing this because the history is wrong, and a part of patch that shouldn't get merged is still in there.

Issue #4030, opened 2015-07-15
Investigate spike in logger container during load test


jchauncey:
While load testing my deis cluster yesterday I noticed the logger container was always using more cpu than the rest of the containers in the cluster. It would be a good exercise to determine why that might be. 

Attached is the scap file which. You can use sysdig to examine the dump.
[capture.scap](https://s3-us-west-2.amazonaws.com/jchauncey-artifacts/capture150714030443.scap)

mboersma:
This was Deis v1.8-era when @jchauncey noticed and faithfully reported this, but nothing similar has cropped up since so I'm closing it.

PR #4034, opened 2015-07-15
fix(logger): use one ticker for all time-related operations

bacongobbler:
In a separate goroutine, we were sleeping for the exact same time as
publishInterval was set to. As @kalbasit mentioned in [a comment][1],
it's better to use a ticker anyways. This removes one extra underlying
system call to clock_gettime() by making both publishKeys() and the
log drain poll share the same ticker event.

refs #4030, as I found this using the .scap dump. I don't think this will
drop the CPU usage significantly, but it's a start.

[1]: https://github.com/deis/deis/pull/3795#discussion_r31878237

### DIFF
--- a/logger/main.go
+++ b/logger/main.go
@@ -44,7 +44,7 @@ func main() {
 	flag.Parse()
 
 	client := etcd.NewClient([]string{"http://" + publishHost + ":" + publishPort})
-
+	ticker := time.NewTicker(time.Duration(publishInterval) * time.Second)
 	signalChan := make(chan os.Signal, 1)
 	drainChan := make(chan string)
 	stopChan := make(chan bool)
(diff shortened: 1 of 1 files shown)


technosophos:
Code LGTM

mboersma:
Code LGTM.

`Showing  1 changed file  with 12 additions and 29 deletions.` :smile:

logger/main.go
```diff
@@ -76,12 +80,6 @@ func main() {
...
 			}
-			time.Sleep(time.Duration(publishInterval) * time.Second)
-		}
-	}()
-
-	for {
-		select {
 		case <-signalChan:
 			close(exitChan)
 			stopChan <- true
```
kalbasit: You should also stop the ticker. Github doesn't allow me to add a comment on a line that is hidden.

```
 		case <-cleanupChan:
 			return
```

To

```
 		case <-cleanupChan:
 			t.Stop()
 			return
```

to stop the ticket (avoid leaking goroutines)

logger/main.go
```diff
@@ -97,21 +95,6 @@ func publishKeys(client *etcd.Client, host, etcdPath, port string, ttl uint64) {
...
-// something on exitChan.
-func publishService(exitChan chan bool, client *etcd.Client, host string, etcdPath string, port string, ttl uint64) {
-	publishKeys(client, host, etcdPath, port, ttl)
-	t := time.NewTicker(time.Duration(publishInterval) * time.Second)
-	for {
-		select {
-		case <-t.C:
-			publishKeys(client, host, etcdPath, port, ttl)
-		case <-exitChan:
-			return
```
kalbasit: probably my mistake here I forgot to stop the ticker :(

logger/main.go
bacongobbler: good catch. Thanks!

bacongobbler:
done. now it's calling `ticker.Stop()` on `cleanupChan` as well as I removed an unused variable (`stopChan`) which was probably from an older refactor.

mboersma:
LGTM++.

PR #4042, opened 2015-07-15
chore(*): update lists of supported providers

carmstrong:
We added Linode support. I updated the lists so they are the same
everywhere.


client/setup.py
```diff
@@ -41,7 +41,7 @@
       url='https://github.com/deis/deis',
       keywords=[
           'opdemand', 'deis', 'paas', 'cloud', 'coreos', 'docker', 'heroku',
```
mboersma: Who left that first keyword in there? :open_mouth:

mboersma:
Code LGTM.

bacongobbler:
docs LGTM

PR #4043, opened 2015-07-15
chore(contrib/ec2): rename ec2 to aws

carmstrong:
We were incorrectly using "AWS" and "EC2" interchangeably, when in
fact EC2 is one of the services that compose AWS. Deis clusters on
AWS make use of EC2 and CloudFormation and other AWS services,
so calling it EC2 is incorrect, and using both terms is inconsistent.


bacongobbler:
:+1: haven't reviewed this but the language change makes sense.

carmstrong:
Oops... looks liek we can't use `test-ec2` for this since it won't find the old script. I could symlink it to test this PR, but not sure how to go about testing it. Deferring to @mboersma

mboersma:
> We were incorrectly using "AWS" and "EC2" interchangeably

Old habits die hard. You're right--this is better.

I'll see if I can rig the Jenkins job to find either test script location.

carmstrong:
Test failed with:
```
test.test-775fc7b80e.deis.works
ok
sleeping for 60 seconds until certs are generated...
ok
SIGQUIT: quit
PC=0x43ec61
```

Seems like a transient error?

bacongobbler:
Yes, looks transient. Let's re-run it once more to make sure but this should be good if it passes.

mboersma:
Code LGTM.

bacongobbler:
code LGTM!

Issue #4040, opened 2015-07-15
Add shell unit testing harness


sgoings:
We have a bunch of shell scripts in the deis project, and no unit (or integration) tests for them. There are many shell unit testing frameworks and even ones that have mock/stub functionality. Let's pull one of them in and provide a model for shell script tests.

I know it may sound weird to test your shell scripts with more shell scripts, but take for example #4023. Wouldn't it have been cool to have a way to test all those conditions during the development of that script? Additionally, it would help someone in the future verify that future changes have not broken existing behavior.

bacongobbler:
:+1: for more tests

sgoings:
Also see parallel discussion happening in https://github.com/deis/rigger/issues/14.

mboersma:
Good stuff here, but this isn't going to be a development priority as Deis v1 LTS is focused on stability and Deis v2 has minimized the use of shell scripts.

PR #4044, opened 2015-07-15
test(style): run bash scripts through shellcheck

### LINKED ISSUE #3347: Proposal chore(scripts): use shellcheck to verify shell scripts
http://www.shellcheck.net/about.html

an example:

```
aledbf:bin aledbf$ shellcheck test-integration-ec2.sh

In test-integration-ec2.sh line 11:
export THIS_DIR=$(cd $(dirname $0); pwd)
                     ^-- SC2046: Quote this to prevent word splitting.
                               ^-- SC2086: Double quote to prevent globbing and word splitting.


In test-integration-ec2.sh line 14:
source $THIS_DIR/test-setup.sh
       ^-- SC2086: Double quote to prevent globbing and word splitting.


In test-integration-ec2.sh line 25:
    aws cloudformation delete-stack --stack-name $STACK_NAME
                                                 ^-- SC2086: Double quote to prevent globbing and word splitting.


In test-integration-ec2.sh line 26:
    python $DEIS_ROOT/tests/bin/route53-wildcard.py delete $DEIS_TEST_DOMAIN $ELB_DNS_NAME
           ^-- SC2086: Double quote to prevent globbing and word splitting.
                                                           ^-- SC2086: Double quote to prevent globbing and word splitting.
                                                                             ^-- SC2086: Double quote to prevent globbing and word splitting.


In test-integration-ec2.sh line 151:
        | grep InService | wc -l)
          ^-- SC2126: Consider using grep -c instead of grep|wc.
......
```

Joshua-Anderson:
Notes:

1. Jenkins now needs to have shellcheck installed. I'm defering to @mboersma and @sgoings on this.
2. I made a **lot** of changes to get these tests to make things pass, so this needs to be well manually tested to make sure I didn't break anything.

closes #3347

### DIFF
--- a/contrib/aws/provision-aws-cluster.sh
+++ b/contrib/aws/provision-aws-cluster.sh
@@ -12,14 +12,14 @@ fi
 
 set -e
 
-THIS_DIR=$(cd $(dirname $0); pwd) # absolute path
-CONTRIB_DIR=$(dirname $THIS_DIR)
+THIS_DIR=$(cd "$(dirname "$0")"; pwd) # absolute path
+CONTRIB_DIR=$(dirname "$THIS_DIR")
 
-source $CONTRIB_DIR/utils.sh
+source "$CONTRIB_DIR/utils.sh"
 
 # check for AWS API tools in $PATH
 if ! which aws > /dev/null; then
-  echo_red 'Please install the AWS command-line tool and ensure it is in your $PATH.'
+  echo_red 'Please install the AWS command-line tool and ensure it is in your PATH.'
   exit 1
 fi
 
@@ -40,18 +40,20 @@ if [ -n "$VPC_ID" ]; then
 fi
 
 # check that the CoreOS user-data file is valid
-$CONTRIB_DIR/util/check-user-data.sh
+"$CONTRIB_DIR/util/check-user-data.sh"
 
 # Prepare bailout function to prevent us polluting the namespace
 bailout() {
-  aws cloudformation delete-stack --stack-name $STACK_NAME
+  aws cloudformation delete-stack --stack-name "$STACK_NAME"
 }
 
 # create an AWS cloudformation stack based on CoreOS's default template
+# Don't warn about EXTRA_AWS_CLI_ARGS being unquoted
+# shellcheck disable=SC2086
 aws cloudformation create-stack \
-    --template-body "$($THIS_DIR/gen-json.py --channel $COREOS_CHANNEL --version $COREOS_VERSION)" \
-    --stack-name $STACK_NAME \
-    --parameters "$(<$THIS_DIR/cloudformation.json)" \
+    --template-body "$("$THIS_DIR/gen-json.py" --channel "$COREOS_CHANNEL" --version "$COREOS_VERSION")" \
+    --stack-name "$STACK_NAME" \
+    --parameters "$(<"$THIS_DIR/cloudformation.json")" \
     $EXTRA_AWS_CLI_ARGS
 
 # loop until the instances are created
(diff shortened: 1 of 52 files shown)


Joshua-Anderson:
ref #4040

Joshua-Anderson:
BTW Shellcheck caught 2 major errors.

Typo:
```
#/bin/bash
```

> 1 arg not supported:
```
#!/usr/bin/env bash -e
```


Is there a standard about whether to use `#!/bin/bash` or `#!/usr/bin/env bash`

bacongobbler:
it's a recommendation: http://unix.stackexchange.com/questions/29608/why-is-it-better-to-use-usr-bin-env-name-instead-of-path-to-name-as-my

Joshua-Anderson:
Is this something that should be standardized throughout all the shellscripts? They are a mixture of both.

bacongobbler:
Yes, we've had PRs in the past that have migrated to /usr/bin/env before: https://github.com/deis/deis/pull/2538

mboersma:
Here's how I got current `shellcheck` installed on CI master and nodes:
```
# Use cabal (Haskell installer) to build and install ShellCheck
apt-get install -yq cabal-install
cabal update
cd /tmp
git clone --single-branch https://github.com/koalaman/shellcheck.git
cd shellcheck
cabal install --global
apt-get purge cabal-install
```

@Joshua-Anderson could you add those lines to [setup-node.sh](https://github.com/deis/deis/blob/master/tests/bin/setup-node.sh) as part of this PR?

Joshua-Anderson:
@mboersma Done!

Joshua-Anderson:
Rebased and made sure that `$EXTRA_AWS_CLI_ARGS` isn't quoted.

contrib/aws/provision-aws-cluster.sh
```diff
@@ -102,10 +110,12 @@ until [ `wc -w <<< $INSTANCE_STATUSES` -eq $DEIS_NUM_INSTANCES ]; do
...
 
     if [ $COUNTER -ne 1 ]; then sleep $SLEEPTIME; fi
-    echo "Waiting for instances to pass initial health checks ($(expr 61 - $COUNTER)0s) ..."
+    echo "Waiting for instances to pass initial health checks ($((61 - COUNTER))0s) ..."
+    # Don't warn about EXTRA_AWS_CLI_ARGS being unquoted
+    # shellcheck disable=SC2086
     INSTANCE_STATUSES=$(aws ec2 describe-instance-status \
         --filters Name=instance-status.reachability,Values=passed \
-        --instance-ids $INSTANCE_IDS \
+        --instance-ids "$INSTANCE_IDS" \
```
mboersma: This shouldn't be quoted: it makes all the instance IDs appear as one arg, which confused the provisioning script:
```
A client error (InvalidInstanceID.Malformed) occurred when calling the DescribeInstanceStatus operation: Invalid id: "i-661ddcaf
i-d4809923
i-ca80993d"
```

mboersma:
D'oh, I think my diagnosis was wrong. CI still gets the same error after your most recent change, so probably we mistakenly quoted `INSTANCE_IDS` previously in the script.

Joshua-Anderson:
@mboersma I think I found the guilty line in the `test-integration-aws.sh` script.

controller/Makefile
```diff
@@ -4,6 +4,8 @@ include ../includes.mk
 
 all: build run
 
+SHELL_SCRIPTS = build.sh $(wildcard bin/*) $(wildcard migrations/data/*.sh)
```
bacongobbler: Something easier and more dynamic is to use `$(shell find $PWD -name '*.sh')`. That way we won't have to update this if we introduce new directories. Lazy searching ftw

Joshua-Anderson: Not every shell script is `*.sh`. `bin/boot` for example.

Joshua-Anderson: I'd be happy to add that in addition to the others ones though.

bacongobbler: good point. It'll remove the wildcard for the etcd data migrations, at the very least :)

Joshua-Anderson: Your command won't work because `$PWD` is very fragile. So if I ran

Joshua-Anderson: Fixed `$(shell find "." -name '*.sh')`

bacongobbler: That works! I didn't know if `.` would demonstrate the same behaviour as $PWD. I thought Makefiles were smarter than that. Good catch!

Joshua-Anderson:
Added shellcheck to mesos and used @bacongobbler's suggestion for searching for shell files.

carmstrong:
Looks like this needs a rebase, @Joshua-Anderson :(

Joshua-Anderson:
Rebased!

Joshua-Anderson:
I'm going to break this into smaller chunks so that it is easier to review.

carmstrong:
> I'm going to break this into smaller chunks so that it is easier to review.

:+1: Going to close this PR, then. One less thing "awaiting review".

PR #4048, opened 2015-07-16
fix(contrib): warn if commit subject is too long

mboersma:
[skip ci]


Joshua-Anderson:
:+1: Do we want to use yellow for warnings?

mboersma:
Yes, I'll change it. :yellow_heart:

bacongobbler:
LGTM. The line length was more of a guideline than an enforced rule. I'll update the docs to be less harsh. :)

technosophos:
LGTM

PR #4049, opened 2015-07-16
fix(commit-msg): ignore comments in line count

Joshua-Anderson:
`git commit ammend` has comments about the files changed. If you had a comment longer than the limit the hook would throw an error. Now the hook ignores comments.


mboersma:
Code LGTM.

PR #4050, opened 2015-07-16
Install go tools in root Makefile from golang.org

mboersma:
[code.google.com is going away](http://google-opensource.blogspot.com/2015/03/farewell-to-google-code.html), and there were multiple versions of `make setup-gotools` in the project. This also moves the location of the `websocket` package used by logspout.


Godeps/Godeps.json
```diff
@@ -345,6 +340,10 @@
 			"Rev": "f7445b17d61953e333441674c2d11e91ae4559d3"
 		},
 		{
+			"ImportPath": "golang.org/x/net/websocket",
+			"Rev": "d9558e5c97f85372afee28cf2b6059d7d3818919"
```
bacongobbler: How come we're bumping the rev here to master? https://github.com/golang/net/commits/4231557d7c726df4cf9a4e8cdd8a417c8c200bdb doesn't seem to exist so I'm not sure what we were currently pointing at, but if you're confident that this won't break stuff then I'm ok with this

bacongobbler:
code LGTM.

Godeps/Godeps.json
mboersma: I couldn't match up the SHAs between hosts, so I just vendored from master and looked at the diff. I agree, it looks like safe changes that logspout probably wouldn't care about anyway. :four_leaf_clover:

carmstrong:
Code LGTM.

Issue #2819, opened 2014-12-30
database: Better understand the window of loss (and loss scenarios) for database backups


carmstrong:
The database component uses [wal-e](https://github.com/wal-e/wal-e) to ship backups and WAL logs to the deis-store component. Full backups occur by default every 3 hours, and WAL logs are shipped continuously. When a database comes up fresh on a host, it first asks WAL-e if there are any backups to restore. If so, it restores the backup and then begins to replay the WAL logs that were written since the backup.

There is potentially a small window of loss here if the database component should exit before the last WAL log is written to disk. Additionally, users (including @lorieri) have occasionally reported loss of some data when upgrading Deis or otherwise performing a database restore. We test Deis platform upgrades for every Deis release, and this includes a database restore. However, the amount of data we're storing is not large. There is likely either a larger loss window than we think, or other causes of data loss. 

Ceph doesn't seem to be the blame, as registry data (which is also stored in Ceph) is reported to be fully intact. This points to a combination of the database backup loop logic and wal-e itself. We should so some investigation to better attempt to replicate what users are sometimes seeing in real-world environments.

Somewhat related is the fragility of the backup loop in the bin/boot script: #2741

lorieri:
Today I restarted the cluster, I started ceph in order (monitor, daemon, metadata, volume, gateway)

same happened, last app created did not appears in ```deis apps```

carmstrong:
This should be re-evaluated after #2741 -- that reimplementation of the /bin/boot scripts is something we want to do as soon as possible, and that work will affect the findings from this.

lorieri:
Hi,

I was thinking in something

lets say I start my database in machine A, it will create a database volume on A.
then A goes down, database will be migrate to B and gets backups from the gateway.
then all machines goes down and database starts on A before the gateway.

will it get the old volume and ignore backups ?

bacongobbler:
yes, that's what is described in #3128.

lorieri:
:)

mboersma:
The fixes mentioned have been merged into Deis, and several other boundary cases relating to DB restore were found and fixed by the v1.10.1 release. I'm going to close this, but please open a new issue with specifics if something like this recurs with the current [Deis v1.11 release](https://github.com/deis/deis/releases/tag/v1.11.0).

PR #4056, opened 2015-07-17
feat(database): backup the db on unit stop

iancoffey:
When performing an upgrade, any db records added since the last backup are lost. This becomes obvious when testing upgrades, but should occur anytime the database is stopped and started.

The thought here is that a backup be done proactively on stop to avoid it. Maybe theres a better place for it, but this does "work".


bacongobbler:
perhaps this is something that could be added to `on_exit()`? That way we can cover this with tests!

iancoffey:
@bacongobbler makes sense, ill change this around.

carmstrong:
This is related to #2819. I don't understand why this is happening, though, since we are continuously shipping WAL logs...

carmstrong:
@iancoffey Let me know if I can help with this. This is definitely a good fix.

iancoffey:
So the term signal doesnt get actually trapped on container stop - not sure why. Anyway, putting the backup stuff in the existing on_exit() wont do the job since it doesnt get called when database is stopped.

carmstrong:
@iancoffey Are we sure the ExecStopPosts get called in order? As in, systemd waits for the first to complete before moving onto the second?

If so, code LGTM.

carmstrong:
Kicked off a new CI build: https://ci.deis.io/job/test-ec2/2088/

iancoffey:
@carmstrong yes, from what I am able to grok from the man page, they are executed sequentially.

carmstrong:
Code LGTM.

bacongobbler:
code LGTM

PR #4057, opened 2015-07-17
fix(Makefiles): silence static binary check

mboersma:
The [`check-static-binary`](https://github.com/deis/deis/blob/master/includes.mk#L55) Makefile function echoes itself to stdout with every invocation. This silences it unless it actually finds an error.

ping @technosophos


technosophos:
LGTM and HUZZAH!

carmstrong:
LGTM.

PR #4059, opened 2015-07-17
feat(client-go): add domains endpoint

Joshua-Anderson:


client-go/controller/models/domains/domains.go
```diff
@@ -0,0 +1,77 @@
...
+
+	domains := api.Domains{}
+	if err = json.Unmarshal([]byte(body), &domains); err != nil {
+		return []api.Domain{}, err
+	}
+
+	return domains.Domains, nil
+}
+
+// New adds domains to an app.
```
mboersma: Only one though, so maybe the comment should be singular.
```
// New adds a domain to an app.
```

client-go/controller/models/domains/domains.go
```diff
@@ -0,0 +1,77 @@
...
+
+	res := api.Domain{}
+	if err = json.Unmarshal([]byte(resBody), &res); err != nil {
+		return api.Domain{}, err
+	}
+
+	return res, nil
+}
+
+// Delete removes domains registered with an app.
```
mboersma: Same as above, I think singular is more clear:
```
// Delete removes a domain registered with an app.
```

Joshua-Anderson:
@mboersma Done ( I was already fixing it as you commented)!

client-go/parser/domains.go
```diff
@@ -0,0 +1,107 @@
+package parser
+
+import (
+	"fmt"
+
+	"github.com/deis/deis/client-go/cmd"
+	docopt "github.com/docopt/docopt-go"
+)
+
+// Domains routes domain commands to the specific function
```
mboersma: Ideally all comments should be full sentences, ending with a period, so:
```
// Domains routes a domain command to the specific function implementing it.
```

mboersma:
Code LGTM, my comments about comments notwithstanding.

client-go/parser/domains.go
Joshua-Anderson: Already fixed, refresh!

technosophos:
Code LGTM

PR #4062, opened 2015-07-17
feat(client-go): add builds endpoint

Joshua-Anderson:


client-go/cmd/builds.go
```diff
@@ -0,0 +1,64 @@
+package cmd
+
+import (
+	"fmt"
+
+	"gopkg.in/yaml.v2"
```
mboersma: I was going to squawk about multiple vendoring techniques here, but I see we [already import this way](https://github.com/deis/deis/blob/master/builder/utils.go#L7) in the builder. We could use `godep` to vendor from `gopkg.in` for consistency, but I'm curious if other maintainers see this as an issue.

Joshua-Anderson: It is already in godeps, that's why I'm using it :smiley:.

mboersma: I should have looked! Nevermind, problem solved.

mboersma:
Code LGTM.

technosophos:
Code LGTM

Joshua-Anderson:
Rebased!

PR #4068, opened 2015-07-20
fix(client-go): defer deferred closing of body until after error check.

Joshua-Anderson:
If the request aborted due to an error the body didn't exist and
the deferred function would throw a nil pointer exception.

Reported by yebyen on IRC.

Example panic: http://pastebin.com/NsjfvEWf


bacongobbler:
LGTM

technosophos:
Code LGTM

PR #4071, opened 2015-07-20
fix(controller): convert config to correct types

### LINKED ISSUE #4070: Dynamic typing for config response
```
POST /v1/apps/<app id>/config/

{
	"cpu" {
		"web": 1024
	}
}

Response:

{
	...
	"cpu" {
		"web": 1024
	}
}

POST /v1/apps/<app id>/config/

{
	"cpu" {
		"web": "1024"
	}
}

Response:

{
	...
	"cpu" {
		"web": "1024"
	}
}
```

I think that dynamic parsing for the request is fine, but having a dynamic type for a response make it very difficult to parse.

Joshua-Anderson:
config.cpu will always be a integer

config.memory, config.values, and config.tags will always be
a string

Previously the controller would take and save any type,
provided it passed validation. This was problematic because
it could lead to mixed or unexpected types.

Examples:

This config would be valid:

```
{
  "values": {
    "TEST1": true,
    "TEST2": 123,
    "TEST3": "test"
  }
}
```

You could get a response like this:

```
{
  "cpu": {
    "web": 1024,
    "worker": "512"
  }
}
```

Fixes #4070

### DIFF
--- a/controller/api/serializers.py
+++ b/controller/api/serializers.py
@@ -24,15 +24,56 @@ TAGVAL_MATCH = re.compile(r'^\w+$')
 
 
 class JSONFieldSerializer(serializers.Field):
+    """
+    A Django REST framework serializer for JSON data.
+    """
+
     def to_representation(self, obj):
+        """Serialize the field's JSON data, for read operations."""
         return obj
 
     def to_internal_value(self, data):
+        """Deserialize the field's JSON data, for write operations."""
         try:
             val = json.loads(data)
         except TypeError:
             val = data
         return val
+
+
+class JSONIntFieldSerializer(JSONFieldSerializer):
+    """
+    A JSON serializer that coerces its data to integers.
+    """
+
+    def to_internal_value(self, data):
+        """Deserialize the field's JSON integer data."""
+        field = super(JSONIntFieldSerializer, self).to_internal_value(data)
+
+        for k, v in field.viewitems():
+            if v is not None:  # NoneType is used to unset a value
+                try:
+                    field[k] = int(v)
+                except ValueError:
+                    field[k] = v
+                    # Do nothing, the validator will catch this later
+        return field
+
+
+class JSONStringFieldSerializer(JSONFieldSerializer):
+    """
+    A JSON serializer that coerces its data to strings.
+    """
+
+    def to_internal_value(self, data):
+        """Deserialize the field's JSON string data."""
+        field = super(JSONStringFieldSerializer, self).to_internal_value(data)
+
+        for k, v in field.viewitems():
+            if v is not None:  # NoneType is used to unset a value
+                field[k] = unicode(v)
+
+        return field
 
 
 class ModelSerializer(serializers.ModelSerializer):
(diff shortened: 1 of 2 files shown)


Joshua-Anderson:
@mboersma It's an outlier! I think you might want the controller label, not the client one, unlike the other 3 or so PR's I opened today. :smiley:

bacongobbler:
We should also add a data migration to convert existing values into ints as well, or else we're still gonna end up with mixed types from existing clusters :)

Joshua-Anderson:
@bacongobbler Unless there are current users who were doing weird things with the api, nobody should be seeing this. If for some reason users were type mixing with the config the next time they change anything with the config this code will run and fix it.

controller/api/tests/test_config.py
```diff
@@ -133,6 +133,31 @@ def test_response_data(self):
         self.assertDictContainsSubset(expected, response.data)
 
     @mock.patch('requests.post', mock_import_repository_task)
+    def test_response_data_types_converted(self):
+        """Test that config data is converted into the correct type."""
+        body = {'id': 'test'}
+        response = self.client.post('/v1/apps', json.dumps(body),
+                                    content_type='application/json',
+                                    HTTP_AUTHORIZATION='token {}'.format(self.token))
+        url = "/v1/apps/test/config".format()
```
mboersma: This `format` call is a no-op. :question:

Joshua-Anderson:
@mboersma Fixed!

mboersma:
Code LGTM.

controller/api/models.py
```diff
@@ -672,6 +672,14 @@ def save(self, **kwargs):
                 setattr(self, attr, data)
         except Config.DoesNotExist:
             pass
+
+        for k, v in getattr(self, 'cpu').copy().viewitems():
+            self.cpu[k] = int(v)
```
bacongobbler: what if someone enters a bad value like `foo` here? I think we should be adding tests for bad values here as well so we're handling both good and bad values properly.

bacongobbler: Note that this brittle code may be something that the [serializers](http://www.django-rest-framework.org/api-guide/serializers/) can handle better. See https://github.com/deis/deis/blob/master/controller/api/serializers.py#L26

Joshua-Anderson:
@bacongobbler I'm now using serializers as you suggested. Thanks for the tip, the code is much cleaner now!

As for your comment about the results if somebody gives a string to `cpu`, the serializer already validates that cpu is a integer, but since python is dynamically typed `"1234"` would pass the integer check but would still be saved as a string.

controller/api/serializers.py
```diff
@@ -35,6 +35,42 @@ def to_internal_value(self, data):
         return val
 
 
+class JSONIntFieldSerializer(serializers.Field):
```
mboersma: There should be python docstrings for these classes and methods. Also, I think subclassing `JSONFieldSerializer` could provide more code reuse. Something like this (untested!):
```python
class JSONFieldSerializer(serializers.Field):
    """
    A Django REST framework serializer for JSON data.
    """

    def to_representation(self, obj):
        """Serialize the field's JSON data, for read operations."""
        return obj

    def to_internal_value(self, data):
        """Deserialize the field's JSON data, for write operations."""
        Transform lists of primitive values dicts of native values to
        try:
            val = json.loads(data)
        except TypeError:
            val = data
        return val


class JSONIntFieldSerializer(JSONFieldSerializer):
    """
    A JSON serializer that coerces its data to integers.
    """

    def to_internal_value(self, data):
        """Deserialize the field's JSON integer data."""
        field = super(JSONIntFieldSerializer, self).to_internal_value(data)

        for k, v in field.viewitems():
            if v is not None:  # NoneType is used to unset a value
                field[k] = int(v)

        return field


class JSONStringFieldSerializer(JSONFieldSerializer):
    """
    A JSON serializer that coerces its data to strings.
    """

    def to_internal_value(self, data):
        """Deserialize the field's JSON string data."""
        field = super(JSONStringFieldSerializer, self).to_internal_value(data)

        for k, v in field.viewitems():
            if v is not None:  # NoneType is used to unset a value
                field[k] = unicode(v)

        return field
```

bacongobbler: +1

Joshua-Anderson:
@mboersma Good idea! Done.

mboersma:
Code LGTM.

bacongobbler:
would you mind adding a unit test for poorly formatted values? This is just for sanity's sake so the system won't blow up on those values. We should expect the API to handle it and give a correct error response if we ever decide to work on this code again.

Joshua-Anderson:
@bacongobbler I'm now testing a bad value when converting to int. Good idea too, because I inadvertently broke validation behavior for cpu values, resulting in a 500 instead of a nice 400.

bacongobbler:
bwahaha

code LGTM

PR #4074, opened 2015-07-20
feat(client-go): add tags endpoint

Joshua-Anderson:


mboersma:
This code looks good, but there aren't any tests...

Joshua-Anderson:
@mboersma The `tags` code primarily uses the `config` models package which is heavily tested. This code is responsible for the command line parsing and STDOUT. The `cmd` and `parser` packages, the packages that this PR touched, are very lightly tested as I figured these packages were covered by the integration tests as they only dealt with command line parsing and output.

mboersma:
Code LGTM.

bacongobbler:
code LGTM

PR #4078, opened 2015-07-21
fix(deisctl): update mesos code to use io.Writer

mboersma:
The mesos start/stop/install code needed to change from using channels to io.Writers in order to compile.

Tested manually with `make -C deisctl/ build && deisctl install|start|etc mesos`.

Refs #4025.

### DIFF
--- a/deisctl/cmd/mesos.go
+++ b/deisctl/cmd/mesos.go
@@ -2,90 +2,80 @@ package cmd
 
 import (
 	"fmt"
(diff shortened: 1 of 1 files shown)


carmstrong:
Code LGTM.

bacongobbler:
code LGTM!

technosophos:
This almost certainly needs to be done on the K8s code, too, @smothiki

Issue #4000, opened 2015-07-09
Stateless logger


carmstrong:
Users who have elected to run a Ceph-less cluster (http://docs.deis.io/en/latest/managing_deis/running-deis-without-ceph/) give up application logs since ``deis-controller`` has nowhere to store them. To alleviate this, we plan to implement an in-memory ring buffer for the ``deis-logger`` component, so that a reasonable period of recent logs is stored and can be queried with ``deis logs``.

bacongobbler:
related: #3963

Issue #3963, opened 2015-07-03
API / controller logging should be via logspout


krancour:
I haven't dug into the code too deeply yet, but conversation in #3947 exposed that `deis-controller` logs are written _directly_ to a volume that it shares with `deis-logger`.  This means that the route `deis-controller` logs take to their resting place on that volume is _different_ than the route taken by application logs making their way to the same place.  In the case of an application, it's via `STDOUT` --> `deis-logspout` --> `deis-logger` --> the volume.

There's a slight problem that arises from this discrepancy.  In the increasingly common scenario where an operator runs "Ceph-less," Deis is installed sans the storage subsystem components and adequate, external alternatives are provided for those components _and_ any other Deis component that expresses a direct dependency on the storage subsystem (e.g. `deis-database`).  When opting to run this way, `deis-logger` is among the components not installed.  A consequence of this is that `deis-logspout` is configured to forward logs _directly_ to an external log service.  Since application logs are forwarded via `deis-logspout`, those logs all make it to the external service.  `deis-controller` logs, on the other hand, since they take a different route, would never make it to the external log service.

It seems this could be easily corrected by transitioning `deis-controller` to writing its logs via `STDOUT` and `logspout` just as any application does.  This would increase consistency and guarantee that whether running with or without Ceph and `deis-logger`, all logs ultimately land in one common place via a consistent path.

In #3947 @bacongobbler has noted some possible problems with this approach, but we should probably move that discussion here.

krancour:
I think the right way to tackle this is to have the controller prefix log messages for application events with some "tag" like `[kents-awesome-app]` and then give our customized logspout the smarts to understand that such messages "belong" to the app and not the controller.

I did briefly think about putting those smarts in logger instead of logspout, but logspout makes more sense.  Those running the "stateless platform" (cephless) are already not necessarily using logger-- which is part of the reason for this PR in the first place.  This will ensure log messages from the controller (including app events) will be handled correctly regardless of whether they're being forwarded to logger or to somewhere else, like Papertrail.

PR #4083, opened 2015-07-22
feat(logger) Stateless logger implementation through in memory ring buffer

rvadim:
For testing:
 * deisctl config logs set handlertype=ringbuffer
 * deisctl restart logger

Implementing #4000


deis-admin:

mboersma:
Jenkins, add to whitelist.

carmstrong:
@rvadim This is some awesome work! We'll need to dive in and play with this. :+1:

rvadim:
I actually have no idea why one of request to controller "apps:info --app={{.AppName}}", failed with ('Connection aborted.', BadStatusLine("''",))

Maybe problem in slow environment? I will try to reproduce this.

rvadim:
There are some problems in logs implementation in controller. I will try to fix this in next few hours.

technosophos:
A little bit of a description of how this works might speed up the review.

logger/handlers/ringbuffer.go
```diff
@@ -0,0 +1,134 @@
...
+func (h *Handler) ringBufferLoop() {
+	for {
+		m := h.Get()
+		if m == nil {
+			break
+		}
+		if h.DrainURI != "" {
+			drain.SendToDrain(m.String(), h.DrainURI)
+		}
+		err := writeToBuffer(m)
```
technosophos: nit: Should be `if err := writeToBuffer(m); err != nil {` ...

bacongobbler:
+1 for usage, architecture and maintenance docs so the user knows what they're getting into

logger/handlers/ringbuffer.go
```diff
@@ -0,0 +1,134 @@
...
+		fmt.Fprintf(w, "Could not get application name from url: %s", r.RequestURI)
+		return
+	}
+
+	log_lines, err := strconv.Atoi(match[2])
+	if err != nil {
+		fmt.Fprintln(w, "Unable to get log lines parameter from request")
+		return
+	}
+	data, err := getFromBuffer(match[1], log_lines)
```
technosophos: What if log_lines > length of ring buffer?

logger/handlers/ringbuffer.go
```diff
@@ -0,0 +1,134 @@
+package handlers
```
technosophos: Why is this `handlers` (plural) instead of `handler`? I think we tend to stay with singular terms (`model`, not `models`, `encoder`, not `encoders`...)

I don't actually like the name `handler` here at all, since it isn't descriptive, and it seems at odds with the usage of `handler` in `net/http`. But I don't really care enough about that to make a big deal of it.

logger/handlers/ringbuffer.go
```diff
@@ -0,0 +1,134 @@
...
+		projectBuffer[name] = r
+	}
+}
+
+func RingBufferHandler(bufferSize int, webServicePort int) *Handler {
+	ringBufferSize = bufferSize
+	h := Handler{
+		BaseHandler: syslog.NewBaseHandler(5, filter, false),
+	}
+	go h.ringBufferLoop()
```
technosophos: Do we need a cancelation channel here? Are there cases where we want to stop ringBufferLoop other than having m nil'ed out.

Really, this is an open question. I don't know what the answer is. I see that the existing syslogger does this too. I'm good with leaving this as is unless we think it's possible that loggers need to be toggled at runtime, in which case it might make sense to externally cancel the goroutine.

logger/handlers/ringbuffer.go
```diff
@@ -0,0 +1,134 @@
...
+		data += fmt.Sprintln(line)
+	}
+
+	currentRing.Next().Do(getLine)
+	return data, nil
+}
+
+func httpHandler(w http.ResponseWriter, r *http.Request) {
+
+	if r.Method == "POST" {
```
technosophos: What is using the POST request? I can't find a place here where something is sending to this URL, and I'm concerned about TCP backpressure if too much traffic is submitted over this POST handler, since the client blocks until the request is complete.

technosophos:
I am enthusiastically in favor of this PR. I left a few questions and a few nit-picky comments. I would strongly prefer to at least change the package name to `handler`, not `handlers`. I'd also encourage adding Go-style comments to every public struct and function. I think we've gotten a little remiss on that one.

I'm not totally sure what the purpose of the POST request on the HTTP service is for. Maybe I didn't read carefully enough. But because TCP and HTTP both introduce blocking latency, I wouldn't want to use this as the main way of putting logs into the ring. Otherwise, a high amount of log traffic could actually deny service not just here, but on the nodes that are sending log data to the logger (since they will have TCP sessions tied up).

Again, I'm enthusiastically in favor of this PR!

*Note:* I did not review the controller code, other than just giving it a quick glance to see what it did.

logger/handlers/ringbuffer.go
```diff
@@ -0,0 +1,134 @@
...
+		regex := regexp.MustCompile(`^/([-a-z0-9]+)/.*`)
```
technosophos: Since the regexp is fixed, we could probably precompile them, store them outside of this function, and re-use the matcher repeatedly. Same with other regexp compile.

logger/handlers/ringbuffer.go
```diff
@@ -0,0 +1,134 @@
...
+		r.Value = message
+		projectBuffer[name] = r
+	} else {
+		r := currentRing.Next()
+		r.Value = message
+		projectBuffer[name] = r
+	}
+}
+
+func RingBufferHandler(bufferSize int, webServicePort int) *Handler {
```
technosophos: Rings are not goroutine-safe, are they? If we have one goroutine reading and another writing, we could end up with some weird race conditions. Have you run this with `go run -race -cpu=2`? Some unit tests would be helpful here.

logger/handlers/ringbuffer.go
```diff
@@ -0,0 +1,134 @@
...
+
+	"github.com/deis/deis/logger/drain"
+	"github.com/deis/deis/logger/syslog"
+)
+
+var projectBuffer = make(map[string]*ring.Ring)
+var ringBufferSize int
+
+func addToBuffer(name string, message string) {
+	currentRing, ok := projectBuffer[name]
```
technosophos: I think there may be a race condition here, too. It is possible for projectBuffer to be read/written-to out of order. (viz. A reads, ok = false. B reads, ok= false. A creates a new ring ring and adds to buffer. B creates new ring and adds to buffer. A's ring is lost.

logger/handlers/ringbuffer.go
rvadim: Nothing happening. Ring.Do() stop working first. If you request for 10000 lines, but buffer configured for 1000 items and store only 10 items - only 10 lines will be returned.

logger/handlers/ringbuffer.go
rvadim: I think cancel of ringBufferLoop needed for example in case of changing options or changing working mode of working w/o restarts. But i don't known any priority user scenario for this.
In case of cephless cluster user(admin) may want to change size of buffer in real time, but i think importance of this scenario is low and implementation of this functionality will have more complexity, then i would like in this PR.

logger/handlers/ringbuffer.go
rvadim: POST implemented for controller:
https://github.com/deis/deis/pull/4083/files?diff=unified#diff-211dbf56ab4ba80bf5d2b2a028b749daR195

Only controller on requests like 'deis run' or 'deis push' write some data to logs directly(through POST). So i think this is very low amount of http traffic to logger, in case of hosting for 100 - 1000 applications i think.

rvadim:
@technosophos thanks for feedback. For now i was fix some issues and add some docs. In Saturday i will try to write some tests.
Access to map of ring buffers have been rewritten for now to make it thread-safe.

Please notify me here, if i forgot something.

technosophos:
@rvadim I was thinking that a `sync.RWMutex` would solve the synchronization issue, but I think your solution should work, too.

I am going to mark this LGTM so we can move it toward a merge into master. Feel free to keep working, and I'll keep an eye on the issue.

Thanks again! This is something we've been talking about adding for a long time.

rvadim:
Can someone help me with tests? There are some questions:
* In logger_test.go only checks for publishing entry points to etcd. I don't known what tests needed here. Maybe switch mod and parse stdout?
* In deis/tests catalog there are different functional tests, but most of them can be run on in memory logger storage. So maybe we need to create stateless cluster configuration for tests, and run it on demand?

technosophos:
I think we'll try to merge as-is and then see what we can do about adding tests.

rvadim:
All tests failed due to yesterday docker hub maintenance

rvadim:
All tests failed due to yesterday docker hub maintenance

technosophos:
I'm tracking a copy of this in my repo now, doing some testing there. I'm going to try to write some unit tests for the ringbuffer.
https://github.com/technosophos/deis/tree/rvadim-stateless-logger

technosophos:
Alright, so I added a simple unit test to see if the ringbuffer is working. But running it fails. It's possible that the test was set up incorrectly, or that there's a data race or some other bug.

Here's the branch:
https://github.com/technosophos/deis/tree/rvadim-stateless-logger/logger/handler

Here's the commit:
https://github.com/technosophos/deis/commit/5068d60217cb4faa7f23b352be2703d139ae844c
(Note that almost all of the changes were done automatically by `gofmt`. I only fixed a small spelling error and then added the test.)

Here's the output of the test:

```
go test ./...
Web service started on 1031 port
--- FAIL: TestRingBufferHandler (0.01s)
	ringbuffer_test.go:41: Expected 'junk test[1] Test 1
		junk test[2] Test 2
		junk test[3] Test 3
		junk test[4] Test 4', Got 'junk test[1] Test 1
		junk test[3] Test 3
		junk test[3] Test 3
		junk test[4] Test 4
		'
FAIL
FAIL	github.com/deis/deis/logger/handler	0.019s
```

rvadim:
Hello, i will try to check code and tests in next few days, thank you.

carmstrong:
> Hello, i will try to check code and tests in next few days, thank you.

Thanks @rvadim! Much appreciated.

rvadim:
Problem was in writing pointers to channels. I rework this part to sync.Mutex

logger/handler/ringbuffer.go
```diff
@@ -0,0 +1,149 @@
...
+	"github.com/deis/deis/logger/syslog"
+)
+
+var logStorage = make(map[string]*ring.Ring)
+var ringBufferSize int
+
+var regexpForPost = regexp.MustCompile(`^/([-a-z0-9]+)/.*`)
+var regexpForGet = regexp.MustCompile(`^/([-a-z0-9]+)/([0-9]+)/.*`)
+
+var mu sync.Mutex
```
kalbasit: replace this section with

```
var (
  logStorage = make(map[string]*ring.Ring)
  ringBufferSize int

  regexpForPost = regexp.MustCompile(`^/([-a-z0-9]+)/.*`)
  regexpForGet = regexp.MustCompile(`^/([-a-z0-9]+)/([0-9]+)/.*`)

  mu sync.Mutex
)

File: logger/handler/ringbuffer.go [hunk 10]
```diff
@@ -0,0 +1,149 @@
...
+// Add log message to main map with ring byffers by project name
+func addToStorage(name string, message string) {
+	mu.Lock()
+	defer mu.Unlock()
+	currentRing, ok := logStorage[name]
+	if !ok {
+		r := ring.New(ringBufferSize)
+		r.Value = message
+		logStorage[name] = r
+	} else {
```
kalbasit: Add a return statement to the end of the `if !ok { }` block then drop the else statement in here.

logger/handler/ringbuffer.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+}
+
+// Get specific amount of log lines for application name from main map with ring buffers
+func getFromStorage(name string, lines int) string {
+	mu.Lock()
+	defer mu.Unlock()
+	currentRing, ok := logStorage[name]
+	if !ok {
+		return fmt.Sprintf("Could not find logs for project '%s'", name)
```
kalbasit: This function must return an error, it feels that you are returning the error as the value format.

logger/handler/ringbuffer.go
```diff
@@ -0,0 +1,149 @@
...
+}
+
+func RingBufferHandler(bufferSize int, webServicePort int) *Handler {
+	ringBufferSize = bufferSize
+	h := Handler{
+		BaseHandler: syslog.NewBaseHandler(5, filter, false),
+	}
+	go h.ringBufferLoop()
+	go startWebService(webServicePort)
+	fmt.Printf("Web service started on %d port\n", webServicePort)
```
kalbasit: `log` instead of `fmt`? @bacongobbler ?

logger/handler/ringbuffer.go
```diff
@@ -0,0 +1,149 @@
...
+	return &h
+}
+
+// Main loop for this handler, each log line will be sended to drain (if DrainURI specified) and copied to log storage
+func (h *Handler) ringBufferLoop() {
+	var err error
+	var projectName string
+	var message string
```
kalbasit: you don't need any of these declared as they are overwritten on each loop. remove them and add the `:=` to where they are being assigned to below.

logger/handler/ringbuffer.go
```diff
@@ -0,0 +1,149 @@
...
+	h.End()
+}
+
+// Only one http handler which process requests
+func httpHandler(w http.ResponseWriter, r *http.Request) {
+	if r.Method == "POST" {
+		match := regexpForPost.FindStringSubmatch(r.RequestURI)
+
+		if match == nil {
+			fmt.Fprintf(w, "Could not get application name from url: %s", r.RequestURI)
```
kalbasit: log as well?

logger/handler/ringbuffer.go
```diff
@@ -0,0 +1,149 @@
...
+
+	logLines, err := strconv.Atoi(match[2])
+	if err != nil {
+		fmt.Fprintln(w, "Unable to get log lines parameter from request")
+		return
+	}
+
+	getFromStorage(match[1], logLines)
+
+	if err != nil {
```
kalbasit: this `err` is always `nil` here. the `err` is being assigned a value on line 127 after which you are returning immediately if it's not a nil. so at this point it's always nil.

logger/handler/ringbuffer.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+}
+
+// Start web service which serve controller request for get and post logs
+func startWebService(port int) {
+	http.HandleFunc("/", httpHandler)
+	err := http.ListenAndServe(fmt.Sprintf(":%d", port), nil)
+	if err != nil {
+		log.Println(err)
+	}
```
kalbasit: As a rule of thumb if you are assigning only to an `err` and you are using it only in a if statement, do the assignment in the if statement.

```
	if err := http.ListenAndServe(fmt.Sprintf(":%d", port), nil); err != nil {
		log.Println(err)
	}
```

logger/handler/ringbuffer_test.go
```diff
@@ -0,0 +1,43 @@
...
+	"testing"
+	"time"
+
+	"github.com/deis/deis/logger/syslog"
+)
+
+func TestGetProjectName(t *testing.T) {
+	pname, err := getProjectName("junk test[junk] junk junk junk")
+	if err != nil {
+		t.Errorf("Failed to get project name: %s", err)
```
kalbasit: please remove capital letters from the error messages..

logger/handler/ringbuffer_test.go
```diff
@@ -0,0 +1,43 @@
...
+	}
+
+	if pname != "test" {
+		t.Errorf("Expected project name 'test', got '%s'", pname)
+	}
+
+}
```
kalbasit: maybe this is better re-written like so (more idiomatic when it comes to testing style)

```
func TestGetProjectName(t *testing.T) {
	testStr := "junk test[junk] junk junk junk"
	got, err := getProjectName(testStr)
	if err != nil {
		t.Fatalf("getProjectName(%q) got an expected error: %s", testStr, err)
	}

	if want := "test"; want != got {
		t.Errorf("getProjectName(%q): want %s got %s", want, got)
	}
}
```

So basically, there's no need to continue if there was an error, also the format of the message is far more interesting if it's function(args) and return. Also I know this isn't really Go style specific but I personally prefer not to see empty line right after `{` or right before `}` feels like a wasted space on my screen :trollface:

logger/handler/ringbuffer.go
rvadim: log is a bad idea, due to this is logger component, and when i try using log, messages loops infinite time.

logger/handler/ringbuffer_test.go
rvadim: This file i was commit by mistake.

logger/handler/ringbuffer.go
kalbasit: Oh I see, good point.

aj-may:
We will need to add the stateless-logger unit file to the deisctl commands.

https://github.com/rvadim/deis/pull/1

krancour:
@aj-may I don't believe that's necessary.  A separate command shouldn't be needed because there isn't a separate Fleet unit for the stateless logger.  There's just _one_ `deis-logger` which can simply be configured via etcd to use one log handler or another-- e.g. use the stateless log handler.

aj-may:
@krancour part of this PR is a stateless-logger unit file (https://github.com/deis/deis/pull/4083/files#diff-04e525eab27a07a07290e8d42aa06717R1), which is the same as the logger unit file except it doesn't require the deis-store units.  We can possible just update the logger unit file from require to wants.  Thoughts?

carmstrong:
> We can possible just update the logger unit file from require to wants.

:+1:

aj-may:
@krancour @carmstrong Here is the second option

https://github.com/rvadim/deis/pull/2

krancour:
> We can possible just update the logger unit file from require to wants.

Also :+1: 

@rvadim consider accepting @aj-may's PR so that commit gets bundled in with yours.

aj-may:
Ideally, shouldn't api logs always be passed through the logger component, regardless if the cluster is stateless or not?

krancour:
> shouldn't api logs always be passed through the logger component

The issue raised by #3963 isn't really about logger.  It's about how app events logged by the controller get _to_ the logger-- assuming the logger is even in play.

I've been working on updating logspout (which is not to be confused with the logger) as well as providing a plugin for it that adapts it to Deis.  I'll be sure to address #3963, but IMO, this PR shouldn't be dependent on that issue being solved first.

aj-may:
@krancour fair enough

technosophos:
@krancour You should probably be the owner of this ticket now, since I'm way out in other-stuff-world right now, and you're way more familiar than I am.

krancour:
@technosophos I'm cool with that.  It's mine now.

controller/api/models.py
```diff
@@ -526,6 +567,21 @@ def logs(self, log_lines=str(settings.LOG_LINES)):
         data = subprocess.check_output(['tail', '-n', log_lines, path])
         return data
 
+    def logs(self, log_lines=str(settings.LOG_LINES)):
+        if self._get_logging_mode() == 'standard':
```
aj-may: Is there a reason why we would want to get logs from the file system over fetching them from the logger component?

Why wouldn't we just always fetch from the logger?

bacongobbler: it's how deis works today (logs are mounted into the controller from cephfs)

aj-may: Yes, I guess my question is if there is any reason why we shouldn't get the logs from the logger component wither ceph is available or not?  It would make the controller logic more simple.

bacongobbler: not really, though that's out of scope for this PR. I think this already exists as a proposal/issue at https://github.com/deis/deis/issues/3963 to fix this hack.

aj-may: ok

aj-may:
LGTM

When can this be released?  We can really use this feature.

krancour:
@aj-may this still needs to go through our code review process and speaking for myself, besides reviewing the code, I also like to personally take these things for a test drive before I consider LGTM'ing them.  We need LGTMs from two core maintainers before anything can be merged into master.  Even once that's done, minor/feature releases are monthly.

If you want this sooner and you're willing to take on the risk, there is no reason you can't build the controller and logger components from source in this branch and deploy them to your cluster now.

aj-may:
@krancour I'll give it a shot.

rvadim:
JFI: We use this logger on production cluster with 3 apps, each load 1 to 10 RPS, in case of daytime for last 2 weeks. Also we have dev. cluster with 20+ apps for staging and inner projects in every day use.

controller/api/models.py
```diff
@@ -518,7 +541,25 @@ def _default_scale(self, user, release):
...
+                self._etcd_client = etcd.Client(host=settings.ETCD_HOST,
+                                                port=int(settings.ETCD_PORT))
+            except etcd.EtcdException:
+                logger.log(logging.WARNING,
+                           'Cannot synchronize with etcd cluster')
+        return self._etcd_client
+
+    def _get_logging_mode(self):
+        try:
+            return self._get_etcd_client().get('/deis/logs/handlertype').value
```
krancour: To match conventions used elsewhere in the platform, this key should use lower-camel-case-- i.e. the key should be `/deis/logs/handlerType`

krancour: Obviously that applies to all uses of that key throughout the code and doc.

controller/api/views.py
```diff
@@ -207,7 +207,8 @@ def logs(self, request, **kwargs):
         try:
             return Response(app.logs(request.query_params.get('log_lines',
                                      str(settings.LOG_LINES))),
-                            status=status.HTTP_200_OK, content_type='text/plain')
+                            status=status.HTTP_200_OK,
```
krancour: We should avoid the temptation to apply formatting fixes in this sort of PR.  If formatting fixes to existing code are needed, they should be applied in their own separate chore PR.

controller/api/models.py
krancour: Also, I'd like to see some of this code simplified and even eliminated.  confd already runs in the background and is used to set values for a number of constants in `settings.py` whenever relevant values in etcd change.  Upon changing, the gunicorn master process for the controller is signaled to reload configuration.  You should take advantage of this existing process to detect configuration changes for "free" without needing to add extra etcd-related code to `models.py`.

bacongobbler: ideally we should be reaching out to a setting like it's done [here](https://github.com/deis/deis/blob/master/controller/api/models.py#L95) and [here](https://github.com/deis/deis/blob/master/controller/deis/settings.py#L292), which can be templated by confd [here](https://github.com/deis/deis/blob/master/controller/templates/confd_settings.py). That way we can extract this bit away so we're not reliant as much on etcd if we were to migrate to something else.

logger/handler/ringbuffer.go
krancour: I will look into that a little more.  An infinite loop when the logger writes a log message isn't what _should_ be happening.  (I'm not saying it can't happen-- in fact, I think I've seen it.  Only saying it _shouldn't_.)  When the logger writes log messages, they should be going to `STDOUT` and as in the case of all other containers, the `deis-logspout` agent running on the host `deis-logger` runs on should see that message and forward it _to_ `deis-logger`.  `deis-logger` will process that messages (that originated from itself), but it doesn't output (re-output in this case) log messages that is receives... the message should be treated just as is a message from any other component or app.

All that being said, there may be an opportunity to short-circuit that process because it's really inefficient.

controller/api/models.py
```diff
@@ -216,6 +218,17 @@ def _get_command(self, container_type):
             return '' if container_type == 'cmd' else 'start {}'.format(container_type)
 
     def log(self, message):
+        if self._get_logging_mode() == 'standard':
```
krancour: It's probably not worth changing how the controller writes application event logs (scaling events, for instance) since that is / will be addressed in #3963 and #4347 

The only real change to the controller may be the ability to _retrieve_ logs from the logger via HTTP-- no need to write them via HTTP.

logger/handler/ringbuffer.go
```diff
@@ -0,0 +1,146 @@
...
+		fmt.Fprintln(w, err)
+	} else {
+		fmt.Fprint(w, data)
+	}
+}
+
+// Start web service which serve controller request for get and post logs
+func startWebService(port int) {
+	http.HandleFunc("/", httpHandler)
+	if err := http.ListenAndServe(fmt.Sprintf(":%d", port), nil); err != nil {
```
krancour: Consistent with some other comments I made (and I think @aj-may was trying to get at this earlier and @bacongobbler and I were misunderstanding him), let's _always_ serve logs from the logger over HTTP-- not _just_ when the ring buffer is in use.

By doing this, controller can be updated to _always_ just get the logs from the logger via HTTP and not need to concern itself at all with which log handler is in use.

krancour: i.e. The logger will be the component that knows whether to fetch the requested logs from memory or from disk.  The controller won't know or care.

aj-may: correct

controller/api/models.py
rvadim: Yes, got it. I will rework controller in next few days.

aj-may:
@krancour do we now want to start the logger with stateless platform?

https://github.com/rvadim/deis/pull/6

aj-may:
this should get the build to pass:
https://github.com/rvadim/deis/pull/7

krancour:
> do we now want to start the logger with stateless platform?

Yes.

krancour:
Replaced by #4435 

Need to get this across the finish line for v1.11.0.

Credit to the original authors is preserved in the commit history.

PR #4084, opened 2015-07-22
chore(*): use etcd2 inside a container

mboersma:
This PR cherry-picks the essential `etcd2` changes from #4024 to make testing and reviewing easier. All credit to @aledbf.


aledbf:
Code LGTM.

contrib/coreos/user-data.example
```diff
@@ -19,7 +9,48 @@ coreos:
...
+      LimitNOFILE=40000
+      EnvironmentFile=/etc/environment
+      Environment="ETCD_VERSION=v2.0.13"
+      Environment="ETCD_ELECTION_TIMEOUT=2000"
+      Environment="ETCD_HEARTBEAT_INTERVAL=400"
+      Environment="ETCD_DATA_DIR=/var/lib/etcd2"
+      Environment="ETCD_NAME=%m"
+      Environment="DOCKER_HOST=unix:///var/run/early-docker.sock"
+      ExecStartPre=-/usr/bin/docker kill etcd2
+      ExecStartPre=-/usr/bin/docker rm etcd2
```
carmstrong: Is it necessary to rm and re-pull the container every time the service is started (i.e. on each reboot)?

aledbf: > Is it necessary to rm and re-pull the container every time the service is started (i.e. on each reboot)?

No, it can follow the same pattern that the rest of the deis services. Sorry for the change in the format.
```
ExecStartPre=/bin/sh -c "ETCD_IMAGE=quay.io/coreos/etcd:$ETCD_VERSION && docker history $ETCD_IMAGE >/dev/null 2>&1 || docker pull $ETCD_IMAGE"
ExecStartPre=/bin/sh -c "docker inspect etcd2 >/dev/null 2>&1 && docker rm -f etcd2 || true"
```

mboersma: > Is it necessary to rm and re-pull the container every time...?

I don't see a reason to. `ETCD_VERSION` isn't really dynamic, and this always hits quay.io. @aledbf am I missing the reason for something? Could we remove the `docker rm`, and could the ExecStartPre be:
`IMAGE=quay.io/coreos/etcd:${ETCD_VERSION} docker history $IMAGE >/dev/null 2>&1 || docker pull $IMAGE`

krancour:
What is the upgrade path for this?  `deis-logspout` and `deis-publisher` units now explicitly require `etcd2.service`.  cloud-config for nodes in an existing cluster would not have set that up.

mboersma:
> What is the upgrade path for this?

An excellent question I was just puzzling over. systemd doesn't give us the tools to say "requires one or the other" AFAICTL...

aledbf:
I think that using `RequiresOverridable=etcd.service etcd2.service` instead of `Requires` help with that

http://www.freedesktop.org/software/systemd/man/systemd.unit.html#RequiresOverridable=

krancour:
@mboersma is it deliberate that the cherry-picking doesn't include upgrading etcdctl on each `deis-x` image to 2.0.x?

mboersma:
I tested upgrade from v1.8.0 to this PR after adding `RequireOverride=` sections as suggested. This appears to be backward-compatible.

mboersma:
> doesn't include upgrading etcdctl

Keep it simple? I didn't want to change more than was strictly necessary, but I can upgrade `etcdctl` in the containers if there's a good reason to.

carmstrong:
> is it deliberate that the cherry-picking doesn't include upgrading etcdctl on each deis-x image to 2.0.x?

That would certainly break backwards-compatibility, no? Users on etcd 0.4 would upgrade components to 1.9 and they wouldn't be able to connect to etcd. See https://github.com/deis/deis/pull/4024#issuecomment-123406457

krancour:
> Users on etcd 0.4 would upgrade components to 1.9 and they wouldn't be able to connect to etcd.

So etcdctl 2.0.x strictly uses the IANA ports?

carmstrong:
> So etcdctl 2.0.x strictly uses the IANA ports?

Not 100%, but given backwards-incompatibility issues in the past, we're probably best keeping them in-sync?

krancour:
Also wanted to add I'm not totally in favor of the `RequireOverride=` hack.  As I understand it, CoreOS, by convention, runs etcd 2.0.x as `etcd2.service` so that users can (easily) choose to run either that or legacy 0.4.x-- which is simple `etcd.service.`  We could choose, near-term, to defy that convention and simply run our containerized etcd 2.0.x _as_ `etcd.service`, thereby requiring very few other changes-- some of which would have the potential to complicate the upgrade path as noted in some previous comments.  When we hit Deis 2.0 and breaking changes are expected, we could step back in line with the convention of running etcd 2.0.x _as_ `etcd2.service` at that time.

krancour:
> we're probably best keeping them in-sync

Hard to say what "in-sync" is when users could be running 0.4.x _or_ 2.0.x.  :)

My vote would be that the 0.4.x client appears to work with both old and new versions of the server, we probably shouldn't mess with it.

carmstrong:
> My vote would be that the 0.4.x client appears to work with both old and new versions of the server, we probably shouldn't mess with it.

That's what I was trying to say :) That's why I think we skip upgrading the clients for now.

mboersma:
The components use v0.4.9 since it's the latest of the v0.x series, and has the /v2/migration/snapshot endpoint if needed. I agree that's the safe choice for now, and leaving the pre-IANA ports in use. The components can be upgraded and the old ports purged in a future major release.

krancour:
Ok.  I'm with ya.  I just noticed that client upgrades were included in #4024 but not here, so just wanted to ask.

krancour:
Thoughts on naming the containerized etcd 2.0.x as `etcd.service` to avoid having to tamper with the systemd dependency graph to avoid breaking changes?  It actually reduces the size and scope of this PR by quite a bit as well.

mboersma:
> simply run our containerized etcd 2.0.x as `etcd.service`

That could make this a simpler diff, since we are deferring updating `etcdctl` and requiring only the `etcd2` service. But it ignores CoreOS' convention for the etcd -> etcd2 transition. So I'm on the fence.

krancour:
> But it ignores CoreOS' convention for the etcd -> etcd2 transition. So I'm on the fence.

I'm normally reticent to defy conventions as well, but I think the reasons for doing so in this case are defensible.  And I'd be 100% in favor of getting back on board with that convention when Deis reaches its own 2.0-- when breaking changes are to be expected.

mboersma:
Closed in favor of #4090.

PR #4087, opened 2015-07-22
fix(mesos): build images with prefix and tag

mboersma:
Ensure that the mesos-* and zookeeper images honor the same env vars used in other "make build" targets, and add the `DEIS_RELEASE` variable.

The `IMAGE_PREFIX` variable contains a trailing slash, so I used `@` as an alternate separator for `sed`.

ping @smothiki


carmstrong:
Code LGTM.

smothiki:
Code LGTM

PR #4088, opened 2015-07-22
style(mesos): check for static go binaries

mboersma:
The mesos binaries are built as static, so this adds the silent double-check that is done in all other cases.


carmstrong:
Code LGTM.

bacongobbler:
code LGTM

PR #4090, opened 2015-07-22
chore(*): use etcd2 inside a container, extended dance remix

mboersma:
This PR does the inverse of #4084 by using `etcd.service` to host etcd2 and masking `etcd2.service` as suggested by @krancour. This makes the final diff smaller and avoids changing any systemd dependencies, but deviates from CoreOS' service names.

Comments welcome.

### DIFF
--- a/contrib/aws/gen-json.py
+++ b/contrib/aws/gen-json.py
@@ -73,7 +73,7 @@ PREPARE_ETCD_DATA_DIRECTORY = '''
   Description=Prepares the etcd data directory
   Requires=media-etcd.mount
   After=media-etcd.mount
-  Before=etcd2.service
+  Before=etcd.service
   [Service]
   Type=oneshot
   RemainAfterExit=yes
--- a/contrib/aws/gen-json.py
+++ b/contrib/aws/gen-json.py
@@ -73,7 +73,7 @@ PREPARE_ETCD_DATA_DIRECTORY = '''
   Description=Prepares the etcd data directory
   Requires=media-etcd.mount
   After=media-etcd.mount
-  Before=etcd.service
+  Before=etcd2.service
   [Service]
   Type=oneshot
   RemainAfterExit=yes
@@ -101,11 +101,12 @@ data = yaml.load(file(os.path.join(CURR_DIR, '..', 'coreos', 'user-data'), 'r'))
 # are started
 data['coreos']['units'] = new_units + data['coreos']['units']
 
-# configure etcd to use its EBS volume
-data['coreos']['etcd']['data-dir'] = '/media/etcd'
-
 header = ["#cloud-config", "---"]
 dump = yaml.dump(data, default_flow_style=False)
+
+# configure etcd to use its EBS volume
+dump = dump.replace('-data-dir /var/lib/etcd2', '-data-dir /media/etcd')
+dump = dump.replace('--volume=/var/lib/etcd2', '--volume=/media/etcd')
 
 template = json.load(open(os.path.join(CURR_DIR, 'deis.template.json'), 'r'))
 
--- a/contrib/aws/provision-aws-cluster.sh
+++ b/contrib/aws/provision-aws-cluster.sh
@@ -136,17 +136,29 @@ ELB_NAME=$(aws elb describe-load-balancers \
     $EXTRA_AWS_CLI_ARGS | grep -F $ELB_DNS_NAME | head -n1 | cut -f2)
 echo "Using ELB $ELB_NAME at $ELB_DNS_NAME"
 
-echo_green "Your Deis cluster has been successfully deployed to AWS CloudFormation and is started."
-echo_green "Please continue to follow the instructions in the documentation."
-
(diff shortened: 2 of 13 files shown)


krancour:
+1.  Apart from my crazy suggestion, all credit belong to @aledbf

mboersma:
I tested upgrade from v1.8.0 to this PR as I did for #4084: this also appears to be backward-compatible. Deis components all run whether it's actually etcd or etcd2 behind `etcd.service`.

krancour:
I took the upgrade process for a spin with good results as well.

LGTM

contrib/aws/gen-json.py
```diff
@@ -87,12 +87,13 @@
...
 data['coreos']['units'] = new_units + data['coreos']['units']
 
-# configure etcd to use its EBS volume
-data['coreos']['etcd']['data-dir'] = '/media/etcd'
-
 header = ["#cloud-config", "---"]
 dump = yaml.dump(data, default_flow_style=False)
 
+# configure etcd to use its EBS volume
+dump = dump.replace('-data-dir /var/lib/etcd2', '-data-dir /media/etcd')
```
bacongobbler: can't we use the old code for this? That would make this one less change :)

contrib/aws/provision-aws-cluster.sh
```diff
@@ -147,6 +147,21 @@ FIRST_INSTANCE=$(aws ec2 describe-instances \
 export DEISCTL_TUNNEL=$FIRST_INSTANCE
 echo_green "Enabling proxy protocol"
```
bacongobbler: should we move this after the etcd2 check?

contrib/coreos/user-data.example
```diff
@@ -20,6 +10,46 @@ coreos:
   units:
   - name: etcd.service
     command: start
+    content: |
```
bacongobbler: If only there was a way we could deploy this with `deisctl`... We'll never be able to tune this in future releases, which could be a concern.

contrib/aws/gen-json.py
aledbf: > can't we use the old code for this?

no because the values this is inside the ExecStart section of the service. In the actual version is possible because is possible to get using the parsed yaml file

mboersma:
I moved the `deisctl list` loop before "enabling proxy protocol" as @bacongobbler suggested (in both PRs).

I slightly prefer this approach--using `etcd.service` to host etcd2--because I think we may want to use CoreOS' `etcd2.service` as-is in future versions. But I'm still undecided, so please help me out by being opinionated. It would be great if we could merge this or #4084 today.

bacongobbler:
I vote for this pr

krancour:
> because I think we may want to use CoreOS' etcd2.service as-is in future versions

:+1:

contrib/aws/gen-json.py
```diff
@@ -101,9 +101,6 @@
 # are started
 data['coreos']['units'] = new_units + data['coreos']['units']
 
-# configure etcd to use its EBS volume
-data['coreos']['etcd']['data-dir'] = '/media/etcd'
```
carmstrong: Are we setting this somewhere else? We still need to use the volume mounted at `/media/etcd` for etcd, even if it's bind-mounted into the container.

aledbf: @carmstrong yes, here https://github.com/deis/deis/pull/4024/files#diff-7b5fafac1ab3baf928f663c2b0a1a445R93

@mboersma I think that is one of the commits not selected

mboersma: Good point, I'll grab that.

contrib/coreos/user-data.example
```diff
@@ -20,6 +10,46 @@ coreos:
...
+      LimitNOFILE=40000
+      EnvironmentFile=/etc/environment
+      Environment="ETCD_IMAGE=quay.io/coreos/etcd:v2.0.13"
+      Environment="ETCD_ELECTION_TIMEOUT=2000"
+      Environment="ETCD_HEARTBEAT_INTERVAL=400"
+      Environment="ETCD_DATA_DIR=/var/lib/etcd2"
+      Environment="ETCD_NAME=%m"
+      Environment="DOCKER_HOST=unix:///var/run/early-docker.sock"
+      ExecStartPre=/bin/sh -c "docker history $ETCD_IMAGE >/dev/null 2>&1 || docker pull $ETCD_IMAGE"
+      ExecStartPre=/bin/sh -c "docker inspect $ETCD_NAME >/dev/null 2>&1 && docker rm -f $ETCD_NAME || true"
```
carmstrong: Should we remove this line in favor of `docker run --rm` as we did in #3918?

mboersma: I don't think so--we still have the same "docker inspect || docker rm -f" line in the `ExecStartPre` of other services after #3918, and we use `--rm` with `docker run` just below this line.

carmstrong: Oops, missed it - sorry!

carmstrong:
Code LGTM.

PR #4091, opened 2015-07-22
feat(client-go): add perms endpoint

Joshua-Anderson:


mboersma:
Also needs rebasing. :circus_tent:

Joshua-Anderson:
Rebased!

technosophos:
Code LGTM

client-go/parser/perms.go
```diff
@@ -0,0 +1,129 @@
...
+	if err != nil {
+		return err
+	}
+
+	admin := args["--admin"].(bool)
+
+	return cmd.PermsList(safeGetValue(args, "--app"), admin)
+}
+
+func permCreate(argv []string) error {
```
bacongobbler: the function names seem inconsistent here. To match with the call, let's make this `permsCreate()`

client-go/parser/perms.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	app := safeGetValue(args, "--app")
+	username := args["<username>"].(string)
+	admin := args["--admin"].(bool)
+
+	return cmd.PermCreate(app, username, admin)
+}
+
+func permDelete(argv []string) error {
```
bacongobbler: same here; `permsDelete()`

bacongobbler:
small nitpicks, otherwise LGTM

client-go/parser/perms.go
Joshua-Anderson: For my code I went with singular versions when dealing with a single object (creating, removing, getting) and plural with commands that deal with multiple objects (listing). I'm happy to change if it is confusing, but is probably a client-go wide change.

bacongobbler: nah as long as we're consistent

Issue #4033, opened 2015-07-15
Automate upgrade test


sgoings:
We'd like to add the ability test the deis upgrade process in an automated fashion. This process currently is:

- spin up the previous release of deis
- deploy an app to it
- verify that the app is running properly (smoke test)
- tear down deis around the app
- deploy the current version of deis
- verify that the app is running properly (smoke test)

Ideally this test should be able to run on all providers, but for the first round of implementation I'm only looking to support Vagrant and EC2.

carmstrong:
> Ideally this test should be able to run on all providers,

I think EC2, Azure, and GCE are ones we should shoot for. Other providers don't have separate volumes and etcd would almost certainly fall to its knees under heavy IO. See https://github.com/deis/deis/issues/3648#issuecomment-121672023

PR #4097, opened 2015-07-23
test(*): add simple upgrade test

sgoings:
In order to introduces better bash code reuse, I've brought in [rerun](http://rerun.github.io/rerun/) and implemented:

- simple upgrade test (see: #4033) (runs `make test-smoke` after a deis install/start of <old version> and then checks to make sure that the same app is responding after an in place Deis upgrade)
- simple provision test (runs `make test-smoke` after a deis install/start)

The coolest bit about this is that you can use rerun and plug different providers in on the command line:

```
./rerun.sh tests:upgrade --from <version/git tag/hash/branch of Deis> [ --to <version/git tag/hash/branch of Deis> --provider [vagrant, aws] --skip-cleanup <true|false> ]
```

I've also introduced a `provision` command, which does just a provision on a provider:

```
./rerun.sh tests:provision --version <version/git tag/hash/branch of Deis> [ --provider [vagrant, aws] --skip-cleanup <true|false> ]
```

If you take a look at the `tests/commands/[upgrade, provision]/script` files you'll notice that they're really quite easy to read and are intended to use composable parts of the process of provisioning a cluster, installing/starting deis, and then destroying the cluster. I took the general procedures laid out in all the test scripts and made their parts easier to share.

There are a few things I'd like to improve about this new way of constructing acceptance tests, but I wanted to get some feedback while I had something working (for both Vagrant and AWS). I'm also doing this work with the other test-infrastructure issues and our goal for making provisioning easier on other providers.


sgoings:
Closing temporarily so I don't fill up Jenkins' queue with PR builds.

PR #4099, opened 2015-07-23
feature(deisctl) allow for graceful in-place upgrades

iancoffey:
First pass at adding upgrade functionality to deisctl.

The whole thing needs :eyes:, but the following things especially.
- [ ] The service keys are persisted by a workaround which resets them with a 30 minute TTL. Is this the best way to keep the services up while the publishers cycle?
- [x] TestUpgradeTakeover fails due to the etcd stuff. If we keep that stuff in there, how should it be tested, a mocked etcd endpoint?
- [x] It would be nice if the command names were better, but its the best Ive come up with so far.
- [x] The docs indicate a deisctl install command that does not yet work - https://github.com/deis/deis.io/pull/114

### DIFF
--- a/deisctl/client/client.go
+++ b/deisctl/client/client.go
@@ -28,6 +28,9 @@ type DeisCtlClient interface {
 	Status(argv []string) error
 	Stop(argv []string) error
 	Uninstall(argv []string) error
+	UpgradePrep(argv []string) error
+	UpgradeTakeover(argv []string) error
+	RollingRestart(argv []string) error
 }
 
 // Client uses a backend to implement the DeisCtlClient interface.
@@ -55,6 +58,49 @@ func NewClient(requestedBackend string) (*Client, error) {
 		return nil, errors.New("invalid backend")
 	}
 	return &Client{Backend: backend}, nil
+}
+
+// UpgradePrep prepares a running cluster to be upgraded
+func (c *Client) UpgradePrep(argv []string) error {
+	usage := `Prepare platform for graceful upgrade.
+
+Usage:
+  deisctl upgrade-prep [options]
+`
+	if _, err := docopt.Parse(usage, argv, true, "", false); err != nil {
+		return err
+	}
+
+	return cmd.UpgradePrep(c.Backend)
+}
+
+// UpgradeTakeover gracefully restarts a cluster prepared with upgrade-prep
+func (c *Client) UpgradeTakeover(argv []string) error {
+	usage := `Complete the upgrade of a prepped cluster.
+
+Usage:
+  deisctl upgrade-takeover [options]
+`
+	if _, err := docopt.Parse(usage, argv, true, "", false); err != nil {
+		return err
+	}
+
+	return cmd.UpgradeTakeover(c.Backend)
+}
+
+// RollingRestart attempts a rolling restart of an instance unit
+func (c *Client) RollingRestart(argv []string) error {
+	usage := `Perform a rolling restart of an instance unit.
+
+Usage:
+  deisctl rolling-restart <target>
+`
+	args, err := docopt.Parse(usage, argv, true, "", false)
+	if err != nil {
+		return err
+	}
+
+	return cmd.RollingRestart(args["<target>"].(string), c.Backend)
 }
 
 // Config gets or sets a configuration value from the cluster.
--- a/deisctl/cmd/cmd_test.go
+++ b/deisctl/cmd/cmd_test.go
@@ -13,6 +13,8 @@ import (
 	"testing"
 
 	"github.com/deis/deis/deisctl/backend"
+	"github.com/deis/deis/deisctl/etcdclient"
+	"github.com/deis/deis/deisctl/test/mock"
 	"github.com/deis/deis/deisctl/units"
 )
 
@@ -21,6 +23,7 @@ type backendStub struct {
 	stoppedUnits     []string
 	installedUnits   []string
 	uninstalledUnits []string
+	restartedUnits   []string
 	expected         bool
 }
 
@@ -46,6 +49,10 @@ func (backend *backendStub) Scale(component string, num int, wg *sync.WaitGroup,
 		backend.expected = false
 	}
 }
+func (backend *backendStub) RollingRestart(target string, wg *sync.WaitGroup, out, ew io.Writer) {
+	backend.restartedUnits = append(backend.restartedUnits, target)
+}
+
 func (backend *backendStub) ListUnits() error {
 	return nil
 }
@@ -267,6 +274,57 @@ func TestStartSwarm(t *testing.T) {
 
 	if !reflect.DeepEqual(b.startedUnits, expected) {
 		t.Error(fmt.Errorf("Expected %v, Got %v", expected, b.startedUnits))
+	}
+}
+
+func TestRollingRestart(t *testing.T) {
+	t.Parallel()
+
+	b := backendStub{}
+	expected := []string{"router"}
+
+	RollingRestart("router", &b)
+
+	if !reflect.DeepEqual(b.restartedUnits, expected) {
+		t.Error(fmt.Errorf("Expected %v, Got %v", expected, b.restartedUnits))
+	}
+}
+
+func TestUpgradePrep(t *testing.T) {
+	t.Parallel()
+
+	b := backendStub{}
+	expected := []string{"database", "registry@*", "controller", "builder", "logger", "logspout", "store-volume",
+		"store-gateway@*", "store-metadata", "store-daemon", "store-monitor"}
+
+	UpgradePrep(&b)
+
+	if !reflect.DeepEqual(b.stoppedUnits, expected) {
+		t.Error(fmt.Errorf("Expected %v, Got %v", expected, b.stoppedUnits))
+	}
+}
+
+func TestUpgradeTakeover(t *testing.T) {
+	t.Parallel()
+	testMock := mock.Client{Expected: []*etcdclient.ServiceKey{{Key: "/deis/services/app1", Value: "foo", TTL: 10},
+		{Key: "/deis/services/app2", Value: "8000", TTL: 10}}}
+
+	b := backendStub{}
+	expectedRestarted := []string{"router"}
+	expectedStarted := []string{"publisher", "store-monitor", "store-daemon", "store-metadata",
+		"store-gateway@*", "store-volume", "logger", "logspout", "database", "registry@*",
+		"controller", "builder", "publisher", "router@*", "database", "registry@*",
+		"controller", "builder", "publisher", "router@*"}
+
+	if err := doUpgradeTakeOver(&b, testMock); err != nil {
+		t.Error(fmt.Errorf("Takeover failed: %v", err))
+	}
+
+	if !reflect.DeepEqual(b.restartedUnits, expectedRestarted) {
+		t.Error(fmt.Errorf("Expected %v, Got %v", expectedRestarted, b.restartedUnits))
+	}
+	if !reflect.DeepEqual(b.startedUnits, expectedStarted) {
+		t.Error(fmt.Errorf("Expected %v, Got %v", expectedStarted, b.startedUnits))
 	}
 }
 
--- a/deisctl/cmd/upgrade.go
+++ b/deisctl/cmd/upgrade.go
@@ -1,0 +1,110 @@
+package cmd
+
+import (
+	"fmt"
+	"sync"
+
+	"github.com/deis/deis/deisctl/backend"
+	"github.com/deis/deis/deisctl/etcdclient"
+)
+
+// UpgradePrep stops and uninstalls all components except router and publisher
+func UpgradePrep(b backend.Backend) error {
+	var wg sync.WaitGroup
+
+	b.Stop([]string{"database", "registry@*", "controller", "builder", "logger", "logspout"}, &wg, Stdout, Stderr)
+	wg.Wait()
+	b.Destroy([]string{"database", "registry@*", "controller", "builder", "logger", "logspout"}, &wg, Stdout, Stderr)
+	wg.Wait()
+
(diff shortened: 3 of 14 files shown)


Joshua-Anderson:
> TestUpgradeTakeover fails due to the etcd stuff. If we keep that stuff in there, how should it be tested, a mocked etcd endpoint?

I mocked etcd [here](https://github.com/deis/deis/blob/master/deisctl/config/config_test.go), which may be a good starting point. I used an interface to access etcd and then mocked the interface for my tests.


Great Stuff!

deisctl/client/client.go
```diff
@@ -56,6 +59,50 @@ func NewClient(requestedBackend string) (*Client, error) {
...
+func (c *Client) UpgradePrep(argv []string) error {
+	usage := `Prepare platform for graceful upgrade.
+
+Usage:
+  deisctl upgrade-prep [options]
+`
+	_, err := docopt.Parse(usage, argv, true, "", false)
+	if err != nil {
+		return err
+	}
```
Joshua-Anderson: I think deis tries to go with the 

``` go
if _, err := foo(); err != nil {
  return err
}
```
syntax when possible, though admitingly I may have introduced some of this before I knew the coding style myself. :smiley:

deisctl/cmd/cmd_test.go
```diff
@@ -260,6 +265,55 @@ func TestStartSwarm(t *testing.T) {
...
+	t.Parallel()
+
+	b := backendStub{}
+	expectedRestarted := []string{"router"}
+	expectedStarted := []string{"database", "registry@*", "controller", "builder", "logger", "logspout", "store-volume",
+		"store-gateway@*", "store-metadata", "store-daemon", "store-monitor", "publisher", "router"}
+
+	err := UpgradeTakeover(&b)
+	if err != nil {
+		fmt.Printf("ERRERRERR %s", err)
```
carmstrong: Is this intentional, or leftover cruft from debugging?

deisctl/cmd/upgrade.go
```diff
@@ -0,0 +1,109 @@
...
+	wg.Wait()
+
+	b.Stop([]string{"store-monitor"}, &wg, Stdout, Stderr)
+	wg.Wait()
+	b.Destroy([]string{"store-monitor"}, &wg, Stdout, Stderr)
+	wg.Wait()
+
+	fmt.Fprintln(Stdout, "Backend services stopped, but your services should still be serving traffic as normal.")
+
+	fmt.Fprintln(Stdout, "Your cluster is now ready for uprade, use a new deisctl binary and run deisctl take-over to complete this upgrade.")
```
carmstrong: s/uprade/upgrade/

deisctl/cmd/cmd_test.go
iancoffey: oopsies :grin:

deisctl/deisctl.go
```diff
@@ -41,6 +41,9 @@ Commands, use "deisctl help <command>" to learn more:
   refresh-units     refresh unit files from GitHub
   ssh               open an interacive shell on a machine in the cluster
   help              show the help screen for a command
+  upgrade-prep      prepare a running cluster for upgrade
+  upgrade-takeover  allow an upgrade to gracefully takeover a running cluster
+  rolling-restart   perform a rolling restart for instance units
```
carmstrong: What's the meaning of "instance unit" here? The term is unfamiliar to me in this context.

iancoffey: The term I cribbed from systemds [docs](http://www.freedesktop.org/software/systemd/man/systemd.unit.html) , its just referring to units created like unit@1.service - so the router, registry and gateway are examples. Its intended to indicate that the restart could only function for units created this way - global or one off units couldn't work.

As for better wording - maybe it should say something more generic? "perform a rolling restart for certain valid Deis components" or something.

carmstrong: Yeah, maybe something like "perform a rolling restart of a Deis component (currently only router is supported)"

deisctl/cmd/upgrade.go
```diff
@@ -0,0 +1,109 @@
...
+	wg.Wait()
+	b.Destroy([]string{"store-daemon"}, &wg, Stdout, Stderr)
+	wg.Wait()
+
+	b.Stop([]string{"store-monitor"}, &wg, Stdout, Stderr)
+	wg.Wait()
+	b.Destroy([]string{"store-monitor"}, &wg, Stdout, Stderr)
+	wg.Wait()
+
+	fmt.Fprintln(Stdout, "Backend services stopped, but your services should still be serving traffic as normal.")
```
carmstrong: "The platform has been stopped, but applications are still serving traffic as normal.

deisctl/cmd/upgrade.go
carmstrong: "Your cluster is now ready for upgrade. Install a new deisctl version and run `deisctl take-over`. "

carmstrong: We should also link to the documentation section pertaining to Graceful Upgrades.

iancoffey:
I moved the etcd and etcd mock stuff to its own packages outside of config, since they are not specific to config and other things now need to use them. Im not sure what the best pattern for the namespacing and locations would be, would love input on that and everything else.

iancoffey:
Also, the mock.MockClient lines result in go vet complaining about "mock.MockStore composite literal uses unkeyed fields" - which doesnt appear to be reality to me. Maybe go vet confuses slice literal with omitted types? Im not sure how to resolve those fails.

carmstrong:
@iancoffey Looks like this needs a rebase.

mboersma:
> the mock.MockClient lines result in go vet complaining...

Yep, and there isn't a workaround AFAICT: golang/go#9171 :disappointed:

iancoffey:
fwiw - I tried the go vet workarounds we talked about (as far as I understood them) and had the same unfortunate result.

carmstrong:
Can we have `go vet` ignore that line?

iancoffey:
this I do not know - i havent found a nice way to do that yet without skipping all the structtags

iancoffey:
This should go green now :four_leaf_clover:

docs/managing_deis/upgrading-deis.rst
```diff
@@ -63,6 +63,56 @@ Finally, update ``deisctl`` to the new version and reinstall:
     enable PROXY protocol for ports 80 and 443 on the ELB, add a ``TCP 443:443`` listener, and
     change existing targets and health checks from HTTP to TCP.
 
+
+Graceful Upgrade
```
krancour: @iancoffey since Deis < 1.9.0 won't support `deisctl upgrade-prep` and 1.9.0 will if this is merged, the instructions documented here won't actually _work_ for anyone until upgrading from m >= 1.9.0 to n > m.

Would it make sense to _not_ include this documentation until the release after this feature is rolled out?  Otherwise we may have people attempting an upgrade procedure that is guaranteed to fail.

krancour: I wouldn't throw the docs away either-- just break them out into another PR that we can sit on for a while.

carmstrong: Good point. I think we could just add a note here stating that this can be used for 1.9.0 moving forward.

krancour: :+1:

krancour: Should it also be noted somewhere that this won't work across major versions either?

deisctl/config/api.go
```diff
@@ -1,8 +0,0 @@
-package config
```
krancour: @iancoffey I'm not sure about relocating this interface to the `etcdclient` package.  This interface looks like / was previously a generic "config client" interface of which many implementations are possible-- your new mock implementation of that being one of them.  But moving this interface to the `etcdclient` means users of the interface will be coding to an abstraction (the interface) whose package name is polluted with the name of a specific underlying implementation.

Was there a specific reason for moving it?

deisctl/cmd/upgrade.go
```diff
@@ -0,0 +1,110 @@
...
+	b.Destroy([]string{"store-daemon"}, &wg, Stdout, Stderr)
+	wg.Wait()
+
+	b.Stop([]string{"store-monitor"}, &wg, Stdout, Stderr)
+	wg.Wait()
+	b.Destroy([]string{"store-monitor"}, &wg, Stdout, Stderr)
+	wg.Wait()
+
+	fmt.Fprintln(Stdout, "The platform has been stopped, but applications are still serving traffic as normal.")
+	fmt.Fprintln(Stdout, "Your cluster is now ready for upgrade. Install a new deisctl version and run deisctl take-over.")
```
krancour: @iancoffey Doesn't match the actual command that does the take-over.  Should be `upgrade-takeover`.

Also, for maximum clarity and consistency, you might consider using backticks to demarcate the boundaries of the command you're asking them to run.  That's done elsewhere in deisctl.  As a for instance, this is the message provided after a `deisctl install platform` reads as follows:

```
Please run `deisctl start platform` to boot up Deis.
```

docs/managing_deis/upgrading-deis.rst
```diff
@@ -63,6 +63,56 @@ Finally, update ``deisctl`` to the new version and reinstall:
...
+----------------
+
+Alternatively, an experimental feature exists to provide the ability to perform a graceful upgrade. Unlike the
+in-place process above, this process keeps the platforms routers and publishers up during the upgrade process.
+This means that there should only be a maximum of around 1-2 seconds of downtime while the routers boot up. Many
```
krancour: > This means that there should only be a maximum of around 1-2 seconds of downtime while the routers boot up

Is this on account of how health checks from the load balancers are configured?  I can imagine a scenario where even with the rolling restarts, the timing of the health checks that determine a router is down and then back up could have all _n_ routers out of commission for a while.  This would happen if health checks were frequent and quick to consider a router down after a couple bad responses, but slow to consider it back up again by requiring some substantial number of healthy responses first.

deisctl/cmd/upgrade.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	if err := doUpgradeTakeOver(b, etcdClient); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func doUpgradeTakeOver(b backend.Backend, etcdClient etcdclient.Client) error {
```
krancour: @iancoffey should this function also update the value of `/deis/platform/version` in etcd?  Otherwise, won't the same versions of each component just come right back up (despite a newer version of deisctl having been used to do so)?

Referencing the inner-workings of the `get_image` function:

```
#!/bin/bash
# usage: get_image <component_path>
IMAGE=`etcdctl get $1/image 2>/dev/null`

# if no image was set in etcd, we use the default plus the release string
if [ $? -ne 0 ]; then
  RELEASE=`etcdctl get /deis/platform/version 2>/dev/null`

  # if no release was set in etcd, use the default provisioned with the server
  if [ $? -ne 0 ]; then
    source /etc/deis-release
    RELEASE=$DEIS_RELEASE
  fi

  IMAGE=$1:$RELEASE
fi

# remove leading slash
echo ${IMAGE#/}
```

Come to think of it, it needs to be noted that the upgrade won't upgrade any components for which a custom image has already been specified via etcd key `/deis/<component>/image`

krancour:
@iancoffey despite my nitpicking... I think this is great!  Locally, this worked just as advertised and the only downtime I experienced was when the router hosted on `172.17.8.100` recycled and that was to be expected since there's no LB in play and `*.local3.deisapp.com` just points to that IP.

mboersma:
This did pass despite the mixed checks indicated above: https://ci.deis.io/job/test-ec2/2417/

deisctl/cmd/upgrade.go
carmstrong: I think we ask the user to explicitly set the new platform version in the Graceful Upgrade docs.

iancoffey: Yeah, setting the version is part of the procedure. IMO thats fine for the time being.

krancour: I see that now.  If the consensus is that's sufficient, I'm happy to go along with it, but for my own edification, why would't we want to set that for them as part of `deisctl upgrade-takeover`?

docs/managing_deis/upgrading-deis.rst
iancoffey: This is based on the actual availability of the routers - there will almost certainly be at least one functional router up at all times. I originally had some text about configuring LBs to avoid that second or two of downtime, but it ended up being a bit much. Ill re-add something short and generic to explain the concern.

krancour: > Ill re-add something short and generic to explain the concern

Got it.  My only concern was that the following sounds a bit non-deterministic, so a brief explanation would help:

> This means that there should only be a maximum of around 1-2 seconds of downtime while the routers boot up. Many times, there will be no downtime at all.

deisctl/cmd/upgrade.go
iancoffey: To me, its just a clearer workflow and less code in this already huge PR. We already have the logic to do this without more magic - I suggest we start there and add more logic down the road. I can create an issue if you want to keep track of that. Otherwise, Ill defer to maintainers consensus on if its worth it - Ill gladly add it if it is.

krancour: Good enough for me.

docs/managing_deis/upgrading-deis.rst
iancoffey: Yes indeed.

deisctl/config/api.go
iancoffey: Yeah - I likely confused the config interface with an etcd specific interface since the terms are similar.

krancour:
Replaced by #4175

PR #4101, opened 2015-07-24
ref(controller): share common scheduler code

mboersma:
Defines a `SchedulerClient` class to represent the interface that schedulers should implement, and subclasses existing schedulers from that. `attach` was unimplemented anywhere, so I removed it.

Also normalizes some code comments and removes extra code.

### DIFF
--- a/controller/scheduler/__init__.py
+++ b/controller/scheduler/__init__.py
@@ -1,0 +1,35 @@
+
+class AbstractSchedulerClient(object):
+    """
+    A generic interface to a scheduler backend.
+    """
+
+    def __init__(self, target, auth, options, pkey):
+        self.target = target
+        self.auth = auth
+        self.options = options
+        self.pkey = pkey
+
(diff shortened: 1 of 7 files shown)


bacongobbler:
LGTM. Everything was based from MockScheduler but it's better to have something like this. Perhaps to be clear, should this be called AbstractSchedulerClient? What's the convention here?

mboersma:
> Everything was based from MockScheduler

Actually only ChaosScheduler subclassed that (and still does); all other schedulers derived from `object`. I can call it `AbstractSchedulerClient`, that's probably more conventional.

carmstrong:
Code LGTM. Note that merging this will require yet another rebase for #3975, so we may want to merge that PR first and then add k8s to this PR.

mboersma:
Agreed--I'll update this after k8s is merged, which will provide still more reason for sharing common scheduler code and defining the interface.

PR #4103, opened 2015-07-24
feat(platform): add overlay network

krancour:
This is a minimal subset of #3868 with one critical difference: `--iptables=false` has been omitted from the drop-in for `docker.service`.  The result is that docker manages iptables rules _in addition_ to those that are managed by flannel.  The result is that port mappings continue to function as they always have-- the benefit of which is that no changes are required to any Deis components.  With no changes to components required, upgrades should work smoothly.

I'd appreciate any help I can get in verifying / validating those finding and / or making swiss cheese out of this approach.  :)

Majority credit to @aledbf!

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -90,7 +90,13 @@ Vagrant.configure("2") do |config|
   end
 
   config.trigger.before :up do
-    if !File.exists?(CLOUD_CONFIG_PATH) || File.readlines(CLOUD_CONFIG_PATH).grep(/\s*discovery #DISCOVERY_URL/).any?
+    if File.exists?(CLOUD_CONFIG_PATH) && !File.readlines(CLOUD_CONFIG_PATH).grep(/\s*discovery #DISCOVERY_URL/).any?
+      # Vagrant binds the VMs IP in VirtualBox's bridge network to the eth1 interface instead of eth0.
+      # This necessitates the substitution below, which is not required anywhere except in Vagrant.
+      user_data = File.read(CLOUD_CONFIG_PATH)
+      new_userdata = user_data.gsub("--iface=eth0", "--iface=eth1")
+      File.open(CLOUD_CONFIG_PATH, "w") {|file| file.puts new_userdata }
+    else
       raise Vagrant::Errors::VagrantError.new, "Run 'make discovery-url' first to create user-data."
     end
   end
--- a/contrib/coreos/user-data.example
+++ b/contrib/coreos/user-data.example
@@ -63,6 +63,41 @@ coreos:
       Service=docker.service
       [Install]
       WantedBy=sockets.target
+  - name: flanneld.service
+    command: start
+    content: |
+      [Unit]
+      Description=Network fabric for containers
+      Documentation=https://github.com/coreos/flannel
+      Requires=early-docker.service etcd.service
+      After=etcd.service early-docker.service
+      Before=early-docker.target
+
+      [Service]
+      Type=notify
+      Restart=always
+      RestartSec=5
+      Environment="TMPDIR=/var/tmp/"
+      Environment="DOCKER_HOST=unix:///var/run/early-docker.sock"
+      Environment="FLANNEL_VER=0.5.1"
+      LimitNOFILE=40000
+      LimitNPROC=1048576
+      ExecStartPre=/sbin/modprobe ip_tables
+      ExecStartPre=/usr/bin/mkdir -p /run/flannel
+      ExecStartPre=/usr/bin/touch /run/flannel/options.env
+      ExecStartPre=-/usr/bin/etcdctl mk /coreos.com/network/config '{"Network":"10.244.0.0/16", "SubnetLen": 24, "SubnetMin":"10.244.0.0", "Backend": {"Type": "vxlan"}}'
+      ExecStart=/usr/libexec/sdnotify-proxy /run/flannel/sd.sock \
+        /usr/bin/docker run --net=host --privileged=true --rm \
+        --volume=/run/flannel:/run/flannel \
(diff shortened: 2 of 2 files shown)


krancour:
Ping @aledbf @carmstrong @smothiki @kmala

carmstrong:
Awesome stuff! Can you explain the Vagrantfile interface change?

krancour:
@carmstrong that's from @aledbf, but I know the reason is because Vagrant VM's bind the machine's IP on the VB bridge network to `eth1` instead of `eth0`.  `eth0` is suitable for any cloud provider, but in Vagrant, this substitution needs to be made.

carmstrong:
replaces #3868

carmstrong:
> eth0 is suitable for any cloud provider, but in Vagrant, this substitution needs to be made.

Understood.

aledbf:
> Can you explain the Vagrantfile interface change?

By default flannel uses the interface with the default gateway and in Vagrant eth0 uses the same ip in all the nodes.

carmstrong:
@smothiki made a comment on #3868: "should the docker opts contain -ip-masq=false as we already started flannel with --ip-masq=true"

krancour:
At @smothiki's request, I am trying `docker.service` with `--iptables=false --ip-masq=false` locally.  I'll report back how that goes.

krancour:
The suggestion above doesn't work out as @smothiki was hoping.  Nevermind the `--ip-masq=false` setting... when `--iptables=false`, Docker doesn't do anything at all to iptables and as far as I can tell (@aledbf understands it a bit better than me) that leads to a break down in port mappings-- which in turn leads to needing to modify more things-- like changing the port used within the `deis-builder` and using `--net=host`.  Some of these changes promote the complexities that make it difficult to run both with or without flannel.  This is what I was trying to avoid with this more minimal implementation.

As it stands, this PR _is_ working for me.  (I haven't re-pushed with the suggestion above that didn't work.)  Do we know for certain that k8s is not working on top of this?

kmala:
I had tried before with this change of just adding the flannel and it was not working. I will try again now.

carmstrong:
Run looks promising so far! https://ci.deis.io/job/test-ec2/2190/console

krancour:
@kmala do you know what exactly didn't work?  I don't understand kubernetes well yet, but I'm aware of it's requirement that each container has its own IP.  I think this PR does satisfy that (in a new cluster at least).

Of course I'm willing to accept that that alone isn't sufficient and/or allowing Docker to mess with iptables interferes with k8s in some other way that I don't yet understand.

carmstrong:
Code LGTM. Would like review from @smothiki and @aledbf specifically, too.

smothiki:
The IP gets assigned from docker bridge. Which is a flannel subnet and uses IPtables set by flannel. Port forwarding will be handled by flannel.

krancour:
Right.  I'm not understanding though why Docker can't _also_ define some iptables rules for port forwarding.  There should be no conflict afaik.  Rules set by Docker are forwarding traffic where dest was the host.  Rules set by flannel are forwarding traffic where dest was an IP in the machine's internal subnet.  Aren't those mutually exclusive?

aledbf:
Code LGTM but it must be tested in conjunction with #3975 because Kubernetes creates it own set of rules to using a virtual IP to proxy traffic to the pods

Vagrantfile
```diff
@@ -12,6 +12,11 @@ end
 CLOUD_CONFIG_PATH = File.join(File.dirname(__FILE__), "contrib", "coreos", "user-data")
 CONFIG = File.join(File.dirname(__FILE__), "config.rb")
 
+user_data = File.read(CLOUD_CONFIG_PATH)
```
mboersma: Could you add a comment about why we have to sub in `eth1` here? I had to ask @smothiki and @carmstrong also asked, so others will wonder when they see this code.

krancour: Yes.  I'll do that.

carmstrong:
Looks like an unlucky run: https://ci.deis.io/job/test-ec2/2190/consoleFull#682192135502066c60b2236c09cc9bf8
```
remote: [1G/home/git/builder: line 175: cd: /home/git/pssample.git: No such file or directory[K
fatal: Unable to read current working directory: No such file or directory
To ssh://git@deis.test-afe94c3665.deis.works:2222/pssample.git
 ! [remote rejected] master -> master (pre-receive hook declined)
error at command wait
--- FAIL: TestPs (42.65s)
	itutils.go:294: exit status 1
FAIL
exit status 1
FAIL	github.com/deis/deis/tests	42.663s
godep: go exit status 1
make[1]: Leaving directory `/var/lib/jenkins/jobs/test-ec2/workspace@2/src/github.com/deis/deis/tests'
```

krancour:
Note: I added a comment about the `eth0` / `eth1` swap per @mboersma's request.  I also fixed an issue in the Vagrantfile wherein that swap was failing (and not gracefully) if `contrib/coreos/user-data` hadn't already been generated by `make discovery-url`.  I moved that bit of code down into the trigger that checks for the existence of that file.

kmala:
I had tested kubernetes with this overlay changes and it is working fine.Will be doing more testing.

smothiki:
Working, tested on virtual box cluster . Will test on aws for one more time probably by tomorrow EOD.

mboersma:
Code LGTM.

carmstrong:
This is good to ship as far as I'm concerned, unles @kmala or @smothiki ran into any issues.

smothiki:
We just wanna test on AWS once before merging

smothiki:
Code LGTM tested with only the changes from PR. @kmala has issues with his changes.

krancour:
@smothiki indicated that @kmala's issues aren't dire enough to block merging this PR.

PR #4104, opened 2015-07-24
feat(client-go): add certs endpoint

Joshua-Anderson:
After this PR, client-go will be feature complete. \o/

### DIFF
--- a/client-go/cmd/certs.go
+++ b/client-go/cmd/certs.go
@@ -1,0 +1,143 @@
+package cmd
+
+import (
+	"fmt"
+	"io/ioutil"
+	"strings"
+
+	"github.com/deis/deis/pkg/prettyprint"
+
+	"github.com/deis/deis/client-go/controller/client"
+	"github.com/deis/deis/client-go/controller/models/certs"
+)
+
+// CertsList lists certs registered with the controller.
+func CertsList() error {
+	c, err := client.New()
+
+	if err != nil {
+		return err
+	}
+
+	certList, err := certs.List(c)
+
+	if err != nil {
+		return err
+	}
+
+	if len(certList) == 0 {
+		fmt.Println("No certs")
+		return nil
+	}
+
+	certMap := make(map[string]string)
+	nameMax := 0
+	expiresMax := 0
+	for _, cert := range certList {
+		certMap[cert.Name] = cert.Expires
+
+		if len(cert.Name) > nameMax {
+			nameMax = len(cert.Name)
+		}
+		if len(cert.Expires) > nameMax {
+			expiresMax = len(cert.Expires)
+		}
+	}
+
+	nameHeader := "Common Name"
+	expiresHeader := "Expires"
+	tabSpaces := 5
+	bufferSpaces := tabSpaces
+
+	if nameMax < len(nameHeader) {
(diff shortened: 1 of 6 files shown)


client-go/cmd/certs.go
```diff
@@ -0,0 +1,143 @@
...
+
+	if err != nil {
+		return err
+	}
+
+	fmt.Println("done")
+	return nil
+}
+
+func processCertsAdd(c *client.Client, cert string, key string, cN string, sans string) error {
```
mboersma: I would shorten the func signature this way, although this is just personal taste:
```go
func processCertsAdd(c *client.Client, cert, key, cN, sans string) error
```

client-go/cmd/certs.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	fmt.Printf("%s%s%s\n", nameHeader, strings.Repeat(" ", bufferSpaces), expiresHeader)
+	fmt.Printf("%s%s%s\n", strings.Repeat("-", nameMax), strings.Repeat(" ", 5),
+		strings.Repeat("-", expiresMax))
+	fmt.Print(prettyprint.PrettyTabs(certMap, tabSpaces))
+	return nil
+}
+
+// CertAdd adds a certs to the controller.
```
mboersma: Minor: "certs" should be "cert".

client-go/cmd/certs.go
```diff
@@ -0,0 +1,143 @@
...
+
+	if err != nil {
+		return err
+	}
+
+	_, err = certs.New(c, string(certFile), string(keyFile), commonName)
+	return err
+}
+
+// CertRemove deletes a cert from the controller
```
mboersma: Go comments should be full sentences, so this should end with a period. :black_circle:

mboersma:
Code LGTM, my minor comments notwithstanding.

Joshua-Anderson:
Done!

client-go/controller/api/certs.go
```diff
@@ -0,0 +1,26 @@
+package api
+
+// Cert is the definition of the cert object.
+type Cert struct {
+	Updated string `json:"updated,omitempty"`
+	Created string `json:"created,omitempty"`
+	Name    string `json:"common_name"`
+	Expires string `json:"expires"`
+	Owner   string `json:"owner,omitempty"`
```
bacongobbler: are these `omitempty` fields here strictly to workaround tests or something? Certs should always have an owner.

Joshua-Anderson: API fun! When listing certificates only the fields `common_name` and `expires` are returned. When getting or creating a certificate, all of these fields are returned. I figured since they share fields I could use the same object if I omitted fields that would be empty on listing. Is that too confusing? Should I add a comment about that?

bacongobbler: Adding a comment to explain this rationale would be a nice touch, thanks :)

bacongobbler:
code LGTM

client-go/controller/api/certs.go
Joshua-Anderson: Done!

carmstrong:
Running this as https://ci.deis.io/job/test-ec2/2242/

PR #4105, opened 2015-07-24
feat(deisctl): extend SSH to allow exec and docker

technosophos:
This feature extends 'deis ssh' to allow for an arbitrary command.
If specified, the command will be run as an SSH Exec call, instead
of opening a new shell.

Further, this adds 'deis dock', which is a convenience wrapper
for 'deis ssh THING docker exec -it THING COMMAND'. This now
becomes 'deis dock THING COMMAND'.

Testing:

After installation, you should be able to run all of the following commands:

- `deis ssh deis-builder`: Works as normal.
- `deis ssh deis-builder ls -la`: Runs `ls -la` on the host that is running builder.
- `deis dock deis-builder sh`: Starts a shell on the `deis-builder` container.

### DIFF
--- a/docs/troubleshooting_deis/index.rst
+++ b/docs/troubleshooting_deis/index.rst
@@ -24,6 +24,33 @@ To open a interactive shell on a machine in your cluster:
 
     $ deisctl ssh <unit>
 
+For example, to open a shell session on the machine that is running Controller,
+you can run this:
+
+.. code-block:: console
+
+    $ deisctl ssh controller
+
+You can execute just a single command instead of opening a shell:
+
+.. code-block:: console
+
+    $ deisctl ssh <unit> <command>
+
+You can also connect directly to the Docker instance of that unit:
+
+.. code-block:: console
+
+    $ deisctl dock <unit> <command>
+
+For example, to start a Bash session on the Builder Docker container, you can
+run the following command:
+
+.. code-block:: console
+
+    $ deisctl dock builder bash`
+
+
 Troubleshooting etcd
 --------------------
 
--- a/deisctl/backend/backend.go
+++ b/deisctl/backend/backend.go
@@ -13,6 +13,7 @@ type Backend interface {
 	Stop([]string, *sync.WaitGroup, io.Writer, io.Writer)
 	Scale(string, int, *sync.WaitGroup, io.Writer, io.Writer)
 	SSH(string) error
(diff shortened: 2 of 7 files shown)


Joshua-Anderson:
s/deis/deisctl/ in description. :smiley:

carmstrong:
Code LGTM. Looks like we have an added test for this, too. :+1:

technosophos:
I think I will fix it to run 'sh' by default when someone does `deisctl dock CONTAINER` with no command. It won't work on all Docker containers, but it will give a decent base experience. And it's consistent with @Joshua-Anderson 's `deisctl ssh` behavior.

krancour:
I haven't looked at the code yet.  I am just working off the description in the OP... for consistency with other `deisctl` commands, should components be called out, for instance, as `builder` instead of `deis-builder`?

Joshua-Anderson:
@krancour I believe that both `builder` and `deis-builder` should work, as `deisctl` matches units that contain the target string .

Joshua-Anderson:
@technosophos This is a really useful feature, you might want to document it with alongside the segment on `deisctl ssh`.

technosophos:
The last commit makes `sh` the default command if you run `deisctl dock foo` (without a fourth arg).

bacongobbler:
> The last commit makes sh the default command

+1. Most of the controller components are on alpine linux and may or may not have `bash` installed.

technosophos:
And, at @Joshua-Anderson 's suggestion, I added some documentation alongside the normal `deisctl ssh` docs in the troubleshooting guide.

I think I'm done with this now. Anything else, or can we LGTM and merge?

docs/troubleshooting_deis/index.rst
```diff
@@ -24,6 +24,34 @@ To open a interactive shell on a machine in your cluster:
...
 
+For example, to open a shell session on the machine that is running Controller,
+you can run this:
+
+.. code-block:: console
+
+    $ deisctl ssh controller
+
+You can execute just a single command instead of opening a shell:
+
```
carmstrong: *Super* minor, but there's an extra space here.

technosophos: Fixed

carmstrong:
Code LGTM.

bacongobbler:
code LGTM

PR #4107, opened 2015-07-25
fix(vagrant): fix too many redirects problem

krancour:
This fixes the "too many redirects" problem that seems to occur (particularly on Yosemite) during `docker pull`-- particularly when `deis-builder` is building the slugbuilder.

This issue discusses the problem and solution with respect to `boot2docker`:

https://github.com/boot2docker/boot2docker/issues/451

This PR merely applies the same fix to all our Vagrant-managed VMs via the `Vagrantfile`.

Ping @jchauncey-- I think you said you were having this issue too.  Can you confirm this helps you as well?

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -146,6 +146,12 @@ Vagrant.configure("2") do |config|
       ip = "172.17.8.#{i+99}"
       config.vm.network :private_network, ip: ip
 
+      # Use the same nameserver as the host machine in order to avoid the "too many redirects" problem.
+      config.vm.provider :virtualbox do |vb|
+        vb.customize ["modifyvm", :id, "--natdnshostresolver1", "off"]
+        vb.customize ["modifyvm", :id, "--natdnsproxy1", "off"]
+      end
+
       # Uncomment below to enable NFS for sharing the host machine into the coreos-vagrant VM.
       #config.vm.synced_folder ".", "/home/core/share", id: "core", :nfs => true, :mount_options => ['nolock,vers=3,udp']
       $shared_folders.each_with_index do |(host_folder, guest_folder), index|


carmstrong:
@krancour We should probably fix this upstream, too - looks like they've seen it as well. https://github.com/coreos/coreos-vagrant/issues/124#issuecomment-49481032

carmstrong:
Code LGTM.

krancour:
> We should probably fix this upstream, too

PR opened: https://github.com/coreos/coreos-vagrant/pull/254

mboersma:
This exact code used to be in our Vagrantfile, but we removed it last year after "upstream" coreos-vagrant did: see 
deis/deis@241932c702bca2616a49f79425334da2d904d680 and coreos/coreos-vagrant@4d81118f8feb021b3dd37cef923bc2144f255ce4. I don't see a specific justification for removing it in either case; I assume both authors thought they were removing cruft.

technosophos:
So can we add it back, @mboersma ? Code LGTM, and anything to speed up builder development is a welcome change.

mboersma:
I think we can add it back in, but I guess I was hoping for some confirmation from the CoreOS folks that there wasn't a good reason for it having been removed. I wouldn't want to merge this and then hear "Oh yeah, we had to take that out because xyz..."

carmstrong:
I think we're good:
```
11:35:09 AM <marineam> carmstrong: looks like I moved it to the Vagrantfile that is bundled into our .box https://github.com/coreos/coreos-overlay/blob/master/coreos-base/oem-vagrant/files/box/Vagrantfile
11:35:24 AM <carmstrong> marineam: interesting. okay, thanks. we'll add it back into ours
```

PR #4109, opened 2015-07-26
ref(deisctl): replace if/else trees with switch

### LINKED ISSUE #3973: Change if/else if/else to switch in deisctl
There are multiple places in `deisctl` where we use `if/else if/else` trees rather than `switch` statements. The Go idiom (http://golang.org/doc/effective_go.html#switch) is to prefer switch over if/else trees.

mboersma:
Closes #3973.


carmstrong:
Code LGTM.

PR #4114, opened 2015-07-27
fix(client-go): make output from apps:destroy match python client

### LINKED ISSUE #4113: client-go does not return time to destroy on apps:destroy
```
$ deis destroy -a appssample
 !    WARNING: Potentially Destructive Action
 !    This command will destroy the application: appssample
 !    To proceed, type "appssample" or re-run this command with --confirm=appssample

>appssample
Destroying appssample...$
```

The old client behaviour was like so:

```
$ python ~/go/src/github.com/deis/deis/client/deis.py destroy -a foo

 !    WARNING: Potentially Destructive Action
 !    This command will destroy the application: foo
 !    To proceed, type "foo" or re-run this command with --confirm=foo

> foo
Destroying foo... 
done in 0s
$
```

There also seems to be some formatting issues as well, both when destroying the git remote and without.

Joshua-Anderson:
Fixes #4113

### DIFF
--- a/client-go/cmd/apps.go
+++ b/client-go/cmd/apps.go
@@ -5,6 +5,7 @@ import (
 	"net/url"
 	"os"
 	"strings"
+	"time"
 
 	"github.com/deis/deis/pkg/prettyprint"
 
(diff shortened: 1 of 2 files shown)


carmstrong:
I think we report elapsed time in several places in the Python client - would probably be prudent to do an audit and see if we're still doing that everywhere.

Joshua-Anderson:
The other places where time is printed is `ps:scale` and `ps:scale`, and the new client matches this behavior. I just forgot to add it in `apps:destroy`.

bacongobbler:
code lgtm

mboersma:
Code LGTM.

PR #4117, opened 2015-07-27
fix(store): block metadata start until confd runs

### LINKED ISSUE #4115: store-metadata: mds.-1.0 ERROR: failed to authenticate: (1) Operation not permitted
This error has cropped up a bit lately, and has been seen by @crigor and @mdolian on users' clusters. Full logs from one cluster are as follows:

```
Jul 27 18:05:25 host1 sh[1094]: starting mds.host1 at :/0
Jul 27 18:05:25 host1 sh[1094]: 2015-07-27 18:05:25.768406 7fed4c852800 -1 auth: error parsing file /var/lib/ceph/mds/ceph-host1/keyring
Jul 27 18:05:25 host1 sh[1094]: 2015-07-27 18:05:25.789818 7fed4c852800 -1 mds.-1.0 log_to_monitors {default=true}
Jul 27 18:05:28 host1 sh[1094]: 2015-07-27 18:05:28.784466 7fed4c852800 -1 mds.-1.0 ERROR: failed to authenticate: (1) Operation not permitted
Jul 27 18:05:28 host1 sh[1094]: 2015-07-27 18:05:28.784495 7fed4c852800  1 mds.-1.0 suicide.  wanted down:dne, now up:boot
```

Note that in this instance, this didn't happen on first metadata boot, but only after the host had rebooted. I suspect a race condition in which confd has not yet run and templated a valid keyring file. Note that while we [block for confd](https://github.com/deis/deis/blob/e4a54d031280d708d962a2dbde6d11e0755fc6a6/store/metadata/bin/boot#L43-L46) during initial metadata setup, we do *not* block on subsequent starts before checking to see if the keyring file exists: https://github.com/deis/deis/blob/e4a54d031280d708d962a2dbde6d11e0755fc6a6/store/metadata/bin/boot#L50-L61

An example of where we do this properly is in the monitor: https://github.com/deis/deis/blob/e4a54d031280d708d962a2dbde6d11e0755fc6a6/store/monitor/bin/boot#L67-L70

While this is an assertion as to the root cause here, the proposed fix should be straightforward and strictly correct, so I'll implement it.

carmstrong:
We faced a race condition where the MDS could start and initialize
with an invalid keyring file, since it hadn't yet been templated
by confd.

closes #4115

### DIFF
--- a/store/metadata/bin/boot
+++ b/store/metadata/bin/boot
@@ -47,6 +47,11 @@ if ! etcdctl --no-sync -C $ETCD get ${ETCD_PATH}/filesystemSetupComplete >/dev/n
   fi
 fi
 
+until confd -onetime -node $ETCD --confdir /app --log-level error; do
+  echo "store-metadata: waiting for confd to write initial templates..."
+  sleep 5
+done
+
 # Check to see if we are a new MDS
 if [ ! -e /var/lib/ceph/mds/ceph-$MDS_NAME/keyring ]; then
   mkdir -p /var/lib/ceph/mds/ceph-${MDS_NAME}


store/metadata/bin/boot
```diff
@@ -47,6 +47,11 @@ if ! etcdctl --no-sync -C $ETCD get ${ETCD_PATH}/filesystemSetupComplete >/dev/n
   fi
 fi
 
+until confd -onetime -node $ETCD --confdir /app --log-level error; do
```
bacongobbler: isn't this covered by lines 11-14? If that passes, the keyring file should be templated. Odd...

carmstrong: Yep, came to the same conclusion a few minutes ago...

PR #4120, opened 2015-07-27
test(*): add provision/upgrade test

### LINKED ISSUE #4033: Automate upgrade test

sgoings:
- use rerun for reusability of scripts

fixes #4033


tests/tests/clients-1-test.sh
```diff
@@ -0,0 +1,93 @@
...
+  rm -r "${TEST_ROOT}"
+}
+
+trap cleanup EXIT
+
+it_sets_up_released_deisctl() {
+
+  function download-client {
+    [ ${1} == "deisctl" ] &&
+    [ ${2} == "1.8.0" ]
```
carmstrong: Is it okay to hard-code 1.8.0 here?

sgoings: This is a test, so the line below

`setup-deisctl-client "1.8.0"`

eventually I want to see download the 1.8.0 deisctl client (which I intercept in the inner `download-client` function here).

carmstrong: :+1:

rerun
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
```
bacongobbler: If this is part of the test infrastructure, I feel that this should be moved to `tests/contrib` or `tests/bin`. Is there a compelling reason to keep it at the project root?

tests/bin/vagrant.sh
```diff
@@ -0,0 +1,4 @@
+function clean-vagrant {
+  "${THIS_DIR}/halt-all-vagrants.sh"
+  vagrant destroy --force
+}
```
bacongobbler: missing newline here

tests/tests/clients-1-test.sh
bacongobbler: tests for the tests. So meta ;) :+1:

rerun
sgoings: I'm ok with moving it, my plan at one point in the process was maybe use rerun as the entry point to a lot of other things, but I think it's going to become the wrapper around provisioning scripts (which may end up in a different repository). I'm ok with moving it.

mboersma: I agree, it would be better to put `rerun` in tests/bin unless and until it is being used elsewhere. We will just confuse people by dropping a binary in the root of the repo.

mboersma:
This looks great. I reviewed it and didn't find anything to change (although it's a lot of shell to read). I'll test this out manually, and hopefully we can put it into service for the next release.

carmstrong:
Code LGTM.

carmstrong:
@sgoings This could probably make use of the new graceful upgrade logic in #4175. It'd be good to test that so we can certify it as the preferred upgrade workflow and drop the "experimental" tag ASAP.

sgoings:
@carmstrong I saw that land while I was away. Is there any reason we should enable automated testing of pre 1.9.0 -> something new (because we can't use the graceful upgrade for those if that's the case).

carmstrong:
> Is there any reason we should enable automated testing of pre 1.9.0 -> 1.9.0 or greater

Would be interesting to see where upgrades break down (i.e. can I upgrade Deis 1.4 to 1.9?) but that's not nearly as important as validating n-1 can upgrade to n.

sgoings:
I've added another option to the upgrade acceptance test:

`./rerun accept:upgrade --from 1.9.0 --to master --provider aws --upgrade-style [graceful,inplace]`

Which is available to configure on the Jenkins job as well.

sgoings:
I'd like to get see this merged sometime today. What needs to happen to improve the consistency of the test-ec2 job?

mboersma:
LGTM. I've run this against vagrant and AWS with the new graceful upgrade default.

> What needs to happen to improve the consistency of the test-ec2 job?

Some kind of sacrifice to higher powers undoubtedly. :cookie:

PR #4121, opened 2015-07-27
chore(tests): bump etcd to v2.0.13

carmstrong:
We bumped the platform to etcd 2 in #4090. We should probably be testing against this version of etcd.

/cc @mboersma @sgoings


carmstrong:
@mboersma I'm wondering if the changes in #1351 are problematic here, since the test-etcd container will exist but it will be old, and the components only pull test-etcd if it doesn't already exist.

carmstrong:
Jenkins won't test this, but it can be verified by building locally.

carmstrong:
Refs #4135

carmstrong:
Replaced by #4145

PR #4122, opened 2015-07-27
fix(client): remove nonexistant certs:update

### LINKED ISSUE #4102: `deis certs:update` doesn't exist
See [here](https://github.com/deis/deis/blob/master/client/deis.py#L1096). There is no code to implement the command.

I'd vote for entirely removing the command, since it presumably was never functional and nobody caught that it was broken yet.

Joshua-Anderson:
Fixes #4102


carmstrong:
Code LGTM.

mboersma:
Code LGTM.

PR #4125, opened 2015-07-27
test(database): run shell scripts through shellcheck

Joshua-Anderson:


mboersma:
Code LGTM. Changing even trivial things in the shell scripts worries me, but this is almost restricted to bin/boot, so it's being tested that way.

bacongobbler:
code LGTM

PR #4127, opened 2015-07-28
feat(platform): add http liveness and readiness health checks

### LINKED ISSUE #3813: Application healthcheck endpoint for checks during deploy and publishing

### LINKED ISSUE #4045: Design Doc: HTTP Healthcheck on Deploy
refs the new proposal in #4019. All input/feedback on the doc is welcome.
refs #3813

Please note that this document is being maintained at https://gist.github.com/bacongobbler/d2f0ef2311d1e3e28997 to keep a living history of any changes made.

During a typical deployment of an application release, the controller performs a status check from Fleet after it has been deployed to ensure that it is running. However, Fleet is a glorified systemd wrapper, which means that it only checks systemd if the process is running (in this case, `docker run`). If a container's web process takes some time to boot, this may cause a false positive status check from the user's perspective. This proposal introduces a new concept into Deis known as a health check, which will make the controller check Deis' Publisher component for real-time stats on the app to ensure the container is running before recycling the old release.

Goals
-----

The main goals of this proposal is to

 - ensure a zero-downtime deployment between deployments of new releases
 - allow the user to configure the URL path, the health check delay, and the health check request timeout value.

Code Changes
------------

In order for this change to occur, there needs to be a way for users to configure both the expected health check URL as well as the initial delay and timeout value of the request. Since the `Config` object currently exists and the health checks are considered part of an application's configuration, the code changes in the controller is fairly minimal:

 - `deis config:set HEALTHCHECK_URL=/` will configure a health check URL for the application
 - the expected status code from the health check is a 200 OK. This is non-configurable.
 - `deis config:set HEALTHCHECK_PORT=5432` will configure a health check port for the application (default: 5000)
 - `deis config:set HEALTHCHECK_INITIAL_DELAY=10` will tell Deis' Publisher to wait 10 seconds before performing a health check (default: 0)
 - `deis config:set HEALTHCHECK_TIMEOUT=10` will tell Deis' Publisher that the request may take up to 10 seconds before the app is considered "down" (default: 5).

These keys will be exposed to Etcd at /Deis/config/appname/health check_URL=/ so deis's Publisher can retrieve these values to perform its job.

Once `HEALTHCHECK_URL` is set, the controller will do the typical application release workflow. However, once the controller is notified by Fleet that the new version of the application is "up", the controller will wait a minimum of x+y seconds, where x is equal to the value of HEALTHCHECK_INITIAL_DELAY, and y is equal to the value of HEALTHCHECK_TIMEOUT. By default this is 5 seconds. Once the waiting is over, the controller will then perform an API request against the Publisher to inspect the container's health (more on that below).

On the Publisher's side, before it publishes a container to Etcd, it will check Etcd for these keys. If health check_URL is set, then it will perform the health check. It will also expose an API which the controller can now ping. The API will be very simple:

Example request:

	GET /up

Example Response:

	[
		"foo_v2.web.1",
		"foo_v2.web.3",
		"bar_v3.cmd.1"
	]

This response is a list of all containers which Publisher has exposed to Etcd and are considered "up". It will be the responsibility of the controller to parse this response to check if the containers deployed are up based on the container ID. If the health check passes, the rest of the deployment workflow will then continue, recycling old containers and returning a successful response to the client. If only a few are down (few being anything less than all of the new containers), a warning will be raised, but the operation will be considered "passed". However, if health check failed, the controller will respond with an appropriate error and roll back to the old version of the application.

Issues to Note
--------------

With this approach, We have some legacy code in the Publisher which will only publish the latest version of the container at any given time (see #2413 for a history lesson). This was put in place to prevent two existing versions of the same app to be published to the router. We'll have to be extra careful not to break this functionality, but we also have to make sure that we're still publishing the older version of the controller to the load balancer (Deis-router) before the new container passes the health check. For example, if the user has set an initial delay that is higher than the TTL of the Etcd keys, this could cause the old version of the containers to come down, causing downtime. This can be mitigated with proper unit tests within Publisher to ensure that an older version of the container is still getting published at this time.

Testing
-------

In order for this proposal to work, we need to ensure a zero-downtime deployment, as well as proper errors to be returned when an application fails to respond in time or with an incorrect status code. Since the controller has a great suite for unit tests, most of the workflow for the client will be covered by unit tests to ensure that the proper errors are propagated when certain cases occur. However, the zero-downtime approach will be covered by a combination of unit tests in Deis-Publisher as well as some comprehensive integration tests. How this will be accomplished is by invoking a goroutine that will continuously ping the app URL. If a 404 response shows up (due to #3774), then the deployment failed as a zero-downtime deployment.

bacongobbler:
More documentation is forthcoming.

closes #3813 
closes #4045

### DIFF
--- a/controller/api/models.py
+++ b/controller/api/models.py
@@ -31,7 +31,7 @@ from OpenSSL import crypto
 import requests
 from rest_framework.authtoken.models import Token
 
-from api import fields, utils
+from api import fields, utils, exceptions
 from registry import publish_release
 from utils import dict_diff, fingerprint
 
@@ -116,6 +116,20 @@ def validate_certificate(value):
         crypto.load_certificate(crypto.FILETYPE_PEM, value)
     except crypto.Error as e:
         raise ValidationError('Could not load certificate: {}'.format(e))
+
+
+def get_etcd_client():
+    if not hasattr(get_etcd_client, "client"):
+        # wire up etcd publishing if we can connect
+        try:
+            get_etcd_client.client = etcd.Client(
+                host=settings.ETCD_HOST,
+                port=int(settings.ETCD_PORT))
+            get_etcd_client.client.get('/deis')
+        except etcd.EtcdException:
+            logger.log(logging.WARNING, 'Cannot synchronize with etcd cluster')
+            get_etcd_client.client = None
+    return get_etcd_client.client
 
 
 class AuditedModel(models.Model):
@@ -353,6 +367,58 @@ class App(UuidAuditedModel):
         if set([c.state for c in to_add]) != set(['up']):
             err = 'warning, some containers failed to start'
             log_event(self, err, logging.WARNING)
+        # if the user specified a health check, try checking to see if it's running
+        try:
+            config = self.config_set.latest()
+            if 'HEALTHCHECK_URL' in config.values.keys():
+                self._healthcheck(to_add, config.values)
+        except Config.DoesNotExist:
+            pass
+
+    def _healthcheck(self, containers, config):
+        # if at first it fails, back off and try again at 10%, 50% and 100% of INITIAL_DELAY
+        intervals = [1.0, 0.1, 0.5, 1.0]
+        # HACK (bacongobbler): we need to wait until publisher has a chance to publish each
+        # service to etcd, which can take up to 20 seconds.
+        time.sleep(20)
+        for i in range(0, 4):
+            delay = int(config.get('HEALTHCHECK_INITIAL_DELAY', 0))
+            try:
+                # sleep until the initial timeout is over
+                if delay > 0:
+                    time.sleep(delay * intervals[i])
+                self._do_healthcheck(containers, config)
+                break
+            except exceptions.HealthcheckException as e:
+                try:
+                    new_delay = delay * intervals[i+1]
+                    msg = "{}; trying again in {} seconds".format(e, new_delay)
+                    log_event(self, msg, logging.WARNING)
+                except IndexError:
+                    log_event(self, e, logging.WARNING)
+        else:
+            self._destroy_containers(containers)
+            msg = "aborting, app containers failed to respond to health check"
+            log_event(self, msg, logging.ERROR)
+            raise RuntimeError(msg)
+
+    def _do_healthcheck(self, containers, config):
+        path = config.get('HEALTHCHECK_URL', '/')
+        timeout = int(config.get('HEALTHCHECK_TIMEOUT', 1))
+        if not _etcd_client:
+            raise exceptions.HealthcheckException('no etcd client available')
+        for container in containers:
+            try:
+                key = "/deis/services/{self}/{container.job_id}".format(**locals())
+                url = "http://{}{}".format(_etcd_client.get(key).value, path)
+                response = requests.get(url, timeout=timeout)
+                if response.status_code != requests.codes.OK:
+                    raise exceptions.HealthcheckException(
+                        "app failed health check (got '{}', expected: '200')".format(
+                            response.status_code))
+            except (requests.Timeout, requests.ConnectionError, KeyError) as e:
+                raise exceptions.HealthcheckException(
+                    'failed to connect to container ({})'.format(e))
 
     def _restart_containers(self, to_restart):
         """Restarts containers via the scheduler"""
@@ -452,7 +518,7 @@ class App(UuidAuditedModel):
 
         self.scale(user, structure)
 
-    def logs(self, log_lines):
+    def logs(self, log_lines=str(settings.LOG_LINES)):
         """Return aggregated log data for this application."""
         path = os.path.join(settings.DEIS_LOG_DIR, self.id + '.log')
         if not os.path.exists(path):
@@ -513,7 +579,7 @@ class Container(UuidAuditedModel):
 
     @property
     def state(self):
-        return self._scheduler.state(self._job_id).name
+        return self._scheduler.state(self.job_id).name
 
     def short_name(self):
         return "{}.{}.{}".format(self.app.id, self.type, self.num)
@@ -526,15 +592,10 @@ class Container(UuidAuditedModel):
         get_latest_by = '-created'
         ordering = ['created']
 
-    def _get_job_id(self):
-        app = self.app.id
-        release = self.release
-        version = "v{}".format(release.version)
-        num = self.num
-        job_id = "{app}_{version}.{self.type}.{num}".format(**locals())
-        return job_id
-
-    _job_id = property(_get_job_id)
+    @property
+    def job_id(self):
+        version = "v{}".format(self.release.version)
+        return "{self.app.id}_{version}.{self.type}.{self.num}".format(**locals())
 
     def _get_command(self):
         try:
@@ -566,45 +627,41 @@ class Container(UuidAuditedModel):
         kwargs = {'memory': self.release.config.memory,
                   'cpu': self.release.config.cpu,
                   'tags': self.release.config.tags}
-        job_id = self._job_id
         try:
             self._scheduler.create(
-                name=job_id,
+                name=self.job_id,
                 image=image,
                 command=self._command,
                 **kwargs)
         except Exception as e:
-            err = '{} (create): {}'.format(job_id, e)
+            err = '{} (create): {}'.format(self.job_id, e)
             log_event(self.app, err, logging.ERROR)
             raise
 
     @close_db_connections
     def start(self):
-        job_id = self._job_id
         try:
-            self._scheduler.start(job_id)
+            self._scheduler.start(self.job_id)
         except Exception as e:
-            err = '{} (start): {}'.format(job_id, e)
+            err = '{} (start): {}'.format(self.job_id, e)
             log_event(self.app, err, logging.WARNING)
             raise
 
     @close_db_connections
     def stop(self):
-        job_id = self._job_id
         try:
-            self._scheduler.stop(job_id)
+            self._scheduler.stop(self.job_id)
         except Exception as e:
-            err = '{} (stop): {}'.format(job_id, e)
+            err = '{} (stop): {}'.format(self.job_id, e)
             log_event(self.app, err, logging.ERROR)
             raise
 
     @close_db_connections
     def destroy(self):
-        job_id = self._job_id
         try:
-            self._scheduler.destroy(job_id)
+            self._scheduler.destroy(self.job_id)
         except Exception as e:
-            err = '{} (destroy): {}'.format(job_id, e)
+            err = '{} (destroy): {}'.format(self.job_id, e)
             log_event(self.app, err, logging.ERROR)
             raise
 
@@ -614,7 +671,6 @@ class Container(UuidAuditedModel):
             raise EnvironmentError('No build associated with this release '
                                    'to run this command')
         image = self.release.image
-        job_id = self._job_id
         entrypoint = '/bin/bash'
         # if this is a procfile-based app, switch the entrypoint to slugrunner's default
         # FIXME: remove slugrunner's hardcoded entrypoint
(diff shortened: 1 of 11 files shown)


bacongobbler:
just goes to show that static counters are not great in tests. I'll refactor that.

controller/api/tests/test_config.py
```diff
@@ -509,3 +551,125 @@ def test_unauthorized_user_cannot_modify_config(self):
         response = self.client.post(url, json.dumps(body), content_type='application/json',
                                     HTTP_AUTHORIZATION='token {}'.format(unauthorized_token))
         self.assertEqual(response.status_code, 403)
+
+    def _test_app_healthcheck(self,):
```
mboersma: Extra comma?

mboersma:
Please keep in mind the potential need for more docs as you review this one.

carmstrong:
Test failure:
```
======================================================================
FAIL: test_app_backoff_interval (api.tests.test_config.ConfigTest)
----------------------------------------------------------------------
  File "/home/jenkins/workspace/test-ec2/src/github.com/deis/deis/controller/venv/local/lib/python2.7/site-packages/mock.py", line 1201, in patched
    return func(*args, **keywargs)
  File "/home/jenkins/workspace/test-ec2/src/github.com/deis/deis/controller/api/tests/test_config.py", line 614, in test_app_backoff_interval
    self.assertEqual(mock_time(), 5)
AssertionError: 7 != 5
```

bacongobbler:
yeah, this was something I fixed in an older commit but I must've overwritten it when working across 2 machines. :sheep:

controller/api/models.py
```diff
@@ -299,6 +313,58 @@ def _start_containers(self, to_add):
...
+        except Config.DoesNotExist:
+            pass
+
+    def _healthcheck(self, containers, config):
+        # HACK (bacongobbler): we need to wait until publisher has a chance to publish each
+        # service to etcd, which can take up to 20 seconds.
+        time.sleep(20)
+        for i in range(0, 4):
+            # if at first it fails, back off and try again at 10%, 50% and 100% of INITIAL_DELAY
+            intervals = [1.0, 0.1, 0.5, 1.0]
```
mboersma: `intervals` is invariant and could be moved outside the loop.

docs/using_deis/config-application.rst
```diff
@@ -91,6 +91,28 @@ appname to the old one:
     an application, however you can work around this by pointing the @ record to the
     address of the load balancer (if any).
 
+Custom Health Checks
+--------------------
+
+By default, Deis checks if the container is running, performing no health checks. You can add a
+healthcheck by adding a healthcheck URL, the initial delay and the timeout value by setting custom
+config:
```
mboersma: This could be phrased more tersely:
> By default, Deis only checks that a container is running. You can add a healthcheck by configuring a URL, initial delay, and timeout value:

controller/api/models.py
```diff
@@ -118,6 +118,20 @@ def validate_certificate(value):
...
+        # wire up etcd publishing if we can connect
+        try:
+            get_etcd_client.client = etcd.Client(
+                host=settings.ETCD_HOST,
+                port=int(settings.ETCD_PORT))
+            get_etcd_client.client.get('/deis')
+        except etcd.EtcdException:
+            logger.log(logging.WARNING, 'Cannot synchronize with etcd cluster')
+            get_etcd_client.client = None
+    return get_etcd_client.client
```
mboersma: Is this a way of ensuring `get_etcd_client.client` is a singleton?

bacongobbler: Yes. Functions have instance variables as well. Reading up more on this, the more pythonic way is to just use a global module variable. I can change it back if it looks awkward :)

mboersma: Right, I would expect a module-level variable usually for a singleton, but this works.

mboersma:
Code LGTM.

carmstrong:
Code LGTM other than the potentially extraneous comma that @mboersma identified.

carmstrong:
@bacongobbler This is consistently seeing a 502 in the integration tests, and isn't mergeable until this is fixed: https://ci.deis.io/job/test-ec2/2435/consoleFull#682192135502066c60b2236c09cc9bf8

publisher/server/publisher.go
```diff
@@ -115,6 +116,24 @@ func (s *Server) publishContainer(container *docker.APIContainers, ttl time.Dura
 			port := strconv.Itoa(int(p.PublicPort))
 			hostAndPort := s.host + ":" + port
 			if s.IsPublishableApp(containerName) && s.IsPortOpen(hostAndPort) {
+				configKey := fmt.Sprintf("/deis/config/%s/", appName)
+				// check if the user specified a healthcheck URL
+				healthcheckURL := s.getEtcd(configKey + "healthcheck_url")
+				delay, err := strconv.Atoi(s.getEtcd(configKey + "healthcheck_initial_delay"))
+				if err != nil {
+					log.Println(err)
```
aledbf: this will only show something like `strconv.ParseInt: parsing "blablabla": invalid syntax` in the logs. 
Why not showing this only in debug mode?

bacongobbler:
job 2444 passed: https://ci.deis.io/job/test-ec2/2444/console

For background, this works great on vagrant and on another EC2 cluster (100% success rate), but jenkins doesn't seem to like it 80% of the time.

carmstrong:
Latest logs from the test failure: https://www.dropbox.com/s/3b2uiy28tx4zjza/deis-test-failure-2015-07-31-175430.tar.gz?dl=0

krancour:
Should title begin with "feat(platform)" maybe since scope isn't actually limited to `controller`?

bacongobbler:
indeed. the PR title is based off the commit. Since it used to be just a controller change. I'll change it.

controller/api/models.py
```diff
@@ -353,6 +367,58 @@ def _start_containers(self, to_add):
...
+            delay = int(config.get('HEALTHCHECK_INITIAL_DELAY', 0))
+            try:
+                # sleep until the initial timeout is over
+                if delay > 0:
+                    time.sleep(delay * intervals[i])
+                self._do_healthcheck(containers, config)
+                break
+            except exceptions.HealthcheckException as e:
+                try:
+                    new_delay = delay * intervals[i+1]
```
Blystad: Won't this be out of bounds on the 4th health check?

`range(0, 4)` results in `[0, 1, 2, 3]`, right?

However, when you are at `i = 3`, and you try to refer `intervals[(3) + 1]` then that will be out of bounds, right?

carmstrong: This does look to be a bug. 

/cc @bacongobbler

controller/api/models.py
```diff
@@ -353,6 +367,58 @@ def _start_containers(self, to_add):
...
+            config = self.config_set.latest()
+            if 'HEALTHCHECK_URL' in config.values.keys():
+                self._healthcheck(to_add, config.values)
+        except Config.DoesNotExist:
+            pass
+
+    def _healthcheck(self, containers, config):
+        # if at first it fails, back off and try again at 10%, 50% and 100% of INITIAL_DELAY
+        intervals = [1.0, 0.1, 0.5, 1.0]
+        # HACK (bacongobbler): we need to wait until publisher has a chance to publish each
```
Blystad: Instead of hacking a `time.sleep()`, could we perhaps just use the etcd client to watch for the service to be registered, and have a timeout at 20-30 seconds?

With this, all deploys with healthchecks will add at least 20s of time to deploy. Even if the container is up and running, and passes health checks after 3-4 seconds.

carmstrong: Unfortunately, there is a bug in go-etcd that causes the watch event to fire whenever a key is updated *or* a TTL of a key is updated. This causes a flood of requests to etcd, which has hurt us in the past. Hoping that using etcd2 and a fixed go-etcd library will rectify this, but for now I think we have to avoid using watch.

It'd be great if we could come up with another alternative than sleeping, but not sure if we can.

bacongobbler: etcd has some current issues with watching keys which have given us issues in the past, which is why we aren't using it in `confd` or in other places like the logger: #3505 #3517 #3795

mboersma: Is this fixed in newer go-etcd? Because we have that in #4168 which I think should be merged for this release so that swarm isn't broken.

Blystad: Hmm, how about just implementing a very simple polling based solution then? A simple while loop which counts down, does a call to etcd checking if the key exists, if it does it returns, if not it sleeps for a second.

controller/api/models.py
```diff
@@ -353,6 +367,58 @@ def _start_containers(self, to_add):
...
+        # service to etcd, which can take up to 20 seconds.
+        time.sleep(20)
+        for i in range(0, 4):
```
Blystad: Wouldn't it be better to use something like

`range(0, len(intervals))`

instead of hardcoding it to four values? (assuming that works of course)

bacongobbler: yes, that would be a better implementation. This was something I overlooked after changing it for https://github.com/deis/deis/pull/4127#discussion_r35889077

controller/api/models.py
bacongobbler: @mboersma I don't think so. @aledbf's issue on etcd remains open: https://github.com/coreos/etcd/issues/2679

publisher/server/publisher.go
bacongobbler: These should be ignored if the key is unset, but I think it makes sense to log them if they're an invalid integer.

controller/api/models.py
carmstrong: For context, discussion on our side is in #3505

bacongobbler: @Blystad for the sake of getting this PR through the door I'll keep the current implementation as-is, but I'll follow up in a future PR to use a polling solution. Does that sound fair to you?

controller/api/models.py
bacongobbler: That's why I catch an IndexError. It's very sloppy coding, but it works.

controller/api/models.py
Blystad: Yep, sounds fair to me :)

controller/api/models.py
bacongobbler: @Blystad to explain with a unit test run:

```
$ cat logs/newish-zoologer.log
2015-08-03T19:26:07UTC deis[api]: autotest created initial release
2015-08-03T19:26:07UTC deis[api]: autotest deployed autotest/example
2015-08-03T19:26:07UTC deis[api]: autotest scaled containers cmd=1
2015-08-03T19:26:07UTC deis[api]: autotest added HEALTHCHECK_URL
2015-08-03T19:26:07UTC deis[api]: app failed health check (got '404', expected: '200'); trying again in 0.0 seconds
2015-08-03T19:26:07UTC deis[api]: app failed health check (got '404', expected: '200'); trying again in 0.0 seconds
2015-08-03T19:26:07UTC deis[api]: app failed health check (got '404', expected: '200'); trying again in 0.0 seconds
2015-08-03T19:26:07UTC deis[api]: app failed health check (got '404', expected: '200')
2015-08-03T19:26:07UTC deis[api]: aborting, app containers failed to respond to health check
```

the try/catch block just wraps the call to `log_event`. The fourth healthcheck will still occur as expected. It's just for that fancy `trying again in x seconds` message.

bacongobbler:
https://ci.deis.io/job/test-ec2/2497/ passed earlier this morning, so this looks good to merge. I'll follow up with the comments with another PR so we can start the release dance :)

Merging!

controller/api/tests/test_config.py
bacongobbler: surprisingly flake8 doesn't care about this, but it doesn't break anything. I'll follow up on this one.

PR #4128, opened 2015-07-28
Put user-data on a diet

mboersma:
Subtitle: "this is why we can't have nice things." :smiley:

Removes some optional services, the `nse` func, lots of whitespace, and shortens command flags and variable names. Reduces user-data.example from 12654 to 10785 bytes. Other suggestions for trimming user-data are welcome.

Refs #3975, which seems to create user-data over the 16K limit.


aledbf:
@mboersma using `print json.dumps(template, separators=(',', ':'))` in `gen-json.py` reduces the size in least 512Kb just removing the spaces

aledbf:
@mboersma externalizing some services using a github url like [deis-builder](https://raw.githubusercontent.com/deis/deis/master/deisctl/units/deis-builder.service) and using `ConditionPathExists` like in `install-deisctl.service` can help to remove a lot of code (`preseed` is a good example)

krancour:
> Subtitle: "this is why we can't have nice things."

lol at the sentiment, but may be true... I use `nse` frequently when troubleshooting.  I know we can `docker exec` instead, but the convenience if less typing was always nice.  If I'm a minority, then goodbye `nse` but I'm wondering if others were using it as much as I and will miss it also.

bacongobbler:
@krancour I used it often but `deisctl dock` will do the same thing moving forward :)

krancour:
I haven't tried that yet... I know that's brand new functionality.  How does it respond in the case of a global unit?  That's where I see `nse` as still useful.

mboersma:
`nse` doesn't have to go. I'm just looking for any way I can to shorten user-data, but the practical issue for the k8s PR is only a few hundred bytes.

mboersma:
Closing this, although I have noted the great suggestions. I'm going to make bite-size PRs instead. See #4130.

PR #4130, opened 2015-07-28
ref(user-data): remove load-overlay-module

mboersma:
Current versions of CoreOS (including 647.2.0) don't need any help: docker starts with overlayfs as its backend storage driver regardless. I verified this on Vagrant and AWS.

Replaces #4128. It's easier to do this as smaller PRs.

Refs #3975, since we need to shrink user-data a bit to accomodate those changes. *Edit*: never mind, #3975 shrunk itself enough to pass without other changes.


carmstrong:
LGTM.
```console
core@deis-01 ~ $ lsmod | grep overlay
overlay                40960  2
```

bacongobbler:
LGTM!

PR #4134, opened 2015-07-28
feat(client-go): decode json error messages

### LINKED ISSUE #3739: deis login - fix "bad request" message to be user readable
As a user, I would like a clearer message when my username and/or password fail on login.

Today reads as follows:
 "400 BAD REQUEST
{u'non_field_errors': [u'Unable to log in with provided credentials.']}

Recommended:
"Unable to login: 
The username and password you entered do not match"

SEE image: http://cl.ly/image/143z3J2W202o
deis be kind

Joshua-Anderson:
Fixes #3739

### DIFF
--- a/client-go/controller/client/auth.go
+++ b/client-go/controller/client/auth.go
@@ -129,14 +129,8 @@ func Passwd(username string, password string, newPassword string) error {
 		return err
 	}
 
-	resBody, status, err := client.BasicRequest("POST", "/v1/auth/passwd/", body)
-
-	if err != nil {
+	if _, err = client.BasicRequest("POST", "/v1/auth/passwd/", body); err != nil {
 		return err
-	}
-
-	if status != 200 {
-		return fmt.Errorf("Password change failed: %s", resBody)
 	}
 
 	fmt.Println("Password change succeeded.")
@@ -151,10 +145,8 @@ func Cancel() error {
 		return err
 	}
 
-	body, status, err := client.BasicRequest("DELETE", "/v1/auth/cancel/", nil)
-
-	if status != 204 {
-		return fmt.Errorf("Cancellation failed: %s", body)
+	if _, err = client.BasicRequest("DELETE", "/v1/auth/cancel/", nil); err != nil {
+		return err
 	}
 
 	if err = deleteSettings(); err != nil {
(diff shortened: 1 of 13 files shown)


carmstrong:
Does this close #3739?

Joshua-Anderson:
@carmstrong Yes, it does.

bacongobbler:
From the logs so nobody has to grok:

> FAIL	github.com/deis/deis/client-go/controller/client [build failed]

Joshua-Anderson:
I fixed up the error and added additional tests.

client-go/controller/client/http.go
```diff
@@ -81,20 +83,57 @@ func (c Client) Request(method string, path string, body []byte) (*http.Response
...
-		return "", -1, err
+		return "", err
 	}
-	return string(resBody), res.StatusCode, nil
+	return string(resBody), checkForErrors(res, string(resBody))
+}
+
+func checkForErrors(res *http.Response, body string) error {
+	switch res.StatusCode {
+	case http.StatusOK, http.StatusCreated, http.StatusNoContent:
```
bacongobbler: we should probably change this to be anything in the 200 status range

Joshua-Anderson:
Unfortunately, I discovered that some of the errors now printed are too generic to be useful: `Error: This field must be unique.` So I had to add in the key so now errors are printed like this:


```
Error:
email: This filed must be unique
```

client-go/controller/client/auth.go
```diff
@@ -129,16 +129,12 @@ func Passwd(username string, password string, newPassword string) error {
 		return err
 	}
 
-	resBody, status, err := client.BasicRequest("POST", "/v1/auth/passwd/", body)
+	_, err = client.BasicRequest("POST", "/v1/auth/passwd/", body)
```
mboersma: This could be collapsed with the `if err != nil` statement below, same for lines 150 and 182.

mboersma:
This works well:
```console
$ deis create deis
Creating Application... Error: 
id: deis is a reserved name.
400 BAD REQUEST
$ deis login http://deis.local3.deisapp.com --username=fake --password=fake
Error: 
non_field_errors: Unable to log in with provided credentials.
400 BAD REQUEST
```

Code LGTM, modulo the couple of comments outstanding.

Joshua-Anderson:
I compacted error checks (they were all over the code, as many methods now only check for errors) and cleaned up a flakey tests (iteration over the json map is random).

bacongobbler:
code LGTM

client-go/controller/client/http.go
mboersma: Should this still be changed so we're not just catching a few specific HTTP codes?

bacongobbler: I think so. Anything the API throws at us in the future should work. Checking the range of non-error response codes would be an easy enhancement.

PR #4137, opened 2015-07-28
feat(commit-msg): don't prevent a commit, only warn

Joshua-Anderson:
That way you don't loose an entire commit msg if you make a mistake.

### DIFF
--- a/contrib/util/commit-msg
+++ b/contrib/util/commit-msg
@@ -39,7 +39,7 @@ if ! [[ $SUBJECT =~ $subject_regex ]]; then
 	echo "chore"
 	echo ""
 	echo "Read more at http://docs.deis.io/en/latest/contributing/standards/$NORMAL"
-	exit 1
+	exit 0
 fi
 
 if ! [[ $SUBJECT =~ $capital_regex ]]; then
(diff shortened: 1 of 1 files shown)


bacongobbler:
LGTM

mboersma:
I prefer the strict version actually, because with this one I'm likely to see the warnings but just commit things anyway. I was already in the habit of copying my commit message in case it got shot down...

technosophos:
Could we at least echo back the message on error so that we can copy/paste it back in and then edit?

carmstrong:
> Could we at least echo back the message on error so that we can copy/paste it back in and then edit?

:+1: I think failing hard and outputting the offending commit message is probably a reasonable middle ground.

mboersma:
I'm ok with this change if this is what others would prefer. I can always tweak my local commit hook to be stricter.

Joshua-Anderson:
My logic for making everything a warning was I figured a big block of red text was pretty obvious, and it is easier to amend a commit message if you mess it up than it is to copy and paste it and then fix it.

I don't have a strong opinion either way, as copying and pasting the commit message isn't ***too*** much work :trollface:.

mboersma:
Code LGTM.

PR #4138, opened 2015-07-29
feat(client-go): add minimal windows support

Joshua-Anderson:
This PR adds very minimal windows support to client-go.

1. Add a batchfile so it can be built on windows.
2. `apps:open` launches the default web browser.
3. Uses `%HOMEDRIVE%` and `%HOMEPATH%` on windows instead of `$HOME`.

Current issues:

1. Test will not pass on windows, as tests expect unix behavior.

### DIFF
--- a/client-go/cmd/keys.go
+++ b/client-go/cmd/keys.go
@@ -3,7 +3,6 @@ package cmd
 import (
 	"fmt"
 	"io/ioutil"
-	"os"
 	"path"
 	"regexp"
 	"strconv"
@@ -129,7 +128,7 @@ func chooseKey() (api.KeyCreateRequest, error) {
 }
 
 func listKeys() ([]api.KeyCreateRequest, error) {
-	folder := path.Join(os.Getenv("HOME"), ".ssh")
+	folder := path.Join(client.FindHome(), ".ssh")
 	files, err := ioutil.ReadDir(folder)
 
 	if err != nil {
--- a/.gitignore
+++ b/.gitignore
@@ -27,6 +27,7 @@ cache/image/bin/
 client/dist/
 client/makeself/
 client-go/deis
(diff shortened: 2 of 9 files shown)


client-go/cmd/keys.go
```diff
@@ -129,7 +128,7 @@ func chooseKey() (api.KeyCreateRequest, error) {
 }
 
 func listKeys() ([]api.KeyCreateRequest, error) {
-	folder := path.Join(os.Getenv("HOME"), ".ssh")
+	folder := path.Join(client.FindHome(), ".ssh")
```
mboersma: I guess we can't use `os.user.Current().HomeDir` since we build this as a static binary.

mboersma:
Code LGTM.

bacongobbler:
I'm all for adding support to more platforms, even if all the features won't work on it just yet.

Would you mind documenting the workaround for the current issues by writing up new issues for those? :)

Joshua-Anderson:
I fixed the issue with password input on windows (windows doesn't use 0 for stdin), so everything should work now. Somebody else with a windows computer should also manually test a bit if possible as well.

bacongobbler:
I have a windows box but it's currently upgrading to windows 10 in the background, so I won't be able to test this until later.

mboersma:
Code still LGTM, but I don't have access to a Windows box currently to test this. :floppy_disk:

carmstrong:
Code LGTM. Do we absolutely need to manually test this ourselves? Our official position is that the client isn't supported on Windows, but *may* run. I'd be okay with letting users give this a shot and providing feedback.

mboersma:
> Do we absolutely need to manually test this ourselves?

Not in this case, since it's a preview release and Windows isn't officially supported anyway.

PR #4143, opened 2015-07-29
docs(managing_deis): add section for registrationMode

mdolian:
This commit adds a section to the managing_deis/production_deployment documentation
that explains how to disable or set registrations to adminOnly.

### DIFF
--- a/docs/managing_deis/production_deployments.rst
+++ b/docs/managing_deis/production_deployments.rst
@@ -53,6 +53,25 @@ Back up data
 
 Backing up data regularly is recommended. See :ref:`backing_up_data` for steps.
 
+Change Registration Mode
+------------------------
+
+Changing the registration process is highly recommended in production.  By default, registrations for a
+new cluster are open to anyone with the proper URL.  Once the admin user has registered on a new cluster,
+it is recommended that you either turn off registrations entirely or enable the adminOnly registration feature.
+
+Turn off registrations:
+
+.. code-block:: console
+
+   $ deisctl config platform set registrationMode=disable
+
+Turn on admin only registrations:
+
+.. code-block:: console
+
+   $ deisctl config platform set registrationMode=adminOnly
+
 Configure logging and monitoring
 --------------------------------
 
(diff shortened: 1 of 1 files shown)


docs/managing_deis/production_deployments.rst
```diff
@@ -53,6 +53,25 @@ Back up data
...
+
+Changing the registration process is highly recommended in production.  By default, registrations for a
+new cluster are open to anyone with the proper URL.  Once the admin user has registered on a new cluster,
+it is recommended that you either turn off registrations entirely or enable the adminOnly registration feature.
+
+Turn off registrations:
+
+.. code-block:: console
+
+   $ deisctl config platform set registrationMode=disable
```
Joshua-Anderson: Should be `disabled`: http://docs.deis.io/en/latest/customizing_deis/controller_settings/#changing-the-registration-mode

carmstrong: :+1: We should probably link to that page at the bottom of this documentation, anyhow: "For details on changing hte registration mode, see: http://docs.deis.io/en/latest/customizing_deis/controller_settings/#changing-the-registration-mode"

krancour:
@mdolian thanks for the contribution!  The existing instructions on modifying the reg mode are buried just deep enough to easily overlook.  Calling out the need for this from a more central place in the documentation is a big improvement.

You mention the admin only registration mode, but don't show how to enable that.  Maybe instead of adding that detail, you could consider striking the code block altogether since it partially repeats documentation that already exists elsewhere in the tree.  That would leave you with a blurb calling out the need to restrict registration to which you can add a link to the relevant technical instructions (the doc @Joshua-Anderson referenced above).

mdolian:
I took @krancour suggestion and removed the code blocks and just referenced the controller documentation.  Let me know what you guys think.

mboersma:
LGTM. Thanks @mdolian--this is definitely worth making more visible in the docs.

carmstrong:
LGTM!

krancour:
@mdolian, thanks for this!

PR #4145, opened 2015-07-29
Bump to etcd 2.1.1; properly test it with Jenkins

### LINKED ISSUE #4135: Test fixtures should be built locally with the git-asdf1234 tags
Currently, `test-postgresql` and `test-etcd` are built as an automated build on Docker Hub. `mock-store` is built and pushed manually.

In #1351, we added an optimization to only pull these images if they don't already exist. For changes to `test-postgresql` or `test-etcd`, this means that we'll continue to use the old images until someone manually nukes the images on the test boxes. 

All three of these test fixtures should be refactored to:
- Not push to Docker Hub (disable automated builds)
- Build locally with the same `git-asdf1234` hash as the rest of the components
- Use the `git-asdf1234` image name in test runs, ensuring we always use the latest version

carmstrong:
replaces #4121
closes #4135

### DIFF
--- a/builder/Makefile
+++ b/builder/Makefile
@@ -67,7 +67,7 @@ test-unit:
 	$(GOTEST) .
 
 test-functional:
-	@docker history deis/test-etcd >/dev/null 2>&1 || docker pull deis/test-etcd:latest
+	@$(MAKE) -C ../tests/ test-etcd
 	godep go test -v ./tests/...
 
 test-style:
--- a/cache/Makefile
+++ b/cache/Makefile
@@ -67,5 +67,5 @@ test-style:
 	$(GOLINT) ./...
 
 test-functional:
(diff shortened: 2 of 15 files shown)


mboersma:
Code LGTM.

carmstrong:
@mboersma I'll defer to you as to whether we ship this for 1.9. I'd argue that we should bump to the latest etcd version when we're making the big etcd version jump (especially since there are substantial improvements), but 2.0.13 has admittedly been vetted for a bit longer.

slack:
One argument for v2.1.1 is that it has fixed a few bugs related to leaking fds.

carmstrong:
> One argument for v2.1.1 is that it has fixed a few bugs related to leaking fds.

Yeah. I'm leaning on the side of shipping this in 1.9.

mboersma:
:ship:

aledbf:
Code LGTM.

PR #4153, opened 2015-07-30
fix(deisctl): allow help for ssh and dock

mboersma:
"deisctl ssh" and "deisctl dock" need open-ended command-line parsing to allow unquoted commands to be run. But this short-circuits the normal way docopt would print the usage string.

Currently `deisctl help dock` or `deisctl --help ssh` or variants won't display their detailed usage strings. This fixes that, but still allows those two commands to swallow everything that follows on the command line.

The code compares `target` only to "--help" because earlier parsing code translated "help" and "-h" to "--help"

### DIFF
--- a/deisctl/client/client.go
+++ b/deisctl/client/client.go
@@ -252,12 +252,19 @@ Usage:
 		return err
 	}
 
+	target := args["<target>"].(string)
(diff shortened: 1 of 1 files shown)


carmstrong:
Code LGTM.

bacongobbler:
code LGTM

PR #4157, opened 2015-07-31
fix(tests): use command line flags to test auth:passwd

Joshua-Anderson:
gexpect was unreliable and causing errors with the go client. The exact reason for the error is uncertain, but it seems like the go client exited before gexpect read the final output.


Joshua-Anderson:
Fixed up the Godep issue.

mboersma:
Code LGTM.

krancour:
Code LGTM

PR #4163, opened 2015-07-31
fix(client-go): paginate over lists

### LINKED ISSUE #3911: bug(client): deis apps only shows 100 apps
The client is not using the pagination to return all the apps

Joshua-Anderson:
Right now, if you have over 100 apps, deis only lists the first 100, now all of the apps are displayed.

Fixes #3911

### DIFF
--- a/client-go/cmd/apps.go
+++ b/client-go/cmd/apps.go
@@ -55,20 +55,24 @@ func AppCreate(id string, buildpack string, remote string, noRemote bool) error 
 }
 
 // AppsList lists apps on the Deis controller.
-func AppsList() error {
+func AppsList(results int) error {
 	c, err := client.New()
 
 	if err != nil {
 		return err
 	}
 
-	apps, err := apps.List(c)
+	if results == defaultLimit {
+		results = c.ResponseLimit
+	}
+
+	apps, count, err := apps.List(c, results)
 
 	if err != nil {
 		return err
 	}
 
-	fmt.Println("=== Apps")
+	fmt.Printf("=== Apps%s", limitCount(len(apps), count))
 
 	for _, app := range apps {
 		fmt.Println(app.ID)
--- a/client-go/controller/client/client.go
+++ b/client-go/controller/client/client.go
@@ -26,13 +26,21 @@ type Client struct {
 
 	// Username is the name of the user performing requests against the API.
 	Username string
+
+	// ResponseLimit is the number of results to return on requests that can be limited.
+	ResponseLimit int
 }
+
+// DefaultResponseLimit is the default number of responses to return on requests that can
+// be limited.
+var DefaultResponseLimit = 100
 
 type settingsFile struct {
 	Username   string `json:"username"`
 	SslVerify  bool   `json:"ssl_verify"`
 	Controller string `json:"controller"`
 	Token      string `json:"token"`
+	Limit      int    `json:"response_limit"`
 }
 
 // New creates a new client from a settings file.
@@ -62,15 +70,23 @@ func New() (*Client, error) {
 		return nil, err
 	}
 
+	if settings.Limit <= 0 {
+		settings.Limit = DefaultResponseLimit
+	}
+
 	return &Client{HTTPClient: CreateHTTPClient(settings.SslVerify), SSLVerify: settings.SslVerify,
-		ControllerURL: *u, Token: settings.Token, Username: settings.Username}, nil
+		ControllerURL: *u, Token: settings.Token, Username: settings.Username,
+		ResponseLimit: settings.Limit}, nil
 }
 
 // Save settings to a file
 func (c Client) Save() error {
-	settings := settingsFile{Username: c.Username,
-		SslVerify:  c.SSLVerify,
-		Controller: c.ControllerURL.String(), Token: c.Token}
+	settings := settingsFile{Username: c.Username, SslVerify: c.SSLVerify,
+		Controller: c.ControllerURL.String(), Token: c.Token, Limit: c.ResponseLimit}
+
+	if settings.Limit <= 0 {
+		settings.Limit = DefaultResponseLimit
+	}
 
 	settingsContents, err := json.Marshal(settings)
 
--- a/client-go/controller/client/client_test.go
+++ b/client-go/controller/client/client_test.go
@@ -8,7 +8,7 @@ import (
 	"testing"
 )
 
-const sFile string = `{"username":"t","ssl_verify":false,"controller":"http://d.t","token":"a"}`
+const sFile string = `{"username":"t","ssl_verify":false,"controller":"http://d.t","token":"a","response_limit": 50}`
 
 func createTempProfile(contents string) error {
 	name, err := ioutil.TempDir("", "client")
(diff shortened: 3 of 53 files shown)


bacongobbler:
coed LGTM

krancour:
Hypothetically, if an app had 1,000 releases (not at all unrealistic in the era of continuous deployments), this means the client makes 10 HTTP requests to stitch together the full list of 1,000 releases?  This doesn't seem ideal and seems as if it may defeat the purpose of paginating those responses in the first place.

What would be your thoughts on allowing users to interactively scroll through the results, triggering new page requests only when they scroll past the end of what's already been received?  (Similar to what happens, for instance, with large result sets in `psql`.)  I know this is easier said than done, but does it make more sense for the users?

carmstrong:
> What would be your thoughts on allowing users to interactively scroll through the results

I think that would be difficult for API clients. I think it's up the the client what it wants - it should be able to specify a "results per page". So an API client could request 10,000 results per page to get everything in one request.

krancour:
> So an API client could request 10,000 results per page to get everything in one request.

But that doesn't answer the question of how _this_ client should deal with large result sets.  Regardless of whether it makes 1 request to get 1,000 results, or 10 requests to get 1,000 results, the user probably doesn't want to _see_ 1,000 results-- which means in either case, we pulled down too much stuff.

carmstrong:
Maybe for this case, cap it to 100 or something by default, and display a message like "More than 100 apps exist." That cap could be toggled by a parameter to `deis ps`.

krancour:
@carmstrong :+1: same goes for `deis apps`, `deis releases`, etc.

krancour:
```
Options:
  -l --limit=<limit>
    the maximum number of results to display; defaults to 1000
```

carmstrong:
:+1:

Joshua-Anderson:
I'm now using the limit based approach to paginate!

client-go/controller/models/apps/apps_test.go
```diff
@@ -291,7 +291,7 @@ func TestAppsList(t *testing.T) {
 
 	client := client.Client{HTTPClient: httpClient, ControllerURL: *u, Token: "abc"}
 
-	actual, err := List(&client)
+	actual, err := List(&client, 100)
```
bacongobbler: I feel like this [magic number](https://en.wikipedia.org/wiki/Magic_number_%28programming%29) should be part of the client config with a default of 100. That way users can tune this number by modifying ~/.deis/client.json

Joshua-Anderson: Good Idea! I implemented it. In this particular location these are just the tests and I just gave a random value as it doesn't affect the response.

krancour:
I like the implementation overall.  My only outstanding concern is that with the limit-based approach, the frequent use of "paginate" or "paginated" in the code (e.g. in function names) strongly hints at the existence of a feature that is _not_ there-- the ability to actually "turn the page" so-to-speak (which would require handling an offset in addition to the limit).  I don't think that feature needs to be added or anything, but adapting the function names to no longer imply responses are "pageable" might be a good thing.

Joshua-Anderson:
@krancour I agree. I renamed the functions to be less ambiguous.

krancour:
@Joshua-Anderson on the controller end, would `limit` be a better query param than `page_size`?  Notion of pages still appears in some comments, too.

Joshua-Anderson:
@krancour The reason I went with `?page_size` is because it is a standard and used by almost everything that uses pagination. Most of the time it is used like `\paginate?page=2&page_size=80`.

> Notion of pages still appears in some comments, too.

If so, it's accidental and please leave a comment at the location so I can fix it.

krancour:
> The reason I went with ?page_size is because it is a standard and used by almost everything that uses pagination.

So is the idea here that the controller will allow for pagination, even though the client really only supports retrieval of one page of specifiable size?

If so, I think reference to "page" or "paginated" make sense on the controller side, and I'll look for just the errant references to "page" or "paginated" on the CLI side.

client-go/controller/client/client.go
```diff
@@ -26,13 +26,20 @@ type Client struct {
 
 	// Username is the name of the user performing requests against the API.
 	Username string
+
+	// ResponseLimit is the number of results to return on a paginated requests.
```
krancour: "paginated" might confuse here.  The controller supports it, but the CLI can only ever fetch one "page," even though its size is specifiable.  Same comment applies on line 34.

Joshua-Anderson:
@krancour Done!

krancour:
LGTM

client-go/cmd/apps.go
```diff
@@ -55,14 +55,18 @@ func AppCreate(id string, buildpack string, remote string, noRemote bool) error
...
-func AppsList() error {
+func AppsList(results int) error {
 	c, err := client.New()
 
 	if err != nil {
 		return err
 	}
 
-	apps, err := apps.List(c)
+	if results == -1 {
```
mboersma: Should this magic number `-1` be a constant instead, like `DEFAULT_LIMIT`?

client-go/controller/client/client_test.go
```diff
@@ -8,7 +8,7 @@ import (
 	"testing"
 )
 
-const sFile string = `{"username":"t","ssl_verify":false,"controller":"http://d.t","token":"a"}`
+const sFile string = `{"username":"t","ssl_verify":false,"controller":"http://d.t","token":"a", "response_limit": 50}`
```
mboersma: Minor, but could you omit the spaces in this JSON string? It just reads strangely to have inconsistent spacing here.

mboersma:
Needs a rebase against master now. :repeat_one:

Joshua-Anderson:
@mboersma Rebased!

carmstrong:
@bacongobbler @mboersma Would you mind re-reviewing this? I don't feel comfortable enough with the code.

mboersma:
LGTM.

krancour:
I've been banging pretty hard on the k8s scheduler.  Something I can easily do now that wasn't previously realistic is scale an app to hundreds of containers very quickly.  In one case, I scaled to 200.  `deis ps` only shows me the first hundred (and I know that's because 100 is the default limit we discussed above), but I see no indication that I'm viewing a partial list.  Would it possibly make sense, for certain queries, to return and display a summary of the full result set?

Joshua-Anderson:
@krancour Could you give some example output for what you would like to see?

krancour:
So, immediately after scaling, I see this:

```
[kent@mbp example-go]$ deis scale web=200
done in 94s
=== jicama-jamboree Processes

--- web: 
web.1 up (v2)
... omitted for brevity...
web.100 up (v2)
```

This shows me 100 processes, even though I just scaled to 200.  For anyone who doesn't know about limits, this looks very odd.  I see the same result as above when using `deis ps`.

It seems a lot of clarity could be added by having the output resemble something like this:

```
[kent@mbp example-go]$ deis scale web=200
done in 94s
=== jicama-jamboree Processes (100 of 200)

--- web: 
web.1 up (v2)
... omitted for brevity...
web.100 up (v2)
```

That sort of `(m of n)` info could be generally useful and insightful for other queries as well.  e.g. `deis apps` or `deis users`.  In the case of `deis ps`, it would actually be possible to also summarize the status of the processes with an additional line that shows the count in each status-- how many are up, how many are down, etc., but that seems like it's probably not totally necessary.  If we're at least making sure users _know_ they're seeing only a fragment of the results, they can at least choose to query again with a higher limit and maybe `grep` it themselves for whatever it is they are looking for.

Joshua-Anderson:
@krancour  :+1: Like the idea.

Joshua-Anderson:
@krancour I added the `(100 of 200)` message to every limited request (expect certs, because it has a different layout.

krancour:
:+1: I'll re-review as soon as I've got a chance, but let me preemptively say-- nice work on this.

bacongobbler:
manually tested, works as intended:

```
><> deis ps:list -a go
=== go Processes (100 of 101)
```

LGTM!

krancour:
LGTM

PR #4164, opened 2015-07-31
feat(godeps): cleanup godeps

Joshua-Anderson:
I just ran `godep save ./...` and let it cleanup some unused packages.


krancour:
@Joshua-Anderson this keeps failing CI.  Mysteriously, I am unable to replicate the failure locally, but it does look related to `chrootarchive ` being removed from the godeps.  This also is in need of a rebase... let's rebase and let the tests have another go.

mboersma:
We have also updated go-etcd and removed gexpect recently, so this PR will need to catch up with those changes.

Joshua-Anderson:
I've tried everything I can think off and can't reproduce this error locally. I even deleted the `docker/docker` src in my local gopath, so I know go isn't falling back on it. I updated the code to remove the latest packages. Maybe this is due to the node using a older version of godeps?

bacongobbler:
Hmm, I wonder if there's a stray `Godeps` directory in `deis/deis/builder` on that node. That could explain the failures. It passed locally for me too with a `make -C builder build test-functional`.

PR #4165, opened 2015-07-31
fix(client-go): accept all non error HTTP codes

### LINKED ISSUE #4162: new deis CLI should check for entire range of http errors
The new go CLI code checks for some specific HTTP error codes, but should be expanded to consider everything from 200-399. See https://github.com/deis/deis/blob/master/client-go/controller/client/http.go#L104.

Joshua-Anderson:
Fixes #4162


carmstrong:
Code LGTM.

mboersma:
Code LGTM.

PR #4168, opened 2015-07-31
Fix swarm preview

### LINKED ISSUE #4159: bug: swarm preview broken by etcd2
I looked briefly... it doesn't seem the swarm preview gets tested during integration tests.  Am I mistaken?

When using it locally (from the master branch), I see errors like those below repeated over and over for _both_ `deis-swarm-manager` and `deis-swarm-node`.

```
Jul 31 01:36:12 deis-02 systemd[1]: Starting deis-swarm-node...
Jul 31 01:36:12 deis-02 systemd[1]: Started deis-swarm-node.
Jul 31 01:36:12 deis-02 sh[9830]: ./deis-swarm join --addr=172.17.8.101:2375 etcd:///deis/scheduler/swarm/node
Jul 31 01:36:12 deis-02 sh[9830]: error at command wait
Jul 31 01:36:12 deis-02 sh[9830]: 2015/07/31 01:36:12 exit status 1
Jul 31 01:36:12 deis-02 systemd[1]: deis-swarm-node.service: main process exited, code=exited, status=1/FAILURE
Jul 31 01:36:12 deis-02 systemd[1]: Unit deis-swarm-node.service entered failed state.
Jul 31 01:36:12 deis-02 systemd[1]: deis-swarm-node.service failed.
Jul 31 01:36:17 deis-02 systemd[1]: deis-swarm-node.service holdoff time over, scheduling restart.
```

mboersma:
The etcd -> etcd2 migration broke the Swarm scheduler preview, which @aledbf had already anticipated and fixed. This updates the logic used to obtain the etcd leader. Tested on vagrant. Thanks to @smothiki for all the debugging and help.

Note that this isn't backward-compatible, so users upgrading to the next Deis release who were already using the Swarm scheduler preview will need to continue using the :v1.8.0 swarm units in order to talk to etcd v1. I think that's acceptable for a tech preview, but let me know if you disagree--we could keep the old discovery strategy around in branching code.

Closes #4159.


carmstrong:
Code LGTM. This reminds me that the `debug-etcd` service likely broke with etcd2 too. I'mma take a look now.

smothiki:
Code LGTM

carmstrong:
Latest test run: https://ci.deis.io/job/test-ec2/2506/

swarm/swarm.go
```diff
@@ -51,19 +58,32 @@ func run(cmd string) {
...
+	resp, _ = client.Get("http://" + os.Getenv("HOST") + ":2379/v2/members")
+	defer resp.Body.Close()
+
+	body, _ = ioutil.ReadAll(resp.Body)
+	json.Unmarshal(body, &etcdCluster)
+
+	for _, node := range etcdCluster.Members {
+		if node.ID == etcdLeaderID {
+			u, err := url.Parse(node.ClientURLs[0])
+			if err == nil {
```
bacongobbler: should we be throwing an error in this case? Discarding the error sounds like we could run into trouble later on when we can't figure out why it's not booting.

mboersma: Yes, we should. However, the code already discarded such errors and expected an empty string in those cases, so I didn't attempt to rewrite everything, just strip the "http://" off the host:port string.

PR #4169, opened 2015-07-31
ref(client-go): break up controller/client package

Joshua-Anderson:
This makes the code easier to follow and gives go bindings for auth functions


Joshua-Anderson:
Previously this package did many thing and was very ugly. This way all `controller/client` does is setup session based on the preferences file.

bacongobbler:
code LGTM

mboersma:
Code LGTM.

Issue #4161, opened 2015-07-31
k8s: Pods should not be deployed into default namespace.


technosophos:
Each pod in a namespace is given information about every other pod in that namespace. This applies to the default namespace. So if I deploy two apps, each one has env vars like this:

```
APP_14334_PORT=tcp://10.100.119.52:80
APP_14334_PORT_80_TCP=tcp://10.100.119.52:80
APP_14334_PORT_80_TCP_ADDR=10.100.119.52
APP_14334_PORT_80_TCP_PORT=80
APP_14334_PORT_80_TCP_PROTO=tcp
APP_14334_SERVICE_HOST=10.100.119.52
APP_14334_SERVICE_PORT=80
APP_28146_PORT=tcp://10.100.37.242:80
APP_28146_PORT_80_TCP=tcp://10.100.37.242:80
APP_28146_PORT_80_TCP_ADDR=10.100.37.242
APP_28146_PORT_80_TCP_PORT=80
APP_28146_PORT_80_TCP_PROTO=tcp
APP_28146_SERVICE_HOST=10.100.37.242
APP_28146_SERVICE_PORT=80
DEIS_APP=kabuki-quadrant
DEIS_RELEASE=v13
GIT_SHA=3e1d7e90c8ceec6ecbc34a1e1edc31c5ff892d9e
HOME=/root
...
```

It would be better practice for the preview release if we just put every _app_ into its own namespace (so `APP_14334` is also in the namespace `APP_14334`).

This would require making hopefully simple changes to the RC and Service definitions. But it also requires explicitly creating a new namespace.

gabrtv:
@technosophos wouldn't this prevent inter-app discovery/communication?

technosophos:
Currently, if you deploy an app using any other scheduler in Deis, that app has env vars that pertain to itself.

The k8s preview, though, sets 7 environment variables for _every single app_ that is deployed in Deis. Every single app gets information on how to connect to every other app throughout Deis.

To make k8s work the way current Deis works, we should give each app its own namespace so that it will (by default) only know about itself.

In the future, I think we can do some fantastic things with the namespacing feature. But for the preview, I feel like we should stay as close to the current Deis scheme as we can.

mboersma:
Fixed by #4170.

PR #4170, opened 2015-07-31
fix(k8s): deploy app into individual namespace

kmala:


kmala:
fixes #4161

aledbf:
@kmala can this be optional?

kmala:
I am not sure. @technosophos can you please answer?

bacongobbler:
code LGTM. The rationale behind this is that we want the tech preview to emulate Deis as it is today, which means that apps must be deployed in their own individual namespace. We can address this in the future with something like #4173, which apps can be deployed to the same team (or organization/environment, whatever makes the most sense) and they can share the same namespace.

technosophos:
Code LGTM.

@aledbf This change is just for the preview release. We're planning on allowing explicit control of namespaces in the future. But what we're after now is what @bacongobbler said: We don't want to surprise people by having k8s behave a lot differently than the others.

PR #4171, opened 2015-07-31
fix(contrib): fix debug-etcd

carmstrong:
Hitting the /v2/stats/leader endpoint would only work on the etcd
leader. Instead, we hit /v2/stats/self which will always work.This
doesn't show stats from the other members, but a user could start
the service on all hosts.


carmstrong:
We'll actually get `jq` on the host once we bump past CoreOS [717.0.0](https://coreos.com/releases/#717.0.0).

mboersma:
Code LGTM.

smothiki:
Code LGTM

PR #4174, opened 2015-07-31
fix(builder): always configure SSH key

### LINKED ISSUE #4172: SSH_KEY is ignored when not specifying BUILDPACK_URL
Per IRC conversation with @Joshua-Anderson (:beers:).

My original issue was that [the documentation](http://docs.deis.io/en/latest/using_deis/using-buildpacks/#using-private-repositories) seems to imply that setting `SSH_KEY` would allow a user to reference private Github repositories over SSH in the build process. In practice, I was met with `Host key verification failed. fatal: Could not read from remote repository.` when attempting to do so.

[The code](https://github.com/deis/deis/blob/34cce8351f928ad01522ecb4548cdd9e1c35eb33/builder/rootfs/usr/local/src/slugbuilder/builder/build.sh#L90) pertaining to `SSH_KEY` is currently nested inside the `BUILDPACK_URL` block and will only be executed if you specify `BUILDPACK_URL`.

As a work-around, a user can set `BUILDPACK_URL` to the official Heroku buildpack repository which causes the `SSH_KEY` code to be run.

Trivial PR to follow.

clayzermk1:
SSH is now configured regardless of whether a custom buildpack was specified.

Fixes #4172.

### DIFF
--- a/builder/rootfs/usr/local/src/slugbuilder/builder/build.sh
+++ b/builder/rootfs/usr/local/src/slugbuilder/builder/build.sh
@@ -69,6 +69,19 @@ export HOME="$app_dir"
 export REQUEST_ID=$(openssl rand -base64 32)
 export STACK=cedar-14
 
+## SSH key configuration
+
+if [[ -n "$SSH_KEY" ]]; then
+    mkdir -p ~/.ssh/
+    chmod 700 ~/.ssh/
+
+    echo $SSH_KEY | base64 -d > ~/.ssh/id_rsa
+    chmod 400 ~/.ssh/id_rsa
+
+    echo 'StrictHostKeyChecking=no' > ~/.ssh/config
+    chmod 600 ~/.ssh/config
+fi
+
 ## Buildpack detection
 
 buildpacks=($buildpack_root/*)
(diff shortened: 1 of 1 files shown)


deis-admin:

Joshua-Anderson:
I don't think that we have tests that cover private buildpacks or ssh keys, we should look into that in the future.

This code is also not touched by @technosophos's builder go rewrite.

technosophos:
Yeah, this code should not impact the new builder, and I think it makes sense. I don't immediately see any nasty side-effects from doing this.

Code LGTM

mboersma:
Jenkins, add to whitelist.

carmstrong:
Change makes sense. Code LGTM.

mboersma:
@clayzermk1 could you amend the commit message to agree with our [style guide](http://docs.deis.io/en/latest/contributing/standards/#commit-style)? Something like:
```
fix(builder): always configure SSH key
```

You can also do `make commit-hook` to install a git script that will warn you if the commit message doesn't comply with project standard.

clayzermk1:
Sure thing. Sorry, I read the commit style guide and took a stab at it. Nice to know about the hook.

carmstrong:
> I read the commit style guide and took a stab at it

:100: Points for reading and taking a stab!

carmstrong:
@clayzermk1 Should be `fix(builder)`, not `fix(4172)` :)

clayzermk1:
Reading - how does it work.

mboersma:
Code LGTM.

bacongobbler:
code LGTM

clayzermk1:
Thanks for the merge and such a killer community and product! Cheers all :beers:

carmstrong:
> Thanks for the merge and such a killer community and product!

We have a killer community thanks to people like you! <3

PR #4175, opened 2015-07-31
feat(deisctl): add graceful upgrade

krancour:
Alternate implementation of #4099 

The main difference here is that deisctl has been significantly refactored to avoid some of the uglier hacks that would have been required to make the feature work and test cleanly.

Most credit to @iancoffey

Ping @carmstrong


krancour:
To proactively address questions that will probably arise, the config package was a significant pain point prior to this refactor.  It contained a model for configuration data, an interface for (potentially) multiple implementations of a config backend, and one implementation of that based on etcd.  Having everything in one package like that made it tricky to code to the interface, agnostic of the implementation and made it difficult to create alternate implementations (e.g. an in-memory implementation intended to aid in testing).  The refactor was aimed at breaking that package up cleanly to allow better reuse and testing.

iancoffey:
@krancour, this might be nitpicky but could you can repush this with your refactor stuff as a separate commit please? As it is, its difficult to review your backend refactor. Also we lose the commit history.

carmstrong:
> could you can repush this with your refactor stuff as a separate commit

:+1: We generally only add a new commit or two onto an existing commit stream when we take over PRs, to preserve the original author's commits.

docs/managing_deis/upgrading-deis.rst
```diff
@@ -63,6 +63,86 @@ Finally, update ``deisctl`` to the new version and reinstall:
...
+
+.. note::
+
+    The process utilizes a rolling restart of the platform's router components.
+    No more than one router instance is ever down at a time, but each stops and
+    restarts in fairly rapid succession.  Therefore, your load balancer
+    configuration is the determining factor in how much downtime, if any, will
+    occur during a successful upgrade.
+
+    Load balancers that are quick to remove a host from service after just one
```
iancoffey: A bit of a run-on. I think the original wording might have been easier to on the eyes.

krancour:
@iancoffey @carmstrong sorry... rushing to get this done.  i might have cut some corners in preserving the commit history.  I'll try to fix this asap.

docs/managing_deis/upgrading-deis.rst
```diff
@@ -63,6 +63,86 @@ Finally, update ``deisctl`` to the new version and reinstall:
...
+
+First, a new deisctl version (reflecting the desired version to upgrade
+to) should be installed to a temporary location. Care should be taken not to
+overwrite the existing deisctl version.
+
+.. code-block:: console
+
+    $ mkdir /tmp/upgrade
+    $ curl -sSL http://deis.io/deisctl/install.sh | sh -s 1.10.0 /tmp/upgrade
+    $ /tmp/upgrade/deisctl --version  # should match the desired version 1.10.0
```
iancoffey: Might be confusing to the folks, since 1.10.0 wont exist yet will it?

krancour: I thought it was potentially confusing to readers that the feature is for upgrading _from_ m >= 1.9.0 to n > m, but the docs showed an upgrade _to_ 1.8.0.

We could add a note that 1.10.0 does not exist yet.

This is why I was originally in favor of holding off on this documentation altogether until there was something to upgrade _to_ from 1.9.0.

krancour:
@iancoffey @carmstrong commit history should be cleaner, doc issues are addressed, and I managed to do a tiny bit more cleanup on the deisctl refactor too.  If tests pass, this is ready for review.

docs/managing_deis/upgrading-deis.rst
```diff
@@ -68,6 +68,77 @@ Upgrade Deis clients
 As well as upgrading ``deisctl``, make sure to upgrade the :ref:`deis client <install-client>` to
 match the new version of Deis.
 
+Graceful Upgrade
+----------------
+
+Alternatively, an experimental feature exists to provide the ability to perform a graceful upgrade. This process is
+available for version 1.9.0 moving foward and is intended to facilitate upgrades within a major version - upgrading
```
carmstrong: "(for example, from v1.9.0 to v1.10.0)". Not everyone knows what semver major/minor/patch means.

docs/managing_deis/upgrading-deis.rst
```diff
@@ -68,6 +68,77 @@ Upgrade Deis clients
...
+.. code-block:: console
+
+    $ mkdir /tmp/upgrade
+    $ curl -sSL http://deis.io/deisctl/install.sh | sh -s 1.10.0 /tmp/upgrade
+    $ /tmp/upgrade/deisctl --version  # should match the desired platform
+    1.10.0
+    $ /tmp/upgrade/deisctl refresh-units
+    $ /tmp/upgrade/deisctl config platform set version=v1.10.0
+
+.. note::
```
carmstrong: While this note is accurate, I think it could be rewritten to be more conversational. Something like "For example: to upgrade from Deis v1.9.0 to v1.10.0...". I think users can understand what would need to be swapped out for various versions.

carmstrong:
LGTM barring my comments. Would like @iancoffey to sign off on this, too.

iancoffey:
:shipit: 

im going to make an issue to re-eval deisctl etcd stuff, separately.

docs/managing_deis/upgrading-deis.rst
krancour: @carmstrong good call.  Guess I take semantic versioning for granted.  Fixing.

krancour:
@carmstrong your latest doc feedback has been addressed.  Just waiting on the tests-- these pass locally, but have not passed in Jenkins at any point yet.

bacongobbler:
haven't reviewed the code, but the UX works according to the docs. I could still ping my app after `upgrade-prep` was finished.

carmstrong:
> I could still ping my app after upgrade-prep was finished.

Did you finish the upgrade? We need to make sure the entire workflow works well.

bacongobbler:
Yes. `upgrade-prep` and `upgrade-takeover` upgraded my cluster from master to this branch. I had to use the new client for both operations obviously, but in practice using the "old" client should be used when calling `upgrade-prep`, then using the new client for `upgrade-takeover` as the docs describe.

bacongobbler:
code LGTM

krancour:
Manual testing of upgrade process shows it works as advertised.  Integration tests also pass locally.

carmstrong:
Latest test run: https://ci.deis.io/job/test-ec2/2509/

PR #4176, opened 2015-08-01
feat(logger, publisher, swarm): switch to scratch base image

Joshua-Anderson:
Since we compile go staticlly, we don't even need a linux distro, we can run using the `scratch` image. I'm not sure if this is the direction we want to go, since other components still depend on and have to download alpine, but it's worth considering.

What we loose:

- No sshing into container (no shell)
- No package installation

What we gain:

- super tiny (only the go executable < 10 mb), one layer images.


publisher/image/Dockerfile
```diff
@@ -1,8 +1,4 @@
-FROM alpine:3.1
-
-# install curl in the image so it is possible to get the runtime
-# profiling information without any additional package installation.
-RUN apk add --update-cache curl && rm -rf /var/cache/apk/*
```
Joshua-Anderson: @aledbf  Does it make sense to keep this in the image just for profiling? Is this something that could be exposed another way? Logs?

bacongobbler: We can't really expose it outside of the container. Some users run deis out on networks that exposes everything to the public (DigitalOcean)

mboersma:
I think it's preferable to continue to use alpine as a base image. AWS Elastic Beanstalk blacklists the `scratch` layer, I have read, and HTTPS requests may fail since there are no root cents in a blank image. Neither of those may be a practical problem for Deis here, but I think for support and consistency we should aim for everything sharing base layers.

carmstrong:
Agreed with @mboersma's points. Appreciate the idea, @Joshua-Anderson, but I think the disadvantages outweigh the advantages here.

Joshua-Anderson:
No problem. I just though it might be interesting to consider! :smiley:

carmstrong:
> I just though it might be interesting to consider!

There are no bad ideas! :)

PR #4179, opened 2015-08-03
Fixed typos

radarhere:


deis-admin:

bacongobbler:
Nice. These typos are actually [part](https://github.com/deis/deis/commit/975ac6bb840bd492d1863564691598ed91233868) of the [commit messages](https://github.com/deis/deis/commit/cb18218d3661ae94854166df759e59853273cd1b). I think in the future we can introduce something like pylint which can catch these spelling errors.

Would you mind amending your commit to follow the style guide? Something like:

    fix(changelog): correct spelling typos

carmstrong:
Thanks for keeping us honest, @radarhere! Unfortunately, as @bacongobbler mentioned, these were typos in the original commits that we didn't catch :(

The changelog should really be an append-only record of changes to the project, so I'm afraid we'll have to live with these typos for now, and be vigilant moving forward to prevent new ones from slipping in! We could certainly use your eyes on the project!

PR #4185, opened 2015-08-03
ref(controller): code quality enhancements

bacongobbler:
Small improvements over the changes made in #4127. These are small enhancements so there's no rush on these.

### DIFF
--- a/controller/api/models.py
+++ b/controller/api/models.py
@@ -381,7 +381,7 @@ class App(UuidAuditedModel):
         # HACK (bacongobbler): we need to wait until publisher has a chance to publish each
         # service to etcd, which can take up to 20 seconds.
         time.sleep(20)
-        for i in range(len(intervals)):
+        for i in xrange(len(intervals)):
             delay = int(config.get('HEALTHCHECK_INITIAL_DELAY', 0))
             try:
                 # sleep until the initial timeout is over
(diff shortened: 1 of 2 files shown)


mboersma:
Code LGTM.

controller/api/models.py
```diff
@@ -381,7 +381,7 @@ def _healthcheck(self, containers, config):
         # HACK (bacongobbler): we need to wait until publisher has a chance to publish each
         # service to etcd, which can take up to 20 seconds.
         time.sleep(20)
-        for i in range(0, 4):
+        for i in range(len(intervals)):
```
carmstrong: I think `len(intervals)` would get evaluated each iteration, even though the length doesn't change... so wouldn't it be better to evaluate this once and then use the result? *super* minor, though.

carmstrong:
Code LGTM.

controller/api/models.py
bacongobbler: `range` creates a copy of the list in-memory to be evaluated on the first iteration, so it won't be continuously evaluated. This also includes any modifications done to the list inside the loop, which will not be evaluated. :)

You do bring up a good point though. I know @mboersma prefers `xrange` since we're on 2.7. `xrange` will avoid creating a copy of the list by lazily evaluating the list, acting as a generator. I'll change `range` to `xrange` for consistency with the rest of the code.

mboersma: `xrange` probably doesn't matter for such a short list, but I use it consistently anyway, so :+1:

carmstrong:
Looks one the most recent one failed.

PR #4193, opened 2015-08-04
feat(controller): allow users to transfer app ownership.

Joshua-Anderson:
This is really useful if a user needs to delete their account, but they have apps that their company depends on attached to it. A user can use `apps:transfer` to transfer ownership of the app before deleting their account.

### DIFF
--- a/controller/api/views.py
+++ b/controller/api/views.py
@@ -224,6 +224,17 @@ class AppViewSet(BaseDeisViewSet):
         return Response(output_and_rc, status=status.HTTP_200_OK,
                         content_type='text/plain')
 
+    def update(self, request, **kwargs):
+        app = self.get_object()
+
+        if request.data.get('owner'):
+            if self.request.user != app.owner and not self.request.user.is_superuser:
+                raise PermissionDenied()
+            new_owner = get_object_or_404(User, username=request.data['owner'])
+            app.owner = new_owner
+        app.save()
+        return Response(status=status.HTTP_200_OK)
+
 
 class BuildViewSet(ReleasableViewSet):
     """A viewset for interacting with Build objects."""
--- a/client/cmd/apps.go
+++ b/client/cmd/apps.go
@@ -240,3 +240,24 @@ func AppDestroy(appID, confirm string) error {
 
 	return nil
 }
+
+// AppTransfer transfers app ownership to another user.
+func AppTransfer(appID, username string) error {
+	c, appID, err := load(appID)
+
+	if err != nil {
+		return err
+	}
+
+	fmt.Printf("Transferring %s to %s... ", appID, username)
+
+	err = apps.Transfer(c, appID, username)
+
+	if err != nil {
+		return err
+	}
+
+	fmt.Println("done")
+
+	return nil
+}
--- a/client/controller/api/apps.go
+++ b/client/controller/api/apps.go
@@ -15,6 +15,11 @@ type AppCreateRequest struct {
 	ID string `json:"id,omitempty"`
 }
 
+// AppUpdateRequest is the definition of POST /v1/apps/<app id>/.
+type AppUpdateRequest struct {
+	Owner string `json:"owner,omitempty"`
+}
+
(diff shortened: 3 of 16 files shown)


controller/api/views.py
```diff
@@ -210,6 +210,17 @@ def run(self, request, **kwargs):
...
+    def update(self, request, **kwargs):
+        app = self.get_object()
+
+        if request.data.get('owner'):
+            if self.request.user != app.owner and not self.request.user.is_superuser:
+                raise PermissionDenied()
+            new_owner = get_object_or_404(User, username=request.data['owner'])
+            app.owner = new_owner
+        app.save()
+        return Response(status=status.HTTP_204_NO_CONTENT)
```
bacongobbler: I think the response should be a 200 OK to say that the operation was successful. What are your thoughts on a 204 response?

Joshua-Anderson: I don't care either way. I just figured that a 204 might make sense because you're removing an app and it no longer exists for you.

Joshua-Anderson:
It might also make sense to add original owner as a collaborator while transferring, but I wanted some opinions on that.

controller/api/views.py
bacongobbler: A 204 means that we should never return a response body for the issuing request. A 200 could open this up by allowing us to send back more information about the request in a future enhancement to this endpoint (for example, if we were to return information about the new owner, the time it took to process the request etc.). For the sake of forwards compatibility, I think we should make this a 200 :)

krancour:
> It might also make sense to add original owner as a collaborator while transferring, but I wanted some opinions on that.

I had mixed feelings on this and here's the best argument I could come up with in its favor:

When you transfer a repository in github or bitbucket, the receiver needs to _accept_ the transfer, but our workflows don't currently support that sort of transaction.  This creates the potential for an interim period without an "active" owner.

Q:  What happens if the new owner is on vacation and something immediately goes wrong, necessitating a release?

A: An admin could intervene.

If I'm defeating my own strongest argument in its favor, I'm narrowly against it, but I'd really like to hear some other opinions on this.

bacongobbler:
I think it's fine to do a dumb app ownership transfer for now. No need to get fancy with it. Users can just extend the controller to their liking if they want to change the policy, which may be something we can document in the future.

Joshua-Anderson:
I rebased and changed the response code to a 200

mboersma:
I think a naive app transfer like this is fine since admin intervention is already Deis' solution for lost apps. It's possible someone could maliciously transfer apps to another user, but that would be a strange, slow-motion attack that's unlikely and easily thwarted.

LGTM.

carmstrong:
Deferring to @bacongobbler for the second LGTM.

bacongobbler:
code LGTM, though would you mind adding some documentation around this new parameter in the API reference docs?

Joshua-Anderson:
> code LGTM, though would you mind adding some documentation around this new parameter in the API reference docs?

Knew I was forgetting something :smile:. Will do.

Joshua-Anderson:
@bacongobbler Done!

bacongobbler:
code LGTM

krancour:
@Joshua-Anderson needs rebase when you've got the time.  No rush.

Joshua-Anderson:
@krancour Rebased!

mboersma:
@Joshua-Anderson there are 762 changed files in this PR now--maybe some of #4164 crept in here? If you could excise anything not related to the app transfer code, we can get this merged. Or just ping me if you're busy; I'd be glad to do so for you.

Joshua-Anderson:
@mboersma Done! I had a stray Godeps folder from when the router copied Godeps.

bacongobbler:
my lgtm still stands.

Joshua-Anderson:
Rebased!

Joshua-Anderson:
Since API v1.6 was released I bumped the API to v1.7

mboersma:
LGTM++.

Issue #4189, opened 2015-08-03
swarm-node process runs after `deisctl stop swarm`


mboersma:
After using the swarm scheduler, I went to test Mesos, but I noticed that the deis-swarm-node processes continue to run on the nodes, despite the services reporting back as "inactive/dead."
```console
core@deis-02 ~ $ docker ps
CONTAINER ID        IMAGE                        COMMAND                CREATED             STATUS              PORTS                                                              NAMES
4ace0cf7e32d        deis/swarm:v1.9.0            "/app/bin/swarm join   48 seconds ago      Up 47 seconds       2375/tcp                                                           deis-swarm-node  
```

mboersma:
Closed by #4194.

PR #4194, opened 2015-08-04
fix(swarm): fix swarm shutdown

krancour:
Addresses #4189 

I can't figure out why this is happening, but in the meantime, putting these lines back guarantees swarm components shut down upon `deisctl stop swarm`.


mboersma:
LGTM. Tested on vagrant.

krancour:
@mboersma should we try to get this into 1.9.0?

mboersma:
I think this can go in v1.9 if someone else approves it, since it is very low-risk to core functionality.

carmstrong:
> I think this can go in v1.9 if someone else approves it

SGTM. Code LGTM.

carmstrong:
Actually, I don't get this - we are passing `--rm` to the `docker run`. Adding this `ExecStartPost` means the swarm components are special snowflakes...

krancour:
@carmstrong ya... intuitively, the `--rm` in the `docker run` should take care of it, but it's not.  Until we can get to the bottom of that, this at least ensures that `deisctl stop swarm` does indeed stop swarm.

carmstrong:
@krancour I wonder if `docker stop` would get us there, and is logically less of an overlap. I'm wondering if the containers are ignoring the sigkill or something.

krancour:
> I wonder if docker stop would get us there

Good thought.  I'll try first thing tomorrow.

krancour:
@carmstrong `docker stop` worked nicely, and I agree, it's less overlap with the `--rm`.  The `--rm` _does_ work once the container actually stops, so `docker stop` is definitely less overlap.  :100: on the suggestion.

Tests are re-running.  Can I get a fresh set of LGTM's?

mboersma:
Code LGTM. `docker stop` is cleaner.

PR #4200, opened 2015-08-04
feat(*): replace /bin/bash with /usr/bin/env bash

Joshua-Anderson:
`/usr/bin/env bash` is more flexible and standardizes the header across the project.


carmstrong:
Code LGTM.

krancour:
Code LGTM

PR #4214, opened 2015-08-05
fix(mesos-marathon): change instances to zero instead of scale to zero

smothiki:
Creating the app by default starts an instance of container. Scaling down to zero doesn't destroy app container. this fixes that by creating an APP with zero instances.


krancour:
LGTM

mboersma:
Code LGTM.

PR #4215, opened 2015-08-05
fix(Makefile): don't build swarm by default

### LINKED ISSUE #4156: swarm build should be optional
At least for consistency with optional Mesos and k8s, where we don't build any components by default, it should probably require `make -C swarm/ build` to build the swarm scheduler components.

mboersma:
The swarm-* components, like mesos-*, are optional and not installed by default. This removes swarm from `COMPONENTS` in the root Makefile, so that it must be built or installed explicitly: `make -C swarm/ build`.

Closes #4156.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ repo_path = github.com/deis/deis
 GO_PACKAGES = version
 GO_PACKAGES_REPO_PATH = $(addprefix $(repo_path)/,$(GO_PACKAGES))
 
-COMPONENTS=builder cache controller database logger logspout publisher registry router store swarm
+COMPONENTS=builder cache controller database logger logspout publisher registry router store
 START_ORDER=publisher store logger logspout database cache registry controller builder router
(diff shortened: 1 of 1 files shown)


krancour:
Code LGTM

bacongobbler:
Do we need to make any changes to our current test-acceptance workflow to publish new releases of swarm or will that still work?

carmstrong:
> Do we need to make any changes to our current test-acceptance workflow to publish new releases of swarm or will that still work?

I think we're definitely still pushing it, but are we building it?

mboersma:
> Do we need to make any changes to our current test-acceptance workflow?

Yes, we have to do `make -C swarm/ build` now as well as `make -C mesos/ build`. I've got it covered.

carmstrong:
Code LGTM.

bacongobbler:
thanks @mboersma for double checking :)

Code LGTM!

PR #4221, opened 2015-08-06
fix(contrib): remove manage_etc_hosts from example user-data

iancoffey:
The manage_etc_hosts cloud config was added so that instances with hostnames that did not resolve correctly could still use the graceful shutdown logic. I personally only observed this issue with vagrant.

This seems to cause issues on GCE though, and should probably be removed.

https://github.com/coreos/bugs/issues/425


carmstrong:
Code LGTM.

mboersma:
Code LGTM. 

Docs for `manage_etc_hosts` are at the bottom of this page: https://coreos.com/os/docs/latest/cloud-config.html

krancour:
> This is helpful when the host does not have DNS infrastructure in place to resolve its own hostname, for example, when using Vagrant.

Couldn't this potentially include some users running bare metal?

carmstrong:
> Couldn't this potentially include some users running bare metal?

Potentially, but arguably users on bare metal have more control over their `/etc/hosts` file than do users on cloud providers, so I don't think it's a big deal to put the burden on them.

iancoffey:
Agree. The more I think about it, the less reasonable it seems to clobber someone hosts file regardless of where they are running. If it makes problems, we should address those.

krancour:
@carmstrong great point.

Issue #2338, opened 2014-11-01
LICENSE needs updating for 2014


Xe:
It still says 2013

PR #4225, opened 2015-08-06
Update copyright year coverage

dwmcallister:
changed copyright  to include 2015

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright 2013, 2014 Engine Yard, Inc.
+Copyright 2013-2015 Engine Yard, Inc.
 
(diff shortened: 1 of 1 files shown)


carmstrong:
There are also licenses in all the subcomponent READMEs, so if copyright dates are updated, they'd have to be updated elsewhere. We asked our attorneys about copyright dates previously, and I believe they said we should be using the first date something was published, and not updating it for every year.

/cc @JoshuaSchnell

mboersma:
Refs #2338, #2622, and #3611.

carmstrong:
Closing. If we want to accept this, the commit will have to be updated and several other files changed as part of this PR.

PR #4226, opened 2015-08-06
[WIP] feat(deisctl): add a verbose start option

Joshua-Anderson:
Running `deisctl start -v` will print component logs while starting up, showing the progress of things loke downloading docker images.

Why is this WIP?

I believe that I'm leaking ssh connections, but can't find the source. For components that take longer to startup, I start to get `ssh: handshake failed: read tcp 172.17.8.100:22: connection reset by peer` and when I panic I see an enormous amount of `gossh` goroutines. If anybody knows what I'm forgetting to clean up I'd be very appreciative.

### DIFF
--- a/deisctl/backend/backend.go
+++ b/deisctl/backend/backend.go
@@ -20,4 +20,5 @@ type Backend interface {
 	ListUnitFiles() error
 	Status(string) error
 	Journal(string) error
+	JournalBackground(string, io.Writer) []chan bool
 }
--- a/deisctl/backend/fleet/fleet.go
+++ b/deisctl/backend/fleet/fleet.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"path"
 	"text/tabwriter"
+	"time"
 
 	"github.com/deis/deis/deisctl/config"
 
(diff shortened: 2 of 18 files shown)


Joshua-Anderson:
I changed the code to poll only the output of the last seconds. This makes the code much cleaner and heavily decreases the amount of journal output. During periods of heavy output however, this has the potential for duplication, due to the fact that `journalctl` isn't accurate about the time (I'm fairly sure it's on journalctl's side on not on deisctl's side, but I could be wrong). I think this is the best approach however, as a little bit of duplication for progress logs doesn't seem to be a huge deal to me, while trying to fetch the entire logs of say, builder, every 0.5 sec could be problematic. It also cleans up a race condition I noticed.

mboersma:
@Joshua-Anderson is leaking SSH connections still a concern here? (Can we remove the [WIP]?)

Joshua-Anderson:
@mboersma unfortunately, yes. I have been doing some looking but still haven't found the source of the leak.  If one of the core maintainers has time, I would appreciate it if somebody else could also give it a glance to see if I missed anything or could suggest an alternate approach for getting the text output of a ssh command.

mboersma:
Of course, thanks for the update. We will definitely get to it at some point, just wanted to make sure we had up-to-date info. Thanks!

kalbasit:
Accidentally closed the PR.

Joshua-Anderson:
Don't think this is going to make it into V1 anymore and this is obsolete in V2, so I'm closing.

PR #4233, opened 2015-08-07
fix(publisher): ignore healthcheck values if unset

### LINKED ISSUE #4227: Deis publisher throwing error after 1.8 to 1.9 upgrade of stateless-platform
I run a stateless deis platform (without ceph).Everything was running fine for a week till now.
But a few hours ago, I updated everything to 1.9 from 1.8, dies-publisher is throwing out a error in logs .
```
deis-publisher[1]: 2015/08/07 04:41:55 strconv.ParseInt: parsing "": invalid syntax
```
This is continuously logging since few hours.

I don't any other visible issue to my app containers from this.

bacongobbler:
publisher would do a lazy check and convert the etcd keys to an integer.
However, these keys are optional, so if they are unset, an empty string
would be returned from getEtcd(). By explicitly checking if the keys are
set, this will circumvent the strconv errors.

closes #4227


krancour:
Code LGTM

mboersma:
Code LGTM.

PR #4241, opened 2015-08-10
fix(client-go): fix webbrowser merge error

Joshua-Anderson:
The webbrowser code was moved to it's own package, but the windows fix pr was created from a point in time before that, so while the PR merged properly it didn't take the move into account.


mboersma:
Code LGTM.

krancour:
Code LGTM

Issue #2959, opened 2015-01-26
AWS instances crashing


mgartner:
I've witnessed 3+ times in the last two weeks where an EC2 instance is terminated by my autoscaling group when it fails an instance status check. These terminations lead to problems with my cluster - putting Ceph into a HEALTH_WARN state and, with enough terminations, losing quorum. 

Given the regularity of the terminations, I think it's unlikely that the cause of the terminations is due to the infrastructure my instances were living on. Either Deis or one of my apps brings down the host machine, or causes an instance status check to fail.

The three apps I'm running on the cluster are fairly simple - one is a Node.js server, and the two other are background workers that pull messages from SQS and perform image manipulation with imagemagick and send out emails via SES. I can't think of any reason that one of these apps would lead to problems for the host, but I'm open to ideas. If other users hosting Deis have never seen unexpected terminations on AWS, then that might be strong evidence that my particular applications may be the source of the problem.

I've tried to look for any clues as to what is exactly happening, but the messaging from AWS is pretty vague. In addition, server metrics via New Relic don't show anything suspicious in terms of CPU, RAM, networking, or disk usage at the time that the instance is terminated.

I've had to move my production applications off of Deis for now, as I was having to manually intervene every-other-day or so, but I've left the cluster of affected machines running in case it will help diagnose the problem (although Ceph has lost quorum and the cluster is in a wonky state).

I'm hoping someone with more experience in this area has an idea of where to look for clues to get to the bottom of this.

carmstrong:
This issue is described a bit in #2877 - we're working on https://github.com/deis/mortician as a solution to automatically handle cases where an autoscaling event removes a node from the cluster, but we really need to get to the bottom of why some users are experiencing frequent failures... I know @ianblenke sees this too.

ianblenke:
I'm pretty sure this is ceph mds (cephfs) related.

In the past, we've had applications eat all available RAM and cause kernels to panic.

We've also had HVM clock drift because we weren't using NTP, which caused some ceph weirdness.

Now that we have both of those things under control, the last time I had a crash I did catch a mention of mds being in the middle of things:

    [534838.718074] ceph: mds0 caps stale
    [535083.710066] ceph: mds0 hung

It is important that we get folks to pull the system console output from AWS when an instances hangs like this, so that we might find  if there is a commonality.

Note: you can still pull the system console even from a TERMINATED instance, so long as you do so before the next periodic AWS cleanup sweep where that instance disappears from your account entirely.

[...]

PR #4247, opened 2015-08-10
Script for setting up a Terraform configuration with AWS

econnell:
This is a shell script that builds a terraform configuration to stand up a Deis cluster.  At present, it only supports the Terraform AWS provider.

The script itself is interactive and guides the user through various configuration settings and deis infrastructure configuration setups (private cluster nodes, public cluster nodes, etc).


alex:
I don't have a review, but I just wanted to chime in to say: thanks for working on this! For me this is a) a very helpful tool when I next go to provision a deis cluster, b) awesome programatic documentation for what a deis topology looks like in a very succinct format.

Thanks :sparkles: :cake: :sparkles:

contrib/terraform/gen-userdata.py
```diff
@@ -0,0 +1,86 @@
...
+
+new_units = [
+  dict({'name': 'format-ephemeral-volume.service', 'command': 'start', 'content': FORMAT_EPHEMERAL_VOLUME}),
+  dict({'name': 'media-ephemeral.mount', 'command': 'start', 'content': MOUNT_EPHEMERAL_VOLUME}),
+  dict({'name': 'prepare-etcd-data-directory.service', 'command': 'start', 'content': PREPARE_ETCD_DATA_DIRECTORY}),
+  dict({'name': 'format-docker-volume.service', 'command': 'start', 'content': FORMAT_DOCKER_VOLUME}),
+  dict({'name': 'var-lib-docker.mount', 'command': 'start', 'content': MOUNT_DOCKER_VOLUME})
+]
+
+data = yaml.load(file(os.path.join(CURR_DIR, '..', 'coreos', 'user-data'), 'r'))
```
alex: You should use `yaml.safe_load`, `yaml.load` can execute arbitrary code (doesn't super matter in this case, but it's a good practice). Also you should use `open` instead of `file`.

contrib/terraform/gen-userdata.py
```diff
@@ -0,0 +1,86 @@
...
+  Where=/var/lib/docker
+  Type=ext4
+'''
+
+new_units = [
+  dict({'name': 'format-ephemeral-volume.service', 'command': 'start', 'content': FORMAT_EPHEMERAL_VOLUME}),
+  dict({'name': 'media-ephemeral.mount', 'command': 'start', 'content': MOUNT_EPHEMERAL_VOLUME}),
+  dict({'name': 'prepare-etcd-data-directory.service', 'command': 'start', 'content': PREPARE_ETCD_DATA_DIRECTORY}),
+  dict({'name': 'format-docker-volume.service', 'command': 'start', 'content': FORMAT_DOCKER_VOLUME}),
+  dict({'name': 'var-lib-docker.mount', 'command': 'start', 'content': MOUNT_DOCKER_VOLUME})
```
alex: `dict` is unnescary here, `{}` already creates a dict.

alex:
@econnell One thing that jumps out to me is that this seems to only provision nodes in a single AZ, is there a reason not to support multi-AZ deployments?

(Also, as a more general question, does this create a deis deployment that is "production ready"?)

bacongobbler:
@alex etcd currently [does not support multi-AZ deployments](https://github.com/coreos/etcd/issues/964). While it's possible to [deploy etcd externally from Deis](http://docs.deis.io/en/latest/managing_deis/isolating-etcd/), by default the provisioning scripts should give a good "Out of the Box" experience for newer users coming to the platform, so we have to stick with that limitation.

bacongobbler:
Sorry! I'm getting DC and AZ mixed up again. :)

After reading through the AWS cloudformation template, we do support [multi-AZ deployments](https://github.com/deis/deis/blob/8f50e7063a4d8669b77592edfef71f910de6e8a3/contrib/aws/deis.template.json#L239-L261) over there, so I don't see a reason why we wouldn't be able to here!

alex:
Cheers! Ok, thanks for validating that assumption.

contrib/terraform/gen-userdata.py
econnell: This file is basically just a ripped apart version of contrib/aws/gen-json.py to generate identical user-data for what is currently used in the CloudFormation config.  I don't have a problem with changing it, but if it should probably be changed in both locations if it's cleaner (or figure out a way to not duplicate the code in the first place)

econnell:
@alex the problem with multi-AZ is that if connectivity drops between the AZs for any reason, you can easily end up with a split brain problem that can be very difficult to resolve.

That is obviously an extremely rare case, but etcd doesn't handle that situation very well.  A lot of thought about infrastructure needs to happen before you go multi-AZ, and it still probably should require 3 AZs/regions (with 1 acting as a simple arbiter even).

It's an extremely complex problem and I'm personally hesitant to recommend that situation because it grants a false sense of security.

Also, there's still the issue of backend data storage and which AZ that's in, how to route the DB requests to the appropriate DB, etc.

Multi-AZ is extremely complicated and I think that should be beyond the scope of this.

I meant this more of a way to build a basic Terraform config that can get users started and to use as a reference, not as something that should be used to build an end-state robust production platform.

krancour:
I haven't gotten into this in great depth yet, but I'd like to say that the _existing_ AWS provisioning process can create a VPC with two subnets-- each in a different availability zone.  It would be bad to lose that ability.

> Multi-AZ is extremely complicated and I think that should be beyond the scope of this.

Multi-_region_ is complicated.  I do think _that_ should be beyond the scope, but multi-_AZ_ is an absolute necessity for anyone wishing to guarantee the availability of the platform and the apps deployed to the platform.

> the problem with multi-AZ is that if connectivity drops between the AZs for any reason, you can easily end up with a split brain problem that can be very difficult to resolve

This can also happen _within_ a single AZ if nodes are unable to talk to one another for any reason, so it might not be impetus not to do multi-AZ.

> and it still probably should require 3 AZs

Yes.  At least three are required if you want to guarantee that an AZ-wide problem in an AZ hosting the majority of your nodes doesn't force you to lose quorum.  By having at least three AZs, you can avoid any one having a _majority_ of your nodes.

> Also, there's still the issue of backend data storage and which AZ that's in, how to route the DB requests to the appropriate DB, etc.

This isn't a problem for those running the "stateless platform" (ceph-less), and opting to use RDS or similar for storage.  This seems to be an increasingly common choice among those concerned about minimizing the platform's own moving parts and maximizing availability.

econnell:
> Multi-region is complicated. I do think that should be beyond the scope, but multi-AZ is an absolute necessity for anyone wishing to guarantee the availability of the platform and the apps deployed to the platform.

It wouldn't be hard to implement at all, but I'd rather see that as a future issue since I don't think that it is critical.  I do, however, disagree that this guarantees availability.  It does improve availability though (with the previously mentioned caveats).

krancour:
That alone doesn't guarantee availability, but it is a necessary step for anyone who's headed in that direction.  Keep in mind, multi-AZ is already available with the existing provisioning scripts.  It would be bad to lose that.

carmstrong:
> That alone doesn't guarantee availability, but it is a necessary step for anyone who's headed in that direction. Keep in mind, multi-AZ is already available with the existing provisioning scripts. It would be bad to lose that.

:+1:

econnell:
Sounds good.  I'll get the multi-AZ stuff in there ASAP.

sstarcher:
@econnell why in the script does it say "1. paravirtual (recommended)".  HVM is normally considered recommended.

carmstrong:
> why in the script does it say "1. paravirtual (recommended)".

We had users report deadlocks and kernel panics when using HVM instances, although we should give that another go to see if it's still an issue.

sstarcher:
@carmstrong we currently use hvm for all of our AMIs except for the Deis cluster.  Our CoreOS cluster has been running kubernetes on HVM amis for a few months now.

carmstrong:
> we currently use hvm for all of our AMIs except for the Deis cluster.

Sorry, I should have clarified. I know HVM is generally stable. We suspect the deadlocks/panics were an interaction of the CephFS client driver and the CoreOS kernel, and potentially exacerbated by clock drift. If you're not running Deis on those hosts, you probably wouldn't have seen what we saw.

sstarcher:
We use STS temporary access keys for our AWS API access.  This gives us a AWS_SESSION_TOKEN which I believe is only supported via terraform when you specify everything as environment variables.  Instead of having aws_access_key and aws_secret_key specified in the terraform file.  Can we have users set the AWS Environment variables.  These tokens also expire after a set amount of time.

https://www.terraform.io/docs/providers/aws/

jgmize:
>>    we currently use hvm for all of our AMIs except for the Deis cluster.

> Sorry, I should have clarified. I know HVM is generally stable. We suspect the deadlocks/panics were an interaction of the CephFS client driver and the CoreOS kernel, and potentially exacerbated by clock drift. If you're not running Deis on those hosts, you probably wouldn't have seen what we saw.

To further clarify, would HVM be preferred when running Deis without Ceph then?

sstarcher:
@jgmize If the problem was with Ceph than we should certainly prefer HVM without Ceph.  HVM is more performant than PV.  http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/virtualization_types.html

carmstrong:
> To further clarify, would HVM be preferred when running Deis without Ceph then?

Generally, yes. With the caveat that the original issue was mainly anecdotes that we've heard from our users (@ianblenke and @mgartner, from memory) - see #2959

sstarcher:
ASG needs the block data to move to the launch_config
https://gist.github.com/sstarcher/248b7c80b5b2928fe796

sstarcher:
When choosing existing VPC in the current state the terraform files will create a new VPC.  I did the following to modify the files to allow it to work with my existing vpc/subnet

*Note the sed syntax is for OSX delete the space between -i and '' for gnu bash
```
SUBNET='subnet-xxxxxxx'
VPC='vpc-xxxxxx'
echo "aws_subnet_deis_public_id = \"$SUBNET\"" >> config.tfvars
echo "aws_vpc_deis_id = \"$VPC\"" >> config.tfvars

echo 'variable "aws_vpc_deis_id" {}' >> variable.tf
echo 'variable "aws_subnet_deis_public_id" {}' >> variable.tf

sed -i '' s/aws_vpc.deis.id/var.aws_vpc_deis_id/g *.tf
sed -i '' s/aws_subnet.deis_public.id/var.aws_subnet_deis_public_id/g *.tf
rm vpc.tf
```

mboersma:
Closing this in favor of #4315 and [`rigger`](https://github.com/deis/rigger).

:v: @econnell

sgoings:
We'll probably still use this, but it'll be wrapped in rigger's swaddling clothes. Don't get rid of it! :smile: I appreciate all the work that has gone into this, @econnell.

PR #4250, opened 2015-08-10
fix(contrib/aws): clarify proxy protocol error

### LINKED ISSUE #4198: provision-aws-cluster.sh vague error
Recently, the provision-aws-cluster.sh script got a new check at the end. It runs `deisctl list` repeatedly, looking for a successful exit. It times out after a few minutes.

https://github.com/deis/deis/blob/master/contrib/aws/provision-aws-cluster.sh#L150

The first few times I got this message, I thought the entire cluster install had failed, and I destroyed and re-installed the cluster multiple times. Later, I realized that this message is actually harmless.

I'm sure this is here for a good reason, but the messaging doesn't make any sense. Is it an error if it times out? If so... what are we supposed to do? If it's a warning, then put it in yellow. But in any event, give some clear indication of what the user needs to do.

In my tests, it appears that the *only* role for this is to delay until running the deisctl command to set the proxy type on nginx. If that's the case, let's make this explicit so that the user knows what to do.

### LINKED ISSUE #4199: Have provision-aws-cluster tell you your DEISCTL_TUNNEL
The new version of `provision-aws-cluster.sh` sets its own DEISCTL_TUNNEL to the IP address of the first EC2 instance in the cluster. This works, and is convenient.

But we should pass this info on to the user, who otherwise is stuck following the vague instructions in the user guide (http://docs.deis.io/en/latest/installing_deis/install-platform/#install-deis-platform).

Can we just emit a message from the script telling the user to set their DEISCTL_TUNNEL to the address we autodiscover?

mboersma:
If ssh auth fails for `deisctl` during AWS provisioning, the user sees a red warning followed by a green success message followed by another red warning, with little help on what actions to take.
This moves the success message last and suggests checking ssh-agent for the proper key.

Closes #4198.
Closes #4199.

### DIFF
--- a/contrib/aws/provision-aws-cluster.sh
+++ b/contrib/aws/provision-aws-cluster.sh
@@ -146,8 +146,10 @@ export DEISCTL_TUNNEL=$FIRST_INSTANCE
 # loop until etcd2 / fleet are up and running
 COUNTER=1
 until deisctl list >/dev/null; do
-    if [ $COUNTER -gt $ATTEMPTS ];
-        then echo_red "Timed out waiting for fleet, giving up"
+    if [ $COUNTER -gt $ATTEMPTS ]; then
+        echo_red "Timed out waiting for fleet, giving up"
+        echo_red "Ensure that the private key in cloudformation.json"
+        echo_red "is added to your ssh-agent."
         break
     fi
     echo "Waiting until fleet is up and running ..."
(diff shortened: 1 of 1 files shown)


bacongobbler:
very descriptive **and** helpful. Code LGTM.

carmstrong:
Code LGTM.

mboersma:
This passed the relevant doc tests, so I'm merging it.

PR #4251, opened 2015-08-11
fix(client-go): unset token before logging in

Joshua-Anderson:
This could cause problem when registering with a fresh cluster before logging out of a old one. Registration doesn't care if the token is valid (it only cares when registration is set to admin only), but login requires no token or a valid token, so it has to be unset.


bacongobbler:
lgtm

mboersma:
LGTM.

PR #4256, opened 2015-08-11
fix(controller): don't check permissions with @permissions_classes

Joshua-Anderson:
This allows finer grained control over permission checking, with the added side effect of allowing users to remove their own app access.

### DIFF
--- a/controller/api/views.py
+++ b/controller/api/views.py
@@ -8,7 +8,6 @@ from django.shortcuts import get_object_or_404
 from guardian.shortcuts import assign_perm, get_objects_for_user, \
     get_users_with_perms, remove_perm
 from rest_framework import mixins, renderers, status
-from rest_framework.decorators import permission_classes
 from rest_framework.exceptions import PermissionDenied
 from rest_framework.permissions import IsAuthenticated
 from rest_framework.response import Response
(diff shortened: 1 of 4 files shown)


controller/api/views.py
```diff
@@ -412,19 +410,28 @@ def list(self, request, **kwargs):
...
+
         user = get_object_or_404(User, username=request.data['username'])
         assign_perm(self.perm, user, app)
         models.log_event(app, "User {} was granted access to {}".format(user, app))
         return Response(status=status.HTTP_201_CREATED)
 
-    @permission_classes([permissions.IsOwnerOrAdmin])
     def destroy(self, request, **kwargs):
-        app = self.get_object()
+        app = get_object_or_404(models.App, id=self.kwargs['id'])
```
bacongobbler: how come this call needs to change? Won't `get_object()` [do the right thing](https://github.com/tomchristie/django-rest-framework/blob/3e1687282ba39b0166b7c17e39cdd3a2be7e2bc3/rest_framework/generics.py#L97-L100) and 404 if it isn't found, or are you trying to circumvent the PermissionDenied exception?

Joshua-Anderson: When I tried `self.get_object()` it was throwing 403 unauthorized for app users.

bacongobbler: Gotcha. That's probably because the BaseDeisViewSet which this class extends from has some permissions set up there only for app owners and administrators, IIRC.

bacongobbler:
code LGTM

mboersma:
LGTM.

PR #4258, opened 2015-08-11
test(*): enable repeated runs of setup-node.sh script

sgoings:
In the process of setting up new nodes for ci.deis.io, I ran across some problems with the setup-node.sh script I took a crack at fixing...

Mainly I wanted the setup-node.sh script to be able to run X times and skip stuff that had already been installed, not choke if a previous run hadn't succeeded, etc.

I also added the aws pip requirements.txt file into the contrib/aws structure. They are necessary on the build slaves due to jobs like... say... test-ec2. :smile:


tests/bin/setup-node.sh
```diff
@@ -9,23 +9,45 @@ set -eo pipefail
...
-sh -c "echo deb https://get.docker.com/ubuntu docker main > /etc/apt/sources.list.d/docker.list"
-apt-get update && apt-get install -yq lxc-docker-1.5.0
+apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 \
+            --recv-keys 36A1D7869245C8950F966E92D8576A8BA88D21E9
+echo deb https://get.docker.com/ubuntu docker main > /etc/apt/sources.list.d/docker.list
+apt-get update
+apt-get install -yq lxc-docker-1.5.0
+
+# install extra extensions (AUFS)
+sudo apt-get -y install "linux-image-extra-$(uname -r)"
```
sgoings: This is new, per @technosophos request. His PR was failing due to the slaves not having aufs support.

carmstrong: Is there a reason we're using aufs on the tests? We default to overlayfs in shipped Deis, so ideally we're using that here as well. I recall some history with this...

mboersma: It's because the CI test boxen are Ubuntu 14.04, so the kernel is too old for overlayfs. We should reprovision them to allow overlayfs, so we don't ever waste time chasing aufs-specific issues.

mboersma: @aledbf suggested we could stick with 14.04 but just update the kernel from http://kernel.ubuntu.com/~kernel-ppa/mainline/ to get overlayfs support. That might be simpler.

sgoings: Why should we stick to 14.04? (or just Ubuntu in general?)

mboersma:
LGTM.

sgoings:
Still waiting on an LGTM x 2. :smile:

carmstrong:
Code LGTM

tests/bin/setup-node.sh
aledbf: @sgoings my suggestion to update the kernel is just because it takes less than 10 minutes and that solves the aufs issues. Is just temporal workaround. I don't have a reason to keep using 14.04 or even Ubuntu.

PR #4259, opened 2015-08-12
fix(contrib) AWS instances may only have a private IP

sbuss:
If you're creating an aws stack in a VPC, then your instances may not
have a publicly accessible IP. This caused the fleet check to hang, and
the proxyProtocol setting to fail.

This fixes the provision-aws-cluster script to prefer the public IP if
available, but also try connecting to the private IP.

### DIFF
--- a/contrib/aws/provision-aws-cluster.sh
+++ b/contrib/aws/provision-aws-cluster.sh
@@ -136,11 +136,24 @@ ELB_NAME=$(aws elb describe-load-balancers \
     $EXTRA_AWS_CLI_ARGS | grep -F $ELB_DNS_NAME | head -n1 | cut -f2)
 echo "Using ELB $ELB_NAME at $ELB_DNS_NAME"
 
-FIRST_INSTANCE=$(aws ec2 describe-instances \
-    --filters Name=tag:aws:cloudformation:stack-name,Values=$STACK_NAME Name=instance-state-name,Values=running \
-    --query 'Reservations[].Instances[].[PublicIpAddress]' \
-    --output text \
-    $EXTRA_AWS_CLI_ARGS | head -1)
+# Instance launched into a VPC may not have a PublicIPAddress
+for ip_type in PublicIpAddress PrivateIpAddress; do
+    FIRST_INSTANCE=$(aws ec2 describe-instances \
+        --filters Name=tag:aws:cloudformation:stack-name,Values=$STACK_NAME Name=instance-state-name,Values=running \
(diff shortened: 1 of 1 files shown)


sbuss:
I wasn't sure if the better behavior would be to `exit 1` when an IP can't be found. Please let me know what you prefer.

sbuss:
@sgoings recommended I `exit 1`, so I added that and amended the commit.

krancour:
The assumption here is that the provisioning script is being run from within the VPC, yes?

sbuss:
@krancour Yes, this happens if you're provisioning a cluster in a VPC and you disable `AssociatePublicIP`. For reference, here's my cloudformation parameters file:

```json
[
    {
        "ParameterKey":     "KeyPair",
        "ParameterValue":   "deis"
    },
    {
        "ParameterKey":     "InstanceType",
        "ParameterValue":   "m3.xlarge"
    },
    {
        "ParameterKey":     "EC2VirtualizationType",
        "ParameterValue":   "HVM"
    },
    {
        "ParameterKey":     "AssociatePublicIP",
        "ParameterValue":   "false"
    },
    {
        "ParameterKey":     "ELBScheme",
        "ParameterValue":   "internal"
    }
]
```

Also the `VPC_ID`, `VPC_SUBNETS`, and `VPC_ZONES` env vars were all set prior to running the `provision-aws-cluster.sh` script.

krancour:
It's occurring to me that there might be a gap in the existing documentation-- that we never point out that when installing into private subnets, you need to run the provisioning script from within the VPC.

krancour:
Also, code LGTM.

sbuss:
>  you need to run the provisioning script from within the VPC

What exactly do you mean by this? When I provision a cluster I do it from my laptop which is connected to our VPN, which has a connection to our VPC in aws. Are you suggesting that I should actually be running this from an aws machine provisioned in the VPC? It's been working fine for me (except for this bit around the public IP)

krancour:
@sbuss you have infrastructure in place that not everyone has-- both client to site and site to site VPN are in play by the sounds of it.  Absent that, someone provisioning Deis onto machines that do not have public IPs would need to do so from within the VPC.

carmstrong:
I'm confused where we left this. @krancour are you still okay with this, or is there something left to discuss based on your latest comment? 

Either way, I'll kick off a new Jenkins run: https://ci.deis.io/job/test-ec2/2839/

krancour:
@carmstrong my LGTM stands.

carmstrong:
Code LGTM.

Issue #4254, opened 2015-08-11
In router boot program, cleanup never fires on SIGINT or SIGTERM


krancour:
I'm not sure how important the cleanup that is to be performed is, but I'm pretty sure it's not happening.

I haven't looked yet, but this is likely happening in other spots as well.

https://github.com/deis/deis/blob/master/router/boot.go#L88

```
func main() {
  // ...
  exitChan := make(chan os.Signal, 2)
  signal.Notify(exitChan, syscall.SIGTERM, syscall.SIGINT)
  go cleanOnExit(exitChan)
  <-exitChan
}
```

The issue is that the main goroutine can potentially exit (and does) before the `cleanOnExit()` function performs cleanup.

Here's a trivial program that demonstrates the techniques does not work.

```
package main

import (
	"fmt"
	"os"
	"os/signal"
	"syscall"
)

func main() {
	exitChan := make(chan os.Signal, 2)
	signal.Notify(exitChan, syscall.SIGTERM, syscall.SIGINT)
	go cleanOnExit(exitChan)
	fmt.Println("Waiting for interrupt...")
	<-exitChan
}

func cleanOnExit(exitChan chan os.Signal) {
	for _ = range exitChan {
		fmt.Println("Cleaning up...")
	}
}
```

bacongobbler:
This is okay. This is going to be dead code once I'm done with what I'm currently working on. This should have used a syncgroup or something to make sure that the cleanup function did finish up, but it will all be gone soon :)

bacongobbler:
We should investigate the other components, but I think this is the only component that has a post-cleanup goroutine AFAICR

krancour:
You're probably right, but let's keep this open until we've had a chance to review those others spots.

krancour:
Closed by #4260

PR #4260, opened 2015-08-12
ref(router): shift things around

bacongobbler:
This follows Docker's "best practices" for structuring projects. This
also simplifies some of the build process by building everything in one
`docker build` rather than compiling nginx, then injecting it into another
chained `docker build`. The image size remains the same. This is the
same as what occurred over in builder a few months ago.

This also removes the log tailing feature to stdout in boot.go by using
symbolic links to /dev/stdout and /dev/stderr as is expected. git.log
is nonexistent and unused.

refs #4254, which may be closed by the latter point.


carmstrong:
@bacongobbler This needs a rebase.

mboersma:
LGTM.

Issue #4158, opened 2015-07-31
possible bug in client-go: error when scaling app w/ mesos as the scheduler


krancour:
About 50% of the time the following happens when I scale using the mesos/marathon scheudler:

```
[kent@mbp example-ruby-sinatra]$ deis scale web=2
Scaling processes... but first, coffee!
done in 10s
Error: Get http://deis.local3.deisapp.com/v1/apps/walkup-espresso/containers/: EOF
```

mboersma:
@krancour are you using the new go `deis` CLI? I've seen the same, but it doesn't happen for me with the traditional python CLI.

krancour:
Yep.  Good call.

Joshua-Anderson:
I'll look into this, as I've seen this as well. I love `EOF` errors!

krancour:
Reporting that this happens with k8s as the scheduler as well.

Joshua-Anderson:
Related #4269

mboersma:
I can't reproduce this currently against master with the go CLI. Maybe nginx 1.9.4 (#4306) or maybe go 1.5?

krancour:
I'll re-attempt just to corroborate.

krancour:
I can no longer re-create this.  Closing.

krancour:
No sooner did I close this than it reared it's ugly head again.  @mboersma has observed this in Jenkins today and just now it _finally_ bit me locally after _many_ recent failures to duplicate.

```
[kent@mbp example-go]$ deis scale web=100
Scaling processes... but first, coffee!
done in 81s
Error: Get http://deis.krancour.deis.ninja/v1/apps/lively-kerchief/containers/?page_size=100: EOF
```

bacongobbler:
This looks like a manifestation of an older bug we've had for a very long time. More information on this bug here: https://github.com/deis/deis/pull/4261#issuecomment-135552944

mboersma:
We haven't seen an EOF since #4261 was merged, so I assume `DisableKeepAlives=true` fixed this. (Please re-open this issue if I'm wrong and you've seen any EOF errors with the new go CLI.)

PR #4261, opened 2015-08-12
ref(tests): default to new golang deis CLI

mboersma:
This makes client-go/deis the default `deis` binary used in tests. 

Once it is merged, we can decide how to deprecate the old CLI and promote the new one within the codebase to make it obvious--that will be a separate PR. (See #4341.)

### DIFF
--- a/pkg/prettyprint/colorizer_test.go
+++ b/pkg/prettyprint/colorizer_test.go
@@ -37,6 +37,7 @@ func ExampleColorize() {
 	out := Colorize("{{.Red}}Hello {{.Default}}World{{.UnderGreen}}!{{.Default}}")
 	fmt.Println(out)
 }
+
 func ExampleColorizeVars() {
 	vars := map[string]string{"Who": "World"}
 	tpl := "{{.C.Red}}Hello {{.C.Default}}{{.V.Who}}!"
@@ -67,16 +68,13 @@ func TestPrettyTabs(t *testing.T) {
 		"foo":  "bar",
 	}
 
-	expected := `test testing
-foo  bar
-`
-	otherExpected := `foo  bar
+	expected := `foo  bar
 test testing
 `
 
 	output := PrettyTabs(test, 1)
 
-	if !(output == expected || output == otherExpected) {
+	if output != expected {
 		t.Errorf("Expected '%s', Got '%s'", expected, output)
 	}
 }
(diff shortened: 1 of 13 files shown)


mboersma:
I see this in the integration tests, so that's at least one thing remaining to fix:
```
limits:set --app=configsample cmd=256MB
cmd=256MB doesn't fit format type=#unit or type=#
Examples: web=2G worker=500M web=300
Applying limits... done
```
It should have been (from the old python client):
```
limits:set --app=configsample cmd=256MB
Applying limits... done, v10
```

Joshua-Anderson:
@mboersma That's due to the fact, I thought we only supported `B`, `M`, or `G`, for memory units. Is there a complete list of all the memory units we support?

Joshua-Anderson:
The tests failed with:

```
Error: Get http://deis.test-46b84a697e.deis.works/v1/apps/healthchecksample/containers/: EOF
```

For some reason the go client frequently throws this error when running `ps:list`.

bacongobbler:
perhaps that's the equivalent of the python client's `BadStatusLine` error we keep seeing in the tests? Not sure...

pkg/prettyprint/colorizer_test.go
```diff
@@ -67,16 +68,13 @@ func TestPrettyTabs(t *testing.T) {
...
-`
-	otherExpected := `foo  bar
+	expected := `foo  bar
 test testing
 `
 
 	output := PrettyTabs(test, 1)
 
-	if !(output == expected || output == otherExpected) {
+	if !(output == expected) {
```
carmstrong: Wouldn't this read better as `if output != expected`?

carmstrong:
Code LGTM.

mboersma:
I've now seen two EOF errors during test_ps.go, although I have tried and failed to reproduce this using the same commands interactively. So there is still something to fix here... I can try `DisableKeepalives=true` but it's difficult to know if that fixes things unless I can reproduce it locally.

krancour:
Related to #4158... which I just closed, but we can re-open that if this warrants it.

mboersma:
So several of us have been testing this, as well as Jenkins. EOF errors are rare but still there. We think this is a different manifestation of the `BadStatusLine` error that cropped up sometimes using the old python CLI. @bacongobbler got some logs from nginx with debug on that seem to point to this: http://serverfault.com/questions/716862/recv-not-ready-11-resource-temporarily-unavailable-nginx-1-1-19-supervisor

tl:dr, still trying to pin down the EOF, but its impact appears to be similar to a problem we already had, so this is probably ready to :ship:.

bacongobbler:
is this superseded by #4341 or is this different?

mboersma:
> is this superseded by #4341 or is this different?

I intended them to be separate in order to facilitate reviewing, but at this point it's probably simpler to close this and focus on #4341.

Issue #3836, opened 2015-06-09
Rewrite Builder shell scripts in Go


technosophos:
The current builder is written in shell scripts. While some of the shell should remain, the majority of it should be reimplemented in Go for the following reasons:

- We get finer grained control over interoperability with etcd and other services
- We can write good tests!
- The platform becomes more consistent.

This is part of the overarching milestone described in #3758.

technosophos:
_I'm going to paste in here a slightly revised snippet of a comment I made in a great conversation with @kalbasit. I think it's important to bubble this up._

_Background: My current prototype uses a chain-of-command pattern with dependency injection to create a composable architecture. Or, in plain English, it maps a named task ("boot") to a series of individual (reusable) steps. What I'm pasting below explains my rationale._

[L]et me hit on the big design goals of this rewrite:
1. Replace a bunch of shell with Go
2. Make everything really, really testable
3. Hopefully work toward a component-based system that we can re-use

The first is easy to accomplish.

For the other two, here's what I am thinking: Basically, builder's boot process runs a sequence of individual steps. Some of these steps are really specific, but most are fairly generic. Read a bunch of environment variables. Send a request to etcd. Start a system service.

So what I'm doing is taking a system that provides the ability to run a sequence of commands, and writing a bunch of re-usable commands. Since every command follows the same API contract, the system is very composable. It's designed for dependency injection, which makes writing good tests very easy as well -- whether unit tests or functional tests. And at the top level, it should be really easy for a new developer to look at the route description and figure out exactly what the process is.

Defining a route (task list, chain) looks like this:
```go
	reg.Route("boot", "Boot the builder.").
		// Get env vars
		Does(env.Get, "vars").
		Using("HOST").WithDefault("127.0.0.1").
		Using("ETCD_PORT").WithDefault("4001").
		Using("ETCD").WithDefault("http://$HOST:$ETCD_PORT").
		Using("ETCD_PATH").WithDefault("/deis/builder").
		Using("ETCD_TTL").WithDefault("20").
		// Are there cases where we need to support more than one URL?
		Does(etcd.Client, "client").
		Using("url").WithDefault("http://localhost:4001").From("cxt:ETCD").
		// See if etcd is awake. This is the legacy way of doing it.
		Does(etcd.Ls, "ls").
		Using("client").From("cxt:client").
		Using("retries").WithDefault(20).
		// Sleep while etcd clears its stale values. Do we really need to do this?
[...]

PR #4273, opened 2015-08-14
feat(builder): reimplement the builder in Go

technosophos:
This is part of issue #3836, where the main builder scripts are
rewritten as a Go executable.

This feature adds an SSH server implementation to Go, which means we
have removed one of the major runtime interops on Builder.

This changes builder's internal port, and should not break other
things, but makes Flannel configuration easier.


builder/builder.go
```diff
@@ -0,0 +1,69 @@
...
+	routes(reg)
+
+	// Bootstrap the background services. If this fails, we stop.
+	if err := router.HandleRequest("boot", cxt, false); err != nil {
+		clog.Errf(cxt, "Fatal errror on boot: %s", err)
+		return StatusLocalError
+	}
+
+	// Set up the SSH service.
+	cxt.Put(sshd.Address, "0.0.0.0:2223")
```
kalbasit: Can we turn this bind address into a flag? I see no reason to have it hardcoded.

builder/commands_test.go
```diff
@@ -0,0 +1,27 @@
...
+
+	start := time.Now()
+	err := router.HandleRequest("test", cxt, true)
+	if err != nil {
+		t.Error(err)
+	}
+	end := time.Now()
+
+	if end.Sub(start) < 3*time.Second {
+		t.Error("Expected elapsed time to be 3 seconds.")
```
kalbasit: Error message should not be capitalized. https://github.com/golang/go/wiki/CodeReviewComments#error-strings

Also, this does not tell us what was the actual time, why don't you use the want, got pattern?

```
if want, got := end.Sub(start), 3*time.Second; want < got {
    t.Errorf("expected elapsed time to be at least %s got %s", want, got)
}
```

builder/confd/confd.go
```diff
@@ -0,0 +1,95 @@
...
+
+	limit := 1200
+	timeout := time.Second * 3
+	var lasterr error
+	start := time.Now()
+	for i := 0; i < limit; i++ {
+		if out, err := exec.Command("confd", dargs...).CombinedOutput(); err == nil {
+			log.Infof(c, "Templates generated for %s on run %d", node, i)
+			return out, nil
+		} else {
```
kalbasit: I usually find `golint` very useful.

```
$ golint ./builder/...
builder/confd/confd.go:51:10: if block ends with a return statement, so drop this else and outdent its block (move short variable declaration to its own line if necessary)
builder/etcd/etcd.go:352:11: if block ends with a return statement, so drop this else and outdent its block (move short variable declaration to its own line if necessary)
builder/git/git.go:212:9: if block ends with a return statement, so drop this else and outdent its block
```

builder/confd/confd.go
```diff
@@ -0,0 +1,95 @@
...
+// Importantly, this blocks until the run is complete.
+//
+// Params:
+// - node (string): The etcd node to use. (Only etcd is currently supported)
+//
+// Returns:
+// - The []bytes from stdout and stderr when running the program.
+//
+func RunOnce(c cookoo.Context, p *cookoo.Params) (interface{}, cookoo.Interrupt) {
+
```
kalbasit: remove the empty line from the beginning of each function, feels a lot like PHP :)

builder/confd/confd.go
```diff
@@ -0,0 +1,95 @@
...
+// reports if the command dies.
+//
+// Params:
+// - node (string): The etcd node to use. (Only etcd is currently supported)
+// - interval (int, default:5): The rebuilding interval.
+//
+// Returns
+//  bool true if this succeeded.
+func Run(c cookoo.Context, p *cookoo.Params) (interface{}, cookoo.Interrupt) {
+	// TODO: What should be done if confd dies at some point?
```
kalbasit: Move the TODO up to the Godoc section so it will be visible in the docs. Hidden TODOs are buried todos.

builder/confd/confd_test.go
```diff
@@ -0,0 +1,37 @@
...
+	reg.Route("test", "Test route").
+		Does(RunOnce, "res").
+		Using("node").WithDefault("localhost:4001")
+
+	/*
+		err := router.HandleRequest("test", cxt, true)
+		if err != nil {
+			t.Error(err)
+		}
+	*/
```
kalbasit: dead code?

builder/confd/confd_test.go
```diff
@@ -0,0 +1,37 @@
...
+		Does(Run, "res").
+		Using("node").WithDefault("localhost:4001").
+		Using("interval").WithDefault(200)
+
+	/*
+		err := router.HandleRequest("test", cxt, true)
+		if err != nil {
+			t.Error(err)
+		}
+	*/
```
kalbasit: dead code?

builder/confd/confd.go
```diff
@@ -0,0 +1,95 @@
...
+import (
+	"github.com/Masterminds/cookoo"
+	"github.com/Masterminds/cookoo/log"
+	"github.com/Masterminds/cookoo/safely"
+
+	"fmt"
+	"os/exec"
+	"strconv"
+	"time"
+)
```
kalbasit: It's idiomatic to specify standard import first and then custom ones. Please see https://github.com/golang/go/wiki/CodeReviewComments#imports

builder/docker/docker.go
```diff
@@ -0,0 +1,350 @@
...
+
+// Path to the Docker unix socket.
+var DockSock = "/var/run/docker.sock"
+
+// Cleanup removes any existing Docker artifacts.
+//
+// Returns true if the file exists (and was deleted), or false if no file
+// was deleted.
+func Cleanup(c cookoo.Context, p *cookoo.Params) (interface{}, cookoo.Interrupt) {
+
```
kalbasit: useless empty line

builder/docker/docker.go
```diff
@@ -0,0 +1,350 @@
...
+		"-d",
+		"--bip=172.19.42.1/16",
+		"--insecure-registry",
+		"10.0.0.0/8",
+		"--insecure-registry",
+		"172.16.0.0/12",
+		"--insecure-registry",
+		"192.168.0.0/16",
+		"--insecure-registry",
+		"100.64.0.0/10",
```
kalbasit: is it OK to have these hardcoded? Maybe have the real IP range (that is the one in use) up in etcd?

builder/docker/docker.go
```diff
@@ -0,0 +1,350 @@
...
+		//"--storage-driver=overlay", // Add this dynamically.
+	}
+
+	// There is probably a better way to do this. I'm not sure whether the
+	// original intent of the mkdir was to accomplish something, or just check
+	// whether overlay is supported.
+	if err := os.MkdirAll("/", 0700); err == nil {
+
```
kalbasit: useless line.

builder/docker/docker.go
```diff
@@ -0,0 +1,350 @@
...
+			dargs = append(dargs, "--storage-driver=overlay")
+		} else {
+			log.Infof(c, "File system type: '%s' (%v)", out, err)
+		}
+	}
+
+	//# spawn a docker daemon to run builds
+	//docker -d --bip=172.19.42.1/16 $DRIVER_OVERRIDE --insecure-registry 10.0.0.0/8 --insecure-registry 172.16.0.0/12
+	// --insecure-registry 192.168.0.0/16 --insecure-registry 100.64.0.0/10 &
+	//DOCKER_PID=$!
```
kalbasit: This is dead code but at the same time quite informative, can we make it a comment for the next block and clean it?

builder/docker/docker.go
```diff
@@ -0,0 +1,350 @@
...
+	})
+
+	for keepon == true {
+		if err := cli.Ping(); err == nil {
+			timer.Stop()
+			log.Infof(c, "Docker is running.")
+			return true, nil
+		}
+		time.Sleep(time.Second)
+	}
```
kalbasit: line 121-133. I feel that this can be done much cleaner having two time channels. A ticker of 1 second (or less) to ping docker, if docker is running stop all timers and return. The other timer is a simple timeout timer. `time.AfterFunc` is cool but I think it makes this code quite unreadable.

builder/docker/docker.go
```diff
@@ -0,0 +1,350 @@
...
+//
+// 	- Waiter: A *sync.WaitGroup that is waiting for the docker downloads to finish.
+//
+// Context:
+//
+// This puts 'ParallelBuild.failN" (int) into the context to indicate how many failures
+// occurred during fetches.
+func ParallelBuild(c cookoo.Context, p *cookoo.Params) (interface{}, cookoo.Interrupt) {
+	images := p.Get("images", []BuildImg{}).([]BuildImg)
+	//client := p.Get("client", nil).(*docli.Client)
```
kalbasit: dead code?

builder/docker/docker.go
```diff
@@ -0,0 +1,350 @@
...
+	var m sync.Mutex
+	var fails int
+
+	for _, img := range images {
+		img := img
+		wg.Add(1)
+		safely.GoDo(c, func() {
+			log.Infof(c, "Starting build for %s (tag: %s)", img.Path, img.Tag)
+			_, err := buildImg(c, img.Path, img.Tag)
+			//err := build(c, img.Path, img.Tag, client)
```
kalbasit: dead code?

builder/docker/docker.go
```diff
@@ -0,0 +1,350 @@
...
+	var wg sync.WaitGroup
+	var m sync.Mutex
+	var fails int
+
+	for _, img := range images {
+		img := img
+		wg.Add(1)
+		safely.GoDo(c, func() {
+			log.Infof(c, "Starting build for %s (tag: %s)", img.Path, img.Tag)
+			_, err := buildImg(c, img.Path, img.Tag)
```
kalbasit: This and the iff block can be combined:

```
if _, err := buildImg(c, img.Path, img.Tag); err != nil {
...
```

builder/docker/docker.go
```diff
@@ -0,0 +1,350 @@
...
+			wg.Done()
+		})
+
+	}
+
+	// Number of failures.
+	c.Put("ParallelBuild.failN", fails)
+
+	//wg.Wait()
+	//log.Infof(c, "Docker images are built.")
```
kalbasit: dead code?

builder/docker/docker.go
```diff
@@ -0,0 +1,350 @@
...
+	registry := host.Node.Value + ":" + port.Node.Value
+	tag := p.Get("tag", "").(string)
+
+	log.Infof(c, "Pushing %s to %s. This may take some time.", tag, registry)
+	rem := path.Join(registry, tag)
+
+	out, err := exec.Command("docker", "tag", "-f", tag, rem).CombinedOutput()
+	if err != nil {
+		log.Warnf(c, "Failed to tag %s on host %s: %s (%s)", tag, rem, err, out)
+		//return nil, err
```
kalbasit: dead code? does it make sense to continue if this command had failed?

builder/docker/docker.go
```diff
@@ -0,0 +1,350 @@
...
+	}
+	out, err = exec.Command("docker", "-D", "push", rem).CombinedOutput()
+
+	if err != nil {
+		log.Warnf(c, "Failed to push %s to host %s: %s (%s)", tag, rem, err, out)
+		return nil, err
+		//return out, err
```
kalbasit: ditto

builder/env/envvar.go
```diff
@@ -0,0 +1,42 @@
...
+// multiple times.
+//
+// For each parameter (`Using` clause), this command will look into the
+// environment for a matching variable. If it finds one, it will add that
+// variable to the context. If it does not find one, it will expand the
+// default value (so you can set a default to something like "$HOST:$PORT")
+// and also put the (unexpanded) default value back into the context in case
+// any subsequent call to `os.Getenv` occurs.
+func Get(c cookoo.Context, params *cookoo.Params) (interface{}, cookoo.Interrupt) {
+	vars := params.AsMap()
```
kalbasit: `vars` is declared but only used once, remove it. One less allocation to worry about.

builder/etcd/etcd.go
```diff
@@ -0,0 +1,524 @@
...
+	"net/http"
+	"os"
+	"os/exec"
+	"strings"
+	"time"
+
+	"github.com/Masterminds/cookoo"
+	"github.com/Masterminds/cookoo/log"
+	"github.com/Masterminds/cookoo/safely"
+
```
kalbasit: remove the empty line

builder/etcd/etcd.go
```diff
@@ -0,0 +1,524 @@
...
+func Get(c cookoo.Context, p *cookoo.Params) (interface{}, cookoo.Interrupt) {
+	cli, ok := p.Has("client")
+	if !ok {
+		return nil, errors.New("No Etcd client found.")
+	}
+	client := cli.(Getter)
+	path := p.Get("path", "/").(string)
+
+	res, err := client.Get(path, false, false)
+
```
kalbasit: this line make the err below hard to pinpoint, remove it?

builder/etcd/etcd.go
```diff
@@ -0,0 +1,524 @@
...
+func IsRunning(c cookoo.Context, p *cookoo.Params) (interface{}, cookoo.Interrupt) {
+	client := p.Get("client", nil).(Getter)
+	count := p.Get("count", 20).(int)
+	for i := 0; i < count; i++ {
+		_, err := client.Get("/", false, false)
+		if err == nil {
+			return true, nil
+		}
+		log.Infof(c, "Waiting for etcd to come online.")
+		time.Sleep(250 * time.Millisecond)
```
kalbasit: I am generally not a big fan of sleep as this will block the goroutine running here. How about a ticker?

builder/etcd/etcd.go
```diff
@@ -0,0 +1,524 @@
...
+func StoreHostKeys(c cookoo.Context, p *cookoo.Params) (interface{}, cookoo.Interrupt) {
+	defaultCiphers := []string{"rsa", "dsa", "ecdsa", "ed25519"}
+	client := p.Get("client", nil).(GetterSetter)
+	ciphers := p.Get("ciphers", defaultCiphers).([]string)
+	basepath := p.Get("basepath", "/deis/builder").(string)
+
+	res, err := client.Get("sshHostKey", false, false)
+	if err != nil || res.Node == nil {
+		log.Infof(c, "Could not get SSH host key from etcd. Generating new ones.")
+		genSSHKeys(c)
```
kalbasit: it feels to me that this function must return the err it has and you have to handle it here. Basically what to do if it failed to generate the host SSH keys?

builder/etcd/etcd.go
```diff
@@ -0,0 +1,524 @@
...
+// This will notify etcd as long as the local sshd is running.
+//
+// Params:
+// 	- base (string): The base path to write the data: $base/host and $base/port.
+// 	- host (string): The hostname
+// 	- port (string): The port
+// 	- client (Setter): The client to use to write the data to etcd.
+// 	- sshPid (int): The PID for SSHD. If SSHD dies, this stops notifying.
+func UpdateHostPort(c cookoo.Context, p *cookoo.Params) (interface{}, cookoo.Interrupt) {
+
```
kalbasit: useless line

builder/etcd/etcd.go
```diff
@@ -0,0 +1,524 @@
...
+	base := p.Get("base", "").(string)
+	host := p.Get("host", "").(string)
+	port := p.Get("port", "").(string)
+	client := p.Get("client", nil).(Setter)
+	sshd := p.Get("sshdPid", 0).(int)
+
+	// If no port is specified, we don't do anything.
+	if len(port) == 0 {
```
kalbasit: why len(port)? why not `if port == ""` ?

builder/etcd/etcd.go
```diff
@@ -0,0 +1,524 @@
...
+	}
+
+	// Update etcd every ten seconds with this builder's host/port.
+	safely.GoDo(c, func() {
+		ticker := time.Tick(10 * time.Second)
+		for range ticker {
+			//log.Infof(c, "Setting SSHD host/port")
+			if _, err := os.FindProcess(sshd); err != nil {
+				log.Errf(c, "Lost SSHd process: %s", err)
+				return
```
kalbasit: you forgot to stop the ticker before returning.

kalbasit: I don't see why retuning from both branches here. Perhaps break out of the `for` block and after it just stop the ticker.

```
		for range ticker {
			if _, err := os.FindProcess(sshd); err != nil {
				log.Errf(c, "Lost SSHd process: %s", err)
				break
			} else if err := setHostPort(client, base, host, port, ttl); err != nil {
					log.Errf(c, "Etcd error setting host/port: %s", err)
					break
			}
		}
        ticker.Stop()
	})
```

builder/etcd/etcd.go
```diff
@@ -0,0 +1,524 @@
...
+	}
+
+	res, err := client.CreateDir(name, ttl)
+	if err != nil {
+		// FIXME: Do we want this to be recoverable?
+		return res, &cookoo.RecoverableError{err.Error()}
+	}
+
+	return res, nil
+
```
kalbasit: empty line

builder/etcd/etcd.go
```diff
@@ -0,0 +1,524 @@
...
+	safely.GoDo(c, func() {
+		<-stop
+		stopwatch <- true
+		stopetcd <- true
+		close(stopwatch)
+		close(stopetcd)
+	})
+
+	return stop, nil
+	*/
```
kalbasit: remove dead code

builder/etcd/etcd.go
```diff
@@ -0,0 +1,524 @@
...
+}
+
+// checkRetry overrides etcd.DefaultCheckRetry.
+//
+// It adds configurable number of retries and configurable timesouts.
+func checkRetry(c *etcd.Cluster, numReqs int, last http.Response, err error) error {
+
```
kalbasit: empty line

builder/git/git.go
```diff
@@ -0,0 +1,228 @@
...
+		hook, err := prereceiveHook(map[string]string{"GitHome": gitHome})
+		if err != nil {
+			return true, err
+		}
+		ioutil.WriteFile(filepath.Join(repoPath, "hooks", "pre-receive"), hook, 0755)
+
+		return true, nil
+	} else if err == nil {
+		return false, errors.New("Expected directory, found file.")
+	} else {
```
kalbasit: you don't need the last else.

builder/rootfs/bin/boot
```diff
@@ -18,153 +18,145 @@ export ETCD_PORT=${ETCD_PORT:-4001}
...
+	#ERROR="$(etcdctl --no-sync -C $ETCD mkdir $1 2>&1 >/dev/null)"
+	#if [[ $? -ne 0 && $(echo $ERROR | grep -ive "key already exists") ]]; then
+		#echo "etcd_safe_mkdir: an etcd error occurred ($ERROR)"
+		#echo "aborting..."
+		#exit 1
+	#fi
+	#set -e
+#}
+
+#etcd_safe_mkdir $ETCD_PATH/users
```
kalbasit: dead code

builder/rootfs/etc/confd/templates/check-repos
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+## DEPRECATED. The Go builder now does this itself.
```
kalbasit: remove it then?

builder/rootfs/etc/confd/templates/gitreceive
```diff
@@ -1,6 +1,10 @@
 #!/usr/bin/env bash
 set -eo pipefail
 
+## !!!!!!!!!!!!!! ##
+## NO LONGER USED ## See builder.go
```
kalbasit: remove it then?

builder/rootfs/usr/local/src/slugbuilder/Dockerfile
```diff
@@ -3,7 +3,7 @@ FROM heroku/cedar:14
 RUN mkdir /app
 RUN addgroup --quiet --gid 2000 slug && \
   useradd slug --uid=2000 --gid=2000 --home-dir /app --no-create-home
-RUN chown -R slug:slug /app
+RUN chown -R 2000:2000 /app
```
kalbasit: why the change? `slu:slug` should exist by now (see one line above).

builder/rootfs/usr/local/src/slugbuilder/builder/build.sh
```diff
@@ -147,6 +148,10 @@ if [[ -s "$build_root/.release" ]]; then
     [[ $default_types ]] && echo_normal "Default process types for $buildpack_name -> $default_types"
 fi
 
+# Fix any wayward permissions. We want everything in app to be owned
+# by slug.
+chown -R 2000:2000 $build_root/*
```
kalbasit: `s/2000/slug`

builder/rootfs/usr/local/src/slugrunner/Dockerfile
```diff
@@ -4,7 +4,7 @@ RUN mkdir /app
 RUN addgroup --quiet --gid 2000 slug && \
     useradd slug --uid=2000 --gid=2000 --home-dir /app --no-create-home \
         --shell /bin/bash
-RUN chown slug:slug /app
+RUN chown 2000:2000 /app
```
kalbasit: ditto.

builder/rootfs/usr/local/src/slugrunner/runner/init
```diff
@@ -11,19 +11,25 @@ export HOME=/app
 mkdir -p $HOME
 
 if [[ $(ls -A $HOME) ]]; then
-	true
+  true
```
kalbasit: now misaligned.

builder/rootfs/usr/local/src/slugrunner/runner/init
```diff
@@ -11,19 +11,25 @@ export HOME=/app
...
 cd $HOME
 
 ## Load profile.d and release config
 
 shopt -s nullglob
 mkdir -p .profile.d
+
+# If a file is created in slugbuilder with the wrong UID, change it. But we
+# don't mess with anything owned by UID 0.
+find . -user 1000 -exec chown 2000:2000 {} \;
```
kalbasit: `s/2000/slug/`

builder/builder.go
technosophos: A flag?

builder/confd/confd.go
technosophos: I intentionally ignored lint on that one. If you look at the code, you'll see why. `out` is not in scope without the else. IIRC, `out` is opened for GC as long as it's in the else block because the sleep call will yield to the scheduler.

I don not tend to blindly follow golint's suggestions, though I do consider them.

builder/confd/confd.go
technosophos: TODO should not show up in the Godoc. It's only for people who are actually reading the code.

builder/docker/docker.go
technosophos: I do not know the history behind these being hard-coded. They were in the original. If anyone wants to provide some informed background, I'm open to hearing it.

builder/docker/docker.go
technosophos: The reason that was commented was that it turns out that docker will return 1 if the label already exists, and the rebuild does not change anything. That happens on occasion, and is not a fatal error. I'll remove the commented code.

builder/env/envvar.go
technosophos: What makes you think the compiler will not inline that? I'll remove it... but I don't think it actually has any adverse impacts on performance.

builder/etcd/etcd.go
technosophos: I don't understand. The point of the sleep is to block the goroutine... we can't move on to the next step until etcd is online. Sleep does yield to the scheduler, so it's not blocking anything other than the goroutine that it's supposed to be blocking.

builder/etcd/etcd.go
technosophos: Is there a reason to favor one over the other? I would have assumed that len(str) is slightly more efficient than == "", but probably both are optimized to the same.

builder/etcd/etcd.go
technosophos: My bad -- time.Tick() doesn't actually return a ticker. That's a big mistake on my part.

builder/etcd/etcd.go
technosophos: As the comment there said, this code is intentionally left in. It's the right way to do things, but we're waiting for upstream Etcd to fix the watch() code. As soon as we switch to etcd2 (which we're doing now) we can re-test this.

builder/git/git.go
technosophos: `err` is out of scope without it.

builder/rootfs/usr/local/src/slugbuilder/Dockerfile
technosophos: There was an error where Docker was assigning slug the UID 1000 instead of 2000 in spite of line 5. I think I fixed the problem elsewhere. I will go ahead and change it back... but there was a reason behind using the UID instead of the name.

builder/builder.go
kalbasit: A flag or an option of cookoo, basically a command-line controlled value.

builder/confd/confd.go
kalbasit: I'll probably just write it like so

```
for i := 0; i < limit; i++ {
	out, err := exec.Command("confd", dargs...).CombinedOutput()
	if err == nil {
		log.Infof(c, "Templates generated for %s on run %d", node, i)
		return out, nil
	}

	log.Debugf(c, "Recoverable error: %s", err)
	log.Debugf(c, "Output: %q", out)
	lasterr = err
	time.Sleep(timeout)
	log.Infof(c, "Re-trying template build. (Elapsed time: %d)", time.Now().Sub(start)/time.Second)
}
```

builder/confd/confd.go
kalbasit: Very true, but also most people reading the code are developers and they might just fix it for you. Anyway, I don't feel strongly about this so I'll leave it to you.

builder/docker/docker.go
kalbasit: This is basically starting the registry allowing it to accept connections from AWS network, bare metal network and other networks. That's fine but it means that we have to maintain it every time a new CIDR shows up. Not sure where the right place for these either. @carmstrong ?

builder/etcd/etcd.go
kalbasit: it blocks not in the sense of next steps but in the sense that it cannot receive a quit (signal or otherwise). Having a ticker for example allows you to also have a quit channel so you can ask the goroutine to die regardless of the ticker blocking.

builder/etcd/etcd.go
kalbasit: Good point.

```
package a

import "testing"

func TestX(t *testing.T) {
}

func BenchmarkEqual(b *testing.B) {
	str := ""
	for i := 0; i < b.N; i++ {
		if str == "" {
			// do something
		}
	}
}

func BenchmarkLen(b *testing.B) {
	str := ""
	for i := 0; i < b.N; i++ {
		if len(str) == 0 {
			// do something
		}
	}
}
```

```
kalbasit@wks-113 ~/code/src/github.com/icza/gowut/gwu [master] ± % go test -v -bench=. -run=XXX /tmp/a_test.go
PASS
BenchmarkEqual-8	2000000000	         0.34 ns/op
BenchmarkLen-8  	2000000000	         0.32 ns/op
ok  	command-line-arguments	1.388s
kalbasit@wks-113 ~/code/src/github.com/icza/gowut/gwu [master] ± % go test -v -bench=. -run=XXX /tmp/a_test.go
PASS
BenchmarkEqual-8	2000000000	         0.33 ns/op
BenchmarkLen-8  	2000000000	         0.34 ns/op
ok  	command-line-arguments	1.413s
kalbasit@wks-113 ~/code/src/github.com/icza/gowut/gwu [master] ± % go test -v -bench=. -run=XXX /tmp/a_test.go
PASS
BenchmarkEqual-8	2000000000	         0.32 ns/op
BenchmarkLen-8  	2000000000	         0.33 ns/op
ok  	command-line-arguments	1.368s
```

I guess it doesn't matter, they perform the same. Go with whichever feel right (style-wize).

builder/etcd/etcd.go
kalbasit: Oh I see, fair enough then :+1:

builder/rootfs/usr/local/src/slugbuilder/Dockerfile
kalbasit: I see, revert it back if it was still acting up then.

builder/builder.go
technosophos: How about an env var? That's consonant with how the rest of the Docker config seems to work.

builder/confd/confd.go
technosophos: So in your revised example above, `out` is now scoped to the block of the entire for loop. Originally, `out` is just scoped to the `if/else` block. Since there's a `time.Sleep` call, `out` remains allocated for that entire time, and the GC can't clean it up.

The reason I did it the way I did is so that the GC can clean up `out` and that memory is not allocated for the entire time during this goroutine's inactive period.

Clearly, this is a micro-optimization, and realistically it is unlikely to ever make a difference in the Builder's performance. But my argument remains that it is preferable to code this way than to simply do what `golint` suggests.

builder/etcd/etcd.go
technosophos: Given that there is no cancellation logic in the existing model, and that we don't currently have a use case for adding one in this context, my inclination is to leave this as is. If at some point we need cancellation logic, we can refactor to use a Ticker.

builder/builder.go
kalbasit: I feel a flag is much better (logger was refactored from env vars to https://github.com/deis/deis/blob/master/logger/main.go#L30-L40) but if you use https://github.com/namsral/flag which is a drop-in replacement for the standard flag package, you'll get both a flag and env var :)

builder/confd/confd.go
kalbasit: Agree on that point and that's the second reason I dislike sleep :) The best would be to re-write this block as `for { select {` block but please move on if you feel it's taking too much of your time. We will revisit afterwards.

builder/etcd/etcd.go
kalbasit: am fine leaving it to later.. we can refactor after the first working release.

builder/builder.go
bacongobbler: Both are fine as long as we're consistent. I lean towards flags simply because the `flag` package deals with usage messages, flag parsing as well as built-in conversion to the proper types (converting envvars from a string to boolean is a little harder than using `flag.Bool()`).

builder/confd/confd.go
bacongobbler: This isn't much of a concern. We let the tooling determine the acceptable styling of the code. As long as it passes `make test-style` (which in turn runs `gofmt`), we're happy with the styling.

builder/confd/confd.go
bacongobbler: I agree with @technosophos. Formal documentation shouldn't have TODOs unless it is a serious design flaw, which then there should be a big fat NOTICE or WARNING header. Developers who are reading the source will still notice the TODO, so I'm fine with it here.

builder/docker/docker.go
bacongobbler: These are the private network ranges from [RFC 1918](https://tools.ietf.org/html/rfc1918) and [RFC 6598](https://tools.ietf.org/html/rfc6598). I think it's fine to hardcode them until another RFC comes up, proposing another address space.

kalbasit: fair enough then.

bacongobbler:
@technosophos is this still ready for more reviews? I see that https://github.com/deis/deis/pull/4273#discussion_r37148119 hasn't been addressed yet but I'm willing to give this a whirl if it's ready :)

technosophos:
@bacongobbler The code functions as intended without that modification, so please keep reviewing.

builder/confd/confd.go
```diff
@@ -0,0 +1,95 @@
...
+
+	cmd := exec.Command("confd", "-log-level", "error", "-node", node, "-interval", interval)
+	if err := cmd.Start(); err != nil {
+		return false, err
+	}
+
+	log.Infof(c, "Watching confd.")
+	safely.Go(func() {
+		if err := cmd.Wait(); err != nil {
+			log.Infof(c, "confd exited with error: %s", err)
```
bacongobbler: I think we should kill execution of the builder here if `confd` exits with an error.

builder/docker/docker.go
```diff
@@ -0,0 +1,337 @@
...
+
+	"github.com/Masterminds/cookoo"
+	"github.com/Masterminds/cookoo/log"
+	"github.com/Masterminds/cookoo/safely"
+	"github.com/deis/deis/builder/etcd"
+	docli "github.com/fsouza/go-dockerclient"
+)
+
+// Path to the Docker unix socket.
+var DockSock = "/var/run/docker.sock"
```
bacongobbler: While not necessary to refactor, in the future docker has this constant defined at https://github.com/docker/docker/blob/51249a3aa09b8e1ade07ed6be0029a190b978b05/opts/opts.go#L27 so we don't have to re-declare this variable :)

builder/docker/docker.go
```diff
@@ -0,0 +1,337 @@
...
+		"--bip=172.19.42.1/16",
+		"--insecure-registry",
+		"10.0.0.0/8",
+		"--insecure-registry",
+		"172.16.0.0/12",
+		"--insecure-registry",
+		"192.168.0.0/16",
+		"--insecure-registry",
+		"100.64.0.0/10",
+		//"--storage-driver=overlay", // Add this dynamically.
```
bacongobbler: how is this added dynamically? Through docker inferring the underlying filesystem type?

builder/docker/docker.go
```diff
@@ -0,0 +1,337 @@
...
+	}
+
+	// There is probably a better way to do this. I'm not sure whether the
+	// original intent of the mkdir was to accomplish something, or just check
+	// whether overlay is supported.
```
bacongobbler: ping @mboersma @carmstrong to see if they know anything about this

builder/docker/docker.go
bacongobbler: ah, I see below. Would you mind removing this? :)

builder/docker/docker.go
technosophos: That is good to know. I hadn't even thought about checking to see if there was already a default const defined.

builder/docker/docker.go
```diff
@@ -0,0 +1,337 @@
...
+	host, err := client.Get("/deis/registry/host", false, false)
+	if err != nil {
+		return nil, err
+	}
+	port, err := client.Get("/deis/registry/port", false, false)
+	if err != nil {
+		return nil, err
+	}
+
+	registry := host.Node.Value + ":" + port.Node.Value
```
bacongobbler: We've seen funny issues before if you don't check the value of `.Node`, as seen [here](https://github.com/deis/deis/commit/e803109cf9bc1bd868aab5c7d3bdcae18da589fd). That was only in the case for etcd's watch where they'd occasionally return a nil Node, but maybe it could happen during a `.Get` as well?

builder/docker/docker.go
```diff
@@ -0,0 +1,337 @@
...
+	tag := p.Get("tag", "").(string)
+
+	log.Infof(c, "Pushing %s to %s. This may take some time.", tag, registry)
+	rem := path.Join(registry, tag)
+
+	out, err := exec.Command("docker", "tag", "-f", tag, rem).CombinedOutput()
+	if err != nil {
+		log.Warnf(c, "Failed to tag %s on host %s: %s (%s)", tag, rem, err, out)
+	}
+	out, err = exec.Command("docker", "-D", "push", rem).CombinedOutput()
```
bacongobbler: Is there a reason we want to run `docker push` with the debug flag?

builder/env/envvar.go
```diff
@@ -0,0 +1,41 @@
...
+		if val = os.Getenv(name); len(val) == 0 {
+			def := def.(string)
+			// We want to make sure that any subsequent calls to Getenv
+			// return the same default.
+			os.Setenv(name, def)
+
+			val = os.ExpandEnv(def)
+		}
+		c.Put(name, val)
+		//c.Logf("info", "Name: %s, Val: %s", name, val)
```
bacongobbler: perhaps we can add this as a debug log? i.e. `c.Logf("debug", ...)`

builder/env/envvar_test.go
```diff
@@ -0,0 +1,60 @@
...
+func TestGet(t *testing.T) {
+	reg, router, cxt := cookoo.Cookoo()
+
+	drink := "DEIS_DRINK_OF_CHOICE"
+	cookies := "DEIS_FAVORITE_COOKIES"
+	snack := "DEIS_SNACK_TIME"
+	snackVal := fmt.Sprintf("$%s and $%s cookies", drink, cookies)
+
+	// Set drink, but not cookies.
+	os.Setenv(drink, "coffee")
```
bacongobbler: +:100:

builder/etcd/etcd.go
```diff
@@ -0,0 +1,530 @@
...
+	}
+	client := cli.(DirCreator)
+
+	if len(name) == 0 {
+		return false, errors.New("Expected directory name to be more than zero characters.")
+	}
+
+	res, err := client.CreateDir(name, ttl)
+	if err != nil {
+		// FIXME: Do we want this to be recoverable?
```
bacongobbler: I think this is fine for now. As long as we log the error, the user can debug from there and check what's happening with etcd.

builder/etcd/etcd.go
```diff
@@ -0,0 +1,530 @@
...
+		return res, &cookoo.RecoverableError{err.Error()}
+	}
+
+	return res, nil
+
+}
+
+// Watch watches a given path, and executes a git cehck-repos for each event.
```
bacongobbler: typo: should be "check-repos"

builder/git/git.go
```diff
@@ -0,0 +1,228 @@
...
+// Returns:
+// 	- nothing
+func Receive(c cookoo.Context, p *cookoo.Params) (interface{}, cookoo.Interrupt) {
+	if ok, z := p.Requires("channel", "request", "fingerprint", "permissions"); !ok {
+		return nil, fmt.Errorf("Missing requirements %q", z)
+	}
+	repoName := p.Get("repoName", "").(string)
+	operation := p.Get("operation", "").(string)
+	channel := p.Get("channel", nil).(ssh.Channel)
+	//req := p.Get("request", nil).(*ssh.Request)
```
bacongobbler: remove dead code?

builder/rootfs/Dockerfile
```diff
@@ -46,7 +46,7 @@ VOLUME /var/lib/docker
 
 ENTRYPOINT ["/bin/entry"]
 CMD ["/bin/boot"]
-EXPOSE 22
+EXPOSE 2223
```
bacongobbler: good catch

builder/rootfs/bin/boot
```diff
@@ -18,153 +18,9 @@ export ETCD_PORT=${ETCD_PORT:-4001}
...
-		sleep $(($ETCD_TTL/2)) # sleep for half the TTL
-	done
-
-	# if the loop quits, something went wrong
-	exit 1
-
-fi
+echo External port $EXTERNAL_PORT
 
+builder
```
bacongobbler: can we remove /bin/boot in lieu of just running /bin/builder directly? Or perhaps an `exec builder` would be useful here? That way the builder can take PID 1 in the process tree.

builder/confd/confd.go
technosophos: So the `cmd.Wait` is running in a goroutine. Basically, we're doing the equivalent of backgrounding `confd`, since it is running as a daemon at that point. Inside that goroutine we could do whatever we want.

I could, from there, kill the entire builder if confd fails. I could also just try restarting confd again. Or I could leave it. The current (shell-based) builder does nothing if confd fails, which is why I wrote the above as is.

Thoughts?

builder/docker/docker.go
technosophos: The reason it's declared, and declared as a var, is so that it can be changed globally at startup time. (flag, net/http and other packages do this). But... I liked the idea of using the Docker-defined one as the default value (`DockSock = opts.DefaultUnixSocket`), but... it appears that that var doesn't exist in our copy of that repo. We're too far back. So... I checked go-dockerclient, and it has that same variable declared, but... it's not exported (`dockerUnixSocket`). Grr...

So... I guess I'll leave it for now, and fix it when we upgrade to a more recent Docker library.

builder/confd/confd.go
bacongobbler: The correct way is to kill the builder. That was a bug in the older builder. In other words...

builder/docker/docker.go
bacongobbler: sounds good :+1:

builder/docker/docker.go
technosophos: Fixed

builder/docker/docker.go
technosophos: I can remove the comment now. I tested it, and the idea is to force an overlay/union-style filesystem to create a layer. I tested with overlay and aufs.

So I'll revise the comment.

builder/docker/docker.go
technosophos: That is good to know. Fixed.

builder/docker/docker.go
technosophos: I did it so that on failures we would get better output in the journal. I added after Jenkins hit a whole bunch of `docker push` errors. I don't mind removing it... but it is only showing the debug data on failure, which seems worthwhile to me. Thoughts?

builder/env/envvar.go
technosophos: Good idea. Fixed.

builder/etcd/etcd.go
technosophos: Okay. Removed comment.

builder/etcd/etcd.go
technosophos: Good catch. Fixed.

builder/git/git.go
technosophos: Removed several.

builder/rootfs/bin/boot
technosophos: As I look at that, I think we could do it pretty easily. I need to see whether there's anything magical in environment_proxy. But if not, it would be pretty straightforward. I'll test out the `exec` method right now, and also look at replacing the entire script.

builder/Makefile
```diff
@@ -11,9 +11,12 @@ COMPONENT = $(notdir $(repo_path))
 IMAGE = $(IMAGE_PREFIX)$(COMPONENT):$(BUILD_TAG)
 DEV_IMAGE = $(REGISTRY)$(IMAGE)
 BINARIES := extract-domain extract-types extract-version generate-buildhook get-app-config get-app-values publish-release-controller yaml2json-procfile
+
 BINARY_DEST_DIR := rootfs/usr/bin
 
 build: check-docker
+	GOOS=linux GOARCH=amd64 CGO_ENABLED=0 godep go build -a -installsuffix cgo -ldflags '-s' -o $(BINARY_DEST_DIR)/builder cli/builder.go || exit 1; \
+		$(call check-static-binary,$(BINARY_DEST_DIR)/builder)
```
mboersma: This has inconsistent spacing and should start with `@` so it's silent, but let's fix that afterward if there aren't other substantive changes remaining to do here.

builder/README.md
```diff
@@ -1,6 +1,9 @@
 # Deis Builder
 
-A Docker image that builds Docker images, for use in the [Deis](http://deis.io/) open source PaaS.
+**WARNING:** This is the experimental Go-based builder.
```
mboersma: Should the **WARNING** line be replaced with the old one above it before merge? Or did we want to leave it in there until closer to the next release?

builder/builder.go
```diff
@@ -0,0 +1,78 @@
...
+)
+
+// Run starts the Builder service.
+//
+// The Builder service is responsible for setting up the local container
+// environment and then listening for new builds. The main listening service
+// is SSH. Builder listens for new Git commands and then sends those on to
+// Git.
+//
+// Run returns on of the Status* status code constants.
```
mboersma: Thank you very much for writing clear and detailed code comments everywhere. :star:

builder/builder.go
```diff
@@ -0,0 +1,78 @@
...
+
+	// Supply route names for handling various internal routing. While this
+	// isn't necessary for Cookoo, it makes it easy for us to mock these
+	// routes in tests. c.f. sshd/server.go
+	cxt.Put("route.sshd.pubkeyAuth", "pubkeyAuth")
+	cxt.Put("route.sshd.sshPing", "sshPing")
+	cxt.Put("route.sshd.sshGitReceive", "sshGitReceive")
+
+	// Start the SSH service.
+	// XXX: We could refactor Serve to be a command, and then run this as
```
mboersma: s/XXX/TODO/ ?

builder/commands_test.go
```diff
@@ -0,0 +1,27 @@
...
+
+func TestSleep(t *testing.T) {
+	reg, router, cxt := cookoo.Cookoo()
+
+	reg.Route("test", "Test route").
+		Does(Sleep, "res").Using("duration").WithDefault(3 * time.Second)
+
+	start := time.Now()
+	err := router.HandleRequest("test", cxt, true)
+	if err != nil {
```
mboersma: This `err != nil` could be collapsed with the previous line, which we seem to prefer whenever possible. (So there's one style thing that isn't enforced by our linting tools--ping @kalbasit --but therefore also isn't strictly required.)

builder/commands.go
```diff
@@ -0,0 +1,58 @@
...
+		log.Info(c, "Builder is running.")
+
+		<-sigs
+
+		c.Log("info", "Builder received signal to stop.")
+		pids := p.AsMap()
+		killed := 0
+		for name, pid := range pids {
+			pid, ok := pid.(int)
+			if ok {
```
mboersma: This typecast could be collapsed with the previous line, which we prefer.

builder/confd/confd.go
mboersma: Agree with @bacongobbler--I expect TODOs to be scattered throughout code comments, which is why many editors have helpful plugins that grep for such lines.

builder/docker/doc.go
```diff
@@ -0,0 +1,7 @@
+/*
+Package docker provides Docker utlities.
+
+Some of the utilities rely upon the presence of the command line client for
+Docker. Others use the Docker API.
+*/
+package docker
```
mboersma: I like this convention of putting the package comment in `doc.go`. I tend to put it in `main.go` or any random source file, so that readers have to search for it. :disappointed:

builder/docker/docker.go
mboersma: We should add a comment to avoid confusion:
```go
// Allow insecure Docker registries on all private network ranges in RFC 1918 and RFC 6598.
```

builder/docker/docker.go
```diff
@@ -0,0 +1,337 @@
...
+//
+// Params:
+// 	- client (*docker.Client): Docker client.
+// 	- timeout (time.Duration): Time after which to give up.
+//
+// Returns:
+// 	- boolean true if the server is up.
+func WaitForStart(c cookoo.Context, p *cookoo.Params) (interface{}, cookoo.Interrupt) {
+	ok, missing := p.RequiresValue("client")
+	if !ok {
```
mboersma: This line could be collapsed into the one above, nitpick++.

builder/docker/docker.go
mboersma: Actually this was one of the first uses of Go closures that I found natural and didn't have to re-read. But if using two timers would be simpler, that's a good argument; let's see some (pseudo-)code.

builder/docker/docker.go
```diff
@@ -0,0 +1,337 @@
...
+// occurred during fetches.
+func ParallelBuild(c cookoo.Context, p *cookoo.Params) (interface{}, cookoo.Interrupt) {
+	images := p.Get("images", []BuildImg{}).([]BuildImg)
+
+	var wg sync.WaitGroup
+	var m sync.Mutex
+	var fails int
+
+	for _, img := range images {
+		img := img
```
mboersma: This line hurt my brain but I understand now, thanks @technosophos.

builder/docker/docker.go
mboersma: If the debug output has already been helpful with debugging, let's leave it in for now. We can silence it later if it is too much spew.

builder/docker/docker_test.go
```diff
@@ -0,0 +1 @@
+package docker
```
mboersma: If this file amounts to just a placeholder for tests, maybe it shouldn't be included. When I scan a go package and see foo_test.go, my assumption is "good, that means foo.go has tests."

builder/env/envvar_test.go
mboersma: This code made me hungry. :cookie:

builder/etcd/etcd.go
```diff
@@ -0,0 +1,529 @@
...
+
+	res, err := client.Get(path, false, false)
+	if err != nil {
+		return res, err
+	}
+
+	if !res.Node.Dir {
+		return res, fmt.Errorf("Expected / to be a dir.")
+	}
+	return res, err
```
mboersma: Shouldn't this be `return res, nil`? I know `err` is probably equal to `nil` if we got here, but go code tends to make it explicit IMHO.

builder/etcd/etcd.go
```diff
@@ -0,0 +1,529 @@
...
+	value := p.Get("value", "").(string)
+	ttl := p.Get("ttl", uint64(20)).(uint64)
+	client := p.Get("client", nil).(Setter)
+
+	res, err := client.Set(key, value, ttl)
+	if err != nil {
+		log.Infof(c, "Failed to set %s=%s", key, value)
+	}
+
+	return res, err
```
mboersma: Same comment as above: `return res, nil` seems more idiomatic.

builder/etcd/etcd.go
```diff
@@ -0,0 +1,529 @@
...
+		}
+	}
+
+	return "", fmt.Errorf("User not found for fingerprint %s", fingerprint)
+}
+
+// StoreHostKeys stores SSH hostkeys locally.
+//
+// First it tries to fetch them from etcd. If the keys are not present there,
+// it generates new ones and then puts the new onces into etcd.
```
mboersma: s/onces/ones/ or just s/the new onces/them/

builder/etcd/etcd.go
```diff
@@ -0,0 +1,529 @@
...
+func checkRetry(c *etcd.Cluster, numReqs int, last http.Response, err error) error {
+
+	if numReqs > retryCycles*len(c.Machines) {
+		return fmt.Errorf("Tried and failed %d cluster connections: %s", retryCycles, err)
+	}
+
+	switch last.StatusCode {
+	case 0:
+		return nil
+	case 500:
```
mboersma: Would the constant [`http.StatusInternalServerError`](http://golang.org/pkg/net/http/#pkg-constants) make this more readable?

builder/etcd/etcd.go
```diff
@@ -0,0 +1,529 @@
...
+		time.Sleep(retrySleep)
+		return nil
+	case 200:
```
mboersma: Ditto for `http.StatusOK`.

builder/etcd/etcd_test.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+func TestCreateClient(t *testing.T) {
+	reg, router, cxt := cookoo.Cookoo()
+
+	reg.Route("test", "Test route").
+		Does(CreateClient, "res").Using("url").WithDefault("localhost:4100")
+
+	err := router.HandleRequest("test", cxt, true)
+	if err != nil {
```
mboersma: Could be a one-liner, yadda yadda. :smiling_imp:

builder/git/git.go
```diff
@@ -0,0 +1,224 @@
...
+	"strings"
+	"sync"
+	"text/template"
+
+	"github.com/Masterminds/cookoo"
+	"github.com/Masterminds/cookoo/log"
+	"golang.org/x/crypto/ssh"
+)
+
+// PrereceiveHookTmpl is a pre-recieve hook.
```
mboersma: s/pre-recieve/pre-receive/

builder/git/git.go
```diff
@@ -0,0 +1,224 @@
...
+//
+// This is overridable. The following template variables are passed into it:
+//
+// 	.GitHome: the path to Git's home directory.
+var PrereceiveHookTpl = `#!/bin/bash
```
mboersma: If there were some way to keep this bash script as a separate file, yet have it be compiled in, we could have `shellcheck` validate it...

builder/sshd/server.go
```diff
@@ -0,0 +1,357 @@
...
+const (
+	// HostKeys is the context key for Host Keys list.
+	HostKeys string = "ssh.HostKeys"
+	// Address is the context key for SSH address.
+	Address string = "ssh.Address"
+	// ServerConfig is the context key for ServerConfig object.
+	ServerConfig string = "ssh.ServerConfig"
+)
+
+// PrereceiveHookTmpl is a pre-recieve hook.
```
mboersma: s/pre-recieve/pre-receive/

builder/routes.go
```diff
@@ -0,0 +1,317 @@
...
+					{Name: "client", From: "cxt:client"},
+				},
+			},
+			/*
+				// Start SSHD.
+				cookoo.Cmd{
+					Name: "sshdstart",
+					Fn:   sshd.Start,
+				},
+			*/
```
mboersma: This commented-out code should be removed.

builder/sshd/server.go
```diff
@@ -0,0 +1,357 @@
...
+
+	// Answer all the requests on this connection.
+	for req := range requests {
+		ok := false
+
+		// I think that ideally what we want to do here is pass this on to
+		// the Cookoo router and let it handle each Type on its own.
+		switch req.Type {
+		case "env":
+			//k, v := parseEnv(req.Payload)
```
mboersma: Commented-out line should be removed.

builder/sshd/server.go
```diff
@@ -0,0 +1,357 @@
...
+			// channel open to take additional requests.
+			log.Infof(s.c, "Received request of type %s\n", req.Type)
+			req.Reply(false, nil)
+		}
+	}
+
+	return nil
+}
+
+/*
```
mboersma: More commented-out code that could be removed.

builder/sshd/server.go
```diff
@@ -0,0 +1,357 @@
...
+// GenericMessage describes a simple string message, which is common in SSH.
+type GenericMessage struct {
+	Value string
+}
+
+// cleanExec cleans the exec string.
+func cleanExec(pay []byte) string {
+	e := &ExecCmd{}
+	ssh.Unmarshal(pay, e)
+	// XXX: Minimal escaping of values in command. There is probably a better
```
mboersma: s/XXX/TODO/ ?

builder/sshd/server.go
```diff
@@ -0,0 +1,357 @@
...
+
+	return key, val
+
+}
+
+// Ping handles a simple test SSH exec.
+//
+// Returns the string PONG and exit status 0.
+//
+// Params:
```
mboersma: Should flesh out (or omit) the Params: and Returns: sections of this comment.

builder/sshd/sshd.go
```diff
@@ -0,0 +1,235 @@
...
+
+}
+
+// ParseHostKeys parses the host key files.
+//
+// By default it looks in /etc/ssh for host keys of the patterh ssh_host_{{TYPE}}_key.
+//
+// Params:
+// 	- keytypes ([]string): Key types to parse. Defaults to []string{rsa, dsa, ecdsa}
+// 	- enabeV1 (bool): Allow V1 keys. By default this is disabled.
```
mboersma: s/enabeV1/enableV1/

builder/sshd/sshd.go
```diff
@@ -0,0 +1,235 @@
...
+func AuthKey(c cookoo.Context, p *cookoo.Params) (interface{}, cookoo.Interrupt) {
+	meta := p.Get("metadata", nil).(ssh.ConnMetadata)
+	key := p.Get("key", nil).(ssh.PublicKey)
+	authorized := p.Get("authorizedKeys", []string{}).([]string)
+
+	auth := new(ssh.CertChecker)
+	auth.UserKeyFallback = func(meta ssh.ConnMetadata, pk ssh.PublicKey) (*ssh.Permissions, error) {
+
+		// This gives us a string in the form "ssh-rsa LONG_KEY"
+		//supplied := strings.TrimSpace(string(ssh.MarshalAuthorizedKey(pk)))
```
mboersma: Death to commented-out code!

builder/sshd/sshd.go
```diff
@@ -0,0 +1,235 @@
...
+			}
+
+			// We use a contstant time compare more as a precaution than anything
+			// else. A timing attack here would be very difficult, but... better
+			// safe than sorry.
+			if allowed.Type() == suppliedType && subtle.ConstantTimeCompare(allowed.Marshal(), supplied) == 1 {
+				log.Infof(c, "Key accepted for user %s.", meta.User())
+				perm := &ssh.Permissions{
+					Extensions: map[string]string{
+						//"key":  ssh.MarshalAuthorizedKey(supplied.Marshal()),
```
mboersma: See previous hyperbole.

builder/sshd/sshd.go
```diff
@@ -0,0 +1,235 @@
...
+// Config sets a PublicKeyCallback handler that forwards public key auth
+// requests to the route named "pubkeyAuth".
+//
+// This assumes certain details about our environment, like the location of the
+// host keys. It also provides only key-based authentication.
+// ConfigureServerSshConfig
+//
+// Params:
+//
+// Returns:
```
mboersma: Comment Params: and Returns: should be fleshed out or trimmed down.

builder/sshd/sshd.go
```diff
@@ -0,0 +1,235 @@
...
+//  An *ssh.ServerConfig
+func Configure(c cookoo.Context, p *cookoo.Params) (interface{}, cookoo.Interrupt) {
+	router := c.Get("cookoo.Router", nil).(*cookoo.Router)
+
+	//hostKeyTypes := []string{"rsa", "dsa", "ecdsa"}
```
mboersma: More commented-out code to be tossed.

builder/tests/builder_test.go
```diff
@@ -70,14 +82,62 @@ func TestBuilder(t *testing.T) {
...
 			imageName)
 	}()
-	dockercli.PrintToStdout(t, stdout, stdoutPipe, "deis-builder running")
+	dockercli.PrintToStdout(t, stdout, stdoutPipe, "Builder is running")
 	if err != nil {
 		t.Fatal(err)
 	}
-	// FIXME: builder needs a few seconds to wake up here!
-	// FIXME: Wait until etcd keys are published
 	time.Sleep(5000 * time.Millisecond)
```
mboersma: We didn't get rid of the `Sleep`, so probably we should keep the old comments about why we're doing it.

deisctl/backend/fleet/stop.go
```diff
@@ -94,6 +94,7 @@ func doStop(c *FleetClient, target string, wg *sync.WaitGroup, out, ew io.Writer
 			fmt.Fprintf(ew, o, target)
 			return
 		}
+		fmt.Println(lastSubState)
```
mboersma: Does this fix something not specific to the PR? I thought `deisctl stop` worked fine...

mboersma:
I couldn't find much more than ticky-tack changes when reviewing this. It would be nice to clean them up, but if there aren't more substantial changes to make and other maintainers are happy with it, we could merge this and I'd be happy to follow on with a cleanup PR.

bacongobbler:
Yep! For the most part this PR looks good to :ship:.

builder/Makefile
technosophos: fixed

builder/README.md
technosophos: fixed

builder/commands_test.go
technosophos: Done. I'm not a huge stickler for it, but I think there is good reason to prefer scoping vars to a smaller block. So, I think it makes (performance) sense to do this where possible. It's not just stylistic.

builder/commands.go
technosophos: Done

builder/docker/doc.go
technosophos: I saw that in one of the core Go packages, and I like it. But I usually forget to do it.

builder/docker/docker_test.go
technosophos: MY TESTS ARE GONE! They must have been a victim of my bad rebase the other day. Hmmm... I'll fix that, or delete the file if there are no tests.

builder/etcd/etcd.go
technosophos: This one is actually not an error. I return an error on purpose, because it may not be `nil`. But I guess I could change line 161 to just return the error there. That's probably easier to read.

builder/etcd/etcd.go
technosophos: That's a good question. Since there's no code 0 constant, is it more readable to use the long constant names and leave 0, or keep them all ints? I prefer for all of the cases to look the same because it requires less thinking.

builder/tests/builder_test.go
technosophos: The comments are outdated. Code was added earlier to handle those things. So there's nothing left to fix. Or at least, that's how I read the comments.

deisctl/backend/fleet/stop.go
technosophos: Oooops...

technosophos:
Alright, I think I got just about everything. I didn't make a huge effort clean up whitespace things as long as gofmt passed, but I did clean them up as I went. As soon as the next Jenkins tests pass...

mboersma:
LGTM. Nice work @technosophos!

bacongobbler:
code LGTM!

krancour:
I haven't had a chance to play with this yet, but @technosophos I just want to tell you how excited I am about a new builder implementation!  Nice work!

kalbasit:
:clap: :clap: very nice work @technosophos

PR #4276, opened 2015-08-14
fix(deisctl): clarify registration message

### LINKED ISSUE #4075: `deis register` -> `deis auth:register` indirection/improvement
After installing and starting the deis platform via deisctl, the deisctl tells you:

```Please use `deis register` to setup an administrator account.```

However when you type:

```deis register```

you get:

```Usage: deis auth:register <controller> [options]```

As a new user of Deis, I'd find this confusing as my first assumption is that `deis register` doesn't exist, but it has moved to `deis auth:register`. Instead, what is actually happening is I just didn't pass in a controller.

A few options:
- improve the output of deisctl after the start of the platform to tell the user to now use the deis cli to register an account (with the correct instructions, maybe even putting in the controller IP/DNS entry for them)
- be more vague on how you register a user via the deis cli from deisctl
- change the usage output of `deis register` to show as `deis register` not `deis auth:register`
- give the user of the deis cli a bit more help on what option/argument they missed (in my case: controller)

Joshua-Anderson:
Fixes #4075

### DIFF
--- a/deisctl/cmd/platform.go
+++ b/deisctl/cmd/platform.go
@@ -54,7 +54,7 @@ func StartPlatform(b backend.Backend, stateless bool) error {
 	wg.Wait()
 
 	fmt.Fprintln(Stdout, "Done.\n ")
-	fmt.Fprintln(Stdout, "Please use `deis register` to setup an administrator account.")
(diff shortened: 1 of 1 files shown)


bacongobbler:
lgtm

krancour:
LGTM

PR #4283, opened 2015-08-17
docs(installing_deis): specify CoreOS for bare metal

### LINKED ISSUE #4282: documentation does not mention incompatibility between deis and coreos shipping buggy docker versions
According the documentation on deis baremetal install (http://docs.deis.io/en/latest/installing_deis/baremetal/ ), "Deis runs on CoreOS version 494.5.0 or later in the Stable channel."

It seems some buggy docker version shipped with coreos stable channel (docker 1.6.2 and 1.7.0 have buggy pull command according to joshua anderson on IRC), cannot run deis.
i've installed the latest stable coreos channel (723.3.0), which ships the potentially buggy docker 1.6.2.
So, the documentation, should not infer that later version of 494.5 coreos releases, can run deis 1.9.
Actually, only the alpha channel ships docker 1.7.1.

best regards, 
Charles.

carmstrong:
closes #4282

### DIFF
--- a/docs/installing_deis/baremetal.rst
+++ b/docs/installing_deis/baremetal.rst
@@ -100,12 +100,12 @@ Start the installation
 
 .. code-block:: console
 
-    coreos-install -C stable -c /tmp/config -d /dev/sda
+    coreos-install -C stable -c /tmp/config -d /dev/sda -V 647.2.0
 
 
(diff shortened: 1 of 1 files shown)


krancour:
Doc LGTM

technosophos:
Code LGTM.

PR #4285, opened 2015-08-17
docs(isolating-etcd): add link to user-data.example

laurrentt:
This confused me at first since I have been using my own cloud-config for a while.

### DIFF
--- a/docs/managing_deis/isolating-etcd.rst
+++ b/docs/managing_deis/isolating-etcd.rst
@@ -18,8 +18,10 @@ cloud-config
 ------------
 
 To realize the topology described above, it is necessary, at the time of
-provisioning, to provide different cloud-config for those hosts that will run
+provisioning, to provide different `cloud-config`_ for those hosts that will run
 etcd and for those that will only run an etcd proxy.
+
+.. _`cloud-config`: ../../contrib/coreos/user-data.example
 
 For the small, fixed number of hosts running full etcd and satisfying the
 "central services" role (as described in the CoreOS documentation), the


carmstrong:
Surprised this worked - our convention is usually to wrap links in backticks. This LGTM, although I'm wondering if we should change this to match our convention.

laurrentt:
@carmstrong added the backticks to match your convention

krancour:
@laurrentt the spot where you put the link is the second reference to cloud-config in that section.  And it's in a spot where we're advocating customizing the cloud-config for one's own purposes.

It might help avoid some confusion if the link were moved up to the _previous_ mention of cloud-config where we mention that out-of-the-box cloud-config is adequate.  This way when we say something's adequate, we link them to it, and when we say something requires customization, we let them figure that out.

Does this make sense?

laurrentt:
@krancour I think you're right. Also, I searched a little and found Wikipedia's guidelines about [linking](https://en.wikipedia.org/wiki/Wikipedia:Manual_of_Style/Linking#Principles): 
> "[...] as a rule of thumb editors should only link the term's **first occurrence** in the text of the article."

So, yeah, I'll move the link to the previous mention.

carmstrong:
LGTM.

carmstrong:
Jenkins passed docs tests.

mboersma:
> our convention is usually to wrap links in backticks.

If there isn't a space in the link name, ReST doesn't care about backticks.

LGTM.

carmstrong:
> If there isn't a space in the link name, ReST doesn't care about backticks.

Ahhhh that makes sense

PR #4289, opened 2015-08-18
docs(installing_deis): clean up old references

carmstrong:
Saying that Deis runs on CoreOS 494.5.0 is a bit misleading,
because a lot of recent improvements require newer versions.

Additionally, we remove the reference for bare metal to
contrib/bare-metal since there are no scripts there.


technosophos:
LGTM

aledbf:
LGTM

carmstrong:
Passed docs tests. Merging...

PR #4290, opened 2015-08-18
feat(router): improve nginx status page

aledbf:
Replace the default status page with one that provides more details using the [nginx-module-vts](https://github.com/vozlt/nginx-module-vts) module.
By default `/status` returns html but is possible to get a json representation using the `/status/format/json` path

status page

### DIFF
--- a/router/build.sh
+++ b/router/build.sh
@@ -63,6 +63,10 @@ get_src 501f299abdb81b992a980bda182e5de5a4b2b3e275fbf72ee34dd7ae84c4b679 \
 get_src 8d280fc083420afb41dbe10df9a8ceec98f1d391bd2caa42ebae67d5bc9295d8 \
         "https://github.com/openresty/set-misc-nginx-module/archive/v$VERSION_SETMISC.tar.gz"
 
+# replace with a release instead a commit once e06a618e5780b4b94b21cab37d61820dcd3bb585 is contained in one.
+get_src 47340cf0c711b10a0231fca9264d73ea791ff8420b823b0295ae3df2d968a205 \
+        "https://github.com/vozlt/nginx-module-vts/archive/22c51e201a550bb94e96239fef541347beb4eeca.tar.gz"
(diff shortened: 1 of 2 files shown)


aledbf:
This is a screenshot from one of the nginx `deis-router` running the CI job.

captura de pantalla 2015-08-18 a las 11 12 27

jchauncey:
@aledbf i could hug you right now. As part of my metrics spike I was actually parsing the nginx log stream to get these stats which is a terrible and brittle way to get that information. With this I can actually hit `/status/format/json` and get that information and it looks like more. 

:+1:

router/build.sh
```diff
@@ -63,6 +63,9 @@ get_src 501f299abdb81b992a980bda182e5de5a4b2b3e275fbf72ee34dd7ae84c4b679 \
 get_src 8d280fc083420afb41dbe10df9a8ceec98f1d391bd2caa42ebae67d5bc9295d8 \
         "https://github.com/openresty/set-misc-nginx-module/archive/v$VERSION_SETMISC.tar.gz"
 
+get_src 47340cf0c711b10a0231fca9264d73ea791ff8420b823b0295ae3df2d968a205 \
+        "https://github.com/vozlt/nginx-module-vts/archive/22c51e201a550bb94e96239fef541347beb4eeca.tar.gz"
```
carmstrong: Is there a fix we need here that isn't in a specific release? We generally prefer that to SHAs.

aledbf: @carmstrong the only reason to use the commit and not a release is because in the last ones there's a fix for the counters and layout of the html

carmstrong: :+1: I think that's fine, then. Could we add a comment with that info above that line?

technosophos:
Code LGTM.

And FWIW, @jchauncey has *never* given me a hug before.

bacongobbler:
code LGTM.

PR #4295, opened 2015-08-18
feature(builder): remove shim Dockerfile

technosophos:
Instead of a shim Dockerfile, use ONBUILD directives in the slugrunner Dockerfile. This should not break backward compatibility.

A new Dockerfile is still generated, but it is now done dynamically, and has only FROM and ENV GIT_COMMIT set.


carmstrong:
Protip for reviewers - review/merge #4273 first and then this one will be cake.

technosophos:
Fixed both of @bacongobbler 's reported issues.

bacongobbler:
Code LGTM!

technosophos:
Superseded by #4324.

PR #4303, opened 2015-08-19
feat(Registry): add support for IAM role authentication

laurrentt:
As described in [AWS best practises](http://docs.aws.amazon.com/IAM/latest/UserGuide/IAMBestPractices.html#use-roles-with-ec2), application running on EC2 should not use hard coded Access Key and Secret Key and instead use  attached IAM roles to authenticate to Amazon web services. 

Here's some fixes I had to made for boto to connect using IAM roles. I've tried to be as backward compatible as possible but I did not test other workflows (help needed :pray:). 

I've also updated the *customizing deis -> registry settings* doc to include the keys that were added with the *Running Deis without Ceph* feature and the new `s3assumeRole` key.

### DIFF
--- a/registry/templates/create_bucket
+++ b/registry/templates/create_bucket
@@ -5,13 +5,22 @@ import sys
 from boto.s3.connection import OrdinaryCallingFormat
 
 conn = boto.connect_s3(
+    {{ if exists "/deis/store/gateway/accessKey" }}
     aws_access_key_id='{{ getv "/deis/store/gateway/accessKey" }}',
     aws_secret_access_key='{{ getv "/deis/store/gateway/secretKey" }}',
+    {{end}}
+    {{ if exists "/deis/store/gateway/host" }}
     host='{{ getv "/deis/store/gateway/host" }}',
     port={{ getv "/deis/store/gateway/port" }},
+    {{ end }}
     is_secure=False,
     calling_format=OrdinaryCallingFormat())
+
+{{ if exists "/deis/registry/s3bucket" }}
+name = '{{ getv "/deis/registry/s3bucket" }}'
+{{ else }}
 name = '{{ getv "/deis/registry/bucketName" }}'
+{{ end }}
 
 if name not in (bucket.name for bucket in conn.get_all_buckets()):
     conn.create_bucket(name)
--- a/docs/customizing_deis/registry_settings.rst
+++ b/docs/customizing_deis/registry_settings.rst
@@ -43,6 +43,21 @@ setting                                   description
 /deis/store/gateway/secretKey             S3 API secret key used to access store-gateway (set by store-gateway)
 ====================================      =================================================================================
 
+If the ``/deis/registry/s3bucket`` key is supplied, the registry
+will use Amazon S3 as its storage backend and use the following values.
+
+====================================      =================================================================================
+setting                                   description
+====================================      =================================================================================
+/deis/registry/s3accessKey                S3 API access key. If not specified, the registry will get it from the instance role
+/deis/registry/s3secretKey                S3 API secret key, required if s3accessKey is specified
+/deis/registry/s3region                   S3 region to connect to, will use boto default if not specified
+/deis/registry/s3bucket                   S3 bucket to store images 
+/deis/registry/s3path                     path in the bucket (default: "/registry")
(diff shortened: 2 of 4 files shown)


carmstrong:
This looks like a great addition, @laurrentt!

sstarcher:
Is an S3Bucket of "" a viable bucket?  If not why is that the default value for S3Bucket and if not why have assumeRole=true instead of assuming if S3Bucket is set to a value you want to use S3 and fall back to IAM Instance access if access/secret key pair is not yet?

laurrentt:
@sstarcher I thought at first that if you didn't specify the S3bucket key a default one was created for you. Since it's not the case, I also think your solution would be better. I'll make the change ASAP.

laurrentt:
@sstarcher Done, let me know what you think.

laurrentt:
I've added a default S3 policy to kickstart people willing to move from Ceph to S3.

sstarcher:
LGTM

carmstrong:
Is there a reason we're using `~` in some places as the default nil value, and `""` in others?

laurrentt:
@carmstrong The difference is that `~` will result in a `None` value in python and `""` will result in an empty string. The docker registry ([here](https://github.com/docker/docker-registry/blob/master/docker_registry/drivers/s3.py#L110)) will try to connect using access key even if the value is empty which is not the case if the value is `None`. I could change almost change every `""` to `~` but I didn't want to fix what's not broken.

mboersma:
This is falling victim in CI to a concurrency issue pointed out by Go 1.5. Once we merge #4335, that problem should go away after a rebase from master.

laurrentt:
@mboersma Done

aj-may:
LGTM

registry/templates/create_bucket
```diff
@@ -5,13 +5,22 @@ import sys
...
+    {{end}}
+    {{ if exists "/deis/store/gateway/host" }}
     host='{{ getv "/deis/store/gateway/host" }}',
     port={{ getv "/deis/store/gateway/port" }},
+    {{ end }}
     is_secure=False,
     calling_format=OrdinaryCallingFormat())
+
+{{ if exists "/deis/registry/s3bucket" }}
+name = '{{ getv "/deis/registry/s3bucket" }}'
```
bacongobbler: can't we keep using `/deis/registry/bucketName` here?

laurrentt: @bacongobbler `s3bucket` is the key used to select the S3 config flavour. Also it's not a new key since it existed prior this PR.

bacongobbler: I think this is what was confusing me:

https://github.com/laurrentt/deis/blob/registry-s3-assume-role/registry/templates/config.yml#L83
https://github.com/laurrentt/deis/blob/registry-s3-assume-role/registry/templates/config.yml#L102

I'm fine with this since it's old behaviour, but it does look like we're using 2 different keys for (generally) the same thing.

laurrentt: @bacongobbler I completely agree, this confused me a lot at first. However changing this would be a braking change.

laurrentt:
@carmstrong, @bacongobbler Any news on this PR ? It almost been a month since it was opened...

sstarcher:
Was this ever merged?

laurrentt:
@sstarcher No but I just noticed every PR was just closed so I think there's something wrong happening right now :open_mouth:

kalbasit:
Accidentally closed the PR.

kalbasit:
@sstarcher @laurrentt my mistake apologies for the inconvenience.

laurrentt:
@kalbasit No problem ! This was so sudden :laughing:, did you mess with the github API or something ?

kalbasit:
No, one of my git aliases thought it's very funny to delete the master branch rather than the branch `issue_4477` which was not supposed to be pushed here anyway :( long day I guess.

laurrentt:
@kalbasit Well no harm, no foul

mboersma:
Jenkins, test this please.

mboersma:
This needs a rebase now. Sorry @laurrentt, we really haven't forgotten this PR. :snail:

laurrentt:
@mboersma Done, and I've included the changes made in 81a8e1f134d8523689153b33329062985e9e5827.

PR #4304, opened 2015-08-20
ref(Godeps): purge unused go dependencies

mboersma:
This is basically the same approach as in #4164: `godep save ./...`, but it preserves the `docker/docker/pkg/chrootarchive` package which is referenced in other Docker packages but was missed by `godep`. Could be a Mac OS X vs. Linux thing.

Update: I recreated this on a Linux box, which is the correct build target, and a couple dependencies that were missed on Mac OS X were now retained, so I think this will pass. We should probably only modify `Godeps.json` when actually on Linux.

Closes #4164.


bacongobbler:
lgtm

carmstrong:
:ship:

PR #4306, opened 2015-08-20
chore(router): update nginx to 1.9.4

mboersma:
See http://nginx.org/en/CHANGES.

### DIFF
--- a/router/rootfs/bin/build
+++ b/router/rootfs/bin/build
@@ -2,7 +2,7 @@
 
 set -eof pipefail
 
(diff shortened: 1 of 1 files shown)


carmstrong:
Code LGTM. This looks potentially valuable:
```
    *) Bugfix: old worker processes might hog CPU after reconfiguration with
       a large number of worker_connections.
```

bacongobbler:
that does indeed, though I wonder what "large" is defined here :)

code LGTM!

Issue #4292, opened 2015-08-18
Proposal: Upgrade logspout


krancour:
Logspout has undergone two major refactors in the time since we forked and enhanced it (mainly for the sake of discovering `deis-logger` host and port from etcd and to hijack the message format).

The master branch of logspout (which is trending toward a v3) is significantly more modular than what we've been using.  The possibility exists that we can bend logspout to our will by creating a module for it, rather than by forking it and hacking on it.

The major motivation for this would be the ability to reap the benefits of upstream bug fixes.

Do other maintainers want to weigh in?

bacongobbler:
If there's a way we can extend logspout without having to maintain the source (keeping the source code upstream), that would be a nice enhancement as well. Otherwise I'm all for a version bump!

carmstrong:
:+1:

krancour:
@bacongobbler @carmstrong 

Trivial example of customizing logspout without forking:

https://github.com/gliderlabs/logspout/tree/master/custom

The major question will be whether the extension points logspout opens are sufficient hooks for us to do all the customizations we need.

krancour:
I took a first swing at such a module here, but it's incomplete, and so far, it crashes my cluster every time I try it.  :astonished:

carmstrong:
Have we reevaluated the need for us to maintain a fork of logspout? Could we ever switch back to using mainline?

kalbasit:
While we're on the subject, I'd love to not change the message format when the message is sent to syslog-compatible (paperclip for instance). It should be changed for the sake of deis logs but I feel it should do so only after it has sent the raw log line to syslog.

krancour:
@carmstrong the very idea here is that we could use mainline logspout and just maintain modules for whatever functionality we need to add... maybe.  Doing so may be easier said then done, but it's the hope.

@kalbasit re: message format... if one is running cephless/stateless and is therefore not running `deis-logger`, or has opted out of `deis-logger` for any other reason, I think we should make it easier to opt-into a syslog-compatible message format.  That being said, if `deis-logger` is the destination for the logs, then the responsibility of emitting syslog-compatible messages, IMO, belongs to `deis-logger`'s drain.

krancour:
See https://github.com/deis/logspout-v3-deis for a possible implementation of a Deis adapter that would allow us to use a newer upstream logspout without forking it.

Also see related upstream logspout issue: https://github.com/gliderlabs/logspout/issues/124

carmstrong:
[...]

PR #4307, opened 2015-08-20
feat(logspout): support sending log via tcp

laurrentt:
I'm using AWS' Elastic Load Balancer in front on my logstash server, however ELBs don't support UDP. So here's a quick fix to add TCP support.

### DIFF
--- a/logspout/logspout.go
+++ b/logspout/logspout.go
@@ -64,14 +64,26 @@ func syslogStreamer(target Target, types []string, logstream chan *Log) {
 			continue
 		}
 		tag, pid := getLogName(logline.Name)
-		addr, err := net.ResolveUDPAddr("udp", target.Addr)
-		assert(err, "syslog")
-		conn, err := net.DialUDP("udp", nil, addr)
-		assert(err, "syslog")
-		// bump up the packet size for large log lines
-		assert(conn.SetWriteBuffer(1048576), "syslog")
+		var conn net.Conn
+		if strings.EqualFold(target.Protocol, "tcp") {
+			addr, err := net.ResolveTCPAddr("tcp", target.Addr)
+			assert(err, "syslog")
+			tcpconn, err := net.DialTCP("tcp", nil, addr)
+			assert(err, "syslog")
+			assert(tcpconn.SetWriteBuffer(1048576), "syslog")
+			conn = tcpconn
+		} else if strings.EqualFold(target.Protocol, "udp") {
+			addr, err := net.ResolveUDPAddr("udp", target.Addr)
+			assert(err, "syslog")
+			udpconn, err := net.DialUDP("udp", nil, addr)
+			assert(err, "syslog")
+			assert(udpconn.SetWriteBuffer(1048576), "syslog")
+			conn = udpconn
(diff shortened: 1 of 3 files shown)


mboersma:
ping @kalbasit, since you asked to be notified about Go code to review. :smile:

laurrentt:
I'm a Go newbie so all comments and critics are welcome.

logspout/logspout.go
```diff
@@ -161,6 +173,14 @@ func httpStreamer(w http.ResponseWriter, req *http.Request, logstream chan *Log,
 	}
 }
 
+func getEtcdValueOrDefault(c *etcd.Client, key string, defaultValue string) string {
+	resp, err := c.Get(key, false, false)
+	if err != nil {
```
bacongobbler: We might want to check the error here, as other errors like if the server are unavailable can be globbed here,

logspout/logspout.go
```diff
@@ -183,9 +203,10 @@ func main() {
 		assert(err, "url")
 		portResp, err := etcd.Get("/deis/logs/port", false, false)
 		assert(err, "url")
+		protocol := getEtcdValueOrDefault(etcd, "/deis/logs/protocol", "upd")
```
bacongobbler: typo: udp

logspout/logspout.go
```diff
@@ -161,6 +173,17 @@ func httpStreamer(w http.ResponseWriter, req *http.Request, logstream chan *Log,
 	}
 }
 
+func getEtcdValueOrDefault(c *etcd.Client, key string, defaultValue string) string {
+	resp, err := c.Get(key, false, false)
+	if err != nil {
+		if strings.Contains(fmt.Sprintf("%v", err), "Key not found") {
```
laurrentt: @bacongobbler: This works, let me know if it's good enough.

bacongobbler: +1, exactly what I was looking for

krancour:
I haven't looked at this too closely yet, but am curious what the effort would be to use TCP over TLS so logs can (optionally) be sent securely.

krancour:
^ Not to imply that we couldn't accept without that.

laurrentt:
@krancour in my case I only need unsecure tcp since none of this traffic is going on the internet. 

I checked and the interface for using TLS is basically the same but it would imply adding certificate configuration, etc. 

I suggest a "Help Wanted" issue could be opened for this.

krancour:
@laurrentt right... when/if this is accepted, we can open or seek out another PR to add TLS.

krancour:
Related #4292 <-- may already support a variety of transport options

laurrentt:
@krancour Yes that would definitely be a good idea. I could start using the third-party module logspout-redis which would buffer my logs in case my logstash is swamped or down.

carmstrong:
@bacongobbler @krancour Where are we with this? Are we okay with this as-is?

bacongobbler:
works for me until we migrate to a newer version of logspout. Code LGTM.

kalbasit:
LGTM for the Go code.

PR #4309, opened 2015-08-21
chore(tests): use go 1.5 toolchain

### LINKED ISSUE #4308: Switch to go 1.5 for product build / test
Improvements in the new Go compiler toolchain suggest significant performance benefits. However, Deis' integration tests use stdout naively and will break if `GOMAXPROCS > 1`. So for the time being we could upgrade CI nodes to Go 1.5 but add `GOMAXPROCS=1` to the CI test jobs.

mboersma:
Since some integration tests will fail with concurrent goroutines, this change also sets GOMAXPROCS=1 when running them. Big improvements to runtime GC should speed up our components somewhat.

Closes #4308.

Note that this will not automagically put Go 1.5 on all CI test boxen; we need to do that manually once we verify this doesn't break things. I've run it locally against vagrant using Go 1.5 without problems.

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -1,6 +1,6 @@
 {
 	"ImportPath": "github.com/deis/deis",
-	"GoVersion": "go1.4.2",
+	"GoVersion": "go1.5",
 	"Packages": [
 		"./..."
 	],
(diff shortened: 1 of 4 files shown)


sgoings:
:+1: Let me know when you want to roll 1.5 out to the slaves and I can help.

bacongobbler:
code LGTM!

carmstrong:
Code LGTM.

PR #4315, opened 2015-08-22
ref(contrib): use Terraform for DO provisioning

carmstrong:
This lets us use Terraform to provision and manage Deis deployments
on Digital Ocean.

~~This is a WIP for two reasons:~~
~~* I haven't yet updated the docs~~
~~* There's seemingly no easy way to get a user's SSH key ID without hitting the API ourselves (hashicorp/terraform#3058)~~

For users who want to try this out, install Terraform and then:
```
$ make discovery-url
$ terraform apply contrib/digitalocean -var 'token=asdf1234' -var 'ssh_keys=1234'
```

The variables are a user's DigitalOcean API token and the ID of their SSH key.


contrib/digitalocean/deis-digitalocean.tf
```diff
@@ -0,0 +1,20 @@
+provider "digitalocean" {
+    token = "${var.token}"
+}
+
+resource "digitalocean_droplet" "deis" {
+    count = "${var.instances}"
+    image = "coreos-stable"
+    name = "deis-${count.index+1}"
```
nathansamson: previously we had an option to pass a prefix to the script. This feature seems to be gone.

Would it be worth to restore this?

nathansamson:
Note that running this seems to work flawlessly (at least for setting up the machines, and having them fleet enabled).

I can also override the region with -var region= (as "advertised").

Setting up is also a lot faster than the old method (as this is more parrellized). Which I think might be a clue with etcd is suddenly working....

Is there a way to silence output of terraform (in the script), having the cloud config printed in unreadble format 3 times isn;t too usefull...

carmstrong:
Thanks for testing, @nathansamson! I agree that it's much faster. How painful was it to get your SSH key id? That's the big blocker here, I think.

> Is there a way to silence output of terraform

I wondered the same thing. I couldn't figure it out, but let me know if you do!

nathansamson:
Quite easy to get my keys. Just a docl Keys away. But Yeah i am the author
of that tool. So i might be slightly biased.

I agree that just installing docl to get your keys is bit cumbersome, but
not sure if terraform can solve that (see comment on that bug report)
Op 22 aug. 2015 20:28 schreef "Chris Armstrong" <notifications@github.com>:

carmstrong:
> So i might be slightly biased.

A bit :) I think docl is great. I think not having people need to query for their keys at all is even better. Still thinking about a solution for this, though. There may not be one :(

carmstrong:
@nathansamson This now supports specifying a prefix, which defaults to "deis".

nathansamson:
You can also use the fingerprint instead of the DO ID. The fingerprint can be viewed on the DO dashboard or locally computed (ssh-keygen -l -E md5 -f ~/.ssh/deis.pub | awk '{print $2}')

~~~
terraform apply -var 'token=obviouslyasecret' -var 'ssh_keys=6b:6e:ac:a8:not:so:secret:but:will:not:post:it:anyway' -var 'prefix=deis-terra'
~~~

worked like a charm

carmstrong:
> The fingerprint can be viewed on the DO dashboard

Aha! That's much more reasonable. I'll whip up some updated documentation.

carmstrong:
This is now ready for review.

carmstrong:
Killing Jenkins run as this won't be tested.

mboersma:
I tried this using the included Sphinx docs but had problems:
```console
$ brew install terraform
$ terraform --version
Terraform v0.6.3

$ make discovery-url
$ terraform apply contrib/digitalocean \
>  -var 'token=b4687fxxxxxx' \
>  -var 'ssh_keys=e2:3f:df:xx:xx:xx' \
>  -var 'region=nyc3' \
>  -var 'prefix=mboersma-deis' \
>  -var 'instances=5' \
>  -var 'size=4GB'
The apply command expects at most one argument.
Usage: terraform apply [options] [DIR]

  Builds or changes infrastructure according to Terraform configuration
  files in DIR.
...

$ terraform apply \
>  -var 'token=b4687fxxxxxx' \
>  -var 'ssh_keys=e2:3f:df:xx:xx:xx' \
>  -var 'region=nyc3' \
>  -var 'prefix=mboersma-deis' \
>  -var 'instances=5' \
>  -var 'size=4GB' \
> contrib/digitalocean
There are warnings and/or errors related to your configuration. Please
fix these before continuing.

Errors:

  * 1 error(s) occurred:

* 5 error(s) occurred:

* file: open /Users/matt/Projects/src/github.com/deis/deis/../coreos/user-data: no such file or directory in:

${file("${path.cwd}/../coreos/user-data")}
* file: open /Users/matt/Projects/src/github.com/deis/deis/../coreos/user-data: no such file or directory in:

...
```

carmstrong:
Ah - looks like the path for user-data I'm using is using the cwd, not the directory the script is in.

I bet if you first changed into `contrib/digitalocean` and then ran `terraform apply` with the vars it'd work.

mboersma:
> I bet if you first changed into `contrib/digitalocean`...

That does seem to work, and then I get the ridiculous amount of textual information you mentioned. (Mine ended with a "401 Unauthorized" error for some reason but I don't think that's the PR's fault necessarily.) 
```console
$ cd contrib/digitalocean/
Rhialto:digitalocean matt$ terraform apply \
>  -var 'token=b4687fxxxxxx' \
>  -var 'ssh_keys=e2:3f:df:xx:xx:xx' \
>  -var 'region=nyc3' \
>  -var 'prefix=mboersma-deis' \
>  -var 'instances=5' \
>  -var 'size=4GB'
digitalocean_droplet.deis.4: Creating...
  backups:              "" => "0"
  image:                "" => "coreos-stable"
  ipv4_address:         "" => "<computed>"
  ipv4_address_private: "" => "<computed>"
  ipv6:                 "" => "0"
  ipv6_address:         "" => "<computed>"
  ipv6_address_private: "" => "<computed>"
  locked:               "" => "<computed>"
  name:                 "" => "mboersma-deis-5"
  private_networking:   "" => "1"
  region:               "" => "nyc3"
  size:                 "" => "4GB"
  ssh_keys.#:           "" => "1"
  ssh_keys.0:           "" => "e2:3f:df:xx:xx:xx"
  status:               "" => "<computed>"
  user_data:            "" => "#cloud-config\n---\ncoreos:\n  fleet:\n  
...

Error applying plan:

5 error(s) occurred:

* digitalocean_droplet.deis.1: Error creating droplet: Error creating droplet: API Error: 401 Unauthorized
* digitalocean_droplet.deis.3: Error creating droplet: Error creating droplet: API Error: 401 Unauthorized
* digitalocean_droplet.deis.2: Error creating droplet: Error creating droplet: API Error: 401 Unauthorized
* digitalocean_droplet.deis.4: Error creating droplet: Error creating droplet: API Error: 401 Unauthorized
* digitalocean_droplet.deis.0: Error creating droplet: Error creating droplet: API Error: 401 Unauthorized

Terraform does not automatically rollback in the face of errors.
Instead, your Terraform state file has been partially updated with
any resources that successfully completed. Please address the error
above and apply again to incrementally change your infrastructure.
```

carmstrong:
Reminder: we need to switch `path.cwd` for something else (maybe `path.root` if it exists?) so that we can use the script root to find user-data.

/cc @sgoings

sgoings:
@carmstrong and @mboersma I've added the files that rigger cares about in the DO script (and handled the path.cwd -> path.module change as requested).

It's possible to use this separate from Rigger as well (but just requires manual setup of Deis after a `terraform apply`)

carmstrong:
Should the `create`, `destroy`, etc. scripts not have the `.sh` extension? Rigger could just see them as executables.

sgoings:
@carmstrong `rigger` actually uses them more like included libraries (they're eval'd). This gives us the ability to call functions that live in rigger itself (like the critical `save-var`).

carmstrong:
> rigger actually uses them more like included libraries (they're eval'd).

Understood. One downside of that is that an internal rewrite (say, to reimplement Rigger in Go), would require changes in the tool *and* all provision scripts. 

We may not care, but it's something to consider.

sgoings:
Agreed. I think for now it leads to the most minimal scripting per provider. Switching them to proper "call-able binaries" is ideal, but I'd have to make some of the functions that provider scripts would need to call also call-able binaries (save-var, logging, etc). This wouldn't be hard, but just more work :-)

carmstrong:
> This wouldn't be hard, but just more work :-)

Let's file an issue for it on deis/rigger

kalbasit:
Accidentally closed the PR.

contrib/digitalocean/README.md
```diff
@@ -1,3 +1,23 @@
...
+    ./rigger destroy
+      # when done with the cluster
+
+Using Terraform:
+
+    terraform apply
+      # will ask you some questions to fill in variables
+      # use your new Deis cluster when this command completes!
+
+Then follow the instructions on [provisioning Deis on Digital Ocean manually](http://docs.deis.io/en/latest/installing_deis/digitalocean/)
```
mboersma: Why wouldn't we include the `rigger` commands in the DigitalOcean instructions at docs.deis.io? 

This README.md is intentionally just a pointer to that, so that there is one source of truth, and also so content isn't split between markdown and restructuredtext.

sgoings: All for it. I wrote this README months back when `rigger` wasn't as blessed as a thing yet.

mboersma:
Following the docs included with this PR doesn't seem to work still:
```console
$ terraform --version
Terraform v0.6.3
$ terraform apply contrib/digitalocean \
>   -var 'token=2acbe548' \
>   -var 'ssh_keys=e1:c3:b0:7f:91:d0' \
>   -var 'region=nyc3' \
>   -var 'prefix=mb-deis' \
>   -var 'instances=3' \
>   -var 'size=4GB'
The apply command expects at most one argument.
Usage: terraform apply [options] [DIR]
```
I think you still need to put `contrib/digitalocean` *after* the dash-prefixed arguments:
```console
$ terraform apply \
>   -var 'token=2acbe548' \
>   -var 'ssh_keys=e1:c3:b0:7f:91:d0' \
>   -var 'region=nyc3' \
>   -var 'prefix=mb-deis' \
>   -var 'instances=3' \
>   -var 'size=4GB' \
>   contrib/digitalocean
digitalocean_droplet.deis.1: Creating...
  backups:              "" => "0"
  image:                "" => "coreos-stable"
  ipv4_address:         "" => "<computed>"
  ipv4_address_private: "" => "<computed>"
  ipv6:                 "" => "0"
  ...
```

docs/installing_deis/digitalocean.rst
```diff
@@ -47,42 +47,29 @@ Generate a New Discovery URL
...
 
-If private networking is not available in your region, swap out ``$private_ipv4`` with
-``$public_ipv4`` in the user-data file.
+    $ terraform apply -var 'token=a1b2c3d3e4f5' \
+                      -var 'ssh_keys=c1:d3:a2:b4:e4:f5' \
+                      -var 'region=nyc3' \
+                      -var 'prefix=deis' \
+                      -var 'instances=3' \
+                      -var 'size=8GB' \
+                      contrib/digitalocean
```
sgoings: This has been updated per @mboersma's findings.

carmstrong: LGTM now

PR #4330, opened 2015-08-25
fix(logger): allow underscores in container names

krancour:
Our custom `deis-logspout` does some work to resolve "application name" and "pid" from the name of each container it's attached to and includes these in the log messages it sends to `deis-logger`.  If the container's name doesn't match the usual format expected for a deis application, then the app name is simply the name of the container and the pid is simply 1.

Where this fails is if a container name has an underscore in it-- and Docker containers, if not explicitly named, use generated names that _do_ container underscores.  In such a case, the app name component of the message coming _from_ `deis-logspout` will contain an underscore, but on the `deis-logspout` side, the message gets rejected because, per a regex, it doesn't appear to contain an app name.

This change updates that regex such that `deis-logger` will recognize app names with underscores in them.

### DIFF
--- a/logger/syslogd/syslogd.go
+++ b/logger/syslogd/syslogd.go
@@ -50,7 +50,7 @@ func fileExists(path string) (bool, error) {
 }
 
 func getLogFile(message string) (io.Writer, error) {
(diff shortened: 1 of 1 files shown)


carmstrong:
Code LGTM.

mboersma:
LGTM.

PR #4331, opened 2015-08-25
docs(azure): add vnet creation instructions

mboersma:
This is required to ensure that the nodes get a static IP address.  If a virtual network is not used, the nodes get a dynamic private IP address that can change when rebooted.  This causes problems with etcd.

From @mdolian.
Closes #4329.


carmstrong:
LGTM.

bacongobbler:
docs LGTM

mboersma:
This passed doc tests, so I'm merging it.

PR #4333, opened 2015-08-25
feat(Router): IP whitelisting for apps and controller

laurrentt:
This is a new security feature for the router. This is useful if you run internal applications on Deis and want to restrict access to specific range of IPs.

### DIFF
--- a/docs/customizing_deis/router_settings.rst
+++ b/docs/customizing_deis/router_settings.rst
@@ -34,9 +34,10 @@ setting                                      description
 =======================================      ==================================================================================================================================================================================================================================================================================================================================
 /deis/builder/host                           host of the builder component (set by builder)
 /deis/builder/port                           port of the builder component (set by builder)
+/deis/config/\*/deis_whitelist               comma separated list of IPs (or CIDR) allowed to connect to the application containers (set by controller) Example: "0.0.0.0:some_optional_label,10.0.0.0/8"
 /deis/controller/host                        host of the controller component (set by controller)
 /deis/controller/port                        port of the controller component (set by controller)
-/deis/domains/*                              domain configuration for applications (set by controller)
+/deis/domains/\*                             domain configuration for applications (set by controller)
 /deis/router/affinityArg                     for requests with the indicated query string variable, hash its contents to perform session affinity (default: undefined)
 /deis/router/bodySize                        nginx body size setting (default: 1m)
 /deis/router/defaultTimeout                  default timeout value in seconds. Should be greater then the frontfacing load balancers timeout value (default: 1300)
@@ -45,7 +46,9 @@ setting                                      description
 /deis/router/controller/timeout/connect      proxy_connect_timeout for deis-controller (default: 10m)
 /deis/router/controller/timeout/read         proxy_read_timeout for deis-controller (default: 20m)
 /deis/router/controller/timeout/send         proxy_send_timeout for deis-controller (default: 20m)
+/deis/router/controller/whitelist            comma separated list of IPs (or CIDR) allowed to connect to the controller (default: not set) Example: "0.0.0.0:some_optional_label,10.0.0.0/8"
 /deis/router/enforceHTTPS                    redirect all HTTP traffic to HTTPS (default: false)
+/deis/router/enforceWhitelist                deny all connections unless specifically whitelisted (default: false)
 /deis/router/firewall/enabled                nginx naxsi firewall enabled (default: false)
 /deis/router/firewall/errorCode              nginx default firewall error code (default: 400)
 /deis/router/errorLogLevel                   nginx error_log level (default: error) Valid options: debug, info, notice, warn, error, crit, alert, emerg
--- a/docs/managing_deis/security_considerations.rst
+++ b/docs/managing_deis/security_considerations.rst
@@ -81,6 +81,24 @@ The :ref:`Router` component includes a firewall to help thwart attacks. It can b
 ``deisctl config router set firewall/enabled=true``. For more information, see the `router README`_
 and :ref:`router_settings`.
 
+IP Whitelist
+------------
+You can enforce cluster-wide IP whitelisting by running ``deisctl config router set enforceWhitelist=true``.
+Then you'll have to manually whitelist IPs to the applications using the config endpoint of the deis
+client. The format is ``{IP_or_CIDR}:{Optional_label},...``. For example:
+
+.. code-block:: console
+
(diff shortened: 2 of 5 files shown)


mboersma:
This is falling victim to a concurrency issue pointed out by Go 1.5. Once we merge #4335, that problem should go away after a rebase from master.

bacongobbler:
@laurrentt would you mind rebasing off master to grab that commit? :)

laurrentt:
@bacongobbler @mboersma Done

docs/customizing_deis/router_settings.rst
```diff
@@ -34,9 +34,10 @@ setting                                      description
 =======================================      ==================================================================================================================================================================================================================================================================================================================================
 /deis/builder/host                           host of the builder component (set by builder)
 /deis/builder/port                           port of the builder component (set by builder)
+/deis/config/\*/whitelist                    comma separated list of IPs (or CIDR) allowed to connect to the application containers (set by controller) Example: "0.0.0.0:some_optional_label,10.0.0.0/8"
```
carmstrong: Looks like a find/replace issue :)

docs/customizing_deis/router_settings.rst
```diff
@@ -34,9 +34,10 @@ setting                                      description
...
 /deis/controller/host                        host of the controller component (set by controller)
 /deis/controller/port                        port of the controller component (set by controller)
-/deis/domains/*                              domain configuration for applications (set by controller)
+/deis/domains/\*                             domain configuration for applications (set by controller)
```
carmstrong: And here

docs/customizing_deis/router_settings.rst
laurrentt: I know it's really weird but it's intended, otherwise the parser thinks it's `/*  */ ` comment. You can try to remove it but it won't compile.

docs/customizing_deis/router_settings.rst
laurrentt: @carmstrong I could remove this one but as soon as somebody adds a `foo/*/bar` it will be parsed as a comment.

docs/customizing_deis/router_settings.rst
carmstrong: Ok... so in the generated docs, it will look okay? That's fine by me.

krancour:
Code LGTM

sstarcher:
Allowing for a default whitelist for the cluster could be useful.  As if we set enforceWhitelist we would lose access to all apps until we specifically set a whitelist for every app.

laurrentt:
@sstarcher I'm not against the idea of having a default whitelist for the cluster, though maybe in another PR.

bacongobbler:
so basically `deis config:set WHITELIST="192.168.0.0/24"` will be picked up by the router as well as `deisctl config controller set whitelist="192.168.0.0/24"`? :+1: 

To be more explicit, would it be better to name it as IP_WHITELIST rather than WHITELIST?

(this is not an LGTM, just a seal of approval)

laurrentt:
@bacongobbler Since it can be one or more IP or CIDR and the user may set a label by entry, I think `IP_WHITELIST` would be a tad misleading since it's a bit more than that. Beside, one or the other, I don't care that much.

sstarcher:
LGTM this solves a pain point on our side @brendangibat should be happy

bacongobbler:
Okay just want to make sure that we don't conflict with a user's application that may rely on an envvar named WHITELIST for something (for example, whitelisting users on a Minecraft server). I think this is completely fine though.

haven't tested this manually but code LGTM.

krancour:
> Okay just want to make sure that we don't conflict with a user's application that may rely on an envvar named WHITELIST for something (for example, whitelisting users on a Minecraft server). I think this is completely fine though.

I did have the same thought.  I almost hate to say it, but the _right_ way to manage a whitelist might be through changes to the app model in the controller and corresponding changes in the client.

bacongobbler:
> but the right way to manage a whitelist might be through changes to the app model in the controller

Isn't that what's happening with the proposed model? A `deis config:set WHITELIST` is how the end-user would use this feature as described in the documentation :)

krancour:
@bacongobbler but that, as you've observed, sets an envar... I'm saying maybe whitelist needs to be an actual attribute of the app model.

laurrentt:
@krancour Well I thought so too at the beginning, and I even have a version of that somewhere. But then I saw how was implemented the HEALTHCHECK feature and I decided to use the same pattern. IMO, I think it's cleaner now. If we're afraid of envar collision we could simply prefix it with `DEIS_` or create a special prefix that we could later ignore in [_construct_env](https://github.com/deis/deis/blob/master/controller/registry/private.py#L162) so it wouldn't be propagated inside the containers.

krancour:
> But then I saw how was implemented the HEALTHCHECK feature and I decided to use the same pattern.

That's a good point.  That being said, I think it feels strange to modify the behavior of the platform (in this case, the router) via _application_ environment variables.

laurrentt:
@krancour I think there's pros and cons on both side. However I don't know if I have the motivation to start over and add it to the app model (assuming I'll have to edit the controller, the router, the Go client, the python client, create a new api version, etc...). Could we wave this one and add it to the Deis 2.0 backlog (along with the health check feature) ?

krancour:
Healthcheck is out there.  We can't very well change how that's managed without breaking backwards compatibility, so I can see giving this a pass since it's at least consistent with how healthchecks are managed.  I agree we should revisit both for the next major release.

nathansamson:
What about prefixing with DEIS_ so at least it won't collide with real app
settings?
Op 2 sep. 2015 21:42 schreef "krancour" <notifications@github.com>:

krancour:
> What about prefixing with DEIS_ so at least it won't collide with real app settings?

That seems reasonable to me.

laurrentt:
@krancour made the necessary changes in d452ebcef7f482790d1890ef419cd089891360ca let me know if that's ok, I'll squash it.

bacongobbler:
:+1:

krancour:
Squash away.

laurrentt:
@krancour squashed

mboersma:
LGTM.

sstarcher:
@bacongobbler  mentioned that the syntax was ```deisctl config controller set whitelist="192.168.0.0/24"``` for setting the controller's whitelist, but the key that is being used is actually /deis/router/controller/whitelist which would instead be  ```deisctl config router/controller set whitelist="192.168.0.0/24"```

Should the key be changed to /deis/controller/whitelist?

PR #4335, opened 2015-08-25
fix(deisctl): avoid stdout issues in tests

mboersma:
Some unit tests will fail with `GOMAXPROCS` > 1, so this change allows `make test-unit` to pass until there is a proper fix.


sgoings:
How is stdout related to GOMAXPROCS?

mboersma:
I haven't actually dug into this one, but in the other similar # TODO: cases I think we were reading stdout from a goroutine with improper channel handling or...mumble... I know empirically in these cases it's `GOMAXPROCS` and therefore a concurrency issue.

carmstrong:
Code LGTM.

technosophos:
Code LGTM

Issue #4293, opened 2015-08-18
Logspout only discovers logger at startup


krancour:
Logspout only discovers the host and port for the `deis-logger` at startup.  If `deis-logger` is rescheduled to another machine, logs are no longer aggregated.

For other components that have a dependency on another component that could move around the cluster, we rely on confd to reconfigure the dependent component.  Since our logspout is a forked and _minimally_ modified version of an older version of logspout, we're not doing any of this.

This also relates to #4292

krancour:
The proposed adapter for logspout v3 fixes this:

https://github.com/deis/logspout-v3-deis

krancour:
The reconciliation loop required to fix this seems like a bigger feature than what we currently are intending to put in the LTS.  However, this _is_ a problem.  Do any other maintainers have thoughts on this?  Ping @bacongobbler and @mboersma

krancour:
Bump @bacongobbler and @mboersma.  This bit someone else today.  Any thoughts on whether this should be fixed in LTS?

bacongobbler:
I feel like this should be fixed for the LTS unless there's a way we can restart the logspout processes through systemd whenever deis-logger is rebooted.

krancour:
Ok.  Tagged it as an LTS issue.

PR #4347, opened 2015-08-26
[WIP] chore(logspout): upgrade logspout

### LINKED ISSUE #4292: Proposal: Upgrade logspout

krancour:
Fixes #4292 #4293 

This WIP PR uses upstream logspout v3 (without forking it!) along with a deis adapter module.

When merged, this addresses many recent, notable log issues related to discovery, deadlock, Kubernetes, and more.  Some (not all) of these issues have _also_ been fixed in our ancient fork of logspout, but the point is I have attempted to keep this PR in feature / fix parity with what currently exists.

Why is this WIP?

* Release of logspout v3 is pending
* The plugin still requires unit and functional tests

~~Ideally, also, it would be good to find a way to keep the adapter code right here in the deis/deis project.  That, however, would require that we deviate from upstream logspout's own documentation on how to produce custom builds that include our own adapter.  This may be worthwhile to figure out since it will mean one less repository to manage, but my early attempts at doing this have resulted in built images that are (somehow) much, much larger than what their documented custom build process produces.  I do intend to solve that.~~


krancour:
I've discovered how to produce adequately compact images from a Dockerfile that combines stock logspout with the custom code from _this_ repo.  This eliminates the need to maintain the separate (and very short-lived) deis/logspout-v3-deis repo.  The downside to this build, however, is that it's slow, so I also provide a second, development-optimized build that produces a less compact image, but does so much more quickly.  This seems reasonable to me.

A few unit and functional test are all that stand between this and a review.

krancour:
Although unit and functional tests are still required, I'm encouraged that this passed integration tests.  :)

mboersma:
This needs a rebase now. See also #2158. :pray:

kalbasit:
Accidentally closed the PR.

krancour:
Unless other maintainers disagree, I don't see this getting done.  Logspout v3 is a blocker for this and has been a long time coming.  Meanwhile, Deis v2 won't use logspout at all.  I can't see this as being something that makes it into a v1.x LTS, not can I see it as being something that makes the leap to V2.

PR #4348, opened 2015-08-26
fix(logger): add unique id for each app name in logs

### LINKED ISSUE #4271: deis logs with kubernetes missing process number in app logs
I found this change to be a little confusing:

```
><> deis logs
[...]
2015-08-13T22:57:19UTC go[web]: 2015/08/13 22:57:19 listening on 5000...
```

With fleet, this used to be `go[web.1]` which was clearer to the user which instance was being hit from incoming requests. This is also how `heroku logs` are formatted, so it seems like this is a regression of sorts for #4222.

smothiki:
fixes #4271 
2015-08-26T23:32:15UTC wiggly-buttress[web.97nsr]: 2015/08/26 23:32:15 listening on 5000...
2015-08-26T23:37:34UTC deis[api]: sm scaled containers web=3
2015-08-26T23:37:48UTC wiggly-buttress[web.po4k7]: 2015/08/26 23:37:48 listening on 5000...
2015-08-26T23:37:49UTC wiggly-buttress[web.7wb4n]: 2015/08/26 23:37:49 listening on 5000...

container name of the application: k8s_wiggly-buttress-web.7ee905fa_wiggly-buttress-v2-web-97nsr_wiggly-buttress_a7ae5889-4c4a-11e5-b18a-06842b0bdcf9_bf8b9b46

### DIFF
--- a/logspout/logspout.go
+++ b/logspout/logspout.go
@@ -102,7 +102,7 @@ func getLogName(name string) (string, string) {
 	if match != nil {
 		return match[1], match[3]
 	}
-	match = getMatch(`^k8s_([a-z0-9-]+)-([a-z]+)\.`, name)
+	match = getMatch(`^k8s_([a-z0-9-]+)-[a-z]+\.[\da-f]+_[a-z0-9-]+-([a-z]+-[\da-z]*)_`, name)
 	if match != nil {
 		return match[1], match[2]
 	}


logspout/logspout.go
```diff
@@ -102,9 +102,11 @@ func getLogName(name string) (string, string) {
 	if match != nil {
 		return match[1], match[3]
 	}
-	match = getMatch(`^k8s_([a-z0-9-]+)-([a-z]+)\.`, name)
```
krancour: I believe if you use `^k8s_([a-z0-9-]+)-[a-z]+\.[\da-f]+_[a-z0-9-]+-([a-z]+-[\da-z]*)_` you can forgo all the additional logic you are adding in subsequent lines.  It would capture two groups (like it does already), and they will both have the right stuff in them.  You'll end up with log messages that look like the following:

```
2015-08-26T23:25:45UTC dashing-picard[web-gw7pp]: 2015/08/26 23:25:45 10.244.48.0:43701 GET /
2015-08-26T23:25:45UTC dashing-picard[web-gw7pp]: 2015/08/26 23:25:45 10.244.48.0:43703 GET /favicon.ico
```

I suppose the dashes in the "pid" could then be converted to periods if we feel that's important, but that's a one liner.

smothiki:
changed the regex as per kent's suggestion 
2015-08-27T17:56:27UTC wiggly-buttress[web-npv8v]: 2015/08/27 17:56:27 listening on 5000...

krancour:
Code LGTM.

mboersma:
LGTM.

PR #4356, opened 2015-08-27
fix(router): revert symlink to /dev/stdout

bacongobbler:
Nginx cannot log to a device for some reason, so this will have to stay. Why this works for `nginx` and not for `deis/router` is a little odd  (perhaps something to do with back-grounding nginx?)

### DIFF
--- a/router/cmd/boot/boot.go
+++ b/router/cmd/boot/boot.go
@@ -11,6 +11,7 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/ActiveState/tail"
 	"github.com/Sirupsen/logrus"
 	"github.com/coreos/go-etcd/etcd"
 
@@ -20,8 +21,10 @@ import (
 var log = logrus.New()
 
 const (
-	timeout time.Duration = 10 * time.Second
-	ttl     time.Duration = timeout * 2
+	timeout        time.Duration = 10 * time.Second
+	ttl            time.Duration = timeout * 2
+	nginxAccessLog string        = "/opt/nginx/logs/access.log"
+	nginxErrorLog  string        = "/opt/nginx/logs/error.log"
 )
 
 func main() {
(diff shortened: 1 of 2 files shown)


bacongobbler:
this was originally introduce by me #4260. Sorry!

krancour:
LGTM

router/cmd/boot/boot.go
```diff
@@ -62,6 +65,9 @@ func main() {
 
 	log.Info("Starting Nginx...")
 
+	go tailFile(nginxAccessLog)
+	go tailFile(nginxErrorLog)
+
 	nginxChan := make(chan bool)
 	go launchNginx(nginxChan)
 	<-nginxChan
```
mboersma: Does this also need to restore the `cleanOnExit` code from #4260?

bacongobbler: yes, but that code didn't work as intended since it never waited for the goroutine to finish. I'll fix that up as well.

bacongobbler: done

router/cmd/boot/boot.go
```diff
@@ -80,6 +86,8 @@ func main() {
 	exitChan := make(chan os.Signal, 2)
 	signal.Notify(exitChan, syscall.SIGTERM, syscall.SIGINT)
 	<-exitChan
+	tail.Stop()
```
mboersma: The compiler failed this because `Cleanup` is a simple func but `Tail` must be called through an actual instance:
```go
func Cleanup() { ...

func (tail *Tail) Stop() error { ...
```

bacongobbler: okay I removed it. I was just briefly reading through tail's test code and saw instances of `tail.Stop` being called so I naively did the same. I'll keep the behavior the same as it once was.

mboersma:
LGTM.

PR #4359, opened 2015-08-28
fix(client-go): pass arguments when using root command to list

### LINKED ISSUE #4357: `deis ps -a www` fails
This fails:

```
deis ps -a www
Found no matching command, try 'deis help'
Usage: deis <command> [<args>...]
```

but with the old client, this was a valid command. `deis ps:list -a www` works as expected.

Joshua-Anderson:
Fixes #4357 

This change got fairly in depth because of the new matching that had to be done to separate invalid commands from arguments passed to `command:list`.

### DIFF
--- a/client-go/deis.go
+++ b/client-go/deis.go
@@ -62,20 +62,19 @@ Shortcut commands, use 'deis shortcuts' to see all::
 Use 'git push deis master' to deploy to an application.
 `
 	// Reorganize some command line flags and commands.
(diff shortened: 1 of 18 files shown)


bacongobbler:
Worked for me! Code LGTM

mboersma:
LGTM. I tested this interactively with as many variations on command-line flags as I could think of.

PR #4365, opened 2015-08-28
ref(contrib/azure): use ARM template for provisioning

### LINKED ISSUE #4334: Azure: Switch provision scripts over to using ARM template
Currently, the Azure provision script uses v1 of the Azure API, which from what I understand is being deprecated in favor of [Azure Resource Manager](https://azure.microsoft.com/en-us/documentation/articles/resource-group-overview/). We should pull in the ARM [template](https://github.com/Azure/azure-quickstart-templates/pull/473) that's already been written for Deis 1.9, polish it, and use it in the project in place of the existing Python script.

I'd also like to use Terraform instead of the Azure CLI to submit the template, but it appears that Terraform doesn't yet support any ARM resources for Azure: https://terraform.io/docs/providers/azure/index.html

Using the ARM template will also address concerns such as #4332 (using an availability set for Deis) and #3323.

/cc @sgoings

carmstrong:
Rewrite `contrib/azure` to use ARM templates instead of the Python provisioning process.

This is a WIP until we:
* [x] Confirm my workaround to address https://github.com/deis/deis/issues/4334#issuecomment-135170325 by assigning public IP addresses to instances
* [x] Update documentation

closes #4334


kalbasit:
Accidentally closed the PR.

contrib/azure/arm-template.json
```diff
@@ -0,0 +1,387 @@
...
+              ]
+            }
+          }
+        },
+        "storageProfile": {
+          "imageReference": {
+            "publisher": "[variables('imagePublisher')]",
+            "offer": "[variables('imageOffer')]",
+            "sku": "[variables('imageSKU')]",
+            "version": "766.4.0"
```
mboersma: CoreOS version looks good, see #4550.

PR #4367, opened 2015-08-28
feat(router): improve healthcheck

### LINKED ISSUE #4344: Do we need a more sophisticated health check for deis-router?
This is unrelated to application health checks.

On each platform we support, load balancers are configured with a health check that just tests whether anything is listening on port 80.  This means any node in a Deis cluster that is running the `deis-router` reports back as "healthy" and is therefore included in that load balancer's pool of instances it can send traffic to.

Where this breaks down is the case where Kubernetes is the scheduler.  If the `deis-kube-proxy` component should fail on any given machine, then the consequence is that even though a router instance might be alive and well on that same machine, application traffic sent to that machine will never reach a pod.

Is it possible that `deis-router` should include a more sophisticated health check endpoint where a 200 is returned if the router is _able_ to route traffic?  Here, "able" is probably defined as the `deis-router` component is up and healthy, _and_, if applicable, `deis-kube-proxy` is as well.

Ping @bacongobbler for thoughts.

krancour:
Fixes #4344 

This improves confd templates for the router's Nginx configuration such that if the k8s scheduler is in use, the _existing_ `/health-check` endpoint on the router's default virtual host is enhanced to proxy `deis-kube-proxy`'s own health check endpoint.

__tl;dr:__ I also added a second healthcheck endpoint specifically for use with AWS ELBs.

Since we use `proxy_protocol` when the routers are behind an AWS ELB and ELBs do not include proxy protocol headers when initiating a healthcheck request over HTTP ([a bug, IMO](https://forums.aws.amazon.com/message.jspa?messageID=633016)) we have historically resorted to simpler healthchecks that only assert something is listening on port 80.  That is inadequate to make use of the enhanced healthcheck endpoint.

To fix this, I wanted to expose the `/health-check` endpoint without using `proxy_protocol`, regardless of whether it was otherwise needed for other traffic.  In theory, that would work, but it seems it may not be possible to have some virtualhosts use `proxy_protocol` and others _not_ use it if they're listening to a common ip:port.

As a workaround, what I have done is added a _second_ `/health-check` endpoint that listens on port 9090 instead and, indeed, does not ever use `proxy_protocol`.

I have tested this on AWS and it all works great.

AFAIK, no other provider's load balancers should have to make use of the secondary healthcheck endpoint on port 9090.

### DIFF
--- a/contrib/aws/deis.template.json
+++ b/contrib/aws/deis.template.json
@@ -300,7 +300,7 @@
         "HealthCheck": {
           "HealthyThreshold": "4",
           "Interval": "15",
-          "Target": "TCP:80",
+          "Target": "HTTP:9090/health-check",
           "Timeout": "5",
           "UnhealthyThreshold": "2"
         },
@@ -367,7 +367,8 @@
           {"IpProtocol": "tcp", "FromPort" : "22",  "ToPort" : "22",  "CidrIp" : { "Ref" : "SSHFrom" }},
           {"IpProtocol": "tcp", "FromPort": "80", "ToPort": "80", "SourceSecurityGroupId": { "Ref": "DeisWebELBSecurityGroup" } },
           {"IpProtocol": "tcp", "FromPort": "443", "ToPort": "443", "SourceSecurityGroupId": { "Ref": "DeisWebELBSecurityGroup" } },
-          {"IpProtocol": "tcp", "FromPort": "2222", "ToPort": "2222", "SourceSecurityGroupId": { "Ref": "DeisWebELBSecurityGroup" } }
+          {"IpProtocol": "tcp", "FromPort": "2222", "ToPort": "2222", "SourceSecurityGroupId": { "Ref": "DeisWebELBSecurityGroup" } },
+          {"IpProtocol": "tcp", "FromPort": "9090", "ToPort": "9090", "SourceSecurityGroupId": { "Ref": "DeisWebELBSecurityGroup" } }
         ],
         "VpcId" : { "Ref" : "VPC" }
       }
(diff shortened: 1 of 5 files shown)


krancour:
For the record, 9090 was selected arbitrarily.  It is not in the usual ephemeral port range that Docker would ever map container ports to on the host and it's not in use by anything else of ours, AFAIK.

mboersma:
Needs a rebase. :recycle:

krancour:
All set.  All your rebase are belong to us.

bacongobbler:
so for users upgrading, should we add documentation somewhere that they'll have to update their cloudformation templates in order to receive this update, or is this a manual process to upgrade?

krancour:
@bacongobbler I was more or less thinking that similar to when we change user-data, this feature is pretty much just for new clusters.  Older clusters will continue to work in the same way they always have.  I hesitate on recommending that people update their CF stacks because unless they've preserved all their environment variables and userdata from when they initially provisioned, they can potentially do more harm than good to their clusters.

kalbasit:
Accidentally closed the PR.

carmstrong:
> Older clusters will continue to work in the same way they always have.

So I thought through this and agree that it's safe, but want to confirm:

In the scenario where users upgrade to this release, their router units and the component itself would be upgraded and start using 9090 *in addition to* existing health checks. It's okay that the AWS cluster isn't yet updated to actually use this healthcheck.

krancour:
@carmstrong not quite...

Anyone who upgrades to this release will get a new (additional) health check endpoint.  It won't mean they're using it.

To use it, they'd have to update their CloudFormation stack-- which I've noted is something that can be dangerous if you don't know what you're doing.  Rigger will help with that.  I don't think I'd necessarily recommend that people do this.

Newly provisioned AWS clusters will get the full benefit of this feature.

Do I get to keep the LGTMs?  :pray:

mboersma:
You did describe this as an "opt-in" feature when we reviewed it previously; understood and I'm still fine with this PR.

carmstrong:
> Anyone who upgrades to this release will get a new (additional) health check endpoint. It won't mean they're using it.

Well, that's what I meant ;) The router will be listening on 9090 in addition to existing ports, even though that may not necessarily be used.

So this looks fine to me.

PR #4381, opened 2015-08-31
fix(firewall): check etcd is running

### LINKED ISSUE #4281: contrib/util/custom-firewall.sh throws increasingly disconcerting messages
https://github.com/deis/deis/blob/master/contrib/util/custom-firewall.sh

I have a five-node cluster on DigitalOcean and I'm following according to the guide

http://docs.deis.io/en/latest/installing_deis/digitalocean/

Apply Security Group Settings

```
... [stuff]
# Accept ssh, http, https and git
-A Firewall-INPUT -m conntrack --ctstate NEW -m multiport -p tcp --dports 22,2222,80,443 -j ACCEPT

# Log and drop everything else
-A Firewall-INPUT -j LOG
-A Firewall-INPUT -j REJECT

COMMIT
Saving firewall Rules
Enabling iptables service
Loading custom iptables firewall
Created symlink from /etc/systemd/system/basic.target.wants/iptables-restore.service to /usr/lib64/systemd/system/iptables-restore.service.
iptables: No chain/target/match by that name.
Bad argument `ACCEPT'
Error occurred at line: 21
Done
Try `iptables-restore -h' or 'iptables-restore --help' for more information.
(stretch)yebyen@localhost:~/deis$ 
```

The command I'm using:
```
    for i in 1 2 3 4 5; do ssh core@deis-$i.[my.cluster.address] 'bash -s' < contrib/util/custom-firewall.sh; done
```

This output looks bad, if I had to guess I'd say my security group settings are probably not being (at least fully) applied.

aledbf:
This also removes `iptables: No chain/target/match by that name.` when the script is executed the first time (is not an error)

closes #4281

### DIFF
--- a/contrib/util/custom-firewall.sh
+++ b/contrib/util/custom-firewall.sh
@@ -1,5 +1,17 @@
 #!/bin/env bash
 
+if [ ! -f /opt/bin/jq ]; then
+  echo "/opt/bin/jq is missing. Downloading..."
+  curl -sSL -o /opt/bin/jq http://stedolan.github.io/jq/download/linux64/jq
+  chmod +x /opt/bin/jq
+fi
+
+# obtain the etcd node members and check that at least there is three
+ETCD_NODES=$(curl -s http://localhost:4001/v2/members | jq '.[] | .[].peerURLs | length' | wc -l)
+if test $ETCD_NODES -lt 3; then
+  echo "etcd is not working correctly. Verify the etcd cluster is running before the execution of this script."
+fi
+
 echo "Obtaining IP addresses of the nodes in the cluster..."
 MACHINES_IP=$(fleetctl list-machines --fields=ip --no-legend | awk -vORS=, '{ print $1 }' | sed 's/,$/\n/')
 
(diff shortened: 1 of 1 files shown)


contrib/util/custom-firewall.sh
```diff
@@ -1,5 +1,11 @@
 #!/bin/env bash
 
+# obtain the etcd node members and check that at least there is three
+ETCD_NODES=$(curl -s http://localhost:4001/v2/members | jq '.[] | .[].peerURLs | length' | wc -l)
```
carmstrong: `jq` is only on CoreOS hosts as of [717.0.0](https://coreos.com/releases/#717.0.0)

We'll have to wait until we bump hosts from 647.2.0 (hopefully soon!)

aledbf:
@carmstrong is ok to do this https://github.com/deis/deis/blob/e3d172fd451892564bcdb267caaa7298c22971a2/contrib/coreos/user-data.example#L144 here in the meantime?

carmstrong:
> is ok to do this

I think that's what we'd have to do :+1:

mboersma:
LGTM.

carmstrong:
Code LGTM.

PR #4382, opened 2015-08-31
fix(client): fix make install

krancour:
This updates the `install` target in the Go-based client's `Makefile` to fulfill the reasonable expectation a contributor may have that `make -C client/ build install` will actually install a `deis` binary.  It does so by copying the built binary to `$GOPATH/bin`.

### DIFF
--- a/client/Makefile
+++ b/client/Makefile
@@ -14,8 +14,8 @@ build:
 	CGO_ENABLED=0 godep go build -a -installsuffix cgo -ldflags '-s' -o deis .
 	@$(call check-static-binary,deis)
 
-install:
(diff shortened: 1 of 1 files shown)


krancour:
For the record, I should also state that _copying_ the binary is to circumvent the limitation that `go install` doesn't permit the name of the installed binary to be specified.

mboersma:
I guess it is possible to set `GOPATH` to contain multiple colon-separated directories, in which case this will fail. But people rarely do that because other tools are easily confused in that case, although go does allow it.

LGTM.

carmstrong:
Code LGTM.

Issue #4377, opened 2015-08-31
Healthcheck problem


vesmi:
My application consists of two parts: web (php) and scheduler (bash script). Health сheck of the second part causes an error because there is no interface, ability to give code 200. Is it possible to more flexibly adjust healthcheck to avoid similar problems?
```bash
2015-08-28T11:55:20UTC deis[api]: failed to connect to container ('Key not found : /deis/services/colibri-api/colibri-api_v11.scheduler.1'); trying again in 1.0 seconds
2015-08-28T11:55:21UTC deis[api]: failed to connect to container ('Key not found : /deis/services/colibri-api/colibri-api_v11.scheduler.1'); trying again in 5.0 seconds
2015-08-28T11:55:26UTC deis[api]: failed to connect to container ('Key not found : /deis/services/colibri-api/colibri-api_v11.scheduler.1'); trying again in 10.0 seconds
2015-08-28T11:55:36UTC deis[api]: failed to connect to container ('Key not found : /deis/services/colibri-api/colibri-api_v11.scheduler.1')
2015-08-28T11:55:41UTC deis[api]: aborting, app containers failed to respond to health check
```
Deis 1.9.1, config: 
HEALTHCHECK_URL="/1.0/service/healthcheck" 
HEALTHCHECK_TIMEOUT=30 
HEALTHCHECK_INITIAL_DELAY=10

Procfile example:
```bash
web: vendor/bin/heroku-php-nginx -C nginx.inc.conf web/
scheduler: while true; do ./yii schedule/run --verbose=0; sleep 3600; done
```

bacongobbler:
Yes. The health check should only apply to web and cmd process types.

laurrentt:
+1

rvadim:
+1

vmakhaev:
Even running web server (which returns 200) in worker process does not work for me.

bacongobbler:
@vmakhaev would you mind checking out #4383? That should fix your issue :)

vmakhaev:
Yes, making healthcheck work only for web processes should solve issue.
I have conceptual question. Could it be useful to have same health check for background processes (running web server inside them)? What are the best practices/recommendations in monitoring background processes?

bacongobbler:
no, I don't think it will be useful. Deis requires that your app only listens on one port due to some limitations in v1 so we can only monitor the process listening on that container's port.

vmakhaev:
[...]

PR #4383, opened 2015-08-31
fix(controller): healthcheck no longer check non-web processes

laurrentt:
This should fix https://github.com/deis/deis/issues/4377.


bacongobbler:
code lgtm

mboersma:
LGTM.

PR #4384, opened 2015-08-31
Remove etcd tweaks

mboersma:
We've seen lots of etcd/fleet hiccups while testing recently, so this attempts to fix that by removing the timing tweaks Deis had in user-data. The [default etcd2 service](https://github.com/coreos/coreos-overlay/blob/master/dev-db/etcd/files/etcd2.service) lets the binary use its default values, so the hope is that will encourage better behavior.

Also updates the etcd version in the graceful-shutdown stanza to 2.1.1.


bacongobbler:
lgtm

pmoust:
we are also experiencing flaky etcd2 behavior recently.
Did switching to 2.1 mitigate the effect?

aj-may:
LGTM

carmstrong:
Code LGTM.

kalbasit:
Accidentally closed the PR.

mboersma:
Closing this in favor of #4444, which also removes these `etcd` timing tweaks.

@pmoust honestly we have not seen version 2.x behaving much differently from 0.4.9 with regard to Deis' behavior, so it's hard to say if would improve stability for you.

PR #4390, opened 2015-09-01
fix(client): show non-JSON HTTP errors

mboersma:
This changes the new CLI to display the raw response body and status code for HTTP 40x and 50x errors if JSON decoding fails. It also moves the HTTP status line first for any error displayed, to match the old python CLI.

### DIFF
--- a/client/controller/client/http.go
+++ b/client/controller/client/http.go
@@ -57,6 +57,10 @@ func (c Client) Request(method string, path string, body []byte) (*http.Response
 		return nil, err
 	}
 
+	if err = checkForErrors(res, ""); err != nil {
+		return nil, err
+	}
+
 	checkAPICompatability(res.Header.Get("DEIS_API_VERSION"))
 
 	return res, nil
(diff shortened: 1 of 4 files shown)


client/controller/client/http.go
```diff
@@ -111,7 +111,7 @@ func checkForErrors(res *http.Response, body string) error {
 	bodyMap := make(map[string]interface{})
 
 	if err := json.Unmarshal([]byte(body), &bodyMap); err != nil {
-		return err
+		return errors.New("\n" + body + "\n" + res.Status + "\n")
```
bacongobbler: The python client was reversed, which would show the status then the raw HTML. Could we change it to display it in that order or is there a reason for this ordering?

mboersma:
> is there a reason for this ordering?
I'm following the convention of the JSON-formatted messages, which have the fields, then the status code. I thought it was preferable to be internally consistent, although I would have put the status code first in both cases...

client/controller/client/http.go
mboersma: > is there a reason for this ordering?

I'm following the convention of the JSON-formatted messages, which have the fields, then the status code. See elsewhere in the same test. I thought it was preferable to be internally consistent, although I would have put the status code first in both cases...

mboersma: You're right, that would be consistent with the python CLI. I'll see about putting the `response.Status` line first in all these cases.

carmstrong:
Code LGTM.

krancour:
Code LGTM

Joshua-Anderson:
:+1: I found this annoying as well!

Issue #3797, opened 2015-06-05
deis-builder stuck in "activating/start-post" after ExecStart failed


mboersma:
In a recent [vagrant test](https://ci.deis.io/job/test-coreos-beta/184/), builder hit an error during its `boot` script that somehow didn't restart the unit:
```console
core@deis-03 ~ $ systemctl status deis-builder
● deis-builder.service - deis-builder
   Loaded: loaded (/run/fleet/units/deis-builder.service; linked-runtime; vendor preset: disabled)
   Active: activating (start-post) (Result: exit-code) since Fri 2015-06-05 19:03:41 ; 24min ago
  Process: 2984 ExecStart=/bin/sh -c IMAGE=`/run/deis/bin/get_image /deis/builder` && docker run --name deis-builder --rm -p 2223:22 --volumes-from=deis-builder-data -c 800 -e EXTERNAL_PORT=2223 -e HOST=$COREOS_PRIVATE_IPV4 --privileged -v /etc/environment_proxy:/etc/environment_proxy $IMAGE (code=exited, status=1/FAILURE)
  Process: 2975 ExecStartPre=/bin/sh -c /sbin/losetup -f (code=exited, status=0/SUCCESS)
  Process: 2962 ExecStartPre=/bin/sh -c docker inspect deis-builder >/dev/null 2>&1 && docker rm -f deis-builder || true (code=exited, status=0/SUCCESS)
  Process: 2660 ExecStartPre=/bin/sh -c IMAGE=`/run/deis/bin/get_image /deis/builder` && docker history $IMAGE >/dev/null 2>&1 || docker pull $IMAGE (code=exited, status=0/SUCCESS)
  Process: 2613 ExecStartPre=/bin/sh -c docker inspect deis-builder-data >/dev/null 2>&1 || docker run --name deis-builder-data -v /var/lib/docker alpine:3.1 /bin/true (code=exited, status=0/SUCCESS)
 Main PID: 2984 (code=exited, status=1/FAILURE);         : 2985 (sh)
   Memory: 748.0K
   CGroup: /system.slice/deis-builder.service
           └─control
             ├─2985 /bin/sh -c echo 'Waiting for builder on 2223/tcp...' && until ncat $COREOS_PRIVATE_IPV4 2223 --exec '/us...
             └─6808 sleep 1

Jun 05 19:03:50 deis-03 sh[2660]: 89a68d4ec32b: Pulling metadata
[...]

PR #4394, opened 2015-09-02
feat(controller): publish etcd keys at controller start

### LINKED ISSUE #4355: Proposal: controller should republish certain things to etcd at startup
Supersedes #3486 and #3539

In the event a new cluster is built as a replacement for an old one and the database is either external (as in the case of running Cephless) or is restored from a backup of the old cluster, certain necessary configuration that _is_ present in the database needs to be republished to etcd.

The controller could do this at startup.

I would propose (at least) the following things be republished:

* Application configuration
* Application domains
* User keys

Relates to #4333

krancour:
Fixes #4355 

Much thanks to @jgmize for the management command.

### DIFF
--- a/controller/api/management/commands/load_db_state_to_etcd.py
+++ b/controller/api/management/commands/load_db_state_to_etcd.py
@@ -1,0 +1,16 @@
+from django.core.management.base import BaseCommand
+
+from api.models import Key, App, Domain, Certificate, Config
+
+
+class Command(BaseCommand):
+    """Management command for publishing Deis platform state from the database
+    to etcd.
+    """
+    def handle(self, *args, **options):
+        """Publishes Deis platform state from the database to etcd."""
+        print "Publishing DB state to etcd..."
+        for model in (Key, App, Domain, Certificate, Config):
+            for obj in model.objects.all():
+                obj.save()
+        print "Done Publishing DB state to etcd."
--- a/controller/api/models.py
+++ b/controller/api/models.py
@@ -1081,9 +1081,10 @@ def _log_config_updated(**kwargs):
 
 
 def _log_domain_added(**kwargs):
-    domain = kwargs['instance']
-    msg = "domain {} added".format(domain)
-    log_event(domain.app, msg)
+    if kwargs.get('created'):
+        domain = kwargs['instance']
+        msg = "domain {} added".format(domain)
+        log_event(domain.app, msg)
 
 
 def _log_domain_removed(**kwargs):
@@ -1093,8 +1094,9 @@ def _log_domain_removed(**kwargs):
 
 
 def _log_cert_added(**kwargs):
-    cert = kwargs['instance']
-    logger.info("cert {} added".format(cert))
+    if kwargs.get('created'):
+        cert = kwargs['instance']
+        logger.info("cert {} added".format(cert))
 
 
 def _log_cert_removed(**kwargs):
@@ -1127,10 +1129,13 @@ def _etcd_purge_user(**kwargs):
         pass
 
 
-def _etcd_create_app(**kwargs):
+def _etcd_publish_app(**kwargs):
     appname = kwargs['instance']
-    if kwargs['created']:
+    try:
         _etcd_client.write('/deis/services/{}'.format(appname), None, dir=True)
+    except KeyError:
+        # Ignore error when the directory already exists.
+        pass
 
 
 def _etcd_purge_app(**kwargs):
@@ -1143,9 +1148,8 @@ def _etcd_purge_app(**kwargs):
 
 def _etcd_publish_cert(**kwargs):
     cert = kwargs['instance']
-    if kwargs['created']:
-        _etcd_client.write('/deis/certs/{}/cert'.format(cert), cert.certificate)
-        _etcd_client.write('/deis/certs/{}/key'.format(cert), cert.key)
+    _etcd_client.write('/deis/certs/{}/cert'.format(cert), cert.certificate)
+    _etcd_client.write('/deis/certs/{}/key'.format(cert), cert.key)
 
 
 def _etcd_purge_cert(**kwargs):
@@ -1167,13 +1171,12 @@ def _etcd_publish_config(**kwargs):
                             prevExist=True, dir=True, recursive=True)
     except KeyError:
         pass
-    if kwargs['created']:
-        for k, v in config.values.iteritems():
-            _etcd_client.write(
-                '/deis/config/{}/{}'.format(
-                    config.app,
-                    unicode(k).encode('utf-8').lower()),
-                unicode(v).encode('utf-8'))
+    for k, v in config.values.iteritems():
+        _etcd_client.write(
+            '/deis/config/{}/{}'.format(
+                config.app,
+                unicode(k).encode('utf-8').lower()),
+            unicode(v).encode('utf-8'))
 
 
 def _etcd_purge_config(**kwargs):
@@ -1187,8 +1190,7 @@ def _etcd_purge_config(**kwargs):
 
 def _etcd_publish_domains(**kwargs):
     domain = kwargs['instance']
-    if kwargs['created']:
-        _etcd_client.write('/deis/domains/{}'.format(domain), domain.app)
+    _etcd_client.write('/deis/domains/{}'.format(domain), domain.app)
 
 
 def _etcd_purge_domains(**kwargs):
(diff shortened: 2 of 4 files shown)


krancour:
@mboersma @bacongobbler can I ask the two of you to review this?  Django is a little new to me.

deisctl/units/deis-controller.service
```diff
@@ -9,7 +9,9 @@ EnvironmentFile=/etc/environment
 TimeoutStartSec=20m
 ExecStartPre=/bin/sh -c "IMAGE=`/run/deis/bin/get_image /deis/controller` && docker history $IMAGE >/dev/null 2>&1 || docker pull $IMAGE"
 ExecStartPre=/bin/sh -c "docker inspect deis-controller >/dev/null 2>&1 && docker rm -f deis-controller || true"
-ExecStart=/bin/sh -c "IMAGE=`/run/deis/bin/get_image /deis/controller` && docker run --name deis-controller --rm -p 8000:8000 -e EXTERNAL_PORT=8000 -e HOST=$COREOS_PRIVATE_IPV4 -v /var/run/fleet.sock:/var/run/fleet.sock -v /var/lib/deis/store:/data $IMAGE"
+ExecStart=/bin/sh -c "IMAGE=`/run/deis/bin/get_image /deis/controller` && docker run --name deis-controller --rm -p 8000:8000 -e EXTERNAL_PORT=8000 -e HOST=$COREOS_PRIVATE_IPV4 -e ETCD=$COREOS_PRIVATE_IPV4:4001 -v /var/run/fleet.sock:/var/run/fleet.sock -v /var/lib/deis/store:/data $IMAGE"
+ExecStartPost=/bin/sh -c "echo 'Waiting for controller on 8000/tcp...' && until ncat $COREOS_PRIVATE_IPV4 8000 --exec '/usr/bin/echo dummy-value' >/dev/null 2>&1; do sleep 1; done"
+ExecStartPost=/usr/bin/docker exec deis-controller ./manage.py load_db_state_to_etcd
```
laurrentt: I think calling the command [here](https://github.com/deis/deis/blob/master/controller/bin/boot#L88) instead would be more consistent.

bacongobbler: indeed, considering other issues we've seen with execstartpost: https://github.com/deis/deis/issues/3797

krancour: > I think calling the command here instead would be more consistent.

:+1: I'll do that.

controller/api/management/commands/load_db_state_to_etcd.py
```diff
@@ -0,0 +1,11 @@
+from django.core.management.base import BaseCommand
+from api.models import Key, App, Domain, Certificate, Config
```
mboersma: [PEP8](https://www.python.org/dev/peps/pep-0008/#imports) says there should be a blank line here, between third-party and local imports. (Although the `flake8` tool doesn't enforce this.)

controller/api/management/commands/load_db_state_to_etcd.py
```diff
@@ -0,0 +1,11 @@
...
+
+
+class Command(BaseCommand):
+    def handle(self, *args, **options):
+        print "Publishing DB state to etcd..."
+        for M in (Key, App, Domain, Certificate, Config):
```
mboersma: I would use `model` and `obj` as local var names instead of `M` and `o`. Brief but descriptive (and lower-case) is pythonic, it's only in Go that we get to use single-char names, except for special cases `i` and `x`.

controller/api/management/commands/load_db_state_to_etcd.py
```diff
@@ -0,0 +1,11 @@
+from django.core.management.base import BaseCommand
+from api.models import Key, App, Domain, Certificate, Config
+
+
+class Command(BaseCommand):
```
mboersma: There should be a docstring here describing what the `Command` class does, and another docstring for `handle`.

krancour:
@laurrentt @bacongobbler @mboersma all feedback received so far has been gratefully received and acted upon.  :)  Feel free to have another gander.

ineu:
Won't it be better to do something like `logger.debug("Failed to ...")` instead of simply swallowing error in the `rescue` block? It would be easier for user to debug issue by just setting controller's loglevel to debug (is there a way to do it btw?) and looking at `journalctl -u deis-controller`.

controller/api/models.py
```diff
@@ -1129,10 +1131,12 @@ def _etcd_purge_user(**kwargs):
         pass
 
 
-def _etcd_create_app(**kwargs):
+def _etcd_publish_app(**kwargs):
     appname = kwargs['instance']
-    if kwargs['created']:
+    try:
         _etcd_client.write('/deis/services/{}'.format(appname), None, dir=True)
+    except KeyError:
```
krancour: @ineu is this the line you were referring to with your previous comment.  There was a specific motivation for this line.  It was late when I did it.  (So I appreciate your spot-checking me here!)  If I recall correctly, this line produces an error if the directory already exists, and that's A-OK.  It was that, specifically, that I was trying to eat.  Let me double-check that.  At the very least, there should be a comment here explaining this.

krancour: I just double-checked myself.  That was exactly the error I was swallowing-- deliberately.  If the directory already exists, the "mkdir" on line 1137 fails.  As it happens, if that particular line failed for any _other_ reason that also throws a `KeyError`, we shouldn't care too much.  The same directory that line is attempting to create would actually be created implicitly by the `deis-publisher` anyway as it goes about its business of making container host IPs and ports discoverable by writing `/deis/services/<app>/<container>/<address>`.

I will add a comment in here to clarify this.

ineu: Thank you.

krancour:
@ineu I added a comment to explain the rationale.

krancour:
Whoops... added the comment in the wrong spot and had to repush.  Should be good now.

controller/api/models.py
mboersma: This could be just `# Ignore error when the directory already exists.`

controller/api/models.py
```diff
@@ -1095,8 +1096,9 @@ def _log_domain_removed(**kwargs):
 
 
 def _log_cert_added(**kwargs):
-    cert = kwargs['instance']
-    logger.info("cert {} added".format(cert))
+    if kwargs.get('created'):
+        cert = kwargs['instance']
+        logger.info("cert {} added".format(cert))
```
mboersma: Just to note here: there was an early convention in the Deis python codebase of using double quotes `"` for format strings but single quotes `'` for all others. This was a nudge toward common standards in a polyglot codebase for previous products built with Ruby and CoffeeScript. 

So that's why you'll see this difference in much of the Deis Python code, but it may not be productive to bother about it now. :snake:

controller/api/models.py
krancour: Now using the more succinct comment as suggested.  Thanks @mboersma !

docs/managing_deis/backing_up_data.rst
```diff
@@ -195,6 +195,12 @@ Note that you'll need to specify the SSH port when using Vagrant:
 Finishing up
 ~~~~~~~~~~~~
 
+.. note::
+
+    As of version 1.11.0, the following steps are no longer necessary because
+    the ``deis-controller`` component will handle them automatically upon
+    starting.
```
mboersma: Should we delete this whole "Finishing up" topic if it's not needed? Then the docs are consistent with the codebase, and there is no need to mention the next unreleased version, which may confuse people looking at the default "latest" documentation.

bacongobbler: I agree. Either that or a TODO to delete this note once v1.11.0 is released since "latest" is what people are usually looking at.

mboersma:
I tested this and it seems to be working fine:
```
Sep 10 22:06:38 deis-03 sh[12861]: [2015-09-10 22:06:38 +0000] [134] [INFO] Booting worker with pid: 134
Sep 10 22:06:38 deis-03 sh[12861]: Publishing DB state to etcd...
Sep 10 22:06:38 deis-03 sh[12861]: Done Publishing DB state to etcd.
Sep 10 22:06:38 deis-03 sh[12861]: deis-controller running...
Sep 10 22:09:09 deis-03 sh[12861]: 172.17.8.100 "GET /v1/apps/?page_size=100 HTTP/1.0" 401 26 "Deis Client v1.11.0-dev"
Sep 10 22:09:16 deis-03 sh[12861]: 172.17.8.100 "GET /v1/ HTTP/1.0" 401 58 "Deis Client v1.11.0-dev"
```
Code looks good once the question about docs is resolved. Nice job @krancour!

kalbasit:
Accidentally closed the PR.

docs/managing_deis/backing_up_data.rst
krancour: I'll remove this doc.

krancour:
@mboersma as discussed here and offline, I went to remove the "finishing up" part of the backup/restore docs.  I decided there for transparency there might be a benefit to noting the automated republish of user-related and app-related etcd keys by the controller.  So instead of deleting it, I slimmed it down quite a bit.  lmk if it looks alright.

mboersma:
LGTM.

carmstrong:
:+1: I like the update to the backup/restore docs rather then removing it entirely.

PR #4395, opened 2015-09-02
feat(Makefile): retry discovery-url five times

mboersma:
We've had some failures contacting discovery.etcd.io, leading to a blank discovery-url entry in user-data. This will try up to five times with a 3-second pause in between before giving up.

Thanks to @sgoings for pairing on this!

```console
$ make discovery-url
Wrote https://discovery.etcd.io/25afd6b8c114cc0e7b67a8f4aaac5405 to contrib/coreos/user-data
$ make discovery-url
Failed to contact https://discovery.etcd.io
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,19 @@ dev-cluster: discovery-url
 	deisctl install platform
 
 discovery-url:
-	sed -e "s,discovery #DISCOVERY_URL,discovery $$(curl -s -w '\n' https://discovery.etcd.io/new?size=$$DEIS_NUM_INSTANCES)," contrib/coreos/user-data.example > contrib/coreos/user-data
+	@for i in 1 2 3 4 5; do \
+		URL=`curl -s -w '\n' https://discovery.etcd.io/new?size=$$DEIS_NUM_INSTANCES`; \
+		if [ ! -z $$URL ]; then \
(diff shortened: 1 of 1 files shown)


bacongobbler:
code LGTM

Makefile
```diff
@@ -31,7 +31,19 @@ dev-cluster: discovery-url
 	deisctl install platform
 
 discovery-url:
-	sed -e "s,discovery #DISCOVERY_URL,discovery $$(curl -s -w '\n' https://discovery.etcd.io/new?size=$$DEIS_NUM_INSTANCES)," contrib/coreos/user-data.example > contrib/coreos/user-data
+	@for i in 1 2 3 4 5; do \
```
carmstrong: We could alternatively use `$(seq 1 5)` here

carmstrong:
Code LGTM.

Makefile
mboersma: We discussed that but left it this way. It seemed like "seq 5" could imply everything is nicely parameterized and that you could change "5" to "10" and get ten retries, but since we had to hard-code "5" a few lines below, it actually seemed safer to hard-code all the values.

carmstrong: Fair point. Ok.

PR #4397, opened 2015-09-02
docs(roadmap): complete production hardening

carmstrong:
This was completed for last release, but I forgot to remove it.


mboersma:
LGTM.

bacongobbler:
lgtm

krancour:
@carmstrong I think one of those completed for 1.9 and the other for 1.10.  Should the whole section really be removed?

carmstrong:
> I think one of those completed for 1.9 and the other for 1.10.

The dates made it look like both had landed for 1.9, but if not, we shouldn't remove it until after tomorrow. Which is which?

krancour:
Isolating etcd landed in 1.9-- it was really just a combination of doc and etcd upgrade that got us there.  Isolating the control plane was more involved and landed in 1.10.0.

carmstrong:
Closing - will address in a post-release planning PR.

PR #4291, opened 2015-08-18
fix(logspout): fix deadlock when attaching

krancour:
Fixes #3633 

This isn't guaranteed to be a magic bullet that solves every logging-related issue, but we have uncovered deadlock that can occur when logspout attempts attaching to containers.  At several points in this process, a lock on a mutex is acquired and released.  At one such point, after acquiring a lock, a message is written to an unbuffered channel.  Since writes to an unbuffered channel will block until another goroutine receives the message, and since sometimes there is no one listening, execution can block indefinitely at this point while holding on to the lock that other goroutines are waiting for.  Using a buffered channel seems to alleviate this issue.

Thanks to @aledbf and @technosophos for help in diagnosing the issue.

N.B.: Longer term, we might consider adopting a newer version of logspout, as it's gone through two major refactors since we forked and enhanced it to work with Deis.  I'll track that in a separate issue.

Issue #3633, opened 2015-05-06
'deis logs' not catching all app logs from docker containers


hivearts:
docker logs
```
Wed May 06 2015 07:44:45 GMT+0000 (UTC): Webhook detected, rebuilding website started..
Wed May 06 2015 07:45:13 GMT+0000 (UTC): Rebuilding website completed.
Wed May 06 2015 07:47:21 GMT+0000 (UTC): Webhook detected, rebuilding website started..
Wed May 06 2015 07:47:50 GMT+0000 (UTC): Rebuilding website completed.
Wed May 06 2015 08:06:08 GMT+0000 (UTC): Webhook detected, rebuilding website started..
Wed May 06 2015 08:06:37 GMT+0000 (UTC): Rebuilding website completed.
Wed May 06 2015 08:08:58 GMT+0000 (UTC): Webhook detected, rebuilding website started..
Wed May 06 2015 08:09:27 GMT+0000 (UTC): Rebuilding website completed.
Wed May 06 2015 08:25:35 GMT+0000 (UTC): Webhook detected, rebuilding website started..
Wed May 06 2015 08:26:02 GMT+0000 (UTC): Rebuilding website completed.
Wed May 06 2015 08:29:41 GMT+0000 (UTC): Webhook detected, rebuilding website started..
Wed May 06 2015 08:30:09 GMT+0000 (UTC): Rebuilding website completed.
Wed May 06 2015 08:41:58 GMT+0000 (UTC): Webhook detected, rebuilding website started..
Wed May 06 2015 08:42:26 GMT+0000 (UTC): Rebuilding website completed.
Wed May 06 2015 08:48:18 GMT+0000 (UTC): Webhook detected, rebuilding website started..
Wed May 06 2015 08:48:46 GMT+0000 (UTC): Rebuilding website completed.
Wed May 06 2015 09:01:14 GMT+0000 (UTC): Webhook detected, rebuilding website started..
Wed May 06 2015 09:01:43 GMT+0000 (UTC): Rebuilding website completed.
Wed May 06 2015 09:05:35 GMT+0000 (UTC): Webhook detected, rebuilding website started..
Wed May 06 2015 09:06:03 GMT+0000 (UTC): Rebuilding website completed.
Wed May 06 2015 09:08:18 GMT+0000 (UTC): Webhook detected, rebuilding website started..
Wed May 06 2015 09:08:47 GMT+0000 (UTC): Rebuilding website completed.
```

deis client logs
```
[...]

PR #4399, opened 2015-09-03
fix(logspout): set GOMAXPROCS=1

krancour:
During v1.10.0 testing, I have been observing issues with `deis-logspout`.  They are most easily replicable when scaling to many containers quickly.  (Using k8s as the scheduler helps.)

The issue seems to be deadlock and is observable when the Docker event stream normally visible in each `deis-logspout` instance's own logs grinds to a halt.  If this sounds familiar, it is different, but similar to the root cause of #3633 which was closed by #4291

After much observation and testing, I realized I was never able to reproduce this issue if I built the `deis-logspout` container myself.  A variable I eventually realized was in play was that I was building `deis-logspout` using Go 1.4, whilst the images used for release testing where built using Go 1.5.

@gabrtv and @mboersma pointed me in the direction of recent changes in Go 1.5 re: concurrency.  See [here](https://golang.org/s/go15gomaxprocs).

As a workaround, this PR sets `GOMAXPROCS ` to 1 and seems to address the problem.

Ultimately, this is a little bit of a band-aid, but given that I expect #4347 to make broader improvements to how we manage logspout, I think this fix is just right.

@carmstrong @mboersma this needs to make it into v1.10.0.

### DIFF
--- a/logspout/logspout.go
+++ b/logspout/logspout.go
@@ -8,6 +8,7 @@ import (
 	"net/url"
 	"os"
 	"regexp"
+	"runtime"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


gabrtv:
LGTM.

carmstrong:
Code LGTM.

bacongobbler:
that might explain why I haven't seen this issue. I still haven't bumped to Go v1.5 locally just yet. I shall do that now!

Issue #4388, opened 2015-09-01
Proposal: update etcdctl on all new CoreOS hosts


krancour:
Currently, we update etcdctl on CoreOS hosts _if_ one has opted into the graceful shutdown unit.

What if we just update etcdctl on _all_ new CoreOS clusters instead?  This would significantly aid troubleshooting efforts since newer versions expose things like `member list` and `cluster-health`.

This shouldn't break anything.  Any containerized component would continue to use the older etcdctl 0.4.6 to ensure backwards compatibility with older clusters.

bacongobbler:
This is not something we'll be able to get to fixing for the LTS release, and provisioning Deis v2 is done through installing a vanilla kubernetes cluster so I'm going to go ahead and close this.

PR #4409, opened 2015-09-04
chore(*): bump etcd to 2.1.3

mboersma:
See https://github.com/coreos/etcd/compare/v2.1.2...v2.1.3 for details of the one bug fix included. (Probably irrelevant to our current testing issues.)


bacongobbler:
yup, completely irrelevant since it only deals with TLS-backed etcd clusters but still always nice to bump revs.

LGTM

carmstrong:
Should we also bump the graceful shutdown script in user-data.example so it checks for etcd 2.1.3?

krancour:
> Should we also bump the graceful shutdown script in user-data.example so it checks for etcd 2.1.3?

IMO, yes.  In fact, see #4388.  I am wondering if we should just upgrade `etcdctl` in _all_ new clusters (not only when graceful shutdown is enabled).  It adds useful stuff like `cluster-health` and `member list`.

mboersma:
> Should we also bump the graceful shutdown script in user-data.example...

Good catch--I did so.

> I am wondering if we should just upgrade `etcdctl` in *all* new clusters...

Perhaps, but that's not in scope for this PR. Also we will hopefully move up to more current CoreOS very soon, in which case we will have `etcdctl` >= v2.1.2 on the box.

krancour:
Code LGTM

kalbasit:
Accidentally closed the PR.

mboersma:
See #4444 instead, which effectively keeps the `etcd` version at 2.1.2 for now.

PR #4410, opened 2015-09-04
fix(database): improve postgres ready checks

mboersma:
The previous check using `netstat` was not specific enough and could return 1 before postgres starts.

### DIFF
--- a/database/bin/boot
+++ b/database/bin/boot
@@ -56,11 +56,13 @@ until confd -onetime -node "$ETCD" -confdir /app --log-level error; do
   sleep $((ETCD_TTL/2))  # sleep for half the TTL
 done
 
+PG_DATA_DIR=/var/lib/postgresql/9.3/main
+
 # initialize database if one doesn't already exist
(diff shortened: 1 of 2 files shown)


bacongobbler:
I like this change. What about using `pg_ctl status -D $PG_DATA_DIR`? That might give us a better representation if the database is running without having to establish a connection, though I think an attempt to connect to the database works just as well. I don't know well enough about the latter command (whether it detects if it's ready to accept connections or not) but it was something I was reading up on while working with governor.

mboersma:
> What about using `pg_ctl status -D $PG_DATA_DIR`?

A previous version of this PR (about 30 minutes ago!) did exactly that, but sadly it reports the server is running well before `psql` can actually connect, which is problematic. So actually using `psql` was the only thing we've found to be reliable in testing.

bacongobbler:
sounds good. Code LGTM!

carmstrong:
Code LGTM.

PR #4413, opened 2015-09-05
fix(Makefile): silence docker-machine warning when it isn't installed

Joshua-Anderson:
Without this change:

```
registry

To use a local registry for Deis development:
/bin/bash: docker-machine: command not found
    export DEV_REGISTRY=:5000
```

This change returns `dev-registry` to the old behavior:

```
registry

To use a local registry for Deis development:
    export DEV_REGISTRY=:5000
```

This is useful on Linux where it's common to use docker natively.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ dev-registry: check-docker
 	@docker inspect registry >/dev/null 2>&1 && docker start registry || docker run --restart="always" -d -p 5000:5000 --name registry registry:0.9.1
 	@echo
 	@echo "To use a local registry for Deis development:"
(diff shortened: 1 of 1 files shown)


mboersma:
I know this was the previous behavior, but is `:5000` actually a legitimate setting for `DEV_REGISTRY`?

bacongobbler:
it is not legit. I've always used `$HOST_IPADDR:5000` on my linux workstation.

Joshua-Anderson:
Switched to printing `HOST_IPADDR`. I was trying to bring the original behavior on linux back (pre-docker machine), but I agree that since people have `HOST_IPADDR` set anyway we should use it.

krancour:
LGTM

bacongobbler:
code LGTM

PR #4414, opened 2015-09-05
fix(*): ensure component containers stop

### LINKED ISSUE #4412: bug: docker containers aren't stopping upon deisctl stop
#4189 looks like it may have been more widespread than we realized.  I can consistently use deisctl or fleetctl to stop services and then observe that the corresponding containers are still running.

I believe the solution is a more widespread application of the fix from #4194

krancour:
Fixes #4412

### DIFF
--- a/deisctl/units/deis-database.service
+++ b/deisctl/units/deis-database.service
@@ -8,8 +8,9 @@ ExecStartPre=/bin/sh -c "docker inspect deis-database-data >/dev/null 2>&1 || do
 ExecStartPre=/bin/sh -c "IMAGE=`/run/deis/bin/get_image /deis/database` && docker history $IMAGE >/dev/null 2>&1 || docker pull $IMAGE"
 ExecStartPre=/bin/sh -c "docker inspect deis-database >/dev/null 2>&1 && docker rm -f deis-database >/dev/null 2>&1 || true"
 ExecStart=/bin/sh -c "IMAGE=`/run/deis/bin/get_image /deis/database` && docker run --name deis-database --rm --volumes-from=deis-database-data -p 5432:5432 -e EXTERNAL_PORT=5432 -e HOST=$COREOS_PRIVATE_IPV4 $IMAGE"
-ExecStopPost=-/usr/bin/docker exec deis-database sudo -u postgres envdir /etc/wal-e.d/env wal-e backup-push /var/lib/postgresql/9.3/main
-ExecStopPost=-/usr/bin/docker exec deis-database sudo service postgresql stop
+ExecStop=-/usr/bin/docker exec deis-database sudo -u postgres envdir /etc/wal-e.d/env wal-e backup-push /var/lib/postgresql/9.3/main
+ExecStop=-/usr/bin/docker exec deis-database sudo service postgresql stop
+ExecStop=-/usr/bin/docker stop deis-database
 Restart=on-failure
 RestartSec=5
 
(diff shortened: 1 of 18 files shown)


deisctl/units/deis-database.service
```diff
@@ -8,8 +8,9 @@ ExecStartPre=/bin/sh -c "docker inspect deis-database-data >/dev/null 2>&1 || do
 ExecStartPre=/bin/sh -c "IMAGE=`/run/deis/bin/get_image /deis/database` && docker history $IMAGE >/dev/null 2>&1 || docker pull $IMAGE"
 ExecStartPre=/bin/sh -c "docker inspect deis-database >/dev/null 2>&1 && docker rm -f deis-database >/dev/null 2>&1 || true"
 ExecStart=/bin/sh -c "IMAGE=`/run/deis/bin/get_image /deis/database` && docker run --name deis-database --rm --volumes-from=deis-database-data -p 5432:5432 -e EXTERNAL_PORT=5432 -e HOST=$COREOS_PRIVATE_IPV4 $IMAGE"
-ExecStopPost=-/usr/bin/docker exec deis-database sudo -u postgres envdir /etc/wal-e.d/env wal-e backup-push /var/lib/postgresql/9.3/main
-ExecStopPost=-/usr/bin/docker exec deis-database sudo service postgresql stop
+ExecStop=-/usr/bin/docker exec deis-database sudo -u postgres envdir /etc/wal-e.d/env wal-e backup-push /var/lib/postgresql/9.3/main
+ExecStop=-/usr/bin/docker exec deis-database sudo service postgresql stop
+ExecStop=/usr/bin/docker stop deis-database
```
krancour: Note that the removed lines were another potential issue.  If not for #4412 (which this PR addresses) this would mean the final backup is triggered _after_ the container stops.  `docker exec` would fail at that point.

Since this PR should ensure containers are stopped properly, this would bring the issue to a head, so it's proactively corrected here.

deisctl/units/deis-mesos-master.service
```diff
@@ -13,7 +13,7 @@ ExecStartPre=-/usr/bin/docker rm deis-mesos-master
 ExecStartPre=/bin/sh -c "docker inspect deis-mesos-master-data >/dev/null 2>&1 || docker run --name deis-mesos-master-data -v /tmp/mesos-master alpine:3.1 /bin/true"
 ExecStartPre=/bin/sh -c "IMAGE=`/run/deis/bin/get_image /deis/mesos-master` && docker history $IMAGE >/dev/null 2>&1 || docker pull $IMAGE"
 ExecStart=/usr/bin/sh -c "IMAGE=`/run/deis/bin/get_image /deis/mesos-master` && docker run --volumes-from=deis-mesos-master-data --name=deis-mesos-master --privileged --net=host -e HOST=$COREOS_PRIVATE_IPV4 $IMAGE"
-ExecStop=-/usr/bin/docker stop deis-mesos-master
+ExecStop=/usr/bin/docker stop deis-mesos-master
```
carmstrong: I think the intention of prefixing with `-` here is that if the container is stopped by the time the `ExecStop` runs, the entire unit won't fail. I think that's something we want. Without this change, if the container is stopped, the `ExecStop` will run and fail.

krancour: Ok.  Good call.  I was trying to follow the example in the Fleet documentation.  But I agree with what you're saying.  I'll fix that.

deisctl/units/deis-database.service
```diff
@@ -8,8 +8,9 @@ ExecStartPre=/bin/sh -c "docker inspect deis-database-data >/dev/null 2>&1 || do
 ExecStartPre=/bin/sh -c "IMAGE=`/run/deis/bin/get_image /deis/database` && docker history $IMAGE >/dev/null 2>&1 || docker pull $IMAGE"
 ExecStartPre=/bin/sh -c "docker inspect deis-database >/dev/null 2>&1 && docker rm -f deis-database >/dev/null 2>&1 || true"
 ExecStart=/bin/sh -c "IMAGE=`/run/deis/bin/get_image /deis/database` && docker run --name deis-database --rm --volumes-from=deis-database-data -p 5432:5432 -e EXTERNAL_PORT=5432 -e HOST=$COREOS_PRIVATE_IPV4 $IMAGE"
-ExecStopPost=-/usr/bin/docker exec deis-database sudo -u postgres envdir /etc/wal-e.d/env wal-e backup-push /var/lib/postgresql/9.3/main
-ExecStopPost=-/usr/bin/docker exec deis-database sudo service postgresql stop
+ExecStop=-/usr/bin/docker exec deis-database sudo -u postgres envdir /etc/wal-e.d/env wal-e backup-push /var/lib/postgresql/9.3/main
+ExecStop=-/usr/bin/docker exec deis-database sudo service postgresql stop
+ExecStop=-/usr/bin/docker stop deis-database
```
krancour: Note that the removed lines were another potential issue. If not for #4412 (which this PR addresses) this would mean the final backup is triggered after the container stops. docker exec would fail at that point.

Since this PR should ensure containers are stopped properly, this would bring the issue to a head, so it's proactively corrected here.

carmstrong:
Code LGTM.

mboersma:
LGTM. Tested with v1.10.0 and saw the problem, then verified that this fixes it.

PR #4417, opened 2015-09-05
[WIP] feat(builder): pull submodules if detected

Joshua-Anderson:
Why is this WIP?
- The builder uses the bare repo format, which is just the `.git` folder of a normal git repository. At the moment, builder uses `git archive` to get the files out of the bare repo, but you need to have the git repository in order to checkout submodules. The only way to get a working directory of a bare git repo is to clone it. I can clone the bare repository with the command in this script when sshing into the builder, but whenever the command runs in the gitrecieve script, I get a warning that I am cloning into a empty repository and checking out the revision fails. I'm not sure what is up with this. Is this some sort of anti-race condition that you cannot checkout a revision still being pushed to the repository? If so, are there any other ways to checkout submodules? Heroku seems like they might parse the gitmodules file themselves and pull the submodules themselves, but that doesn't seem like a great option if there are any alternatives. I'd love any ideas for other ways to checkout the repository that won't hit that strange error.
- We should probably add tests for this, maybe an example app with a submodule in the integration tests?

### DIFF
--- a/builder/rootfs/etc/confd/templates/builder
+++ b/builder/rootfs/etc/confd/templates/builder
@@ -64,10 +64,31 @@ mkdir -p $BUILD_DIR $CACHE_DIR
 TMP_DIR=$(mktemp -d -p $BUILD_DIR)
 
 cd $REPO_DIR
-git archive $GIT_SHA | tar -xmC $TMP_DIR
 
-# switch to app context
-cd $TMP_DIR
+# Check if .gitmodules file exists.
+git ls-files --with-tree $GIT_SHA --error-unmatch .gitmodules 1>/dev/null 2>/dev/null
+
+if [  $? -eq 0 ]; then
+    puts-step "Git submodules detected"
+
+    # HACK: git archive ignores submodules, so we have to clone
+    # and then -rf .git to mimic archive behavior
+
+    git clone --quiet --reference="$REPO_DIR" "file://$REPO_DIR" "$TMP_DIR"
+    cd $TMP_DIR
+    git checkout $GIT_SHA
+
+    # intialize and pull submodules
+    git submodule init --quiet
+    git submodule update --quiet --recursive
+
+    rm -rf .git
+else
+    git archive $GIT_SHA | tar -xmC $TMP_DIR
+
+    # switch to app context
+    cd $TMP_DIR
+fi
 
 USING_DOCKERFILE=false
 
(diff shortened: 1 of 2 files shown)


technosophos:
@sgoings Did your research on this turn up anything?

sgoings:
I didn't look too much, just looked at the difference between bare and normal git repos: http://stackoverflow.com/questions/7861184/what-is-the-difference-between-git-init-and-git-init-bare

As for our other options (not checking out entire histories)... Jenkins has a shallow checkout option which is this:

```
--depth <depth>
Create a shallow clone with a history truncated to the specified number of revisions.
```

As to @Joshua-Anderson's struggle with the bare repository... git clone also has the option --recursive* which explains the behavior he was running into

```
--recursive
--recurse-submodules
After the clone is created, initialize all submodules within, using their default settings. This is equivalent to running git submodule update --init --recursive immediately after the clone is finished. This option is ignored if the cloned repository does not have a worktree/checkout (i.e. if any of --no-checkout/-n, --bare, or --mirror is given)
```

IMO why not try the shallow clone and get rid of the bare repository handling...?

builder/rootfs/etc/confd/templates/builder
```diff
@@ -64,10 +64,31 @@ mkdir -p $BUILD_DIR $CACHE_DIR
 TMP_DIR=$(mktemp -d -p $BUILD_DIR)
 
 cd $REPO_DIR
-git archive $GIT_SHA | tar -xmC $TMP_DIR
 
-# switch to app context
-cd $TMP_DIR
+# Check if .gitmodules file exists.
+git ls-files --with-tree $GIT_SHA --error-unmatch .gitmodules 1>/dev/null 2>/dev/null
```
sgoings: Also note that you can do `&>/dev/null`

Joshua-Anderson:
@sgoings I'm not running submodule init on a bare repository. I attempting am cloning the bare repository on disk first. The problem I am hitting is that I am trying to clone the bare repo with `git clone` to checkout a working copy. Whenever I try to clone the bare repository I get the warning that the repository is empty and running `git checkout

kalbasit:
Accidentally closed the PR.

carmstrong:
@sgoings @technosophos What are we thinking about this? It seems to be something that users need.

mboersma:
Jenkins, test this please.

sgoings:
@carmstrong I think it's something that should be done, but someone just needs to take some time to take it across the line. It's not currently on the list for features to be delivered in 1.11... should we have someone work on it for 1.12 (if it's that important)?

_I regret commenting on this issue_ :wink:

Joshua-Anderson:
The builder is completely rewritten in V2, so I'll look into this there.

PR #4424, opened 2015-09-07
fix(client-go): error deserializing json

### LINKED ISSUE #4423: bug(client-go): error deserializing json
Running `deis config:list` in one application I see this error:
```
panic: interface conversion: interface is float64, not string

goroutine 1 [running]:
github.com/deis/deis/client/cmd.ConfigList(0x8206ee123, 0x12, 0x8206f1c00, 0x0, 0x0)
	/Users/matt/Projects/src/github.com/deis/deis/client/cmd/config.go:50 +0x7d7
github.com/deis/deis/client/parser.configList(0x8206e63b0, 0x1, 0x1, 0x0, 0x0)
	/Users/matt/Projects/src/github.com/deis/deis/client/parser/config.go:67 +0x27a
github.com/deis/deis/client/parser.Config(0x8206e63b0, 0x1, 0x1, 0x0, 0x0)
	/Users/matt/Projects/src/github.com/deis/deis/client/parser/config.go:24 +0x17d
main.Command(0x8206e63b0, 0x1, 0x1, 0x186840)
	/Users/matt/Projects/src/github.com/deis/deis/client/deis.go:89 +0x129b
main.main()
	/Users/matt/Projects/src/github.com/deis/deis/client/deis.go:18 +0x60
```

This application contains one particular configured value: `REDIS_PORT=6381`

aledbf:
closes #4423

### DIFF
--- a/client/cmd/config.go
+++ b/client/cmd/config.go
@@ -47,7 +47,7 @@ func ConfigList(appID string, oneLine bool) error {
 
 		// config.Values is type interface, so it needs to be converted to a string
 		for _, key := range keys {
-			configMap[key] = config.Values[key].(string)
+			configMap[key] = fmt.Sprintf("%v", config.Values[key])
 		}
 
 		fmt.Print(prettyprint.PrettyTabs(configMap, 6))
--- a/client/cmd/limits.go
+++ b/client/cmd/limits.go
@@ -30,7 +30,7 @@ func LimitsList(appID string) error {
 		memoryMap := make(map[string]string)
 
 		for key, value := range config.Memory {
-			memoryMap[key] = value.(string)
+			memoryMap[key] = fmt.Sprintf("%v", value)
 		}
 
 		fmt.Print(prettyprint.PrettyTabs(memoryMap, 5))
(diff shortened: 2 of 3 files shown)


Joshua-Anderson:
I'm :+1: on this, but it's odd that you're seeing this error.

[This](https://github.com/deis/deis/blob/master/controller/api/serializers.py#L173) shows that all variables should be converted to strings server side, so it shouldn't be possible to have it return a integer. Was this variable set on a old cluster that didn't have this conversion code?

aledbf:
@Joshua-Anderson yes (it a cluster running since 0.8)

Joshua-Anderson:
@aledbf I think this same bug also affects `tags` and `limits`.

aledbf:
@Joshua-Anderson done

bacongobbler:
code LGTM

mboersma:
@aledbf I can't reproduce the issue this is supposed to fix. I'm using v1.10. I did `deis config:set REDIS_PORT=6381` with both the old Python and new Go clients but couldn't make `deis config:list` produce this error. Any ideas?

mboersma:
LGTM. I tested this with the 1.10 `deis` CLI against a v1.8 Deis cluster after doing `config:set` and was able to reproduce the issue, then verify that this fixes it.

PR #4432, opened 2015-09-08
fix(controller): write app event logs via logspout

### LINKED ISSUE #3963: API / controller logging should be via logspout

krancour:
Fixes #3963 

This works by allowing `deis-controller` to write log messages on behalf of an application simply by prefixing them with `[app-name]:`.  `deis-logspout` has been enhanced to recognize such messages and treat the application as the source instead of the `deis-controller`.  `deis-controller` will be treated as the "pid" instead.

Given the following sequence of events:

```
$ deis create arrogant-kirk
$ git push deis master
$ deis logs
```

Old output would have looked like this:

```
2015-09-08T22:02:32UTC deis[api]: kent created initial release
2015-09-08T22:03:49UTC deis[api]: kent deployed 14fb753
2015-09-08T22:03:49UTC deis[api]: kent scaled containers web=1
2015-09-08T22:04:07UTC arrogant-kirk[web.1]: 2015/09/08 22:28:17 listening on 5000...
```

New output looks like this:

```
2015-09-08T22:02:32UTC arrogant-kirk[deis-controller]: kent created initial release
2015-09-08T22:03:49UTC arrogant-kirk[deis-controller]: kent deployed 14fb753
2015-09-08T22:03:49UTC arrogant-kirk[deis-controller]: kent scaled containers web=1
2015-09-08T22:04:07UTC arrogant-kirk[web.1]: 2015/09/08 22:04:07 listening on 5000...
```

IMO, the format differences are a positive change.  If logs from the former example were viewed in Papertrail or Splunk, for instance, there's nothing in the first three messages that correlates them to the `arrogant-kirk` application at all.  The logs from the latter example do not exhibit that problem.

### DIFF
--- a/controller/api/models.py
+++ b/controller/api/models.py
@@ -61,7 +61,7 @@ def close_db_connections(func, *args, **kwargs):
 def log_event(app, msg, level=logging.INFO):
     # controller needs to know which app this log comes from
     logger.log(level, "{}: {}".format(app.id, msg))
-    app.log(msg)
+    app.log(msg, level)
 
 
 def validate_base64(value):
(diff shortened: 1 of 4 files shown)


krancour:
Note to self: #4347 will require a similar update in the Deis module it implements for logspout v3.

krancour:
Related to #4083

mboersma:
LGTM.

> IMO, the format differences are a positive change.

I agree. I can see in the [test log](https://ci.deis.io/job/test-ec2/3062/consoleFull) this works well. Nice job writing tests and clean, commented code in both languages!

bacongobbler:
Code LGTM. The rationale behind `deis[api]` was to match heroku's log format of `heroku[api]` for any app-related events from their controller, but it's a cosmetic thing so I'm okay with this.

PR #4433, opened 2015-09-09
feat(controller): add CONTAINER_NAME env

vmakhaev:
The idea is to have docker container name inside container. We need it for monitoring of our apps.


deis-admin:

bacongobbler:
I think you may want to take a look at the [application hostname feature](http://docs.deis.io/en/latest/customizing_deis/controller_settings/#unit-hostname) which will give you what you are describing here. Does that help?

vmakhaev:
It solves it, thanks. Missed this part of documentation.
Btw, why we need to choose one of variants? Why not just add three different env variables? What is the idea?

bacongobbler:
some agents like New Relice identify instances by hostname rather than a Deis-specific envvar, so for backwards compatibility we chose to do so in that fashion.

vmakhaev:
Ok, got it. Thanks

PR #4436, opened 2015-09-09
contrib(aws): add aws support for rigger

sgoings:


contrib/aws/rigger/create.sh
```diff
@@ -0,0 +1,116 @@
...
+  # customize cloudformation.json to use m3.medium instances
+  cat > $DEIS_ROOT/contrib/aws/cloudformation.json <<EOF
+  [
+      {
+          "ParameterKey":     "KeyPair",
+          "ParameterValue":   "deis"
+      },
+      {
+          "ParameterKey":     "InstanceType",
+          "ParameterValue":   "m3.medium"
```
mboersma: Unless this is test-specific provisioning code, we should leave the default instance size of `m3.large`. We try to use smaller instances to save a bit of money when testing, but we recommend users provision with `m3.large` or better.

sgoings: Ok. I'll remove it from here for now.

mboersma:
LGTM.

carmstrong:
I'm thinking we should replace the existing scripts with `contrib`, etc. so that we maintain one set of the provision scripts, that can be called either by rigger or the user directly (and the docs would be updated as such).

Or is that not feasible?

sgoings:
> replace the existing scripts with contrib, etc.

@carmstrong not sure I follow. What scripts are you wanting to replace?

carmstrong:
> What scripts are you wanting to replace?

For AWS, `provision-aws-cluster.sh` and related. So a user could invoke the `create` script himself, or let rigger do it.

sgoings:
Oh I get what you're sayin'. Wholeheartedly agree... however... that'd require: https://github.com/deis/rigger/issues/15

carmstrong:
> I'll take a crack at it.

Ok. Your call on merging this now. Code LGTM.

kalbasit:
Accidentally closed the PR.

contrib/aws/create
```diff
@@ -0,0 +1,114 @@
...
+    aws ec2 import-key-pair --key-name deis \
+        --public-key-material file://${deis_auth_key}.pub \
+        --output text
+  fi
+}
+
+function aws-provision-cluster {
+  local stack_name="${1}"
+
+  # customize cloudformation.json to use m3.medium instances
```
Blystad: 1) This comment makes no sense with regards to what actually happens in the code
2) How are we supposed to customize instance type and other cloudformation parameters when this overwrites the cloudformation.json?

contrib/aws/create
```diff
@@ -0,0 +1,114 @@
...
+  cat > $DEIS_ROOT/contrib/aws/cloudformation.json <<EOF
+  [
+      {
+          "ParameterKey":     "KeyPair",
+          "ParameterValue":   "deis"
```
Blystad: Can we please avoid hardcoding the Keypair?

contrib/aws/create
```diff
@@ -0,0 +1,114 @@
...
+  aws-provision-cluster "${STACK_NAME}"
+)
+
+export ELB_DNS_NAME=$(aws-get-elb-dns-name "${STACK_NAME}")
+export ELB_NAME=$(aws-get-elb-name "${ELB_DNS_NAME}")
+export DEIS_TEST_DOMAIN="${STACK_TAG}.${DEIS_TEST_DOMAIN}"
+
+(
+  cd ${DEIS_ROOT}
+  aws-setup-route53 "${STACK_NAME}" "${DEIS_TEST_DOMAIN}"
```
Blystad: This code seems (based on variable names) to be related to test code, and not general provisioning. 

Can we add this as a configurable argument, and if it is left empty then nothing is done (for those not using Route53 for DNS).

sgoings:
@Blystad I'm in agreement with all of your requests (remove useless comment, customizing instance type, more configurability, and hardcoding keypair). I'm aware of these restrictions, but right now I essentially ported our provisioning logic for test usage to be under the rigger interface and tooling mainly to get better test coverage via CI across the providers that we support (AWS, Vagrant, Digital Ocean, Azure, GCE...).

Since this PR was already merged - could you create issue for these concerns (especially those re: configurability_ on the [rigger](https://github.com/deis/rigger) project?

Blystad:
Alright, will do :)

PR #4438, opened 2015-09-09
chore(*): update CoreOS to 766.3.0

mboersma:
Gets Deis back on the stable channel and reverts #3992. I tested and don't see the issue about multiple simultaneous docker pulls now.


carmstrong:
Does this work just fine with the Docker-in-builder staying at 1.5? And we can bump AWS version by just changing `contrib/utils.sh`? That is awesome!

mboersma:
@carmstrong yes, we have to leave Docker-inside-builder at 1.5.0 for now. It's a bit confusing to have Docker 1.7.1 on the hosts, but there's no conflict.

sgoings:
LGTM. :ship:

mboersma:
> Do these leading hashes not change ever?

I guess not; I copied it afresh from the current CoreOS Azure provisioning doc.

carmstrong:
> Do these leading hashes not change ever?

They haven't changed on us yet, although that's no guarantee.

bacongobbler:
code LGTM

laurrentt:
@mboersma As far as you know, could it work with Deis 1.10 ?

mboersma:
@laurrentt yes, CoreOS 766.3.0 works fine with Deis v1.10--I've tested that configuration many times now.

laurrentt:
@mboersma Thanks !

PR #4444, opened 2015-09-10
ref(user-data): use default etcd2 and flannel services

mboersma:
This refactors and hopefully simplifies the thorny cloud-config Deis has in user-data.example. It uses the native CoreOS flanneld.service and etcd2.service--currently v2.1.2, same as Deis v1.10--and minimizes the differences with a [stock CoreOS user-data](https://github.com/coreos/coreos-vagrant/blob/master/user-data.sample). Also removes `etcd` timing tweaks which seemed to be unneeded based on my testing for v1.10--see #4384.

Also moves Docker `--insecure-registry` config into a unit drop-in and removes the cruft from the graceful-shutdown script. I've tested this against vagrant with Deis v1.10, Deis master, and Deis v1.8 upgraded to v1.10 + this PR. Also tested k8s, v1.9.1 upgrade, and the graceful-shutdown.sh script.

Closes #4384.
Closes #4409.

### DIFF
--- a/contrib/coreos/user-data.example
+++ b/contrib/coreos/user-data.example
@@ -1,161 +1,101 @@
 #cloud-config
----
+
 coreos:
+  etcd2:
+    # generate a new token for each unique cluster from https://discovery.etcd.io/new
+    discovery: #DISCOVERY_URL
+    # multi-region and multi-cloud deployments need to use $public_ipv4
+    advertise-client-urls: http://$private_ipv4:2379
+    initial-advertise-peer-urls: http://$private_ipv4:2380
+    # listen on both the official ports and the legacy ports
+    # legacy ports can be omitted if your application doesn't depend on them
+    listen-client-urls: http://0.0.0.0:2379,http://0.0.0.0:4001
+    listen-peer-urls: http://$private_ipv4:2380,http://$private_ipv4:7001
+    data-dir: /var/lib/etcd2
   fleet:
-    # We have to set the public_ip here so this works on Vagrant -- otherwise, Vagrant VMs
-    # will all publish the same private IP. This is harmless for cloud providers.
-    public-ip: $private_ipv4
-    # allow etcd to slow down at times
-    etcd_request_timeout: 3.0
+    public-ip: $public_ipv4
     metadata: controlPlane=true,dataPlane=true,routerMesh=true
+  update:
+    reboot-strategy: "off"
   units:
-  - name: etcd.service
-    command: start
-    content: |
-      [Unit]
-      Description=etcd2 container
-      Requires=early-docker.service
-      After=early-docker.service
-      Before=early-docker.target
+    - name: etcd2.service
+      command: start
+    - name: fleet.service
+      command: start
+    - name: docker-tcp.socket
+      command: start
+      enable: true
+      content: |
+        [Unit]
+        Description=Docker Socket for the API
 
-      [Service]
-      User=etcd
-      PermissionsStartOnly=true
-      Restart=always
-      RestartSec=10s
-      LimitNOFILE=40000
-      EnvironmentFile=/etc/environment
-      Environment="ETCD_IMAGE=quay.io/coreos/etcd:v2.1.2"
-      Environment="ETCD_ELECTION_TIMEOUT=2000"
-      Environment="ETCD_HEARTBEAT_INTERVAL=400"
-      Environment="ETCD_HOST_DATA_DIR=/var/lib/etcd2"
-      Environment="ETCD_NAME=%m"
-      Environment="DOCKER_HOST=unix:///var/run/early-docker.sock"
-      ExecStartPre=/bin/sh -c "docker history $ETCD_IMAGE >/dev/null 2>&1 || docker pull $ETCD_IMAGE"
-      ExecStartPre=/bin/sh -c "docker inspect $ETCD_NAME >/dev/null 2>&1 && docker rm -f $ETCD_NAME || true"
-      ExecStart=/usr/bin/docker run --net=host --rm \
-        --volume=${ETCD_HOST_DATA_DIR}:/var/lib/etcd2 \
-        --volume=/usr/share/ca-certificates:/etc/ssl/certs:ro \
-        -p 4001:4001 -p 2380:2380 -p 2379:2379 -p 7001:7001 \
-        --name ${ETCD_NAME} \
-        ${ETCD_IMAGE} \
-        -name ${ETCD_NAME} \
-        -data-dir /var/lib/etcd2 \
-        -advertise-client-urls http://${COREOS_PRIVATE_IPV4}:2379,http://${COREOS_PRIVATE_IPV4}:4001 \
-        -listen-client-urls http://0.0.0.0:2379,http://0.0.0.0:4001 \
-        -initial-advertise-peer-urls http://${COREOS_PRIVATE_IPV4}:2380,http://${COREOS_PRIVATE_IPV4}:7001 \
-        -listen-peer-urls http://0.0.0.0:2380,http://0.0.0.0:7001 \
-        --heartbeat-interval ${ETCD_HEARTBEAT_INTERVAL} \
-        --election-timeout ${ETCD_ELECTION_TIMEOUT} \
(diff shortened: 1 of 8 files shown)


deisctl/units/deis-logspout.service
```diff
@@ -1,7 +1,7 @@
 [Unit]
 Description=deis-logspout
-Requires=docker.socket etcd.service
-After=docker.socket etcd.service
+Wants=docker.socket etcd2.service
+After=docker.socket etcd2.service etcd.service
```
carmstrong: Should this still have `etcd.service`?

deisctl/units/deis-logspout.service
```diff
@@ -1,7 +1,7 @@
 [Unit]
 Description=deis-logspout
-Requires=docker.socket etcd.service
-After=docker.socket etcd.service
+Wants=docker.socket etcd2.service
```
carmstrong: Do we really want to change this to `Wants`? The logspout component won't function without `docker.socket`, so I think it'll need to remain `Requires`, no?

deisctl/units/deis-logspout.service
mboersma: Yes, because `After=` is not binding. If `etcd.service` is disabled then systemd won't wait for it before launching logspout. CoreOS uses the same approach to make their own services backward-compatible with v1 etcd:
```console
core@deis-01 ~ $ systemctl cat fleet   
# /usr/lib64/systemd/system/fleet.service
[Unit]
Description=fleet daemon

After=etcd.service
After=etcd2.service

```

carmstrong:
Have we considered the upgrade path for someone using our etcd-in-a-box to this version?

mboersma:
> Have we considered the upgrade path for someone using our etcd-in-a-box to this version?

Yes I did, and I just tested upgrade from v1.9.1 on AWS. Everything is still happy with containerized etcd.service as well.

contrib/coreos/user-data.example
```diff
@@ -1,161 +1,108 @@
...
   fleet:
-    # We have to set the public_ip here so this works on Vagrant -- otherwise, Vagrant VMs
-    # will all publish the same private IP. This is harmless for cloud providers.
-    public-ip: $private_ipv4
-    # allow etcd to slow down at times
-    etcd_request_timeout: 3.0
+    public-ip: $public_ipv4
     metadata: controlPlane=true,dataPlane=true,routerMesh=true
+  flannel:
+    interface: $public_ipv4
```
carmstrong: Do we want the public or private interface here?

mboersma: We didn't specify this when using our own unit for flannel, and I'm using the apparent CoreOS default value here. But we had changed `advertise-client-urls` for etcd2 to $private_ipv4, so perhaps we should restrict this as well?

carmstrong: That was my instinct as well. @smothiki do you know what this should be for k8s?

aledbf: This should be specified only for vagrant. This needs an update too https://github.com/deis/deis/blob/master/Vagrantfile#L84

mboersma:
@aledbf good catch--I updated the Vagrantfile to splice in the flannel $public_ipv4 as CoreOS does, so I hope that addresses your comments.

carmstrong:
ping @smothiki for the Vagrantfile review - I know that was changed for k8s so I want to make sure team upstream is aware of the change.

smothiki:
clarified with @mboersma regarding starting the flannel service. LGTM

contrib/coreos/user-data.example
```diff
@@ -1,161 +1,101 @@
...
-        --volume=/run/flannel:/run/flannel \
-        --env=NOTIFY_SOCKET=/run/flannel/sd.sock \
-        --env-file=/run/flannel/options.env \
-        --volume=/usr/share/ca-certificates:/etc/ssl/certs:ro \
-        quay.io/coreos/flannel:${FLANNEL_VER} /opt/bin/flanneld --ip-masq=true
+        [Install]
+        WantedBy=sockets.target
+    - name: update-engine.service
+      command: stop
+      enable: false
```
carmstrong: We should also have `mask: true` here since we're removing the service which previous had masked it: https://coreos.com/os/docs/latest/cloud-config.html

mboersma: > We should also have `mask: true` here...

I balked at doing that because [CoreOS' description](https://coreos.com/os/docs/latest/cloud-config.html) of what that does makes it sound more destructive than just calling `systemctl mask`:
> mask: Whether to mask the unit file by symlinking it to /dev/null (analogous to systemctl mask &lt;name&gt;). Note that unlike systemctl mask, this will destructively remove any existing unit file located at /etc/systemd/system/<unit>, to ensure that the mask succeeds. The default value is false.

I figured stopping and disabling the service, as well as changing `reboot-strategy=off` above, is sufficient to stop unrequested updating without making the update-engine.service unrecoverable for someone who really wants to enable it.

mboersma:
For further clarification, the default CoreOS `flanneld.service` does already include `--ip-masq=true`, which is what @smothiki was specifically interested in.

contrib/coreos/user-data.example
carmstrong: I suppose that should work as long as the cloud-config is rerun on each boot (are we 100% sure it does on all platforms?). Without masking the service, it will start at boot time until/unless the cloud-config is processed and then the service is stopped.

Also, I'm pretty sure `reboot-strategy=off` doesn't prevent anything from being updated - it just prevents the automatic reboot *after* the OS has applied its update. So on its own, it doesn't do much for us, but in concert with this other change it makes sense. You probably are aware, but just wanted to point it out here.

mboersma: > that should work as long as the cloud-config is rerun on each boot

It should be, according to CoreOS' docs. I'll do some reboot tests just to verify.

You're right about `reboot-strategy=off`, but it seemed like an additional safeguard in this case since we can't literally prevent `update_engine` from starting. Then if somehow it were able to sneak through and download an update before we stopped it, at least the node wouldn't reboot.

mboersma: @carmstrong I verified on VirtualBox, DO, and AWS that rebooting a node with this cloud-config successfully stops and disables update-engine.service after reboots:
```console
core@ip-10-21-2-34 ~ $ systemctl status update-engine
● update-engine.service - Update Engine
   Loaded: loaded (/usr/lib64/systemd/system/update-engine.service; disabled; vendor preset: disabled)
   Active: failed (Result: exit-code) since Mon 2015-09-14 21:23:42 UTC; 1min 6s ago
 Main PID: 551 (code=exited, status=1/FAILURE)

Sep 14 21:23:40 ip-10-21-2-34.us-west-2.compute.internal update_engine[551]: [0914/212340:INFO:main.cc(155)] CoreOS Update...ng
Sep 14 21:23:40 ip-10-21-2-34.us-west-2.compute.internal update_engine[551]: [0914/212340:INFO:update_check_scheduler.cc(8...8s
Sep 14 21:23:40 ip-10-21-2-34.us-west-2.compute.internal systemd[1]: Started Update Engine.
Sep 14 21:23:42 ip-10-21-2-34.us-west-2.compute.internal systemd[1]: Stopping Update Engine...
Sep 14 21:23:42 ip-10-21-2-34.us-west-2.compute.internal systemd[1]: update-engine.service: Main process exited, code=exi...URE
Sep 14 21:23:42 ip-10-21-2-34.us-west-2.compute.internal systemd[1]: Stopped Update Engine.
Sep 14 21:23:42 ip-10-21-2-34.us-west-2.compute.internal systemd[1]: update-engine.service: Unit entered failed state.
Sep 14 21:23:42 ip-10-21-2-34.us-west-2.compute.internal systemd[1]: update-engine.service: Failed with result 'exit-code'.
Hint: Some lines were ellipsized, use -l to show in full.
```
(If anyone can figure out a way to prevent the service from starting in the first place, my hat is off to you. Perhaps it's about time to just leave it update-engine.service enabled and then fix what breaks...)

kalbasit:
Accidentally closed the PR.

aledbf:
Code LGTM

PR #4484, opened 2015-09-15
feat(router): deny paths cluster-wide or per app

### LINKED ISSUE #4477: Deny access to a configurable list of paths applicable for all apps.
I have a use case where, for security reasons, I want to deny access to specific paths for all apps. I propose going as simple as possible for this (list of paths separated by a comma in etcd + range {location} in nginx)

kalbasit:
fixes #4477

### DIFF
--- a/docs/customizing_deis/router_settings.rst
+++ b/docs/customizing_deis/router_settings.rst
@@ -35,6 +35,7 @@ setting                                      description
 /deis/builder/host                           host of the builder component (set by builder)
 /deis/builder/port                           port of the builder component (set by builder)
 /deis/config/\*/deis_whitelist               comma separated list of IPs (or CIDR) allowed to connect to the application containers (set by controller) Example: "0.0.0.0:some_optional_label,10.0.0.0/8"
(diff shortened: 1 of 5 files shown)


benwilber:
I think the syntax of the denied paths should be validated otherwise the router wont reload.

example:

```
$ deis config:set DENY_PATHS=/varz/{foo}
```

where `{foo}` is a really common templating placeholder that could accidentally be written literally, which would cause nginx to fail all subsequent syntax checks (and would be a pain to track down and fix).

mboersma:
I'm still unsure whether denying specific application paths for all apps is a proper responsibility of the platform: i.e., Deis. I see the need for this, but I think it would be more appropriate for the apps themselves to hide their management/auditing/secret URL paths.

benwilber:
I'm tending to agree with @mboersma that this probably shouldn't be the responsibility of Deis itself.

bacongobbler:
+1 on @mboersma's pitch as well. These features should be implemented in the app itself. If we did go forward with this, users may want to customize these routes with more than just a 403 response, so it's not something we can easily support without a lot of spaghetti code (as if the router doesn't have enough spaghetti code).

kalbasit:
given the general tone is to not include this in the router I'll close the pull request. We can revisit this if we get more demand for it.

PR #4486, opened 2015-09-15
ref(cache): remove deis-cache component

mboersma:
The optional deis-cache component has effectively been deprecated for many releases, and the idea of removing it often comes up. This PR does that but retains the `deisctl` logic and the registry templating, so existing uses of deis-cache or an alternative Redis cache will still function.

The motivation here is only that simpler == better. Please speak up if you use deis-cache and this would cause you problems.


aledbf:
Code LGTM

PR #4493, opened 2015-09-16
fix(client): suggest help when git remote creation fails

### LINKED ISSUE #4455: Creating an app when deis remote already exists
'deis create' when a deis remote already exists bubbles up the git error that says the remote called deis already exists. 
It'd be nice to have an more clear error/warning like “Existing deis remote already exists. The app was created, but a git remote couldn’t be created. Remove the git remote and add the new one with <command>"

mboersma:
```console
$ deis create
Creating Application... done, created madras-yachting
fatal: remote deis already exists.
To replace the existing git remote entry, run:
  git remote rename deis deis.old && deis git:remote -a madras-yachting
Error: exit status 128
```

Closes #4455.

cc: @michelleN, let me know if this is what you had in mind.

### DIFF
--- a/client/cmd/apps.go
+++ b/client/cmd/apps.go
@@ -46,7 +46,13 @@ func AppCreate(id string, buildpack string, remote string, noRemote bool) error 
 	}
 
 	if !noRemote {
-		return git.CreateRemote(c.ControllerURL.Host, remote, app.ID)
+		if err = git.CreateRemote(c.ControllerURL.Host, remote, app.ID); err != nil {
+			if err.Error() == "exit status 128" {
+				fmt.Println("To replace the existing git remote entry, run:")
+				fmt.Printf("  git remote rename deis deis.old && deis git:remote -a %s\n", app.ID)
+			}
+			return err
+		}
 	}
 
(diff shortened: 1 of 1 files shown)


client/cmd/apps.go
```diff
@@ -46,7 +46,13 @@ func AppCreate(id string, buildpack string, remote string, noRemote bool) error
 	}
 
 	if !noRemote {
-		return git.CreateRemote(c.ControllerURL.Host, remote, app.ID)
+		if err = git.CreateRemote(c.ControllerURL.Host, remote, app.ID); err != nil {
+			if err.Error() == "exit status 128" {
+				fmt.Println("To replace the existing git remote entry, run:")
```
carmstrong: Should we also say something like "A remote for Deis already exists"?

mboersma: Git itself said that immediately before we print this: `fatal: remote deis already exists.`

carmstrong: :+1:

michelleN:
Thanks @mboersma that's perfect :) :+1:

PR #4497, opened 2015-09-16
fix(deisctl): fix refresh-units as cache is now missing

iancoffey:
refresh-units seems broken on master.

https://gist.github.com/iancoffey/e44456474718c6bbd91d


iancoffey:
yolo

PR #4435, opened 2015-09-09
[WIP] feat(logger): implement in memory ring buffer

krancour:

Issue #4280, opened 2015-08-15
Log size limit settings?


ineu:
Log files stored in /var/lib/deis/store/logs are not limited in size. I've recently had a problem with Ceph deadlock and while investigating, noticed that one of my log files was 18GB large. Looks like there are two improvements that can be made here:

1. There should be configuration option for logger to specify maximum number of log entries (or maximum file size), just like Heroku's 1500 lines limit. This is to avoid huge files on disk. Another option here is some kind of logrotate.

2. There should be configuration option or hardcoded value for maximum logs rate from an app, bytes/second. This is to avoid deadlocks in Ceph MDS when application with multiple containers on several hosts generates high amount of logs.

Please let me know if it makes sense.

krancour:
A dupe of this issue was recently opened and breathed new life into this topic.

I think, at this point in time, we can safely rule out this issue being addressed.  Development efforts are being concentrated on v2.  v1.x is (mostly) receiving just bug fixes now.  Also, the logger was massively overhauled just recently and with that overhaul came the possibility (which we highly recommend) of storing logs in memory.  I'm going to elaborate on that because it's the detail that nearly renders this issue moot.

The logger component was never _intended_ for long-term archiving of logs.  It's a centralized point of aggregation where developers can query _recent_ application logs via CLI -> controller -> logger.  It's only useful for troubleshooting an issue in real time, and that's all it was ever meant for.

If one wants to aggregate and index logs for the sake of audit, analytics, error detection, etc. there are _many_ existing solutions such as Papertrail, Loggly, and Splunk.  Deis-logger doesn't want to compete with those products-- it would rather just feed logs _to_ them.

This being the case, opting for in-memory storage of logs and "draining" those logs to an external archive is an optimal solution that can be achieved today-- and that strategy is likely to carry forward to v2, where file-based log storage _might_ not even be an option.

Hopefully in-memory log storage + drain strikes operators as a reasonable option.  If the community at large doesn't agree, we can re-open this at a later time.

PR #4502, opened 2015-09-17
chore(logger): refactor-- also includes new features

### LINKED ISSUE #4000: Stateless logger

krancour:
Fixes #4000; replaces #4435 

__TL;DR__: A complete rewrite of `deis-logger` aimed at easier extensibility and maintenance.  Some bug fixes and new features are included as well, as are many new unit tests.

Motivations
=========

* #4000 called for the implementation of an in-memory ring buffer as an alternative to file-based log storage (which doesn't work out so well for those running stateless / without Ceph).
* The possibility that in-memory storage is used as an alternative to file-based storage means `deis-controller` can no longer mount a volume in common with `deis-logger` to get logs.  `deis-logger` needed to evolve to serve logs to `deis-controller` over HTTP.
* `deis-logger` was not originally written to _read_ logs; only to write them.
* The implementation offered by the community in #4083 was welcome, but fell short in a few keys areas:
  * Introduced the concept of a "handler" for storing and reading logs, but clashed with `deis-logger`'s existing notion of what a "handler" is.  (Overloaded term.)
  * Introduced some undesired complexity-- e.g. `deis-controller` only read logs from `deis-logger` over HTTP if the in-memory storage option was in use.  (`deis-controller` shouldn't have to know _how_ logs are being stored and shouldn't have to adjust its behavior based on that.  Logs could/should _always_ be retrieved from the `deis-logger` over HTTP.)
  * Only checked the "handler" type to use at startup; couldn't dynamically reconfigure itself.
  * Didn't adequately address failure cases.
* Despite those few problems, #4083 __actually worked great!__  Big shout out to @rvadim and @aj-may for their efforts on this.  Unfortunately, since the implementation was layered on top of existing code that was in rough shape to begin with, this didn't bode well for future maintenance and extension.  `deis-logger` as it stood prior to this PR:
  * Was based on a library that hasn't been updated in three years.
  * Had lots of dead, untested, and hard-to-follow code.
  * Existing support for writing logs to files and draining logs was a hack to begin with.
  * Only supported UDP-based drains.  (TCP needed for use with TLS-- future feature.)

Issues such as #4280 are also considerations that informed the approach.

Weighing the myriad concerns, a scorched earth re-design seemed the way to go.

Design
=====

Here's a brief overview of the design.  (I'll see about adding a use case diagram tomorrow to help make this easier for reviewers.)

Logger
---------

`main.go` handles startup including flag parsing.  It supports all the same flags as the old implementation for complete backwards compatibility.  It initializes and starts four components:

1. A `syslogish.Server`: reads and writes logs.  For writing, it implements a loop that receives logs (1 per packet) over UDP.  A non-blocking write immediately puts these into an internal queue.  This is _exactly_ the same as before.  Another loop ranges over the queue and processes log messages.  It delegates their storage to an underlying `storage.Adapter` (file-based or in-memory) and delegates draining to an underlying `drain.LogDrain` (UDP-based or TCP-based :new:; SSL not supported yet).  For reading, reads are also delegated to the underlying `storage.Adapter`.  A key improvement is that storage write failures (rare) and failures to dial the drain (could happen if Papertrail went down, for instance) are handled _silently_.  This sounds bad, but it helps avoid a logging death spiral that happens when the logger can't handle its own error messages.  (All other failures that won't start the death spiral _are_ logged, of course.)

2. A `weblog.Server`: handles HTTP `GET` and `DELETE` requests.  Delegates all actual work to the `syslogish.Server`.

3. A `Configurer`: uses a ticker and watches etcd for changes to `/deis/logs/storageAdapterType` and `/deis/logs/drain`.  When changes are detected, it uses factories that take the values of those keys to construct appropriate `storage.Adapter` and `drain.LogDrain` instances, respectively.  These are then injected into the `syslogish.Server`.

4. A `Publisher`: uses a ticker and periodically writes the logger's host and port to etcd using the `/deis/logs/host` and `/deis/logs/port`, respectively.

Apart from these four primary components, there are also two implementations each of `storage.Adapter` and `drain.LogDrain`.  These are unit tested rather thoroughly.  (Which is the reason this PR is a net addition of lines to the code base.)

One other notable improvement is that complex cleanup logic has been removed.  It is not needed AFAICT.

The Fleet unit for `deis-logger` no longer `Requires` `deis-store-volume`, but merely `Wants` it.  This allows it to start up in the absence of Ceph.

Controller
-------------

The modifications to `deis-controller` are more modest.  confd is used to look up the `deis-logger` host now.  Combining that with a hard-coded port `8088` permits discovery of the `deis-logger` component's new `weblog.Server`.  The RESTful interface it exposes allows the controller to read and destroy logs.  Failure cases are all accounted for and tests have been added to simulate the failure scenarios and assert that they are handled properly.

The Fleet unit for `deis-controller` no longer expresses a dependency on `deis-store-volume`, nor does it mount the volumes from that container on the `deis-controller` container.  (Let's hear it for decoupling.)

deisctl
---------

`deisctl` has been modified to now install `deis-logger` with the stateless platform (previously omitted) and to also start that component during stateless platform start or restart.


krancour:
@rvadim and @aj-may, I would highly value your feedback on this.

krancour:
refactored logger

carmstrong:
:100: on this write-up

carmstrong:
@krancour Unless I'm mistaken, this also means we can drop deis-store-volume and store-metadata since we're not using them anymore! We'll need to figure out an upgrade path, though.

krancour:
@carmstrong logger will still use `deis-store-volume` if using file-based log storage.

carmstrong:
> logger will still use deis-store-volume if using file-based log storage.

Ah, gotcha. Ok. We can remove that for Deis v2. :+1:

logger/configurer/configurer.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+}
+
+func (c *Configurer) configure() {
+	for {
+		<-c.ticker.C
+		c.manageStorageAdapter()
+		c.manageDrain()
+	}
+}
```
kalbasit: How do you stop a Configurer?

On a side note, I love the name :)

logger/main.go
```diff
@@ -2,102 +2,106 @@ package main
...
+	flag.StringVar(&etcdHost, "publish-host", getopt("HOST", "127.0.0.1"),
+		"service discovery hostname")
+	flag.StringVar(&etcdPort, "publish-port", getopt("ETCD_PORT", "4001"), "service discovery port")
+	flag.StringVar(&etcdPath, "publish-path", getopt("ETCD_PATH", "/deis/logs"),
+		"path to publish host/port information")
+	flag.IntVar(&configInterval, "config-interval", 10, "config interval in seconds")
 	flag.IntVar(&publishInterval, "publish-interval", 10, "publish interval in seconds")
-	flag.StringVar(&publishPath, "publish-path", getopt("ETCD_PATH", "/deis/logs"), "path to publish host/port information")
-	flag.StringVar(&publishPort, "publish-port", getopt("ETCD_PORT", "4001"), "service discovery port")
 	flag.IntVar(&publishTTL, "publish-ttl", publishInterval*2, "publish TTL in seconds")
```
kalbasit: Since we are refactoring this, can you please move all of these to var, this would mean less work for `init()`

logger/configurer/configurer.go
krancour: You don't stop it.  Once it starts, it runs until the program terminates.

logger/main.go
```diff
@@ -2,102 +2,106 @@ package main
...
 	logPort         int
-	drainURI        string
 	enablePublish   bool
-	publishHost     string
-	publishPath     string
-	publishPort     string
+	webAddr         string
+	webPort         int
+	etcdHost        string
+	etcdPort        string
```
kalbasit: Why is etcdPort a `string` and not `int`?

logger/main.go
```diff
@@ -2,102 +2,106 @@ package main
...
+	etcdPath        string
+	configInterval  int
 	publishInterval int
 	publishTTL      int
 )
 
 func init() {
 	flag.StringVar(&logAddr, "log-addr", "0.0.0.0", "bind address for the logger")
+	flag.StringVar(&logHost, "log-host", getopt("HOST", "127.0.0.1"),
```
kalbasit: If you use https://github.com/namsral/flag you won't need `getopt` and all of the flags can also be environment variables.

logger/syslogish/server.go
```diff
@@ -0,0 +1,148 @@
...
+
+// SetStorageAdapter permits a server's underlying storage.Adapter to be reconfigured (replaced)
+// at runtime.
+func (s *Server) SetStorageAdapter(storageAdapter storage.Adapter) {
+	s.storageAdapter = storageAdapter
+}
+
+// SetDrain permits a server's underlying drain.LogDrain to be reconfigured (replaced) at runtime.
+func (s *Server) SetDrain(drain drain.LogDrain) {
+	s.drain = drain
```
kalbasit: There's a possible race here. The goroutine running `configurer.manageDrain()` can be call this function which writes to the `s.drain` while the second goroutine running `syslogish.process()` can be accessing the Send function of `s.drain`.

A [`sync.RWMutex`](http://godoc.org/sync#RWMutex) would be ideal for this situation

logger/syslogish/server.go
```diff
@@ -0,0 +1,148 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return &Server{conn: c, queue: make(chan string, 5)}, nil
+}
+
+// SetStorageAdapter permits a server's underlying storage.Adapter to be reconfigured (replaced)
+// at runtime.
+func (s *Server) SetStorageAdapter(storageAdapter storage.Adapter) {
+	s.storageAdapter = storageAdapter
```
kalbasit: Same as my comment for SetDrain. Use a different mutex though as they represent two different objects.

kalbasit:
I haven't had time to review this in full, I'll finish my review tonight.

logger/main.go
krancour: Not sure I follow.  Can you elaborate?  I should note that I really haven't changed the flag parsing from what was there before.  For backwards compatibility, we should continue to support all the same flags.

logger/main.go
krancour: Because if the `publish-port` flag is not provided, the default value comes from an environment variable.  Environment variables are strings and it is difficult to do `.Atoi()` in line where the flag parsing is defined since that function returns two values.  It bugs me a little too, but this behavior was lifted wholesale from the old implementation.

logger/main.go
krancour: Taking a hint from the old implementation and from other Deis components and trying to avoid external dependencies wherever I can.  (The etcd-go library is the only external dependency for this component.)

logger/syslogish/server.go
krancour: I don't think a race can occur here.  Each time the `process()` function reaches the point where it might send a message to the drain, it's invoking `.Send()` on whichever drain its pointer references at that moment.  Once the instruction pointer is inside the `Send()` function, everything should be fine, regardless of what other goroutines are doing to the server's drain pointer.  The reason is because all the operations that occur within the `Send()` function are executed _not_ against whichever drain the server is pointing to, but rather, they are executed against the drain denoted by the receiver portion of the function signature.

logger/syslogish/server.go
krancour: Same explanation as above.

krancour:
@kalbasit thanks!  I appreciate having your eyes on this.  You're already asking some great questions.  Feel free to fight me on any of it if I am wrong.  :smile:

benwilber:
This looks really cool!  I don't have a ton of visibility into the deis logger subsystem (I'm still new to this project), but I'm wondering if there was any consideration to basing the syslog messaging components on a dedicated syslog server, something like `syslog-ng`?  I'm sure there are good reasons for not using syslog-ng as the logging server (in the same way that nginx is used as the routing layer), but since the deis logger appears to reimplement a number of the core features of `syslog-ng`, I'm wondering what those reasons are.  This is just for my own education on the architecture/design decisions.

krancour:
@benwilber I wasn't around when the decision was initially made, but I'd wager a guess that it had a lot to do with how easy it was to hack on this [exising go-based syslog server](https://github.com/ziutek/syslog) to add in features like custom message parsing and publishing of logger host:port so that all the logspouts can discover it.

logger/main.go
kalbasit: I do know that you didn't change but since it's a rewrite you might as well. Basically when you define flags as (just one flag here but you get the idea)

```
var (
    publishPath = flag.StringVar(&publishPath, "publish-path", getopt("ETCD_PATH", "/deis/logs"), "path to publish host/port information")
)
```

you can access the value using `*publishPath` which is less pretty but the compiler already does the work. When you do it the way it's done now (through `init`) the binary sets up the memory on boot. It's not much of a performance improvement but it's something.

I leave it to you to decide.

logger/main.go
kalbasit: I see, that's fine then.

logger/main.go
kalbasit: could be avoided with the flag replacement but we've already established no external dependencies :)

logger/syslogish/server.go
kalbasit: actually that's not true. I know you don't assign a `nil` but for the sake of the argument say you do. And let's assume that `process` is at the line `if s.drain != nil {` and it was evaluated but right after `s.drain` has become `nil`, now you are calling `Send()` on a `nil`. As a rule of thumb, if you have a variable that is being accessed (read/write) from more than one goroutine than you **must** protect it. Otherwise the behavior is going to be unpredictable and might lead to bugs that are hard to track.

krancour: You're right.  I forgot that the nil check was there.  The drain _could_ be set to `nil` between the time it's checked and seen _not_ to be `nil` and the time the call to `d.Send()` happens.  Nice catch.  (And this really _can_ happen, because `nil` is a "valid" drain-- meaning "no drain.")  I'll fix this.  Same for the call to the storage adapter.

logger/main.go
krancour: Ah.  I do not think that will work.  Here is the function signature for `flag.StringVar` (for example):

```
func StringVar(p *string, name string, value string, usage string)
```

It doesn't return anything, so the example you gave shouldn't compile.

logger/syslogish/server.go
krancour: As mentioned in the convo on the drain up above, yes... there's a possible race condition here.  Nice catch.

benwilber:
Nothing against your implementation, (it's really good), but the the fact the you felt the need to make a diagram of what your code does is a pretty big indicator that it's too complicated.

krancour:
@kalbasit I've fixed the possible race conditions you pointed out.  Thanks again!  I'd appreciate another spot-check if you've got the time.

krancour:
The problem that was interfering with controller's functional tests should be fixed now.

logger/main.go
kalbasit: You would use `flag.String`. I've corrected my comment above, apologies for that.

controller/api/models.py
```diff
@@ -517,12 +519,24 @@ def _default_scale(self, user, release):
         self.scale(user, structure)
 
     def logs(self, log_lines=str(settings.LOG_LINES)):
-        """Return aggregated log data for this application."""
-        path = os.path.join(settings.DEIS_LOG_DIR, self.id + '.log')
-        if not os.path.exists(path):
+        try:
+            url = "http://{}:{}/{}?log_lines={}".format(settings.LOGGER_HOST, 8088, self.id,
```
kalbasit: can we turn `8088` into a constant?

docs/managing_deis/running-deis-without-ceph.rst
```diff
@@ -100,18 +90,18 @@ Deploy an AWS cluster
...
+archival.
 
 .. code-block:: console
 
-    $ HOST=logs.somewhere.com
-    $ PORT=98765
-    $ PROTOCOL=udp # Supported protocols are udp and tcp
-    $ deisctl config logs set host=${HOST} port=${PORT} protocol=${PROTOCOL}
+    $ STORAGE_ADAPTER=memory
+    $ DRAIN=udp://logs.somewhere.com:12345 # Supported protocols are udp and tcp
```
kalbasit: also syslog right? probably syslog should be mentioned as alias for udp.

logger/configurer/configurer.go
```diff
@@ -0,0 +1,137 @@
...
+
+func (c *Configurer) manageStorageAdapter() {
+	newStorageAdapterType, err := c.getEtcd("/storageAdapterType", "file")
+	if err != nil {
+		log.Println("configurer: Error retrieving storage adapter type from etcd.  Skipping.", err)
+		return
+	}
+	if newStorageAdapterType != c.currentStorageAdapterType {
```
kalbasit: -- nitpick mode on --

I'm not a big fan for a long if statement, this would look much nicer the other way around:

```
ffunc (c *Configurer) manageStorageAdapter() {
	newStorageAdapterType, err := c.getEtcd("/storageAdapterType", "file")
	if err != nil {
		log.Println("configurer: Error retrieving storage adapter type from etcd.  Skipping.", err)
		return
	}
	if newStorageAdapterType == c.currentStorageAdapterType {
		return
	}
	newStorageAdapter, err := storage.NewAdapter(newStorageAdapterType)
	if err != nil {
		log.Println("configurer: Error creating new storage adapter.  Skipping.", err)
		return
	}
	c.syslogishServer.SetStorageAdapter(newStorageAdapter)
	c.currentStorageAdapterType = newStorageAdapterType
	log.Printf("configurer: Activated new storage adapter: %s", newStorageAdapterType)
}
```

-- nitpick mode off --

logger/configurer/configurer.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+}
+
+func (c *Configurer) manageDrain() {
+	newDrainURL, err := c.getEtcd("/drain", "")
+	if err != nil {
+		log.Println("configurer: Error retrieving drain URL from etcd.  Skipping.", err)
+		return
+	}
+	if newDrainURL != c.currentDrainURL {
```
kalbasit: -- nitpick mode on --

same nitpick as above :trollface: 

-- nitpick mode off --

logger/drain/factory.go
```diff
@@ -0,0 +1,33 @@
...
+		return nil, nil
+	}
+	if strings.HasPrefix(drainURL, "udp://") || strings.HasPrefix(drainURL, "syslog://") {
+		drain, err := udp.NewDrain(drainURL)
+		if err != nil {
+			return nil, err
+		}
+		return drain, nil
+	}
+	if strings.HasPrefix(drainURL, "tcp://") {
```
kalbasit: How about `else if`? They're related so no need to execute this test if the first one match.

logger/drain/factory_test.go
```diff
@@ -0,0 +1,66 @@
...
+	"fmt"
+	"reflect"
+	"testing"
+)
+
+func TestGetUsingInvalidValues(t *testing.T) {
+	drainStrs := []string{"bogus"}
+	for _, drainStr := range drainStrs {
+		_, err := NewDrain(drainStr)
+		if err == nil ||
```
kalbasit: Can you please remove the newline after `||` here. It is admittedly visible without scrolling on Github but it took my eyes quite some time to adjust to it. See https://github.com/golang/go/wiki/CodeReviewComments#line-length

logger/drain/factory_test.go
```diff
@@ -0,0 +1,66 @@
...
+func TestGetUdpDrain(t *testing.T) {
+	d, err := NewDrain("udp://my-awesome-log-server:514")
+	if err != nil {
+		t.Error(err)
+	}
+	expected := "*udp.logDrain"
+	dType := reflect.TypeOf(d).String()
+	if dType != expected {
+		t.Errorf("Expected a %s, but got a %s", expected, dType)
+	}
```
kalbasit: This all works a lot better with a usual want got block

```
if want, got := "*udp.logDrain", reflect.TypeOf(d).String(); want != got {
  t.Errorf("Expected a %s, but got a %s", want, got)
}
```

also the added benefit of want and got falling out of scope quickly.

Same for the tests below.

logger/drain/tcp/drain_test.go
```diff
@@ -0,0 +1,13 @@
...
+	"fmt"
+	"testing"
+)
+
+func TestInvalidDrainUrl(t *testing.T) {
+	_, err := NewDrain("https://wwww.google.com")
+	if err == nil || err.Error() != fmt.Sprintf("Invalid drain url scheme: %s", "https") {
+		t.Error("Did not receive expected error message")
+	}
+}
```
kalbasit: How about testing Send()? https://coderwall.com/p/wohavg/creating-a-simple-tcp-server-in-go

logger/drain/udp/drain_test.go
```diff
@@ -0,0 +1,13 @@
...
+	"fmt"
+	"testing"
+)
+
+func TestInvalidDrainUrl(t *testing.T) {
+	_, err := NewDrain("https://wwww.google.com")
+	if err == nil || err.Error() != fmt.Sprintf("Invalid drain url scheme: %s", "https") {
+		t.Error("Did not receive expected error message")
+	}
+}
```
kalbasit: Same as testing tcp.Send().

logger/drain/factory.go
krancour: `else if` doesn't buy us anything there.  Any condition that's matched above results in an early return from the function.

logger/drain/factory_test.go
krancour: So, personally, I'm not opposed to longer lines and would agree with you on this.  However, I decided to follow a precedent that was set by the controller's style checks (which is Python, of course) that lines not exceed 99 characters.  There are one or two other spots in the code that break after a logic operator as well.  I'm making a judgement call to follow the precedent instead of hunting down the offending occurrences here.

logger/drain/factory_test.go
krancour: Nice.  I like it.  Applying.

logger/drain/tcp/drain_test.go
krancour: Unfortunately this is more than I have time to code up right now (I'm currently packing for a trip to Italy).  I have manually tested both drain implementations and am able to ship logs to Papertrail with no problem.  Can I suggest that you PR this test improvement after this PR is merged?

krancour:
@kalbasit thanks again for good feedback.  I've either applied or responded to each of your latest suggestions.

krancour:
@bacongobbler fyi, @carmstrong said I can reassign this to you for you to take across the finish line if it's not lgtm'ed and merged before Monday.

logger/main.go
kalbasit: On a second thought, I think it's better to go for https://github.com/namsral/flag as all of our binaries want to support both the environment and the flags. It's fine if you don't want to do the change now, I'll get a PR across all of our binaries.

benwilber:
[`WEBLOG_PORT`](https://github.com/deis/deis/pull/4502/files#diff-211dbf56ab4ba80bf5d2b2a028b749daR255) should probably get moved to `settings` with `LOGGER_HOST` and they should be named consistently.

Django PKs should always be referenced by `self.pk` instead of [`self.id`](https://github.com/deis/deis/pull/4502/files#diff-211dbf56ab4ba80bf5d2b2a028b749daR255), but that's smaller issue.  (But it will likely cause other issues later, namely confusion).

krancour:
@benwilber nice suggestion.  I'll do that.

krancour:
@benwilber applied.

benwilber:
`self.id` is pervasive in the controller app, so it's not a huge deal right now.  This should probably be marked as an `easy-fix` issue to replace all relevant instances of `self.id` -> `self.pk`.  It doesn't *seem* like a real problem but I guarantee that it is.

krancour:
That should probably be PR'ed separately as a chore that fixes it throughout the controller.

logger/drain/factory_test.go
kalbasit: That's Python though :+1: It's not wise to apply style guide of one language on the other.

krancour: You convinced me.  I'm fixing it now.

krancour:
@kalbasit advice on line breaks applied.

logger/main.go
krancour: Ah.  Missed that that was available.  Applied.

krancour:
One last change I might consider making would be to move the parsing of relevant flags closer to where they're used-- i.e. into each package the flags directly influence.  I don't have time to do it right now.  If it's not done before this gets merged, I'll submit that as a cleanup PR afterwards.

krancour:
Also note that all recent failures are in the functional tests and began immediately after I applied @kalbasit 's suggestions on improving log parsing.  Those may be suspect.

krancour:
Ok... more specifically, it's failing because the `publish-host` and `publish-port` flags aren't being parsed properly in `main.go`... not sure why.  If those aren't parsing properly, chances are all the others are not either.

krancour:
Ok.  This is a misunderstanding over how `flag.String()` (for example) works in comparison to `flag.StringVar()`.  The former (suggested by @kalbasit to simplify things) returns a pointer that I immediately dereferenced...

The two misunderstandings...

1. That doesn't let you off the hook for doing `flag.Parse()`.  That should probably be called from `init()`.

2. Immediately dereferencing those pointers is a bad idea.  The pointers initially point to default values which are potentially updated after the call to `flag.Parse()`... so we want to hang onto those pointers!

I believe both of the above need to be fixed in two spots other than `main.go` as well.

I'm on an airport shuttle now and off to Italia with no laptop, so I'll leave it to someone else to straighten this out unless @carmstrong thinks it can wait for my return-- it could still get into v1.11.1 before freeze either way.

bacongobbler:
> I wasn't around when the decision was initially made, but I'd wager a guess that it had a lot to do with how easy it was to hack on this exising go-based syslog server to add in features like custom message parsing and publishing of logger host:port so that all the logspouts can discover it.

Far from it, actually. rsyslog was broken for some time when they were removed from the mainline repositories and we found a project that simulated a syslog server written in Go, so we switched over in [v0.8.0](https://github.com/deis/deis/releases/tag/v0.8.0) (see https://github.com/deis/deis/commit/7701372aa44792ee53311a6ba5dc1d9f4a54d900). We've invested a lot of time into the new logger that I'm not sure how we could bring it back to a proper rsyslog server again (v2?)

cc @technosophos

kalbasit:
@krancour Umm I don't think that's what I was advising to do. Here's the main.go modified.

`publishTTL` is fixed to `20` rather than `2*publishInterval`. If `publishTTL` has to be at least double `publishInterval` then you probably should enforce that right after `flag.Parse` in main.

```
package main

import (
	"flag"
	"log"
	"os"
	"os/signal"
	"strconv"
	"syscall"
	"time"

	"github.com/deis/deis/logger/configurer"
	"github.com/deis/deis/logger/publisher"
	"github.com/deis/deis/logger/syslogish"
	"github.com/deis/deis/logger/weblog"
)

var (
	logAddr       = flag.String("log-addr", "0.0.0.0", "bind address for the logger")
	logHost       = flag.String("log-host", getopt("HOST", "127.0.0.1"), "address of the host running logger")
	logPort       = flag.Int("log-port", 514, "bind port for the logger")
	enablePublish = flag.Bool("enable-publish", false, "enable publishing to service discovery")
	webAddr       = flag.String("web-addr", "0.0.0.0", "bind address for the web service")
	webPort       = flag.Int("web-port", 8088, "bind port for the web service")
	// Support legacy flag names even though the variable names have been refactored
	etcdHost        = flag.String("publish-host", getopt("HOST", "127.0.0.1"), "service discovery hostname")
	etcdPort        = flag.String("publish-port", getopt("ETCD_PORT", "4001"), "service discovery port")
	etcdPath        = flag.String("publish-path", getopt("ETCD_PATH", "/deis/logs"), "path to publish host/port information")
	configInterval  = flag.Int("config-interval", 10, "config interval in seconds")
	publishInterval = flag.Int("publish-interval", 10, "publish interval in seconds")
	publishTTL      = flag.Int("publish-ttl", 20, "publish TTL in seconds")
)

func main() {
	flag.Parse()

	syslogishServer, err := syslogish.NewServer(*logAddr, *logPort)
	if err != nil {
		log.Fatal("Error creating syslogish server", err)
	}
	weblogServer, err := weblog.NewServer(*webAddr, *webPort, syslogishServer)
	if err != nil {
		log.Fatal("Error creating weblog server", err)
	}
	etcdPortNum, err := strconv.Atoi(*etcdPort)
	if err != nil {
		log.Fatalf("Invalid port specified for etcd server.  '%s' is not an integer.", *etcdPort)
	}
	configurer, err := configurer.NewConfigurer(*etcdHost, etcdPortNum, *etcdPath, *configInterval,
		syslogishServer)
	if err != nil {
		log.Fatal("Error creating configurer", err)
	}

	configurer.Start()

	// Give configurer time to run once so we know syslogishServer is ready to rock
	time.Sleep(time.Duration(*configInterval+1) * time.Second)

	syslogishServer.Listen()
	weblogServer.Listen()

	if *enablePublish {
		publisher, err := publisher.NewPublisher(*etcdHost, etcdPortNum, *etcdPath, *publishInterval,
			*publishTTL, *logHost, *logPort)
		if err != nil {
			log.Fatal("Error creating publisher", err)
		}
		publisher.Start()
	}

	log.Println("deis-logger running")

	// No cleanup is needed upon termination.  The signal to reopen log files (after hypothetical
	// logroation, for instance), if applicable, is the only signal we'll care about.  Our main loop
	// will just wait for that signal.
	reopen := make(chan os.Signal, 1)
	signal.Notify(reopen, syscall.SIGUSR1)

	for {
		<-reopen
		if err := syslogishServer.ReopenLogs(); err != nil {
			log.Fatal("Error reopening logs", err)
		}
	}
}

func getopt(name, dfault string) string {
	value := os.Getenv(name)
	if value == "" {
		value = dfault
	}
	return value
}
```

controller/api/models.py
```diff
@@ -251,9 +249,13 @@ def restart(self, **kwargs):
 
     def _clean_app_logs(self):
         """Delete application logs stored by the logger component"""
-        path = os.path.join(settings.DEIS_LOG_DIR, self.id + '.log')
-        if os.path.exists(path):
-            os.remove(path)
+        try:
+            url = 'http://{}:{}/{}/'.format(settings.LOGGER_HOST, settings.LOGGER_PORT, self.id)
+            requests.delete(url)
+        except Exception:
```
bacongobbler: we should be catching specific errors here like `requests.exceptions.RequestException`.

docs/customizing_deis/logger_settings.rst
```diff
@@ -28,7 +28,18 @@ setting                                  description
 
 Settings used by logger
 -------------------------
-The logger component uses no keys from etcd.
+The following etcd keys are used by the logger component.
+
+====================================      ================================================================================
+setting                                   description
+====================================      ================================================================================
+/deis/logs/storageAdapterType             Type of storage adapter to use: ``file`` or ``memory``; if not set, ``file`` is assumed.  It is also possible so specify the size of the in-memory adapter's internal ring buffer using a value like: ``memory:<size>``.  1000 is the default size.
```
bacongobbler: we should be explicit about what the unit of the size represents. e.g. how much is 1000? 1000 bytes? 1000 terabytes?

docs/managing_deis/running-deis-without-ceph.rst
```diff
@@ -35,15 +35,15 @@ the removal of Ceph. The necessary changes are described below.
...
+listens for log events from platform components and running applications.  It
+ships these to the :ref:`logger` component.  By default, the logger writes the
+logs to a distributed Ceph filesystem. These logs can then be fetched by the
+:ref:`controller` component via HTTP.
 
-The Logger component is not necessary in a Ceph-less Deis cluster. Instead of
-using the Logger, we will route all the logs directly to another syslog
-compatible endpoint.
+In a Ceph-less clutser, the Logger component should be configured, instead, to
+use in-memory log storage.  Optionally, a drain may also be configured to forward
```
bacongobbler: This begs the question: why is the in-memory log handler better? We should elaborate a bit on that :)

docs/managing_deis/running-deis-without-ceph.rst
```diff
@@ -100,18 +90,18 @@ Deploy an AWS cluster
...
 DNS" portion.
 
-Configure log shipping
-^^^^^^^^^^^^^^^^^^^^^^
+Configure logger
+^^^^^^^^^^^^^^^^
 
-The :ref:`logspout` component must be configured to ship logs to somewhere other
-than the :ref:`logger` component.
+The :ref:`logger` component must be configured to use in-memory storage. Optionally
```
bacongobbler: s/must/should

krancour:
@kalbasit... yes... I made some mistakes wheb I slammed the flag parsing changes in quickly before leaving.  Thank goodness for CI.  The updated main.go you shared is exactly what I was about to force push when my ride to the airport showed up.  I'll straighten that out once I am back unless you want to take this PR over in the meantime.  I would not object.

docs/customizing_deis/logger_settings.rst
krancour: It's lines.  Max line size is 65k.  You're right that this requires clarification.

controller/api/models.py
krancour: Not sure I understand the rationale.  Any exception that occurs as that spot (regardless of what kind) is only a failure getting the logger to delete logs-- which should be eaten, imo, because that need not result in the app destroy request failing.  At most, I am thinking I should have at least logged the error.  Not doing so was an oversight that I should fix.

bacongobbler: http://c2.com/cgi/wiki?CatchWhatYouCanHandle

In this case, if something unrelated like a BufferError or an IOError occurs, we want to make sure that those errors are not swallowed by this exception and let Django handle the exception.

docs/managing_deis/running-deis-without-ceph.rst
krancour: I think the previous paragraph addresses this already, doesn't it?

controller/api/models.py
krancour: In general, I agree with the catch what you can handle philosophy, but at this particular spot in the code, there are only two lines within the `try` block and they both have only do to with deleting application logs.  No matter what goes wrong here, I specifically do not want Django handling it, because I don't want the CLI to receive a 500 when a delete request against an app resource succeeds except for the deletion of existing app logs.  Does that make any sense?  Disagreement is welcome.  :smile:

krancour: I will, however, start logging any failure that occurs there.

jchauncey: More logging+++

krancour:
Having a tough time spotting the error in recent test failures.

docs/managing_deis/running-deis-without-ceph.rst
bacongobbler: Ah yes, I'm being a little dumb here. Of course you'd want to use in-memory log storage when Ceph doesn't exist. Disregard!

krancour:
The test failures appear to have been unrelated timeouts.  All is well now.  Want to get this into v1.11, so pinging fellow maintainers for another review... @carmstrong @mboersma @bacongobbler @technosophos @kalbasit

kalbasit:
Added to my TODO list, I'll get this reviewed today.

krancour:
Thanks @kalbasit, much appreciated.

bacongobbler:
another rebase is in order. Sorry!

krancour:
@bacongobbler no problem.  Rebased.

logger/configurer/configurer.go
```diff
@@ -0,0 +1,137 @@
+package configurer
+
+import (
+	"fmt"
+	"log"
+	"time"
+
+	"github.com/coreos/go-etcd/etcd"
```
technosophos: To be the bearer of bad news... we should not be using this library anymore. CoreOS deprecated it. Instead, we should be using github.com/coreos/etcd/client.

I don't know whether it makes sense to change it on the 1.x tree. Maybe we should stay with this one for now and switch for version 2. But the new library is definitely much better.

logger/drain/tcp/drain.go
```diff
@@ -0,0 +1,33 @@
...
+	if err != nil {
+		return nil, err
+	}
+	if u.Scheme != "tcp" {
+		return nil, fmt.Errorf("Invalid drain url scheme: %s", u.Scheme)
+	}
+	return &logDrain{uri: u.Host + u.Path}, nil
+}
+
+func (d *logDrain) Send(message string) error {
```
technosophos: Doc line

logger/storage/file/adapter.go
```diff
@@ -0,0 +1,130 @@
...
+	logBytes, err := exec.Command("tail", "-n", strconv.Itoa(lines), filePath).Output()
+	if err != nil {
+		return nil, err
+	}
+	logStrs := strings.Split(string(logBytes), "\n")
+	return logStrs[:len(logStrs)-1], nil
+}
+
+func (a *adapter) Destroy(app string) error {
+	filePath := a.getFilePath(app)
```
technosophos: I think there are race conditions related to the file. It seems that there are a number of places where between fileExists and another operation, the file can be created or destroyed. Calls to Write and Destroy at around the same time could cause unpredictable states, since Destroy has no mutex hold.

And I might be misreading, but it looks like a file can be destroyed, but still be in `a.files`.

logger/storage/file/adapter.go
```diff
@@ -0,0 +1,130 @@
...
+// Write adds a log message to to an app-specific log file
+func (a *adapter) Write(app string, message string) error {
+	// Check first if we might actually have to add to the map of file pointers so we can avoid
+	// waiting for / obtaining a lock unnecessarily
+	f, ok := a.files[app]
+	if !ok {
+		// Ensure only one goroutine at a time can be adding a file pointer to the map of file
+		// pointers
+		a.mutex.Lock()
+		defer a.mutex.Unlock()
```
technosophos: It might be better to push the mutex lock into getFile, which is almost always sensitive to locking.

logger/storage/ringbuffer/adapter.go
```diff
@@ -0,0 +1,102 @@
...
+// Read retrieves a specified number of log lines from an app-specific ringBuffer
+func (a *adapter) Read(app string, lines int) ([]string, error) {
+	rb, ok := a.ringBuffers[app]
+	if ok {
+		return rb.read(lines), nil
+	}
+	return nil, fmt.Errorf("Could not find logs for '%s'", app)
+}
+
+func (a *adapter) Destroy(app string) error {
```
technosophos: You need a mutex.Lock here.

logger/syslogish/server.go
```diff
@@ -0,0 +1,170 @@
...
+	if match == nil {
+		return "", fmt.Errorf("Could not find app name in message: %s", message)
+	}
+	return match[1], nil
+}
+
+// ReadLogs returns a specified number of log lines (if available) for a specified app by
+// delegating to the server's underlying storage.Adapter.
+func (s *Server) ReadLogs(app string, lines int) ([]string, error) {
+	if s.storageAdapter == nil {
```
technosophos: I believe you need a mutex rlock here. Otherwise the storageAdapter can be removed between here and the Read.

logger/syslogish/server.go
```diff
@@ -0,0 +1,170 @@
...
+		return nil, fmt.Errorf("Could not find logs for '%s'.  No storage adapter specified.", app)
+	}
+	return s.storageAdapter.Read(app, lines)
+}
+
+// DestroyLogs deletes all logs for a specified app by delegating to the server's underlying
+// storage.Adapter.
+func (s *Server) DestroyLogs(app string) error {
```
technosophos: Need lock here.

logger/syslogish/server.go
```diff
@@ -0,0 +1,170 @@
...
+	if s.storageAdapter == nil {
+		return fmt.Errorf("Could not destroy logs for '%s'.  No storage adapter specified.", app)
+	}
+	return s.storageAdapter.Destroy(app)
+}
+
+// ReopenLogs delegate to the server's underlying storage.Adapter to, if applicable, refresh
+// references to underlying storage mechanisms.  This is useful, for instance, to ensure logging
+// continues smoothly after log rotation when file-based storage is in use.
+func (s *Server) ReopenLogs() error {
```
technosophos: Ideally, you need a lock here, too.

logger/weblog/requesthandler.go
```diff
@@ -0,0 +1,73 @@
...
+
+func (h requestHandler) serveGet(w http.ResponseWriter, r *http.Request) {
+	match := getRegex.FindStringSubmatch(r.RequestURI)
+	if match == nil {
+		w.WriteHeader(http.StatusNotFound)
+		return
+	}
+	app := match[1]
+	logLines, err := strconv.Atoi(match[2])
+	if err != nil {
```
technosophos: log a warning?

logger/configurer/configurer.go
mboersma: For better or apparently worse, we already have `go-etcd` vendored into the project, but not the newer client.

> Maybe we should stay with this one for now and switch for version 2.

I think that's preferable, then switch over completely to the shinier package either in a 1.12 or 2.0 release.

controller/api/models.py
```diff
@@ -251,9 +249,15 @@ def restart(self, **kwargs):
...
-            os.remove(path)
+        try:
+            url = 'http://{}:{}/{}/'.format(settings.LOGGER_HOST, settings.LOGGER_PORT, self.id)
+            requests.delete(url)
+        except Exception as e:
+            # Ignore errors deleting application logs.  An error here should not interfere with
+            # the overall success of deleting an application, but we should log it.
+            err = 'Error deleting existing application logs: {}'.format(e)
+            log_event(self, err, logging.WARNING)
+            pass
```
mboersma: `pass` is unecessary here.

controller/api/models.py
```diff
@@ -517,12 +521,24 @@ def _default_scale(self, user, release):
         self.scale(user, structure)
 
     def logs(self, log_lines=str(settings.LOG_LINES)):
-        """Return aggregated log data for this application."""
```
mboersma: Need to keep this or write a new docstring, according to [code standards](http://docs.deis.io/en/latest/contributing/standards/#code-standards):
>  Docstrings and tests are required for all public methods, although the `flake8` tool used by Deis does not enforce this.

krancour:
@technosoohos, thanks for the feedback.  I'll respond or apply suggestions before morning.

logger/syslogish/server.go
```diff
@@ -0,0 +1,170 @@
...
+func NewServer(bindHost string, bindPort int) (*Server, error) {
+	addr, err := net.ResolveUDPAddr("udp", fmt.Sprintf("%s:%d", bindHost, bindPort))
+	if err != nil {
+		return nil, err
+	}
+	c, err := net.ListenUDP("udp", addr)
+	if err != nil {
+		return nil, err
+	}
+	return &Server{conn: c, queue: make(chan string, 5)}, nil
```
technosophos: I'd suggest buffering this at closer to 1000. 5 isn't gonna buy you anything.

logger/syslogish/server.go
```diff
@@ -0,0 +1,170 @@
...
+		message := strings.TrimSuffix(string(buf[:n]), "\n")
+		select {
+		case s.queue <- message:
+		default:
+		}
+	}
+}
+
+func (s *Server) process() {
+	for message := range s.queue {
```
technosophos: This is going to cause backpressure. It reads one message at a time off of the queue, and then locks at least three mutex objects. Depending on the backend, it may block until the receive can commit a write. This means that writes will be very slow, and the buffer will fill quickly. Then your sending process will begin to block, which will hold up your socket. In the worst cases, you'll start dropping UDP packets when they expire before they can be queued. If you are backed to something like an EBS volume, this sort of condition could obtain pretty easily because commits to disk are slow.

Strategies for avoiding backpressure are:

- increasing channel buffer size
- writing to an intermediate in-memory cache (ideally, something like a linked list)
- handling multiple messages during one lock/unlock cycle
  - x := len(queue)
  - lock
  - dequeue x items
  - unlock
  - repeat

technosophos:
Alright... one last question, then I'll stop bugging you. As I went through the code in detail, and then re-read your design, I began to wonder why we only allow one adapter, and why we make drain a separate thing. A drain seems to be adaptable, and we could conceivably want multiple adapters. Would it make sense to change that part to allow a list of adapters, and send the same message to each adapter? In that model, would it make sense to treat a drain as an adapter?

logger/drain/tcp/drain.go
krancour: Nice catch.  Not sure why the style check didn't catch this one.

controller/api/models.py
krancour: You're right.  Removing.

controller/api/models.py
krancour: Not sure how that got removed.  I'm putting it back in.

logger/storage/file/adapter.go
krancour: You're right that I was mistakenly not removing the file pointer from `a.files`.  That was an oversight.  Originally, this was a bit more like the old logger and any given file was opened anew anytime a message was written.  (There was no map of file pointers then.)  That seemed absurd to me, so I added that optimization to open files and keep them open in a map, but overlooked the change that had to be made in `Destroy()` to clean that up.  Nice catch.  I'm fixing this.

Re: possible race conditions-- I suppose the possibility exists that a log message for a given application could arrive _after_ logs for that application have been destroyed.  It will just create a new log for that application is all.  I don't think there's much we can do about that.  Although probably not impossible, it's not a likely scenario because when an app is destroyed, the controller destroy's the application's containers _before_ it signals the logger to destroy that app's logs.

logger/configurer/configurer.go
krancour: Ya.  This is not the only component that's still using that.  If we want to get off of it, we should do it across the board as a chore.

logger/storage/file/adapter.go
krancour: I thought about that, but assuming I were to follow the usual lock / defer unlock pattern in `getFile`, I'd be releasing the lock sooner than I'd want.  If the map requires modification, I want to be holding that lock until that is done.

logger/storage/ringbuffer/adapter.go
krancour: Agreed.

logger/syslogish/server.go
krancour: After an adapter has been initialized, a new one can be swapped in, but there isn't really a way for the adapter to get nilled out since (unlike in the case of a drain), there is no notion of a nil adapter.

Still, out of an abundance of caution, I agree with you on this.

logger/syslogish/server.go
krancour: Same as above.  Agreed.

logger/syslogish/server.go
krancour: Again, agreed.

logger/weblog/requesthandler.go
krancour: Sure.

logger/syslogish/server.go
krancour: So this is an interesting thing... while I've refactored most of the code for easier maintenance, some of the mechanics in terms of how this component actually works are inspired by / true to the original implementation.  The queue depth of 5 came straight out of there.

That being said, I do not object to increasing it.

logger/syslogish/server.go
krancour: > It reads one message at a time off of the queue

Unavoidable, unfortunately.  To ensure multiple messages for a given app are stored in the same order they were sent (the same order in which the app produced them), messages have to be handled sequentially.

> and then locks at least three mutex objects

It's not actually that bad.  The locks acquired on adapters and drains are only read locks to ensure those components aren't nilled out by the configurer after we've checked that they aren't already nil.  Read locks aren't deleterious to performance in the way a write lock is since you rarely block waiting for a read lock.  As for the mutexes used within the storage adapters, we always check whether we need a lock before we bother to get one.  The conditions where one is needed are the first message for a new app or the destruction of logs for an app that's also been destroyed.  Both of these are relatively rare events.

> and the buffer will fill quickly. Then your sending process will begin to block...

The write to the internal channel is non-blocking.  If the buffer that is that channel fills up, messages are dropped right there so the UDP listener's own queue doesn't fill up.

> increasing channel buffer size

Bumped to 1000 per your recommendation.

> writing to an intermediate in-memory cache

That's the point of the internal channel / queue, so this is already being done.

krancour:
> I'll stop bugging you

You're not bugging me.  If anything, I appreciate all the nice catches you made.  There's a lot of concurrency going on in here and I think I handled most of it pretty well, but I'm nothing but appreciative for being called out on the spots I missed.  :)

As to the rest of your question... it's a good one.  There _is_ a certain commonality between adapters and drains.  Aren't they are both just "handling" a message after all?  I actually thought about that pretty early on in the refactor, but deliberately decided against it.  Here's my justification:

To start, the interfaces for storage adapters and drains shouldn't actually be the same.  A component handling storage inherently _must_ be able to carry out a very important operation that a drain cannot-- _retrieving_ what's been stored.  So, I believe a storage adapter is its own distinct and specialized thing.

If not for the above we might conclude that storage adapters and drains are just different sorts of "handlers."  Then it would make a great deal of sense to manage a configurable _chain_ of handlers-- all of which can do _something_ with a log message before passing that message on to the next handler in line.  The first would store, the second would drain...

But if we accept my earlier justification for storage adapters and drains being two distinct things, then I ask myself why we'd ever need to concurrently use more than one _storage adapter_ or more than one _drain_.  What would it even mean, for instance, to use more than one storage adapter?  What would be the point of storing log messages on disk _and_ in memory.  When they're retrieved, which of those two would we retrieve them from?  It doesn't make too much sense.  I guess a stronger argument could be made for using multiple drains (even though that's probably an edge case), but configuring a _chain_ of drains would mean some significant and possibly breaking changes to how drains are configured in etcd.  Remember that even though my implementation is new, drains have been around in the old logger component for quite some time and we're obligated to keep the refactored logger backwards compatible with the old logger drain configuration.

With all of the above in mind, I was ultimately content with my conclusion that storage adapters and drains are two distinct sorts of things and multiple implementations of each are possible, but there's never a need to chain storage adapters and rarely, if ever, a need to chain drains-- and that's probably too much of an edge case to justify the complexity of making that possible while _also_ maintaining backwards compatibility with the old, single drain configuration option.

Does any of that make sense?  I'm not sure.  It's late.  :tired_face:

krancour:
@mboersma and @technosophos, all your feedback has either been acted upon or responded to.  You might need to expand the outdated diffs to see some of my responses.  Feel free to LGTM if you are satisfied with where this stands now, but for my own part this latest crop of changes has me wanting to manually test-drive this one more time in addition to the CI and LGTMs.  I'll get that done first thing tomorrow.

logger/syslogish/server.go
technosophos: For now, I think bumping the channel size is probably sufficient.

That said, my suggestion preserved the sequential writes (just reading n off the channel instead of 1).

The problem with lock acquisition -- read or write -- is that it is slow in Go. So reducing the number of lock operations is good just for that reason.

That said, I'm good with going forward with it as as, and maybe doing some performance improvements later.

technosophos:
With all of the concurrency fixes, LGTM.

I think we could talk architecture till the bikeshed gets painted. The bottom line to me is that the present architecture is a HUGE improvement, and definitely gives us flexibility if we want to, say, streamline write operations in the future. So I don't feel like I should belabor that discussion any more. This is a huge bundle of great work. Thanks to @krancour @aj-may and @rvadim

krancour:
Ok.  So one issue cropped up in testing.  In comparing the performance of the new logger implementation to the old, I've determined that the major bottleneck for both is the drain.  In testing this, I've sent many (millions?) of message to my free Papertrail account today.  At some point, it would appear they started throttling me.  This actually exposed an issue...

If a write to the drain blocks until timeout (because of a problem on the other end), the logger's internal queue can get backed up.  That means messages sit around waiting to be stored and drained since this happens sequentially to preserve the order of messages.  If that queue fills up completely, the logger starts dropping new messages since there's no vacancy in the queue.  Bottom line... if the drain starts failing, you're also not storing messages.  Yikes.

(For what it's worth, the old logger implementation panics when this happens.  The unit restarts the logger and you start with an empty queue again, but eventually get right back into the same scenario.  Bottom line is the old implementation flaps when this happens.)

I spoke with @technosophos and we agreed that a good strategy here would be to use two separate queues for storage and drainage.  Additionally, since the drainage queue is the one that's more likely to back up-- especially if there are problems on the other end of the drain, we can make the timeouts on the drain writes inversely proportionate to the queue's current depth... is your queue nearly empty?  Take your time sending the message.  Is the queue nearly full or overflowing?  Move things along faster-- even if it means some writes fail.  This way newly arriving messages aren't as likely to get dropped due to a queue full of messages that probably can't be drained anyway.

krancour:
@technosophos in case it interests you, I think I've determined that Papertrail was not actually rate limiting me.  I added some more debug logging and found that the backpressure during drainage is coming from timeouts on DNS lookups that occur during dialing.

krancour:
But fwiw, I think the strategy we discussed is still required and valid.  So in the case of the UDP drain, I can add a timeout on the dialing that's inversely proportionate to the queue depth and we'll see how that helps.  Haven't started thinking yet about how the TCP drain needs to be modified.  There we probably need to worry about the timeout on the dialing _and_ the write.

krancour:
> ~~the strategy we discussed is still required and valid~~

Strike that.  It doesn't work out very well.  Nevermind whether the timeouts and/or failures have to do with DNS or actually writing to an external log service.  If attempts to drain messages are failing or timing out for any reason, our goal was to more aggressively manage the queue depth by spending progressively _less_ time attempting to drain each message as the queue starts to back up.  The problem with this, however, is that spending less time on each means we're actually increasing the tempo of actions that will probably fail _and_ increasing the likelihood of continuing failure since something that's timing out already won't magically start succeeding with a _lower_ timeout.  The net effect is that this strategy doesn't really help us to recover at all.  It can make things worse, in fact.

What _can_ be done, however, is we can stop dialing so much.  Dialing the drain's URL once per message is a pattern that was imported from the old log implementation.  It has the benefit of catching DNS changes quickly if they should occur, but other than that, constantly closing an re-opening connections is really inefficient.

I'm going to update the drains so that they reuse each connection 100 times before dialing again.  This is vastly more efficient, but reusing a connection _only_ 100 times means we're re-dialing frequently enough that if DNS changes, it's never _too_ long before we're dialing the updated address.  I've already tried this with the UDP drain and it's working nicely.  TCP might be a little harder, but I'm working on it.

krancour:
@technosophos this is ready for your eyes again if you've got a few minutes at some point this weekend.

There's no timeout management or queue depth management in here.  I've just split the one queue into two separate queues like we discussed so storing the next message doesn't need to wait for draining the previous message.  And I improved the drains themselves by dialing less frequently-- which seems to have been a major source of problems.

There is plenty of room to further optimize this later in follow-up PRs if needed, but as is, this is handling an insane amount of load with no problems-- somewhere in the neighborhood of 10,000 log messages a minute.

krancour:
I also left the latest changes unsquashed (for now) so you can more easily see the relevant diffs.

carmstrong:
> There is plenty of room to further optimize this later in follow-up PRs if needed, but as is, this is handling an insane amount of load with no problems-- somewhere in the neighborhood of 10,000 log messages a minute.

:+1: Sounds perfectly acceptable to me!

krancour:
@carmstrong @technosophos the only remaining thing I might be worried about is in the TCP drain, the strategy of reusing each connection 100 times could backfire if a connection sits idle for too long.  I can't seem to find anything in go that allows us to check the connection state tho.

carmstrong:
@krancour Unless @technosophos has any objections, I think we can :ship: this as is. It's a substantial improvement which we can always iterate on.

technosophos:
@krancour Yeah, this looks good as-is. I see what you mean about the TCP connection re-use, but I don't think I'd worry about that just yet.

olalonde:
Is there any doc somewhere to enable the in memory logger in a stateless install? The [docs](http://docs.deis.io/en/latest/managing_deis/running-deis-without-ceph/#logger) mention `In a Ceph-less clutser, the Logger component should be configured, instead, to use in-memory log storage.` but it doesn't say how exactly to do that... I suppose it's something like `deisctl config:set storageAdapterType=memory`?

olalonde:
Oops, nevermind: http://docs.deis.io/en/latest/managing_deis/running-deis-without-ceph/#configure-logger

Should have kept reading.

bacongobbler:
@olalonde it's a little farther down the doc :)

Issue #4503, opened 2015-09-17
Graceful Upgrade: Restart the routers once


crigor:
The experimental [graceful upgrade](http://docs.deis.io/en/latest/managing_deis/upgrading-deis/#graceful-upgrade) currently does a rolling restart of routers and an install and start of *all* components which results in restarting the routers twice. 

This results in downtime.

https://github.com/deis/deis/blob/e3a28b48a451338e12165ab769d9c711f9533805/deisctl/cmd/upgrade.go#L101-L104

Output of `deisctl upgrade-takeover`: https://gist.github.com/crigor/22baefac7ac494b02d17

iancoffey:
@crigor great catch, this shouldnt happen anymore after https://github.com/deis/deis/issues/4506. that commit should also help with stateless cluster upgrades.

Issue #4419, opened 2015-09-06
Stateless install: graceful upgrade re-install Deis in stateful mode


LoicMahieu:
Upgrading from a stateless install in 1.9.1 with [Gracefull upgrade](http://docs.deis.io/en/latest/managing_deis/upgrading-deis/#graceful-upgrade) re-installs Deis in statefull mode.

Steps:
```
$ deisctl --version
1.9.1

$ curl -sSL http://deis.io/deisctl/install.sh | sh -s 1.10.0 /tmp/deis-upgrade/
```

Refresh-units refresh a lot of components. Lot more than a stateless install needs.
```
$ /tmp/deis-upgrade/deisctl refresh-units
Refreshed deis-builder unit from v1.10.0
Refreshed deis-builder decorator from v1.10.0
Refreshed deis-cache unit from v1.10.0
Refreshed deis-cache decorator from v1.10.0
Refreshed deis-controller unit from v1.10.0
Refreshed deis-controller decorator from v1.10.0
Refreshed deis-database unit from v1.10.0
Refreshed deis-database decorator from v1.10.0
Refreshed deis-logger unit from v1.10.0
Refreshed deis-logger decorator from v1.10.0
Refreshed deis-logspout unit from v1.10.0
Refreshed deis-logspout decorator from v1.10.0
Refreshed deis-publisher unit from v1.10.0
Refreshed deis-publisher decorator from v1.10.0
Refreshed deis-registry unit from v1.10.0
Refreshed deis-registry decorator from v1.10.0
Refreshed deis-router unit from v1.10.0
Refreshed deis-router decorator from v1.10.0
Refreshed deis-store-admin unit from v1.10.0
Refreshed deis-store-admin decorator from v1.10.0
Refreshed deis-store-daemon unit from v1.10.0
Refreshed deis-store-daemon decorator from v1.10.0
Refreshed deis-store-gateway unit from v1.10.0
Refreshed deis-store-gateway decorator from v1.10.0
Refreshed deis-store-metadata unit from v1.10.0
Refreshed deis-store-metadata decorator from v1.10.0
Refreshed deis-store-monitor unit from v1.10.0
Refreshed deis-store-monitor decorator from v1.10.0
Refreshed deis-store-volume unit from v1.10.0
Refreshed deis-store-volume decorator from v1.10.0
Refreshed deis-swarm-manager unit from v1.10.0
Refreshed deis-swarm-manager decorator from v1.10.0
[...]

PR #4506, opened 2015-09-17
fix(deisctl): fixes for graceful upgrades.

iancoffey:
- fix stateless upgrade handling
- stop extra router cycling

### DIFF
--- a/deisctl/client/client.go
+++ b/deisctl/client/client.go
@@ -74,13 +74,19 @@ func (c *Client) UpgradePrep(argv []string) error {
 	usage := `Prepare platform for graceful upgrade.
 
 Usage:
-  deisctl upgrade-prep [options]
+  deisctl upgrade-prep [--stateless]
+
+Options:
+  --stateless  Use when the target platform is stateless
 `
-	if _, err := docopt.Parse(usage, argv, true, "", false); err != nil {
+	args, err := docopt.Parse(usage, argv, true, "", false)
+	if err != nil {
 		return err
 	}
 
-	return cmd.UpgradePrep(c.Backend)
+	stateless, _ := args["--stateless"].(bool)
+
+	return cmd.UpgradePrep(stateless, c.Backend)
 }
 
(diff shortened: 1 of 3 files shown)


iancoffey:
fixes #4419
fixes #4503

deisctl/client/client.go
```diff
@@ -74,13 +74,19 @@ func (c *Client) UpgradePrep(argv []string) error {
 	usage := `Prepare platform for graceful upgrade.
 
 Usage:
-  deisctl upgrade-prep [options]
+  deisctl upgrade-prep [--stateless]
```
adamreese: Could we check to see if store services are running to know if they need to be upgraded?

deisctl/cmd/upgrade.go
```diff
@@ -65,16 +69,15 @@ func republishServices(ttl uint64, nodes []*model.ConfigNode, cb config.Backend)
 }
 
 // UpgradeTakeover gracefully starts a platform stopped with UpgradePrep
-func UpgradeTakeover(b backend.Backend, cb config.Backend) error {
-
-	if err := doUpgradeTakeOver(b, cb); err != nil {
+func UpgradeTakeover(stateless bool, b backend.Backend, cb config.Backend) error {
```
adamreese: It would be nice to have the ability to specify version upgrading to and validate unit files are desired version.

michelleN:
:+1:

deisctl/client/client.go
iancoffey: Yeah, auto-detect could be possible down the road. I personally like the explicit flag for now.

deisctl/cmd/upgrade.go
iancoffey: Rolling in a unit refresh has been brought up a few times, it does make sense. For now, I just like allowing the user to use whatever units they want along with documenation, and having a little less machinery in deisctl.

Issue #4512, opened 2015-09-19
config:push with .env does not support multline


sstarcher:
If you have an environment variable with multiple lines config:push fails for that specific environment variable.

benwilber:
@sstarcher can you show an example of a multiline env variable?  from the [dotenv docs](https://github.com/bkeepers/dotenv#usage), you can double-quote multiline strings and use literal `\n` to separate lines:

FOO="bar\nbaz"

Or are you saying that this is what doesn't work?

sstarcher:
@benwilber currently with 1.10.0 multiple problems seems to exist.
With 1.10.0 I could use double quotes and add say an rsa key, but when that is put into a .env via pull only the first line gets set to the variable.

In addition if I would change it to not be a double quoted multi line and I would change it to be \n's the \ns would remain if I attempted to do something along the lines of 
echo "$id_rsa" > /root/.ssh/id_rsa 
the id_rsa would be invalid and contain \n

sstarcher:
I just tested adding a double quote to the .env file that was spit out by deis config:pull, it acts the same was as if no quotes exist and for each line of the multiline string it outputs the line and "does not match the pattern 'key=var', ex: MODE=test"

When the same multiline with double quote is done on the command line it works correctly.

benwilber:
@sstarcher are you able to provide a couple quick examples of what you're seeing from `deis config:pull` and the same thing failing for `deis config:push`?

sstarcher:
When I refer to Deis I'm referring to the platform version installed.  All of the below is using deis client 1.10.0

In Deis 1.10.0 On the command line the following has the expected result of setting a multiline variable.  Although it does complain ```does not match the pattern 'key=var', ex: MODE=test
Creating config... ...```

```deis config:set HEALTHCHECK_INITIAL_DELAY=60 DERPY="HELLO                                                                               
dquote> world how
dquote> are you" -a app
'DERPY=HELLO
world how
are you' does not match the pattern 'key=var', ex: MODE=test
Creating config... done

=== app Config
HEALTHCHECK_INITIAL_DELAY      60
HEALTHCHECK_TIMEOUT            300
HEALTHCHECK_URL                /
```

After doing the above everything finishes correct, but when I do a deis config:pull or a deis config:list I don't see DERPY listed at all.  This previously functioned with Deis 1.9.0.


[...]

PR #4513, opened 2015-09-20
fix(client): accept multi line config vars.

Joshua-Anderson:
The client will now accept multi line configuration variables with `config:set`. Note: this will not work with `config:push` as it uses newlines as a config variable separator.

What now works is: 
```
config:set TEST="MULTI
LINE VAR"`
```

ref #4512

### DIFF
--- a/client/cmd/config.go
+++ b/client/cmd/config.go
@@ -216,7 +216,7 @@ func ConfigPush(appID string, fileName string) error {
 func parseConfig(configVars []string) map[string]interface{} {
 	configMap := make(map[string]interface{})
 
-	regex := regexp.MustCompile("^([A-z_]+[A-z0-9_]*)=(.+)$")
+	regex := regexp.MustCompile(`^([A-z_]+[A-z0-9_]*)=([\s\S]+)$`)
 	for _, config := range configVars {
 		if regex.MatchString(config) {
 			captures := regex.FindStringSubmatch(config)
--- a/tests/config_test.go
+++ b/tests/config_test.go
@@ -11,10 +11,12 @@ import (
 )
 
 var (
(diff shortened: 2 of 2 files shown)


benwilber:
I believe the new [config regex](https://github.com/Joshua-Anderson/deis/blob/config-multi-line/client/cmd/config.go#L219) will remove the ability to use config values like:

```
DATABASE_URL=mysql://user:pass@host/db
```

Joshua-Anderson:
@benwilber Thanks! I think I fixed it.

benwilber:
@Joshua-Anderson I don't think that fixed it.  `(.+)` is the correct regex for config values.  I think there's gotta be a better way to support multiline variables.

Joshua-Anderson:
@benwilber Are you sure? https://play.golang.org/p/c73VjWcOK6 Your example string worked for me. `\s` matches all whitespace and `\S` matches everything that's not whitespace. Is there a better regex to match everything?

benwilber:
@Joshua-Anderson yep I think this is correct.  I was testing my regexs with Python:

```
>>> r = re.compile(r'^([A-z_]+[A-z0-9_]*)=([\s\S]+)$')
>>> print(r.match("mysql://user:pass@host/db"))
None
```

weird difference there between Go's regex and Python's.

benwilber:
@Joshua-Anderson yeah my test was wrong.  This new regex `[\s\S]` works great.  But that's essentially the same as the previous `(.+)`

Joshua-Anderson:
> But that's essentially the same as the previous (.+)

But ever so slightly different! `.` matches everything except newlines. `\S\s` matches newlines as well.

benwilber:
of course!  perfect.

PR #4522, opened 2015-09-22
[WIP] fix([tests, contrib]): use absolute path to $DEIS_TEST_AUTH_KEY

sgoings:
Previous behavior to find the DEIS_TEST_AUTH_KEY:

```
~/.ssh/${DEIS_TEST_AUTH_KEY}
```

Now:

```
${DEIS_TEST_AUTH_KEY} # (so user can specify full path)
```


bacongobbler:
this is a breaking change, but since it only affects testers then it's not a big issue.

sgoings:
Just to be clear about my most recent changes... I set up the defaults to be (in testing):

* deisctl tunnel ssh key for providers other than Vagrant:

  ```
  ~/.ssh/deis-test[.pub]{0,1} # overrideable via DEIS_TEST_SSH_KEY
  ```

* deisctl tunnel for Vagrant

  ```
  ~/.vagrant.d/insecure_private_key
  ```

* deis keys for test app deployment

  ```
  ~/.ssh/deiskey[.pub]{0,1} # overrideable via DEIS_TEST_AUTH_KEY
  ```

Is that ok? Or should we reuse `~/.ssh/deis-test` (instead of `~/.ssh/deiskey`) for the deis keys (for app deployment)?

PR #4529, opened 2015-09-23
feat(logger): add https as a logger drain type

### LINKED ISSUE #4527: Add https log drain type
It would be very convenient to have an https log drain type. Log aggregation providers like SumoLogic do not provider managed syslog collectors, preferring instead https based collectors. While Sumo does offer self-hosted syslog collectors, it seems very awkward to run a non-Deis managed container on each CoreOS cluster node, and pipe journal messages in for each application, or alternatively consume all journal messages, with no application prefixes. Using a drain to an https collector endpoint seems like a much more elegant solution.

nsmith:
Not all log aggregation providers provide a hosted syslog-based
collector. Having a log drain that can use https is therefore desirable.

Closes #4527

### DIFF
--- a/logger/drain/drain.go
+++ b/logger/drain/drain.go
@@ -1,11 +1,14 @@
 package drain
 
 import (
+	"crypto/tls"
 	"fmt"
 	"log"
 	"net"
 	"net/url"
+	"net/http"
 	"os"
+	"strings"
 )
 
 func SendToDrain(m string, drain string) error {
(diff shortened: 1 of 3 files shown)


deis-admin:

logger/drain/drain.go
```diff
@@ -17,12 +20,30 @@ func SendToDrain(m string, drain string) error {
...
 		log.Println(u.Scheme + " drain type is not implemented.")
 	}
 	return nil
 }
 
+func sendToHttpsDrain(m string, drain string) error {
+	buf := strings.NewReader(m)
+
+	tr := &http.Transport{
+  	TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
```
nsmith: We skip certificate verification here since we do not have root certs installed in alpine linux, and can therefore not verify certs issued by public root CAs. I think this is acceptable because we are in control of the endpoint, and since this would presumably be run in a non-hostile network, the chances of a MITM attack are low. The alternative here is to install root certs on the base image, or let the user specify a pinned cert. Specifying a cert to pin would be preferable from a security standpoint, although since the cert of a log aggregator may be out of the user's control, and could be changed at any time, it is somewhat brittle.

krancour:
Blocked by #4502.  This will need a refactor after #4502 is merged... fortunately, #4502 makes just this sort of thing easier to implement.

krancour:
@nsmith needs rebase/refactor because #4502 refactored the logger from the ground up.  This should be easy to refactor though... see drain interface in `logger/drain/drain.go` and examples of implementations in `logger/drain/tcp/drain.go` and `logger/drain/udp/drain.go`

nsmith:
Will do!

nsmith:
Going to close this until the refactor is done so that it isn't hanging around in the PR backlog.

PR #4531, opened 2015-09-23
feat(database): HA postgres

bacongobbler:
Migrating over technosophos/deis#3

This implements an HA postgres implementation with the help of compose/governor and with PR 20 applied (use etcd client bindings instead of urllib).

TODO:

 - [ ] fix functional tests
 - [ ] slim down docker image back to a respectable level (476 MB ew)
 - [x] figure out how to deploy the database in v2 (ping @technosophos, do I just convert the RC/service yml to json and dump them in `deisctl/units/rcs`?)


carmstrong:
deferring reviews for V2 branch merges to @technosophos / @smothiki

PR #4533, opened 2015-09-24
ref(controller): remove REGISTRY_MODULE hack

mboersma:
While working on #3814 I ran into issues trying to use python's `mock` module in controller/registry unit tests. Turns out the local mock.py module there and associated hackery are not needed, so I PR'ed these changes separately.

Removes the REGISTRY_MODULE machinery, adds verbosity to `flake8` errors that are caught, and updates the python `coverage` library.

### DIFF
--- a/controller/api/tests/__init__.py
+++ b/controller/api/tests/__init__.py
@@ -6,6 +6,7 @@ import os
 from django.conf import settings
 from django.test.client import RequestFactory, Client
 from django.test.simple import DjangoTestSuiteRunner
+import requests
 
 
 # add patch support to built-in django test client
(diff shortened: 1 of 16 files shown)


mboersma:
I had some additional mocks I wrote to handle these errors, but for whatever reason the tests are passing for me locally without them. My environment must be in error; I'll add back the mocks and get this passing later today.

carmstrong:
> My environment must be in error; I'll add back the mocks and get this passing later today.

:+1:

controller/api/tests/__init__.py
```diff
@@ -46,18 +47,25 @@ def run_tests(self, test_labels, extra_tests=None, **kwargs):
             test_labels, extra_tests, **kwargs)
 
 
+def mock_status_ok(*args, **kwargs):
```
bacongobbler: +1 for refactoring to a common household

PR #4535, opened 2015-09-24
ref(aws): Improve AWS queries instead of piping output.

helgi:
Fully utilise the power of aws cli query instead of piping various AWS commands into cut, head and other programs


sgoings:
LGTM. Thanks for doing this! I wasn't aware of these AWS powers.

FWIW: Jenkins failed after provisioning succeeded... but you'll still need to retry this job until the build passes.

Issue #117, opened 2013-08-20
Support for interactive `deis run bash` session


gabrtv:
While #70 provided us with the ability to run one-off admin commands, we really need an interactive shell into the ephemeral docker containers.  

This is quite tricky, as it requires a full duplex channel (HTTP Chunked Encoding? WebSockets?) as well as a select loop on the API server to handle routing stdin, stdout and stderr.

https://github.com/paramiko/paramiko/blob/master/demos/interactive.py#L39 provides a good example of a terminal-enabled select loop.

bacongobbler:
for the meantime, this has been the workaround for most users: https://github.com/deis/deis/issues/3449#issuecomment-91845128

jwaldrip:
@bacongobbler I will say that the workaround I have been doing is to just use `docker exec` to get into the running container. Seems to work fine for me (as much as it pains me to do it)

fbjork:
Any update on this? This is as pretty big limitation of Deis compared to Heroku.

gabrtv:
@fbjork I would have liked to ship this earlier, but as we've discussed in the open planning meetings, this feature is on hold pending Deis v2.  Reason being the plumbing we need for this already exists in `kubectl exec`, versus having to write and maintain a bi-directional channel from scratch -- only to throw it away for v2.

We expect to have an RC for Deis v2 available by the end of the year.

hajimenomomo:
Hi everyone,
I wrote a python script for my personal use that opens a rails console directly from your deis app folder, using TLS to authenticate to the docker remote API on your cluster nodes, to use it, you have to enable Docker remote API with TLS, see: https://coreos.com/os/docs/latest/customizing-docker.html 
I personally put it in my PATH directory (/usr/local/bin).
You may find it in my repo: https://github.com/hajimenomomo/deis_rails_console.git
I hope this helps.

kaedroho:
Django are planning on adding a new "channels" feature into the core for 1.10 (http://channels.readthedocs.org/en/latest/integration-plan.html). This should make it possible to serve Websockets from the controller.

rstacruz:
The workaround is to SSH into one of your CoreOS nodes:

```sh
# try the one in DEISCTL_TUNNEL
ssh core@1.2.3.4
```

then check out the docker containers running:

```sh
docker ps
CONTAINER ID        IMAGE                                 COMMAND                  CREATED             STATUS              PORTS                                                                                      NAMES
[...]

PR #4542, opened 2015-09-29
chore(router): update nginx to 1.9.5

mboersma:
See http://nginx.org/en/CHANGES

spdy is gone is favor of http2, so this may be a breaking change. (Team nginx semver foul: fifteen yards and loss of possession.)


helgi:
https://www.nginx.com/blog/nginx-1-9-5/

> Prerequisites Before upgrading, please remove the spdy parameter from all listen directives in  your configuration. This will ensure the upgrade goes smoothly. 

Do all routers auto get bumped to 1.9.5? Is it via new image? Is there any possibility of the new configuration to be applied on top of a setup that has a 1.9.4 still running?

carmstrong:
> Is there any possibility of the new configuration to be applied on top of a setup that has a 1.9.4 still running?

Great point to double-check, but the template is tied to the version of the component. There shouldn't be the possibility of running a template and nginx out-of-sync -- the entire component is upgraded on a new release.

helgi:
Then it should be safe enough - Would recommend keeping an eye on that in the testing. Would be a rather big blow up if it somehow got itself on top of 1.9.4 and older :)

carmstrong:
> Would be a rather big blow up if it somehow got itself on top of 1.9.4 and older :)

Indeed :)

mboersma:
I'm not saying we need to upgrade (yet)--I looked at the changes and thought about it, but if this isn't strictly backward-compatible for our documented installs then it can be targeted for v2. Deis is pretty intimate with nginx, and this is where nginx is going, plus I know we've flirted with HTTP/2 for solving #117 so this might be a forward-looking change.

carmstrong:
> but if this isn't strictly backward-compatible for our documented installs

I think it should be. @aledbf, what are your thoughts on this?

aledbf:
From the [docs](http://nginx.org/en/docs/http/ngx_http_v2_module.html#example) 1.9.5 requires OpenSSL >= 1.0.2 (where "Application-Layer Protocol Negotiation" is available)
Using alpine 3.1 uses 1.0.1

```
old-mbp:~ aledbf$ docker run -it alpine:3.1 /bin/sh
3.1: Pulling from library/alpine

8dd8107abd2e: Pull complete
Digest: sha256:22b8b96ceccf4d484e4ca62363abb527fc5c2fe01a4097642395c4d830c3471f
Status: Downloaded newer image for alpine:3.1
/ # apk -U add openssl
fetch http://dl-4.alpinelinux.org/alpine/v3.1/main/x86_64/APKINDEX.tar.gz
(1/1) Installing openssl (1.0.1p-r0)
Executing busybox-1.22.1-r15.trigger
OK: 6 MiB in 16 packages
/ #
```

I already tried 1.9.5 but I see different behavior depending on the browser. For now I think 1.9.4 is a safer option.

mboersma:
> For now I think 1.9.4 is a safer option.

Totally agree. We can probably fix that with `alpine:3.2`, but let's not rush into it for this release.

mboersma:
We should look at updating to alpine:3.2 across the board.

mboersma:
> I already tried 1.9.5 but I see different behavior depending on the browser.

@aledbf I updated everything to `alpine:3.2` in master, so OpenSSL is 1.0.2d in deis-router. Does that fix the behavior you noticed, or should we play it safe and skip nginx 1.9.5?

aledbf:
> or should we play it safe and skip nginx 1.9.5?

I think would be a good idea to wait for 1.9.6 before switching to http2.

PR #4543, opened 2015-09-29
docs(managing_deis): add caveat for different etcd versions

carmstrong:


docs/managing_deis/upgrading-deis.rst
```diff
@@ -261,6 +261,12 @@ functioning properly.
   Deis. It is always recommended to provision a Deis release with the CoreOS version specified
   in that release's provision scripts or documentation.
 
+.. important::
+
+  Upgrading a cluster can result in simultaneously running different etcd versions,
```
krancour: Possible s/different/two different/.  Otherwise :+1:

carmstrong: Potentially > 2 depending on which versions they're deploying, although unlikely.

Issue #4528, opened 2015-09-23
provisioning should fail on a used discovery URL


mboersma:
A persistent problem is forgetting to run `make discovery-url` before provisioning a cluster. If it hasn't been run, and there's no user-data file, that is caught and handled well, but if a user-data exists with a discovery.etcd.io endpoint that's already been used, that isn't caught and provisioning will proceed and then fail with errors.

Potentially Deis could either:
- make an HTTP request of the discovery URL and fail if it exists, assuming that means it's been used already
- mark the user-data file with some canonical entry that will be caught and handled as an error on reuse, and/or make it invalid YAML
- generate a new disco URL on every provisioning (see #4548)

bacongobbler:
perhaps we could check the timestamp of the user data and warn if it's been at least a half hour since it was last updated?

krancour:
What about an HTTP request to the discovery URL as @mboersma suggested, but then actually parse the JSON response to determine if any nodes have already used it.

mboersma:
> but then actually parse the JSON response

That's probably the most robust way. 

The real trick here is that there are multiple provisioning paths, from `vagrant up` to `provision-aws.sh` and friends. So the solution to this may need to be a standard shell script or binary tool that does what is outlined above, but can be called easily from any of those contexts.

helgi:
What about just running `make discovery-url` on provision each time. That's what I have done.

I did this in python, for ... reasons (import os and subprocess in one of the gen-cluster scripts)

```python
def discovery_url():
    # Ensure the cluster has the latest user-data
    os.chdir(os.path.realpath(os.path.join(CURR_DIR, '..', '..', '..')))  # Just to get to the deis root
    cmd = "make discovery-url"
    _, err = subprocess.Popen(cmd, shell=True, stdout=subprocess.PIPE).communicate()
    if err:
        print err
        raise
    os.chdir(CURR_DIR)
```

Is there a problem with generating a new one on each provision since it is separate from updating? Seems simpler than jumping into HTTP parsing

krancour:
> So the solution to this may need to be a standard shell script or binary tool

[...]

PR #4548, opened 2015-09-29
[WIP] ref(contrib/aws): generate discovery url on each provision

helgi:
Fixes the AWS side of #4528

### DIFF
--- a/contrib/aws-production/cluster/cluster.parameters.json
+++ b/contrib/aws-production/cluster/cluster.parameters.json
@@ -1,0 +1,6 @@
+[
+    {
+        "ParameterKey":     "KeyPair",
+        "ParameterValue":   "deis"
+    }
+]
--- a/contrib/aws-production/cluster/user-data/data-plane-user-data
+++ b/contrib/aws-production/cluster/user-data/data-plane-user-data
@@ -1,0 +1,282 @@
+#cloud-config
+
+coreos:
+  etcd2:
+    # generate a new token for each unique cluster from https://discovery.etcd.io/new
+    discovery: https://discovery.etcd.io/740b815b1abb64db757135451fe4a155
+    # multi-region and multi-cloud deployments need to use $public_ipv4
+    advertise-client-urls: http://$private_ipv4:2379
+    initial-advertise-peer-urls: http://$private_ipv4:2380
+    # listen on both the official ports and the legacy ports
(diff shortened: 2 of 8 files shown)


sgoings:
This will cause 2 `make discovery-url` calls when using `rigger` but I figger that's ok for now.

helgi:
How's that? Is it because you are doing 2 clusters or because of the split up handling and rigger runs both?

sgoings:
https://github.com/deis/rigger/blob/master/commands/provision/script

carmstrong:
If someone is calling an update to an existing stack or making other modifications to a running cluster, chances are they won't want to use a new discovery URL. We should double-check that this doesn't inadvertently result in a split etcd cluster.

helgi:
This only touches provisioning and not updating - No auto thing happening there.

carmstrong:
> This only touches provisioning and not updating

Don't we call `gen-json.py` from within [update-aws-cluster.sh](https://github.com/deis/deis/blob/master/contrib/aws/update-aws-cluster.sh#L35)?

helgi:
Sigh, yeah, brain-fart. Easily resolved. Commit coming

helgi:
Added in a flag for now that tells the template generator if it is in update mode or not.

krancour:
It might be better to apply the same fix to all the different provisioners.  As is, this introduces a feature / safeguard into the AWS provisioner that isn't included in the others.  This puts the burden of that knowledge on the users or perhaps upon rigger.  Longer term, we will be doing ourselves a big favor if we improve overall consistency of use and behavior among each of those.  The result will be the establishment of a defacto interface and spec for any provisioner rigger utilizes.

helgi:
I had this one ready so I threw it in - haven't looked at the other providers. Could have a look when I'm done with my current tickets and see how easy that would be to do.

krancour:
Ok.  I know this feature you're adding is a nice convenience, but let's hold until we can add it for all the provisioners just so we're not adding inconsistencies to the provisioning landscape.

carmstrong:
> but let's hold until we can add it for all the provisioners

Should we close this and file an issue tracking that?

mboersma:
@helgi do you think this could be brought over the finish line in the next week or two so it can go in the Deis v1 LTS release? I know you're busy with Deis v2 tasks, so if this isn't on your near-term roadmap we may want to close it.

PR #4550, opened 2015-09-29
chore(*): update CoreOS to 766.4.0

mboersma:
See https://coreos.com/releases/#766.4.0


sgoings:
Would it be possible to make Vagrant grab the COREOS_VERSION from contrib/utils.sh?

bacongobbler:
@sgoings I think it's possible but not totally required for this to be merged. Would be a nice feature so we have one less version number to bump, though.

mboersma:
> Would it be possible to make Vagrant grab the COREOS_VERSION from contrib/utils.sh?

That would be great; I'll put an issue in to track that.

PR #4552, opened 2015-09-29
feat(registry) add more information on during startup process

slack:
I was confused by registry start because it was blocking on the store to become available. Last thing I saw was "registry running".

So I thought I'd add more information to the startup process to clear up any confusion.

After:
```
Sep 29 23:29:07 ... systemd[1]: Started deis-registry.
Sep 29 23:29:07 ... sh[30838]: registry: environment ETCD=10.21.2.206:4001 ETCD_PATH=/deis/registry HOST_ETCD_PATH=/deis/registry/hosts/10.21.2.206 ETCD_TTL=20 REGISTRY_PORT=5000 EXTERNAL_PORT=5000 BUCKET_NAME=registry
Sep 29 23:29:28 ... sh[30838]: registry: creating bucket
Sep 29 23:29:28 ... sh[30838]: registry: bucket created
Sep 29 23:29:28 ... sh[30838]: registry: starting registry
Sep 29 23:29:28 ... sh[30838]: registry: docker registry started in background with pid: 39
[...]
Sep 29 23:29:29 ... sh[30838]: registry: starting loop to update service discovery
[...]
Sep 29 23:30:13 ... systemd[1]: Stopping deis-registry...
Sep 29 23:30:20 ... sh[30838]: registry: on_exit() called
[...]
```

### DIFF
--- a/registry/bin/boot
+++ b/registry/bin/boot
@@ -21,9 +21,11 @@ export REGISTRY_PORT="${PORT:-5000}"
 
 export BUCKET_NAME="${BUCKET_NAME:-registry}"
 
+echo "registry: environment ETCD=${ETCD} ETCD_PATH=${ETCD_PATH} HOST_ETCD_PATH=${HOST_ETCD_PATH} ETCD_TTL=${ETCD_TTL} REGISTRY_PORT=${REGISTRY_PORT} EXTERNAL_PORT=${EXTERNAL_PORT} BUCKET_NAME=${BUCKET_NAME}"
+
 # wait for etcd to be available
 until etcdctl --no-sync -C "$ETCD" ls >/dev/null 2>&1; do
-	echo "registry: waiting for etcd at $ETCD..."
+	echo "registry: waiting for etcd at ${ETCD}..."
 	sleep $((ETCD_TTL/2))  # sleep for half the TTL
 done
 
@@ -35,8 +37,7 @@ function etcd_set_default {
   ERROR="$(etcdctl --no-sync -C "$ETCD" mk "$ETCD_PATH/$1" "$2" 2>&1 >/dev/null)"
 
   if [[ $? -ne 0 ]] && echo "$ERROR" | grep -iqve "key already exists"; then
-    echo "registry: etcd_set_default: an etcd error occurred ($ERROR)"
-    echo "registry: aborting..."
+    echo "registry: etcd_set_default($1=$2) aborting, an etcd error occurred ($ERROR)"
     exit 1
   fi
   set -e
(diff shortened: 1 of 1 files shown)


krancour:
Why the addition of the "registry:" prefix on all log lines?  Isn't it redundant information?  Whether using `deisctl journal` or `journalctl` to view logs or even when logs undergo forwarding via logspout or aggregation or draining via logger, the component of origin is already known.

bacongobbler:
The database already does this in its logs so there's precedence, though I don't think there was any compelling reason why it was introduced in the first place. We can come back to that and clean 'er up later. I'll file a bug if this is merged :)

registry/bin/boot
```diff
@@ -35,8 +37,7 @@ function etcd_set_default {
   ERROR="$(etcdctl --no-sync -C "$ETCD" mk "$ETCD_PATH/$1" "$2" 2>&1 >/dev/null)"
 
   if [[ $? -ne 0 ]] && echo "$ERROR" | grep -iqve "key already exists"; then
-    echo "etcd_set_default: an etcd error occurred ($ERROR)"
-    echo "aborting..."
+    echo "registry: etcd_set_default($1/$2) aborting, an etcd error occurred ($ERROR)"
```
krancour: This line could be misleading.  Using the path separator / forward slash right there would yield a message that looks like this if, for instance, etcd failed setting _key_ `foo` to _value_ `bar`:

```
registry: etcd_set_default(foo/bar) aborting, an etcd error occurred...
```

That could be very misleading because `foo/bar` looks like a key, when really it's a key and value.  Could I suggest revising the format of that message?

mboersma: I would suggest `etcd_set_default($1=$2)` so we get:
```
etcd_set_default(foo=bar) aborting, an etcd error occurred...
```
Either that or a colon-plus-space `($1: $2)`.

mboersma:
Example output:
```
Oct 09 16:47:45 deis-03 systemd[1]: Starting deis-registry...
Oct 09 16:47:45 deis-03 etcdctl[17541]: Error:  105: Key already exists (/deis/cache) [25197]
Oct 09 16:47:45 deis-03 sh[17563]: deis-registry
Oct 09 16:47:45 deis-03 systemd[1]: Started deis-registry.
Oct 09 16:47:45 deis-03 sh[17576]: registry: environment ETCD=172.17.8.102:4001 ETCD_PATH=/deis/registry HOST_ETCD_PATH=/deis/registry/hosts/172.17.8.102 ETCD_TTL=20 REGISTRY_PORT=5000 EXTERNAL_PORT=5000 BUCKET_NAME=registry
Oct 09 16:48:06 deis-03 sh[17576]: registry: creating bucket
Oct 09 16:48:06 deis-03 sh[17576]: registry: bucket created
Oct 09 16:48:06 deis-03 sh[17576]: registry: starting registry
Oct 09 16:48:06 deis-03 sh[17576]: registry: docker registry started in background with pid: 39
Oct 09 16:48:06 deis-03 sh[17576]: [2015-10-09 16:48:06 +0000] [39] [INFO] Starting gunicorn 19.3.0
Oct 09 16:48:06 deis-03 sh[17576]: [2015-10-09 16:48:06 +0000] [39] [INFO] Listening at: http://0.0.0.0:5000 (39)
```

I agree with @krancour that prefixing everything with `registry:` is redundant, since I can only see these logs by asking for them specifically.

carmstrong:
> I agree with @krancour that prefixing everything with registry: is redundant, since I can only see these logs by asking for them specifically.

Agreed, but as @bacongobbler pointed out, we do this ad-hoc with the other components. I think we merge this and then clean up all the components in one fell swoop.

mboersma:
> merge this and then clean up all the components in one fell swoop

Fine with me, but there needs to be a GitHub issue describing that cleanup work. Also our comments about using slash as a separator should be addressed.

slack:
Updated.

On my cluster that drained to papertrail it wasn't clear which unit was producing output. *shrug*

PR #4553, opened 2015-10-01
chore(tests): update shellcheck to 0.4.1

mboersma:
`brew install shellcheck` now produces version 0.4.1, which finds an additional error that needs to be ignored.


controller/bin/boot
```diff
@@ -65,6 +65,7 @@ etcd_safe_mkdir /deis/services
 echo "controller: running etcd data migrations..."
 for script in /app/migrations/data/*.sh;
 do
+	# shellcheck disable=SC1090
```
sgoings: Nifty! I didn't know that this was possible with shellcheck (disabling a shellcheck check with a bash comment).

Issue #4545, opened 2015-09-29
Graceful Shutdown does not remove from Discovery URL


sstarcher:
The graceful shutdown script will not remove nodes from the Discovery URL.  The effect of this is that the new node that comes up with be a peer always.  So if one of the nodes that is contributing to the quorum goes down and gets replaced the quorum will stay -1.  If you attempt to replace all of your initial nodes or even most it is likely you would lose the cluster.

aj-may:
I also ran into this issue.  I can submit a PR once I test my fix.

iancoffey:
Problem here will be execution order - just adding the proper etcd and discovery url handing to the script wont suffice since the unit will not function unless it can stop without the deis-store-* components being present and active.

Its a little tricky situation, I have a PR forthcoming to address it.

aj-may:
@iancoffey sounds good.  I'll let you address it.  Feel free to tag me on the PR and I'll take a look.

iancoffey:
@aj-may Indeed, your input on the PR would be appreciated https://github.com/deis/deis/pull/4554

PR #4554, opened 2015-10-01
feat(contrib): graceful shutdown for non-ceph nodes 

iancoffey:
This introduces the ability to handle the shutdown and cleanup of a nodes etcd member when the node is not running the Deis store components. Right now, if a node isnt a normal control plane node, the process cant happen.

It does so by making use of separate units for etcd and ceph graceful shutdown. This is because the systemd units need to ``Require`` different units - the Ceph stuff requires all the store stuff and needs to order itself after it while the etcd unit does not.

In the future, when this no longer experimental, we could remove the need for manual start up and enabling of the units by having etcd and/or the store units ``Want`` these units - therefore starting them they activate.

Another change is that these units make use of ``RefuseManualStop``, which should make it significantly harder for someone to accidentally trigger the process.

### DIFF
--- a/contrib/coreos/user-data.example
+++ b/contrib/coreos/user-data.example
@@ -59,21 +59,40 @@ coreos:
         content: |
           [Service]
           ExecStartPre=-/usr/bin/etcdctl mk /coreos.com/network/config '{"Network": "10.244.0.0/16", "SubnetLen": 24, "SubnetMin": "10.244.0.0", "Backend": {"Type": "vxlan"}}'
-    - name: graceful-deis-shutdown.service
+    - name: graceful-ceph-shutdown.service
       content: |
         [Unit]
-        Description=Clean up
+        Description=Ceph node clean up for Deis
         DefaultDependencies=no
-        After=fleet.service etcd2.service docker.service docker.socket deis-store-admin.service deis-store-daemon.service deis-store-volume.service deis-store-monitor.service
-        Requires=fleet.service etcd2.service deis-store-admin.service deis-store-daemon.service deis-store-volume.service docker.service docker.socket deis-store-monitor.service
+        After=fleet.service etcd2.service docker.service docker.socket deis-store-admin.service deis-store-daemon.service deis-store-volume.service deis-store-monitor.service graceful-etcd-shutdown.service
+        Requires=fleet.service etcd2.service docker.service docker.socket deis-store-admin.service deis-store-daemon.service deis-store-volume.service deis-store-monitor.service graceful-etcd-shutdown.service
+        RefuseManualStop=true
 
         [Install]
         WantedBy=shutdown.target halt.target reboot.target
 
         [Service]
-        ExecStop=/opt/bin/graceful-shutdown.sh --really
+        ExecStart=/usr/bin/docker exec deis-store-admin ceph -s
+        ExecStop=/opt/bin/graceful-shutdown.sh --ceph
         Type=oneshot
         TimeoutSec=1200
+        RemainAfterExit=yes
+    - name: graceful-etcd-shutdown.service
+      content: |
+        [Unit]
+        Description=etcd clean up for Deis
+        DefaultDependencies=no
+        After=fleet.service etcd2.service docker.service docker.socket
+        Requires=fleet.service etcd2.service docker.service docker.socket
+        RefuseManualStop=true
+
+        [Install]
+        WantedBy=shutdown.target halt.target reboot.target
+
+        [Service]
+        ExecStop=/opt/bin/graceful-shutdown.sh --etcd
+        Type=oneshot
+        TimeoutSec=120
         RemainAfterExit=yes
     - name: install-deisctl.service
       command: start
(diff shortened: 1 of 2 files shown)


iancoffey:
The unit ordering needs some tweaking before this is ready.

krancour:
@iancoffey can you modify the issue title to reflect that this is coping with isolated planes specifically WRT to graceful shutdown?

iancoffey:
fixes #4554 
fixes #4545

mboersma:
Could you change `Feature(contrib)` to `feat(contrib)` in the commit message? Otherwise this change won't be included properly in CHANGELOG.md.

aj-may:
@iancoffey Instead of creating 2 different services, could we instead put the deis-store components into a wants directive?

iancoffey:
@aj-may I ran into a couple issues when trying to solve this with a single unit, neither of which seemed solvable.

One is that the shutdown unit will still need to have the store components listed in the After, and since they wont exist in a stateless platform or dataplane node, the unit will not execute the ExecStop. This is the behavior I observed, I never found a way to avoid that.

Another issue with a single unit is that once the graceful stop unit is enabled and started, Wants will attempt to start the services listed there, which would not be desirable in a stateless platform.

With two units, the main etcd and store units could Want their cleanup units and start them when they start. That was intended to be part 2 of this, once its been vetted.

It would be great to get your feedback on this approach, thanks much!

carmstrong:
@mboersma Did you happen to test this manually with your LGTM? Code looks good, but I haven't had a chance to manually test it.

mboersma:
@carmstrong no, I just did a code review--it still needs a manual test. :hammer:

crigor:
I tested removing a node from discovery url, booted a new node and the new node didn't add itself to the existing cluster. The steps are

* set up a 3-node etcd cluster (node1, node2, node3)
* remove node3 using etcdctl member remove
* remove node3 from the discovery url using curl -X DELETE
* stop node3 from AWS console
* start a new one (I didn't use a deis cluster for this test so no autoscaling)

node4 used the discovery url, saw it's not full, but created a new etcd cluster with a new id. https://gist.github.com/crigor/d4d701aa1959cdd1d9fa#file-gistfile1-txt-L31-L34

This results in an error since node1 and nod2 already belong to a different etcd cluster.

iancoffey:
@crigor Ive seen that issue as well and it is a bit unexpected. That said, we shouldn't be using the discovery url for adding or replacing nodes anyway - thats long existing a flaw in our provisioning setup. We should only be joining members correctly via the runtime reconfig process - like Deis Pro does. In that light, there is no utility to removing the node from the discovery service - it wont result in anything useful and it shouldn't have any impact on the future of the cluster. Im probably going to remove the discovery service deletion from this PR but would love to hear input if theres any.

aj-may:
@iancoffey +1 for not using discovery url for reconfig.  already got burned on that one.

iancoffey:
This is all set to review now.

carmstrong:
@iancoffey What's the motivation for changing the root `Makefile`? Will that break any of the providers?

Also, re: the discovery service, there is an implicit size passed in on creation (see https://github.com/deis/deis/blob/d830adf1e8d323ff37f23300661c543daebc632b/Makefile#L35). I *believe* this number is used to determine how many nodes should be core members (and the rest remaining as proxies).

carmstrong:
> I believe this number is used to determine how many nodes should be core members (and the rest remaining as proxies).

This seems to be the case. From https://coreos.com/os/docs/latest/cluster-discovery.html -
> Third, if you specified ?size=3 upon discovery URL creation, any other machines that join the cluster in the future will automatically start as etcd proxies.

iancoffey:
@carmstrong Good catch, the makefile change was needed when the node was being deleted from the discovery service. It shouldnt have had any negative impact but I removed that change.

sstarcher:
That is one major issue with etcd discovery service.  If you start it up with 5 nodes and a size of 5 if you terminate 1 node at a time and allow for a new node to join it will always join as a proxy.  So after gracefully terminating 3 nodes your cluster will degrade and break.

bacongobbler:
@sstarcher see https://github.com/deis/deis/issues/4645#issuecomment-151955120 for an explanation on that subject.

sstarcher:
@bacongobbler so the current solution after a few nodes degrade is to create an entirely new cluster to stop the old cluster from degrading and losing consensus?

contrib/coreos/user-data.example
```diff
@@ -190,58 +209,67 @@ write_files:
...
-        until [[ $(docker exec deis-store-admin ceph health) == *"HEALTH_OK"* ]]
+      ceph_shutdown () {
+        # determine osd id
+        OSD_HOSTS=($(etcdctl ls /deis/store/hosts/| awk -F'/' '{print $5}'))
+        for HOST in "${OSD_HOSTS[@]}"
         do
-          if [ $TIMEWAITED -gt "1200" ]
-          then
-            echo "ceph graceful removal timeout exceeded"
+          PUBLIC_IP=$(etcdctl member list|grep `cat /etc/machine-id`| awk '{print $3}'| grep -Eo '[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}')
```
helgi: You could do `etcdctl get /deis/store/osds/$(hostname -i)` (or `--ip-address` for clarity) instead of looping through `etcdctl member list`. Then you are just checking for `$?` success to get the `OSD_ID`

Not entirely sure it would work the way we expect if we throw in some isolation.

contrib/coreos/user-data.example
```diff
@@ -190,58 +209,67 @@ write_files:
...
-      done
-      # if we own an osd and its healthy, try to gracefully remove it
-      if [ ! -z "$OSD_ID" ] && [[ "$CURRENT_STATUS" == *"HEALTH_OK"* ]] && [ ${#OSD_HOSTS[@]} -gt "3" ]; then
-        docker exec deis-store-admin ceph osd out $OSD_ID
-        sleep 30
-        TIMEWAITED=0
-        until [[ $(docker exec deis-store-admin ceph health) == *"HEALTH_OK"* ]]
+      ceph_shutdown () {
+        # determine osd id
+        OSD_HOSTS=($(etcdctl ls /deis/store/hosts/| awk -F'/' '{print $5}'))
```
helgi: If you make the change below then you can technically change this into `etcdctl ls /deis/store/hosts/ | wc -l` and do the comparison on line 224 without treating it as an array

contrib/coreos/user-data.example
```diff
@@ -190,58 +209,67 @@ write_files:
...
+          # remove ceph mon
+          docker stop deis-store-monitor || true
+          docker exec deis-store-admin ceph mon remove `hostname -f` # fixme
+          docker stop deis-store-metadata || true
+        fi
+      }
+      etcd_shutdown () {
+        set -e -x -o pipefail
+        # removing the node from etcd
+        NODE=$(etcdctl member list | grep `cat /etc/machine-id` | cut -d ':' -f 1)
```
helgi: This is not going to work if you use DNS SRV discovery + setting etcd names. I haven't found a reliable way yet to check for that kind of condition

```
etcdctl member list
7c501b0e7a69ecd3: name=node-1 peerURLs=http://node-1.etcd-moo.internal:2380 clientURLs=http://node-1.etcd-moo.internal:2379
9849e1552fbbe9b2: name=node-2 peerURLs=http://node-2.etcd-moo.internal:2380 clientURLs=http://node-2.etcd-moo.internal:2379
b276c45b6b42f883: name=node-3 peerURLs=http://node-3.etcd-moo.internal:2380 clientURLs=http://node-3.etcd-moo.internal:2379
c4d95826ffa38aa9: name=node-4 peerURLs=http://node-4.etcd-moo.internal:2380 clientURLs=http://node-4.etcd-moo.internal:2379
e8c5cab61f1e82b8: name=node-5 peerURLs=http://node-5.etcd-moo.internal:2380 clientURLs=http://node-5.etcd-moo.internal:2379
```

The member id there is not the same as the `fleetctl` `machine-id` so it would only work for `discovery-url` clusters using the `default` name.

helgi: On a vanilla contrib/aws setup you get this output

```
etcdctl member list
16a160897c0bc9ab: name=8cc0e1fea18f4995bc1e693f40f09e3f peerURLs=http://10.21.2.151:2380 clientURLs=http://10.21.2.151:2379
876060a75057748c: name=98660f695d88458ab18a1e0c4f0d70e6 peerURLs=http://10.21.1.116:2380 clientURLs=http://10.21.1.116:2379
9adddc1e295abae3: name=e09bea4147a143bcbb850a0b5fb93cbf peerURLs=http://10.21.1.115:2380 clientURLs=http://10.21.1.115:2379
d070be23d802e98b: name=077417806e3f4cdabf8611c6a2e45163 peerURLs=http://10.21.2.150:2380 clientURLs=http://10.21.2.150:2379
ee94bb3ac2fad3e1: name=35f709ccf298449188c6f365176fbaa5 peerURLs=http://10.21.1.114:2380 clientURLs=http://10.21.1.114:2379
```

Looking at `/run/systemd/system/etcd2.service.d/20-cloudinit.conf` I can see `$ETCD_NAME` is not set.

I'd recommend checking for `$ETCD_NAME` and if that's empty fallback on using `/etc/machine-id` for the matching. Including changing the matching to be `name=<id>` to be a bit more robust. This will make custom `name` and `default` work.

contrib/coreos/user-data.example
iancoffey: Pardon the delay in responding. This part *should* work - if the public ip ends up unset, it should just skip being removed below. That said, this thing definitely could use manual testing.

contrib/coreos/user-data.example
iancoffey: I think setting ETCD_NAME would be outside the scope of this PR. I think it also might be deceptively tricky to implement. Since this system only exists to work with the current vanilla contrib/ setup, the current code should still work (even though its been a while). Does that seem reasonable?

iancoffey:
Rebased, green, ready for manual testing.

docs/managing_deis/add_remove_host.rst
```diff
@@ -288,24 +288,42 @@ This can be achieved by making a request to the etcd API. See `remove machines`_
...
+cluster membership before reboot, shutdown or halt events. The units can be used independently or
+together.
 
-The unit requires that the optional ``deis-store-admin`` component is installed.
+The ``graceful-etcd-shutdown`` unit is useful for any Deis node running its own etcd. To be used, it
+must be enabled and started.
+
+.. code-block:: console
+
+    root@deis-1:/# systemctl enable graceful-etcd-shutdown
```
bacongobbler: this needs to be changed to `geaceful-deis-shutdown`

iancoffey:
@bacongobbler There was a discrepancy in that `graceful-ceph-shutdown` is incorrectly named `graceful-deis-shutdown` in the unit. I fixed that and rebased.

bacongobbler:
So here's how I tested (I missed the name change from deis to ceph but that's just visual):

```
$ deisctl install store-admin && deisctl start store-admin
$ vagrant ssh deis-01
$ sudo systemctl enable graceful-etcd-shutdown
$ sudo systemctl install graceful-etcd-shutdown
$ sudo systemctl enable graceful-deis-shutdown
$ sudo systemctl install graceful-deis-shutdown
$ sudo shutdown now
```

And here's what I end up with:

```
><> vagrant status
Current machine states:

deis-01                   poweroff (virtualbox)
deis-02                   running (virtualbox)
deis-03                   running (virtualbox)

This environment represents multiple VMs. The VMs are all listed
above with their current state. For more information about a specific
VM, run `vagrant status NAME`.
><> vs deis-02
Last login: Fri Dec  4 19:57:37 2015 from 10.0.2.2
 * *    *   *****    ddddd   eeeeeee iiiiiii   ssss
*   *  * *  *   *     d   d   e    e    i     s    s
 * *  ***** *****     d    d  e         i    s
*****  * *    *       d     d e         i     s
*   * *   *  * *      d     d eee       i      sss
*****  * *  *****     d     d e         i         s
  *   *****  * *      d    d  e         i          s
 * *  *   * *   *     d   d   e    e    i    s    s
***** *****  * *     ddddd   eeeeeee iiiiiii  ssss

Welcome to Deis                 Powered by CoreOS
core@deis-02 ~ $ etcdctl cluster-health
member 48c35a6d2dd5edb4 is healthy: got healthy result from http://172.17.8.101:2379
member d41d2beaa8bf8eeb is healthy: got healthy result from http://172.17.8.102:2379
cluster is healthy
core@deis-02 ~ $ docker exec deis-store-admin ceph -s
    cluster 8305ad54-177a-4243-9ef0-2c56051c02f4
     health HEALTH_WARN
            768 pgs degraded
            761 pgs stuck unclean
            768 pgs undersized
            recovery 379/1134 objects degraded (33.422%)
            1/3 in osds are down
            1 mons down, quorum 1,2 deis-02,deis-03
     monmap e3: 3 mons at {deis-01=172.17.8.100:6789/0,deis-02=172.17.8.101:6789/0,deis-03=172.17.8.102:6789/0}
            election epoch 10, quorum 1,2 deis-02,deis-03
     mdsmap e7: 1/1/1 up {0=deis-03=up:active}, 1 up:standby
     osdmap e27: 3 osds: 2 up, 3 in
      pgmap v207: 768 pgs, 12 pools, 471 MB data, 378 objects
            21531 MB used, 23773 MB / 47831 MB avail
            379/1134 objects degraded (33.422%)
                 768 active+undersized+degraded
  client io 1939 B/s wr, 0 op/s
```

So it looks like etcd removed the node, but not ceph. :(

iancoffey:
For what its worth, I tested this twice more and it was successful both times. One reason why @bacongobbler 's test might have failed could have been that ceph wasn't healthy to begin with. Which I tend to see one out of every 3 or 4 installs or so.

Not sure if this is totally tabled at this point, but it does seem to work for me. Maybe I just have the lucky touch!

bacongobbler:
I tested this again and it seemed to work for me. Since this is hidden behind a feature flag I think it's safe to consider adding to the LTS so users can opt into this and we can figure out why failure cases like [this](https://github.com/deis/deis/pull/4554#issuecomment-162068129) happened earlier.

PR #4562, opened 2015-10-03
feat(contrib): vagrantfile sources utils.sh

### LINKED ISSUE #4556: "vagrant up" should source COREOS_VERSION
It would be simpler if vagrant provisioning could reference contrib/utils.sh as AWS does, so we can maintain fewer version number resources in the codebase.

dustinrc:
`COREOS_CHANNEL` and `COREOS_VERSION` are the variables of
particular interest.

fixes #4556

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -11,11 +11,23 @@ end
 
 CLOUD_CONFIG_PATH = File.join(File.dirname(__FILE__), "contrib", "coreos", "user-data")
 CONFIG = File.join(File.dirname(__FILE__), "config.rb")
+CONTRIB_UTILS_PATH = File.join(File.dirname(__FILE__), "contrib", "utils.sh")
+
+# Make variables from contrib/utils.sh accessible
+if File.exists?(CONTRIB_UTILS_PATH)
+  cu_vars = Hash.new do |hash, key|
+    value = `. #{CONTRIB_UTILS_PATH} 2> /dev/null && echo $#{key}`.chomp
+    hash[key] = value unless value.empty?
+  end
+else
+  raise Vagrant::Errors::VagrantError.new, "The file '#{CONTRIB_UTILS_PATH}' is missing."
+end
 
 # Defaults for config options defined in CONFIG
 $num_instances = 1
 $instance_name_prefix = "deis"
-$update_channel = ENV["COREOS_CHANNEL"] || "stable"
+$update_channel = cu_vars["COREOS_CHANNEL"]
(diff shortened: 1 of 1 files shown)


Vagrantfile
```diff
@@ -51,16 +51,29 @@ def vm_cpus
...
+if File.exists?(CONTRIB_UTILS_PATH)
+  cu_vars = Hash.new do |hash, key|
+    value = `. #{CONTRIB_UTILS_PATH} 2> /dev/null && echo $#{key}`.chomp
+    hash[key] = value unless value.empty?
+  end
+else
+  raise Vagrant::Errors::VagrantError.new, "The file '#{CONTRIB_UTILS_PATH}' is missing."
+end
+
+coreos_channel = cu_vars["COREOS_CHANNEL"]
```
mboersma: Could this be preserved as `$update_channel`? One of the (unstated) goals here is to minimize differences with the upstream Vagrantfile in https://github.com/coreos/coreos-vagrant, and if we keep that variable name the diff is just a bit smaller. Otherwise this looks great.

bacongobbler: +1

dustinrc:
There is no good reason to drop `$update_channel` in this case. I have switched the variable names back, as well as made a few other changes to further minimize the difference with upstream.

Vagrantfile
```diff
@@ -52,15 +64,18 @@ def vm_cpus
 end
 
 Vagrant.configure("2") do |config|
-  # always use Vagrant's insecure key
+  # always use Vagrants insecure key
```
bacongobbler: this wasn't a typo. Would you mind fixing this back, please?

Vagrantfile
```diff
@@ -52,15 +64,18 @@ def vm_cpus
...
   config.ssh.insert_key = false
+
```
bacongobbler: extra space is unnecessary

bacongobbler:
tiny nitpicks but LGTM otherwise!

dustinrc:
I debated those changes myself! They seemed silly and unnecessary as I was making them, but that's how it exists in the upstream Vagrantfile. Will have to make the updates later tonight....

mboersma:
I think it's ok to be typo-compatible with the upstream Vagrantfile actually--that was basically what I had asked for by minimizing the diff. But I leave it up to @bacongobbler.

bacongobbler:
If that's what upstream's got, then let's stick with that for now. :)

bacongobbler:
thanks @dustinrc!

PR #4566, opened 2015-10-05
ref(registry/database): change the create_bucket script to avoid allow 'ListAllMyBuckets' on S3 Policy

rvaralda:
Use s3.lookup method to identify bucket name instead of list and expose all bucket names to deis cluster.

### DIFF
--- a/database/templates/create_bucket
+++ b/database/templates/create_bucket
@@ -12,5 +12,5 @@ conn = boto.connect_s3(
     calling_format=OrdinaryCallingFormat())
 name = sys.argv[1]
 
-if name not in (bucket.name for bucket in conn.get_all_buckets()):
+if not conn.lookup(name):
     conn.create_bucket(name)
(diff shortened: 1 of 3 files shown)


carmstrong:
This looks like it should work just fine. For reference: http://boto.readthedocs.org/en/latest/ref/s3.html#boto.s3.connection.S3Connection.lookup

> Attempts to get a bucket from S3. Works identically to S3Connection.get_bucket, save for that it will return None if the bucket does not exist instead of throwing an exception.

@rvaralda If we're making the change for registry, perhaps we should also update database as part of this PR? It does the same thing: https://github.com/deis/deis/blob/2333ccdc19969a6affa1b8b2225b88d186a2af33/database/templates/create_bucket#L15

rvaralda:
That's right, I will update and push it.

rvaralda:
o/

PR #4577, opened 2015-10-07
deisctl list-machines

michelleN:


deisctl/deisctl.go
```diff
@@ -32,6 +32,7 @@ Commands, use "deisctl help <command>" to learn more:
   install           install components, or the entire platform
   uninstall         uninstall components
   list              list installed components
+  list-machines     Enumerate the current hosts in the cluster
```
sgoings: Since we're already a bit different between fleetctl and deisctl... would `machines` be a better choice than `list-machines`?

sgoings: Additionally... you'll probably want lowercase the command's explanation.

bacongobbler: I'd argue for `nodes` as a better choice in prep for v2 (would be handy to make `deisctl nodes` be equivalent to `kubectl get nodes`)

deisctl/deisctl.go
```diff
@@ -95,6 +96,8 @@ Options:
 	switch command {
 	case "list":
 		err = c.List(argv)
+  case "list-machines":
+    err = c.ListMachines(argv)
```
mboersma: This needs `gofmt` run on the source files--looks like Jenkins complained about this whitespace difference. :white_circle:

deisctl/deisctl.go
bacongobbler: We bickered amongst ourselves and found `machines` to be the best candidate for v1. We can change it to `nodes` in v2 if it's too confusing (and as @mboersma pointed out, we can also just make it an alias of `machines`, too).

deisctl/client/client.go
```diff
@@ -240,6 +241,20 @@ Usage:
 	return cmd.ListUnits(c.Backend)
 }
 
+func (c *Client) Machines(argv []string) error {
+	usage := `Enumerate the current hosts in the cluster
```
krancour: This is an awesome PR, so this is nitpicking mode here.  Does "enumerate" actually imply some ordinality to the results that actually isn't present?  Might it be better to just say, "List the current hosts in the cluster?"

deisctl/backend/fleet/list_machines.go
```diff
@@ -0,0 +1,76 @@
...
+			return ms.PublicIP
+		},
+		"metadata": func(ms *machine.MachineState, full bool) string {
+			if len(ms.Metadata) == 0 {
+				return "-"
+			}
+			return formatMetadata(ms.Metadata)
+		},
+	}
+)
```
krancour: This is clever using a map of functions to handle each field.  Looking at the rest of the code and docstrings tho, I don't see any supported options for limiting what fields are returned or requesting additional fields.  Are there plans to do so?  If not, I'm almost wondering if there's some complexity that we could pare down here by just moving this logic right into `printMachines()`

krancour:
This is an awesome PR and very thoroughly tested.  I imagine the many commits are intended to be squashed into one in this case, but make sure that the final commit(s) follow the project's [commit style](http://docs.deis.io/en/latest/contributing/standards/#commit-style).

deisctl/client/client.go
michelleN: We were just copying the fleetctl list-machines cmd description here, but yes list works too.

krancour: Makes sense.  We don't need to be just like them, but this is enough precedent to probably leave it be.  :smile:

deisctl/backend/fleet/list_machines.go
adamreese: I left it open so we could support other args for full machine ids or limiting columns.  I am not sure how much value we are adding by making a clone of the `fleetctl list-machines` command.

PR #4578, opened 2015-10-08
chore(logger): improve drains

krancour:
This replaces the existing "good enough" UDP-based and TCP-based log drains with one drain that can accommodate both and, more importantly, is more performant and resilient.

Cliff notes:

* It reuses connections for up to 5 minutes, but not more than that.  This is to increase efficiency by dialing less, while still dialing often enough that any DNS changes that affect delivery of messages to an external log archive won't go unnoticed for very long.
* It can detect a broken pipe and try resending a message after establishing a new connection.
* Repeated failed dialing attempts will "mute" the drain for five minutes.  This gives the network a break (necessary, especially if the failures may be due to any sort of external DNS or external log service rate limiting) and allows the backed up message queue to be cleared more quickly.  If a drain is muted for any period of time, it is noted in the logger's own journal logs.


krancour:
Ping @jackfrancis.  I think this was of interest to you, so I'd appreciate your feedback.

logger/drain/factory.go
```diff
@@ -15,19 +14,14 @@ func NewDrain(drainURL string) (LogDrain, error) {
...
-		drain, err := udp.NewDrain(drainURL)
-		if err != nil {
-			return nil, err
-		}
-		return drain, nil
-	}
-	if strings.HasPrefix(drainURL, "tcp://") {
-		drain, err := tcp.NewDrain(drainURL)
+	// Any of these three can use the same drain implementation
+	if strings.HasPrefix(drainURL, "udp://") || strings.HasPrefix(drainURL, "syslog://") || strings.HasPrefix(drainURL, "tcp://") {
```
jackfrancis: I like that you've absorbed the drain 'types' into the `simple` package. I wonder if it wouldn't be even better to absorb the conditional prefix string logic into the NewDrain 'constructor' as well. And in fact, the same for future implementations, e.g., TLS/TCP and HTTP(s) as mentioned below.

logger/drain/simple/drain.go
```diff
@@ -0,0 +1,117 @@
...
+	"net"
+	"net/url"
+	"sync"
+	"time"
+)
+
+// This determines how often we'll replace an existing connection with a new one.  This ensures we
+// redial, which in turn will ensure any DNS changes affecting where logs are drained to don't
+// go unnoticed for too long.
+const connRefreshInterval = 5 * time.Minute
```
jackfrancis: How feasible would it be to make the 4 settings starting at this one above user-configurable?

logger/drain/simple/drain.go
```diff
@@ -0,0 +1,117 @@
...
+		return nil
+	}
+	d.mutex.Lock()
+	defer d.mutex.Unlock()
+	conn, err := d.getConnection(false)
+	if err != nil {
+		return err
+	}
+	_, err = fmt.Fprintln(conn, message)
+	if err != nil {
```
jackfrancis: `//TODO wrap this retry logic in an exponential backoff abstraction`

logger/drain/factory.go
krancour: Good question.  Skip a few lines down and there's a TODO with a partial explanation.  I expect other drains to be implemented soon-- either by us or the community.  I think those will differ more significantly from the one drain we have now.  I want the factory to be able to instantiate the correct sort of drain for whatever URL it's given.  It's only right now in the interim where there's only one drain implementation that this doesn't make so much sense, but it will soon.

jackfrancis: Got it. So it sounds like this factory function *is* the convenience abstraction for enabling the platform to simply request a new drain by passing in a URI string to one internal endpoint.

logger/drain/simple/drain.go
krancour: Harder than it sounds.  Currently the configurer just watches `/deis/logs/drain` for changes.  When changes are noticed, that one value is passed to the factory to get the right drain back.  Making all those things configurable means watching more things _and_ expanding the interface for the factory so values for each of those things can be passed in-- which might not be entirely ideal since some of those things may not be relevant to other drain types.

More configuration options are nice, but in this case, I think the hard-coded values are pretty sensible and should suffice near term.  imo, we should re-evaluate this only if the community starts calling for it.

logger/drain/simple/drain.go
krancour: I think this is another thing to come back to if there's a demonstrated need for it.

Currently, there are two things that might fail and thus be retried.  A failed message write is probably on account of a broken pipe.  It's retried with a new connection.  Assuming the new connection was successfully obtained, there's no reason the second attempt shouldn't succeed.  So on that front, I'm not sure if we need to get any fancier... at least for now.

The other thing that gets retried are failed dialing attempts.  That's currently hardcoded through a constant to be retried five times before giving up, at which point there isn't a prayer of firing off a log message to the external service, so the drain is temporarily muted.

Muting a log drain has the benefit of (forgive the overloaded term here) _draining_ the internal message queue very quickly so that when things recover, we're draining recent log messages out to the external log archive instead of working through a backlog of older log messages while newer log messages get dropped because of the backed up queue.

I _think_ this all sounds pretty reasonable, but I am more than willing to hear further challenges to that.

jackfrancis:
Remarks all make sense, and deferring to the community for further enhancements is certainly the appropriate guidepost.

LGTM.

krancour:
In AWS, I'm not seeing this detect bad connections like it had been when I tested in Vagrant.  Trying to pin down why.

krancour:
I'm confused enough about several parts of this that I could really use some help from anyone who understands UDP and TCP better than I apparently do.  Ping @carmstrong.  Who's good with transport protocols?

krancour:
Pinging those who were interested in this issue: @carmstrong @technosophos @jackfrancis 

I added tons of debug statements and a demo/drive in `logger/drain_demo.go`.  There are notes in there explaining its use and enumerating a few interesting / surprising test cases.  If you try, be sure to try on OSX _and_ Linux to see different results for some scenarios.

krancour:
This should be ready for another round of reviews.  The drain now performs exactly as one would expect when the route to the remote end of the drain is interrupted.  With UDP, write errors go undetected and packets are spilled into the ether-- just the way UDP is supposed to work.  With TCP, the write errors also go more-or-less undetected, but that's because we don't reuse a connection for long enough for the OS' underlying TCP stack to give up on retrying writes and waiting for an ACK it will never get (this can actually take a _long_ time and Go doesn't allow timeouts to be set on individual writes).  However, by not using a connection for longer than a minute, the resulting redials can detect broken routes when the TCP handshake fails.  There would, thus, never be more than a one minute gap in logs before a warning about the broken route is recorded in the logger's own journal logs.

logger/drain/simple/drain.go
```diff
@@ -0,0 +1,130 @@
...
+func (d *logDrain) getConnection(forceNew bool) (net.Conn, error) {
+	// If we have a connection, it's not old, and we're not focing a new one...
+	if d.conn != nil && !forceNew {
+		// then return the existing connection
+		return d.conn, nil
+	}
+	// If ANY of those conditions weren't met, it's time for a new connection.
+	// If we have an existing one, close it and nil it out, too for good measure.
+	if d.conn != nil {
+		d.conn.Close()
```
bacongobbler: some connection types return an error if we fail to properly close it. We should log that error if necessary

logger/drain/simple/drain.go
```diff
@@ -0,0 +1,132 @@
...
+	if d.conn != nil && !forceNew {
+		// then return the existing connection
+		return d.conn, nil
+	}
+	// If ANY of those conditions weren't met, it's time for a new connection.
+	// If we have an existing one, close it and nil it out, too for good measure.
+	if d.conn != nil {
+		if err := d.conn.Close(); err != nil {
+			log.Println("drain: Error closing connection.  Drain may be leaking connections.", err)
+		}
```
krancour: @bacongobbler, I've applied your suggestion.

PR #4581, opened 2015-10-08
feat(controller): use Docker for last-mile layer

### LINKED ISSUE #3550: Apps built on Docker 1.6 -- Cannot `deis pull`: Invalid image ID
I get this error on Deis 1.5.0. I used to work just fine a couple of days back.

```
deis pull mozillamffos/masterfirefoxos:ab98aa464cc4eccb1ed8c62145a745d7d2ee95c0 -a masterfirefoxos-dev
Creating build..503 SERVICE UNAVAILABLE
Detail:
GET Image Error (404: {"error": "Invalid image ID"})
```

The tag exists on docker hub and I can manually `docker pull mozillamffos/masterfirefoxos:ab98aa464cc4eccb1ed8c62145a745d7d2ee95c0` from the Deis hosts.

I tried the same command with different deis app, same issue.

### LINKED ISSUE #4378: Issue pulling docker images from local registry.
I have a docker registry set up at `10.10.1.101`, and am trying to deploy the image `example` with the `latest` tag onto a deis cluster.

```
ubuntu@ip-10-10-1-197:~/deis/example$ deis pull 10.10.1.101/example:latest
Creating build... 503 SERVICE UNAVAILABLE
Detail:
GET Image Error (404: {"error": "Tag not found"})
```

I cannot seem to be able to figure out why I'm getting this error, as I'm 100% sure the tag exists in the registry, and I can also manually log into the coreos hosts and do a 'docker pull' and it fetches it fine. I also had to manually install cacert bundle on the coreos hosts - or else docker refuses to pull because of a self signed cert. Is docker used in any of the deis services that would need it installed too?

### LINKED ISSUE #4541: 'deis pull' gives me constant 404 NOT FOUND
I'm running deis using vagrant on my macbook as a POC.  I've followed the documentation and triple checked I have done things correctly wrt the installation.  I'm running the latest Vagrant and latest VirtualBox as its provider.   I have checked out and built your example:

$ git clone https://github.com/deis/example-go.git
$ cd example-go
$ git checkout docker
$ docker build -t ericroubal/example-go .
$ docker tag ericroubal/example-go localhost:5000/ericroubal/example-go
$ docker push localhost:5000/ericroubal/example-go

 and am able to successfully pull :

> docker pull localhost:5000/ericroubal/example-go

I think everything is installed correctly:

$  deisctl list
UNIT				MACHINE				LOAD	ACTIVE	SUB
deis-builder.service		32e0ab49.../172.17.8.102	loaded	active	running
deis-controller.service		32e0ab49.../172.17.8.102	loaded	active	running
deis-database.service		32e0ab49.../172.17.8.102	loaded	active	running
deis-logger.service		32e0ab49.../172.17.8.102	loaded	active	running
deis-logspout.service		32e0ab49.../172.17.8.102	loaded	active	running
deis-publisher.service		32e0ab49.../172.17.8.102	loaded	active	running
deis-registry@1.service		32e0ab49.../172.17.8.102	loaded	active	running
deis-router@1.service		32e0ab49.../172.17.8.102	loaded	active	running
deis-store-daemon.service	32e0ab49.../172.17.8.102	loaded	active	running
deis-store-gateway@1.service	32e0ab49.../172.17.8.102	loaded	active	running
deis-store-metadata.service	32e0ab49.../172.17.8.102	loaded	active	running
deis-store-monitor.service	32e0ab49.../172.17.8.102	loaded	active	running
deis-store-volume.service	32e0ab49.../172.17.8.102	loaded	active	running
 
The problem is that I get 404 NOT FOUND when doing the following

$ deis pull localhost:5000/ericroubal/example-go:latest
Creating build... Error: 
404 NOT FOUND
detail: Not found

I am not logged into docker when using my own registry.   If I try doing things with docker hub with a docker login:

$ docker login
Username: xxxxx
Password:
Email: yyyyy
WARNING: login credentials saved in /Users/ericroubal/.docker/config.json
Login Succeeded

Then I get auth issues:

$ docker push ericroubal/example-go
The push refers to a repository [docker.io/ericroubal/example-go] (len: 1)
0ed2a546eea1: Image push failed 
unauthorized: access to the requested resource is not authorized

Its not important that I access docker hub, we would only realistically use our own registry.  But I added attempts at docker hub in case someone would ask me if that worked.  Any help would be appreciated greatly.   I'm using the latest deis as of today Sept 28th, 2015.

mboersma:
Removes custom code for interacting with the Docker v1 registry in favor of calling Docker itself to do the dirty work. This frees Deis to use any standard v1 or v2 Docker registry internally (which will be a follow-up PR. *Edit*: see #4641), and makes `deis pull` work with images built by Docker 1.6 or later.

Some [quick benchmarking](https://gist.github.com/mboersma/f6d001ee9826d04b7deb) shows this approach to be about 50% slower than the old way for a mix of `deis config:set` and `deis pull` operations. For example `deis config:set FOO=bar` may have taken 18 seconds before and now takes 35 seconds.

Closes #3550.
Closes #4378.
Closes #4541.

~~TODO:~~
- [x] fix controller functional test
- [x] add unit tests
- [x] blacklisting of deis/* images
- [x] error handling (involves `grep`ing the stderr)
- [x] examine other security concerns
- [x] benchmarking--this is probably somewhat slower than the old way
- [x] set default logging back to INFO, not DEBUG
- [x] use `flock` to guard docker-py build and pull operations


deisctl/units/deis-controller.service
```diff
@@ -1,14 +1,14 @@
...
+Requires=docker.socket fleet.socket
+After=docker.socket fleet.socket
 
 [Service]
 EnvironmentFile=/etc/environment
 TimeoutStartSec=20m
 ExecStartPre=/bin/sh -c "IMAGE=`/run/deis/bin/get_image /deis/controller` && docker history $IMAGE >/dev/null 2>&1 || docker pull $IMAGE"
 ExecStartPre=/bin/sh -c "docker inspect deis-controller >/dev/null 2>&1 && docker rm -f deis-controller || true"
-ExecStart=/bin/sh -c "IMAGE=`/run/deis/bin/get_image /deis/controller` && docker run --name deis-controller --rm -p 8000:8000 -e EXTERNAL_PORT=8000 -e HOST=$COREOS_PRIVATE_IPV4 -v /var/run/fleet.sock:/var/run/fleet.sock $IMAGE"
+ExecStart=/bin/sh -c "IMAGE=`/run/deis/bin/get_image /deis/controller` && docker run --name deis-controller --rm -p 8000:8000 -e EXTERNAL_PORT=8000 -e HOST=$COREOS_PRIVATE_IPV4 -v /var/lib/docker:/var/lib/docker -v /var/run/docker.sock:/var/run/docker.sock -v /var/run/fleet.sock:/var/run/fleet.sock $IMAGE"
```
carmstrong: What are we using from `/var/lib/docker`?

mboersma: Nothing actually--it can probably be omitted.

bacongobbler: Yeah if we're using the host's docker socket layers will be saved on that host anyways. Technically bind-mounting /var/lib/docker is pointless in that case. :)

mboersma: The original version of this PR actually used a `docker` binary in deis-controller, so that's where the volume mount came from. Then I realized docker-py was sufficient, but didn't clean this up.

carmstrong: :+1: Concerns about that re: potentially sharing a Docker graph.

mboersma:
I can't rule out the possibility that this code itself could trigger the "simultaneous docker pull" bug, so it needs the equivalent of a `flock` guard on its pull and build methods.

controller/registry/dockerclient.py
```diff
@@ -0,0 +1,116 @@
...
+from simpleflock import SimpleFlock
+import docker
+
+logger = logging.getLogger(__name__)
+
+
+class DockerClient(object):
+    """Use the Docker API to pull, tag, build, and push images to deis-registry."""
+
+    FLOCKFILE = '/tmp/controller-pull'
```
bacongobbler: how come we had to switch from /var/lock to /tmp here? Any particular reason?

controller/registry/dockerclient.py
```diff
@@ -0,0 +1,116 @@
...
+
+    def __init__(self):
+        self.client = docker.Client(version='auto')
+        self.registry = settings.REGISTRY_HOST + ':' + str(settings.REGISTRY_PORT)
```
bacongobbler: would it make sense to re-purpose `settings.REGISTRY_URL` here? We can probably omit the `http://` from the URL anyways since Docker doesn't explicitly care about the protocol.

controller/registry/dockerclient.py
```diff
@@ -0,0 +1,116 @@
...
+    def publish_release(self, source, config, target, deis_registry):
+        """Update a source Docker image with environment config and publish it to deis-registry."""
+        # get the source repository name and tag
+        src_name, src_tag = docker.utils.parse_repository_tag(source)
+        # get the target repository name and tag
+        name, tag = docker.utils.parse_repository_tag(target)
+        # strip any "http://host.domain:port" prefix from the target repository name,
+        # since we always publish to the Deis registry
+        paths = name.split('/')
+        name = '/'.join(p for p in paths if p and '.' not in p and ':' not in p)
```
bacongobbler: are there tests covering this "stripping"? Might be useful so we can cover edge cases in the future if necessary :)

controller/registry/dockerclient.py
```diff
@@ -0,0 +1,116 @@
...
+
+def log_output(stream):
+    for chunk in stream:
+        logger.debug(chunk)
+        # error handling requires looking at the response body
+        if '"error"' in chunk.lower():
+            raise docker.errors.DockerException(chunk)
+
+
+def publish_release(source, config, target, deis_registry):
```
bacongobbler: is there any sense in keeping this around? Might as well just let the controller retrieve a DockerClient and then publish the release so we aren't creating new connections to the socket on each release.

controller/registry/tests.py
```diff
@@ -0,0 +1,89 @@
...
+        self.assertTrue(docker_build.called)
+        args = {"rm": True, "tag": u'localhost:5000/ozzy/embryo:v4', "stream": True}
+        kwargs = docker_build.call_args[1]
+        self.assertDictContainsSubset(args, kwargs)
+        # test that the fileobj arg to "docker build" contains a correct Dockerfile
+        f = kwargs['fileobj']
+        self.assertEqual(f.read(), "FROM ozzy/embryo:git-f3a8020\nENV POWERED_BY='Deis'")
+        # Test that blacklisted image names can't be built
+        with self.assertRaises(PermissionDenied):
+            self.client.build('deis/controller:v1.11.1', {}, 'deis/controller', 'v1.11.1')
```
bacongobbler: We should just blacklist apps being created with these names. We should also blacklist anyone from creating a `deis` user so we fail early in the deployment pipeline. The main problem with doing that is backwards compatibility.

controller/registry/dockerclient.py
mboersma: deis-controller runs `gunicorn` as the "deis" user, so /var/lock isn't writable.

bacongobbler: gotcha. I knew it had to be some form of permissions error. Thanks!

controller/registry/dockerclient.py
mboersma: I was trying to minimize code change elsewhere by preserving a function as the main interface, but your suggestion would also work and might conserve resources. I'll give it a try.

controller/registry/tests.py
mboersma: > We should just blacklist apps being created with these names.

I can move the `_check_blacklist` method up to a module-level function and call it at `deis create` time.

> We should also blacklist anyone from creating a deis user so we fail early

That could be implemented in serializers.py easily. But I agree banning a user named "deis" would, strictly speaking, not be backward-compatible, nor would the previous change. We may want to do so anyway because the Docker deprecation event forces us to implement this PR (or something like it).

controller/registry/dockerclient.py
mboersma: > are there tests covering this "stripping"?

Not explicitly; I'll add some.

controller/registry/dockerclient.py
mboersma: In settings.py we populate all three: `REGISTRY_URL`, `REGISTRY_HOST`, `REGISTRY_PORT`. Joining the latter two seemed simpler than stripping the URL schema prefix off the first, which I thought would be more portable for connecting to the registry. This is what all the scheduler modules do as well: ```self.registry = settings.REGISTRY_HOST+":"+settings.REGISTRY_PORT```

controller/registry/tests.py
bacongobbler: > I can move the _check_blacklist method up to a module-level function and call it at deis create time.

We could implement that in serializers.py as well!

> We may want to do so anyway because the Docker deprecation event forces us to implement this PR (or something like it).

Indeed. Maybe a big fat warning or an opt-out feature flag would be handy for users that rely on a `deis` user along with these app names.

Issue #4585, opened 2015-10-08
Deis should send host header with health checks.


pmclanahan:
Many apps validate the "Host:" header as a security measure (e.g. Django does this by default). As such health check pings using just the IP will return a 400 unless the IP is specifically added to the config. The various health checks should therefore add a host header to the requests with the value being the FQDN for the app (e.g. `Host: friendly-dino.example.com`) so that apps can continue simply including the expected domains in the configuration.

bacongobbler:
good thought.

jgmize:
Note that we'll need to add the host header to the publisher health checks in addition to the controller health checks @pmclanahan has modified in #4586.

bacongobbler:
@jgmize can you confirm if kubernetes health checks perform HTTP health checks as expected? That way we can determine if this is something we need to contribute upstream to fix for v2 or not.

bacongobbler:
closing as something fixed in v2 with kubernetes' health checks.

PR #4586, opened 2015-10-08
feat(controller): add host header to healthcheck

pmclanahan:
re: #4585


deis-admin:

bacongobbler:
Thanks @pmclanahan! Would you mind amending your commit such that it follows the style guide as our bot pointed out? That way your commit will be fetched in the changelog when we cut a new release. Thanks! :)

bacongobbler:
also @jgmize does have a point with https://github.com/deis/deis/issues/4585#issuecomment-146676337 -- we'll need to add the host header there as well.

pmclanahan:
Commit message improved. And yes, @jgmize is correct, but I thought I'd go ahead and get this in since I know how to make this change. I'll look at the other as well, but I've not worked with much golang.

mboersma:
Jenkins, add to whitelist.

bacongobbler:
Unfortunately I don't think there's a way we can get the application's URL from publisher, and I think we need to implement this in both publisher and the controller in one PR. On the plus side, health checks are handled by kubernetes for us in v2, so there is a fix in the future for this as they'll handle all this logic for us, including [HTTP and exec health checks](https://github.com/kubernetes/kubernetes/tree/master/docs/user-guide/liveness) which the healthcheck PR was originally inspired by. :)

@pmclanahan would you be okay for closing this PR in favour of v2? You can follow the docs [here](https://github.com/deis/workflow) if you want to play with v2. It's currently in alpha and we're working on a beta release.

pmclanahan:
Sounds good. Thanks.

PR #4606, opened 2015-10-13
fix(client): read procfile from PWD when pulling

### LINKED ISSUE #4605: bug(client): New deis go client does not read `Procfiles` from current directory during `deis pull`
The [old behavior](https://github.com/deis/deis/blob/v1.9.1/client/deis.py#L1038-L1052) of `deis pull some-image -a myapp --procfile` would read a file named `Procfile` from the current directory. As of `deis-1.10`, the client [no longer reads `Procfile`](https://github.com/deis/deis/blob/v1.10.1/client/cmd/builds.go#L45-L53).

CC @Joshua-Anderson

Joshua-Anderson:
This was the behavior of the python client.

Fixes #4605

### DIFF
--- a/client/cmd/builds.go
+++ b/client/cmd/builds.go
@@ -2,6 +2,8 @@ package cmd
 
 import (
 	"fmt"
+	"io/ioutil"
+	"os"
 
 	"gopkg.in/yaml.v2"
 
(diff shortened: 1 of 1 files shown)


Joshua-Anderson:
Does anybody know what is going on with CI? Every build I try times out with a restart looping`store-monitor`

sgoings:
@Joshua-Anderson we are aware of those issues, and have been trying to resolve. The issue you're seeing with the restart loop on store-monitor is caused by consistent i/o timeouts to Docker Hub (all pulls of deisci/* have this behavior when clusters are created in AWS). :cry: I believe @mboersma has a support ticket in with Docker to figure out why this is happening.

krancour:
@Joshua-Anderson, it doesn't look like you're doing anything with the Procfile.  Is this just a validation step, essentially?

Joshua-Anderson:
@krancour if a Procfile exists in the PWD, it will be read and parsed, allowing users to define custom processes for docker images.

krancour:
I see now.  I had overlooked that the procfileMap was being modified by the parseProcfile function.  This is a total nit, but in my experience, it's better to return things like that than to pass a pointer to simulate pass by reference semantics, simply because the former makes it harder to overlook that a new value has been assigned to something.  You could drop the procfileMap pointer from the argument list, make a new one within the function, and return that.  Just a thought.  Either way, lgtm.

Joshua-Anderson:
@krancour Done! I agree returning things rather than modifying a pointer is a more clear approach.

carmstrong:
Would like @mboersma or @bacongobbler to review this one

PR #4608, opened 2015-10-13
fix(contrib/coreos): allow fleet to infer public IP

carmstrong:
fleet will use the host's default route to set this.

See https://github.com/coreos/fleet/blob/master/Documentation/deployment-and-configuration.md#public_ip

I'd like to test this on the various providers. We've seen cases lately where `fleetctl list-machines` shows the public IP for each member in the cluster (I believe it used to show the private IP), and this address is how the fleet daemon attempts to communicate with the other hosts in the cluster for commands like `fleetctl ssh`.

I'm hoping this change will let fleet do the "right thing" in most cases.

/cc @jgmize @deis/production-team


carmstrong:
I tested the AWS provision with this change and things look OK:

```console
core@ip-10-21-1-232 ~ $ fleetctl list-machines
MACHINE		IP		METADATA
1795a50f...	10.21.1.232	controlPlane=true,dataPlane=true,routerMesh=true
4c1b2f9c...	10.21.2.216	controlPlane=true,dataPlane=true,routerMesh=true
c88176ce...	10.21.2.215	controlPlane=true,dataPlane=true,routerMesh=true
core@ip-10-21-1-232 ~ $ cat /etc/environment
COREOS_PRIVATE_IPV4=10.21.1.232
COREOS_PUBLIC_IPV4=52.91.179.216
```

carmstrong:
Note that coreos-vagrant is still doing this, but I suspect it's primarily for Vagrant clusters: https://github.com/coreos/coreos-vagrant/blob/master/user-data.sample#L15

helgi:
Would be mostly concerned about Azure given how it often rather special things with networking, if you can spin up a cluster and drop the same info in the ticket

carmstrong:
I'd like to find out why CoreOS is setting this by default in all their cloud-config samples -- previously I thought it was only for Vagrant, but it's in their main example too: https://coreos.com/os/docs/latest/cloud-config.html#fleet

@bcwaldon What's your take on this? Do we omit `public-ip` and trust cloud-init/fleetd to detect this based on the default route, or do we set this to `$public_ipv4` everywhere and deal with provider-specific issues by changing this to `$private_ipv4` in some cases?

bcwaldon:
@carmstrong It really depends on what level of routability to fleet you need. Using AWS as an example: if you are comfortable publishing the "private" IP as the address of the fleet machine and having to SSH into the cluster before running fleet commands, then you can rely on fleet autodetecting the machine IP address. It will pick up the private IP in this case. Alternatively, if you want to publish the externally-routable IP, you need to set public-ip in the cloud-config to $public_ipv4.

This documentation did stem from our experience on Vagrant, where the autodetection is completely broken.

carmstrong:
> This documentation did stem from our experience on Vagrant, where the autodetection is completely broken.

Good to know - I figured that's the motivation for this. I think #4635 resolves the immediate concern motivating this PR, so I'll close for now. Thanks @bcwaldon!

PR #4590, opened 2015-10-09
fix(*): use flock to serialize certain pulls

krancour:
Fixes #4584 by serializing `docker pull` among units pulling images with layers in common.  This is applied only to global units that are most likely to start in parallel as the result of a new node being added to a cluster.

Attn: @mboersma @carmstrong @gabrtv

PR #4609, opened 2015-10-13
fix(*) add flocks to additional services

jgmize:


deis-admin:

krancour:
router and database could also need this.

krancour:
Jenkins, add to whitelist.

krancour:
These lines show components that are starting in parallel and may share layers beyond just alpine:

https://github.com/deis/deis/blob/master/deisctl/cmd/cmd.go#L156-L165

carmstrong:
For reference, the root Docker issue is https://github.com/docker/docker/issues/9718. They claim it's fixed in master and will ship in Docker 1.9, but it would be great to confirm that.

jgmize:
> router and database could also need this.

Updated to include flocks for router and database.

krancour:
@jgmize, thank you for this!

@carmstrong one other thing we should consider... is _not_parallelizing the startup of these components in deisctl possibly _better_ than using more flock locks?  Or maybe should we do both?

carmstrong:
> is not parallelizing the startup of these components in deisctl possibly better than using more flock locks

I think doing them in the units is probably safest, as anyone running our components would get that benefit. It's also easier to revert when (and if) we get this fixed in Docker. See also #4610

carmstrong:
For builder and database, I'm wondering if we also need to lock the `*-data` containers as we did in #4590.

jgmize:
> For builder and database, I'm wondering if we also need to lock the *-data containers as we did in #4590.

good catch, @carmstrong. I just added a commit to do that.

krancour:
@jgmize we really appreciate this, but we're now getting reports of other problems with Docker 1.7.1 besides this.  We have a few balls in the air right now in terms of possibly rolling back to Docker 1.5 or shimming in a custom build of Docker.  One way or another, we won't end up using flock locks for this purpose-- and even the ones I had already added to logspout and publisher will certainly be removed as well.

Thanks again for the effort on this.

mboersma:
#4614 is having some issues with `deis run`, so this `flock` approach may still be worth pursuing. I'm going to reopen it to see how it fares in CI--sorry for the whiplash!

deisctl/units/deis-builder.service
```diff
@@ -4,8 +4,9 @@ Description=deis-builder
 [Service]
 EnvironmentFile=/etc/environment
 TimeoutStartSec=30m
-ExecStartPre=/bin/sh -c "docker inspect deis-builder-data >/dev/null 2>&1 || docker run --name deis-builder-data -v /var/lib/docker alpine:3.2 /bin/true"
-ExecStartPre=/bin/sh -c "IMAGE=`/run/deis/bin/get_image /deis/builder` && docker history $IMAGE >/dev/null 2>&1 || docker pull $IMAGE"
+ExecStartPre=/bin/sh -c "IMAGE=alpine:3.2 && docker history $IMAGE >/dev/null 2>&1 || flock -w 1200 /var/run/lock/alpine-pull docker pull $IMAGE"
+ExecStartPre=/bin/sh -c "IMAGE=alpine:3.2 && docker inspect deis-builder-data >/dev/null 2>&1 || docker run --name deis-builder-data -v /var/lib/docker $IMAGE /bin/true"
+ExecStartPre=/bin/sh -c "IMAGE=`/run/deis/bin/get_image /deis/builder` && docker history $IMAGE >/dev/null 2>&1 || flock -w 1200 /var/run/lock/alpine-pull docker pull $IMAGE"
```
bacongobbler: maybe we should change `alpine-pull` to `builder-pull` here?

bacongobbler: oh wait, I see the idea/usage behind `flock` now. Disregard.

bacongobbler:
@krancour you did make a good point:

> These lines show components that are starting in parallel and may share layers beyond just alpine

I know we crafted a few lines in each component so that Jenkins can share these layers between images. I do agree that we share layers beyond alpine. We'll probably have to ensure that the images do **not** share layers somehow, or just `flock` every image pull, not just alpine. That means boot time will be significantly slower, though.

mboersma:
> or just flock every image pull, not just alpine

This PR in conjunction with #4590 should mean that all `docker pull` operations in systemd units are guarded by `flock` now, with one lock for alpine-based images and one for ubuntu-based ones.

bacongobbler:
Ah! Okay, maybe I was confused as to how `flock` worked. With `alpine-pull` being the lock name, I was confused that only the alpine image layers would be locked. I understand now :)

bacongobbler:
manually tested, works as intended.

krancour:
Thanks @jgmize!

Issue #4610, opened 2015-10-13
Consider shipping a custom Docker build


carmstrong:
Lately, we've been bit by several major Docker bugs, including:
* https://github.com/docker/docker/issues/9718 (simultaneous pulls)
* https://github.com/docker/docker/issues/8539 (out-of-memory issues)

Issues/PRs affected:
#4609
#4590
#4238
#2975
#4536

As it stands now, there isn't a version of Docker that we consider stable that is even remotely recent (pre-Docker 1.5). Docker claims that the simultaneous pull issue is fixed in master, but Docker 1.9 won't ship for several more weeks, and it's likely that it won't make it to CoreOS stable for another month or two after that. We should investigate building our own and shipping that on CoreOS.

krancour:
It seems we've elected not do do this.  Closing.

PR #4612, opened 2015-10-13
Revert to Docker 1.5

mboersma:
Reverts the changes in #4438 and #4590.


carmstrong:
This is necessary because of continual issues with the Docker daemon. See #4610.

mboersma:
Without additional changes to reinstall `etcd2` and `flanneld` as before, this won't work. Instead let's try to drop Docker 1.5.0 onto current CoreOS: see #4614.

PR #4614, opened 2015-10-14
fix(user-data): install Docker 1.5.0

mboersma:
Installs Docker 1.5.0 on top of current CoreOS to work around docker/docker#9718.


krancour:
@mboersma, we leave early docker alone?

mboersma:
> we leave early docker alone?

The early-docker.service hasn't been implicated in any bugs that we've found. I think it only bootstraps `flanneld`. (I also assumed it was one of the core services we can't modify through cloud-config, like update-manager.service, but I haven't verified this.)

krancour:
> The early-docker.service hasn't been implicated in any bugs that we've found.

SGTM then.

carmstrong:
Are there any concerns with the upgrade path if someone provisioned a fresh 1.11.1 cluster with the new CoreOS and latest Docker? Will they run into issues if they upgrade to these units?

krancour:
@carmstrong, I think anyone who provisiobed using userdata from 1.11.0 or 1.11.1 has the bad docker and will be error prone for as long as they have that.  Maybe we need to reconsider keeping flocks, and even adding more, as in @jgmize's recent PR.

mboersma:
The current issue preventing this from working is that `deis run` never returns because the stock Docker 1.5.0 binary doesn't return from [this line](https://github.com/deis/deis/pull/4614/files#diff-43358db3bfa05551b90bcf90c48dfe2cR375), which wasn't changed in this PR. I will try CoreOS' own `docker/dockerinit` 1.5.0 and see if that gives better results.

I still think this is a better approach than `flock` everywhere.

krancour:
@mboersma, I have no love for the flocks either, but what happens to anyone who provisioned a cluster using 1.11.x userdata and got stuck with Docker 1.7?  When they upgrade to 1.12 or beyond, they may still have problems-- which is the only reason I wonder if the flocks may still be a necessary evil.

mboersma:
Closing this in favor of the `flock` workaround in #4609. Assuming we could get past the `docker run` hang that keeps this from passing, there are good reasons *not* to roll back to Docker 1.5:
- won't work for existing clusters since the changes are in cloud-config
- `docker pull` will stop working on December 7 for v1.5.0 and earlier clients

jgmize:
> docker pull will stop working on December 7 for v1.5.0 and earlier clients

Would you mind providing a link to more details on this?

gabrtv:
> Would you mind providing a link to more details on this?

I can't find a link on the web which is _very_ strange.  I've attached a screenshot of the email.
screen shot 2015-10-19 at 2 46 43 pm

PR #4635, opened 2015-10-20
fix(user-data): use $private_ipv4 for fleet

### LINKED ISSUE #4594: Coreos user-data file uses public ip for fleet, firewall script brings fleet down
Tested on Deis 1.11.

User-data example file in contrib/coreos/user-data.example is using public ip for fleet. As etcd works with private ip, applying custom_firewall.sh on digital ocean causes fleet to fail with etcd errors.

I was able to work around this, by replacing
```
fleet:
    public-ip: $public_ipv4
```
with 
```
fleet:
    public-ip: $private_ipv4
```
in contrib/coreos/user-data file , after running ```make discovery-url```

mboersma:
Reverts this networking configuration to the interface it used previous to the v1.11.0 release, which was changed in 445dbaecfadb1e2690734413cc43f30abcb13b21.

Closes #4594.
Closes #4608.

### DIFF
--- a/contrib/coreos/user-data.example
+++ b/contrib/coreos/user-data.example
@@ -13,7 +13,7 @@ coreos:
     listen-peer-urls: http://$private_ipv4:2380,http://$private_ipv4:7001
     data-dir: /var/lib/etcd2
   fleet:
-    public-ip: $public_ipv4
+    public-ip: $private_ipv4
     metadata: controlPlane=true,dataPlane=true,routerMesh=true
   update:
     reboot-strategy: "off"


nathansamson:
Is there a migration path for people coming from 1.11? (without rebuilding cluster)

I would like to be able to use the firewall script properly for new nodes etc...

krancour:
I agree with this change.  It reverses something that we probably shouldn't have done.

We need to think about the migration path some more.

carmstrong:
Ahhhhhh this was the change that I was so confused why behavior had changed! I couldn't figure out what had changed where. :100:

carmstrong:
This probably also closes #4608.

mboersma:
I tested this change on DigitalOcean with the [firewall scripts](http://docs.deis.io/en/latest/installing_deis/digitalocean/#apply-security-group-settings) and `fleet` stays working.

> Is there a migration path for people coming from 1.11? (without rebuilding cluster)

Since this is the service definition in cloud-config / user-data, it's not obvious how v1.11.0 and v1.11.1 clusters can incorporate this change (which will be in v1.11.2) without provisioning a new cluster. Can we provide a modified version of the firewall scripts, or is there some way to update the `fleet.service` definition on each box?

carmstrong:
> Would a modified version of the firewall scripts work, or is there some way to update the fleet.service definition on each box?

Anything that changes provisioning, firewalls, or host-level things should require manual intervention, IMO. I'd err on the side of pointing users to this change should they run into the issue.

PR #4641, opened 2015-10-20
[WIP] ref(registry): use docker/distribution 2.1.1+

### LINKED ISSUE #3814: Upgrade registry to docker-registry v2 (Docker Distribution)

mboersma:
Builds on #4581 to get rid of custom [deis-registry V1 code](https://github.com/docker/docker-registry/compare/master...deis:new-repository-import-master) and use the supported V2 registry instead.

The deis-registry code builds from the ["radosgw" branch](https://github.com/docker/distribution/compare/master...deis:radosgw) of deis/distribution, which includes S3 support for Ceph. Deis could use the stock `registry:2` image once docker/distribution#808 is merged and built into a new official release (although this packaging with `alpine:3.2` is smaller).

It was necessary to add `flock` to the systemd app units that deis-controller generates to work around  docker/docker#9718. Otherwise `deis scale web=10` could corrupt the docker graph.

Closes #3814.

TODO:
- [ ] utilize v1 -> v2 migration tool automatically on upgrade--sadly, [docker/migrator](https://github.com/docker/migrator) won't work as-is
- [ ] update the `fsouza/go-dockerclient` library for 1.8.3?
- [ ] extended testing on AWS + Ceph S3
- [ ] benchmarks: reg v1 vs reg v2 vs reg v2 + cache


builder/rootfs/Dockerfile
```diff
@@ -26,7 +26,7 @@ RUN apk add --update-cache \
     && rm -rf /var/cache/apk/*
 
 # the docker package in alpine disables aufs and devicemapper
-RUN curl -sSL https://get.docker.com/builds/Linux/x86_64/docker-1.5.0 -o /usr/bin/docker && \
+RUN curl -sSL https://get.docker.com/builds/Linux/x86_64/docker-1.8.3 -o /usr/bin/docker && \
```
carmstrong: Any potential for running into the simultaneous layer pull issue? It may be time to update this regardless, I suppose, with the EOL for Docker < 1.5.

registry/rootfs/etc/confd/templates/config.yml
```diff
@@ -0,0 +1,59 @@
...
+    tenant: {{ if exists "/deis/registry/swiftTenantName" }}{{ getv "/deis/registry/swiftTenantName" }}{{ else }}~{{ end }}
+    tenantid: {{ if exists "/deis/registry/swiftTenantID" }}{{ getv "/deis/registry/swiftTenantID" }}{{ else }}~{{ end }}
+    domain: {{ if exists "/deis/registry/swiftDomain" }}{{ getv "/deis/registry/swiftDomain" }}{{ else }}~{{ end }}
+    domainid: {{ if exists "/deis/registry/swiftDomainID" }}{{ getv "/deis/registry/swiftDomainID" }}{{ else }}~{{ end }}
+    insecureskipverify: true
+    region: {{ if exists "/deis/registry/swiftRegionName" }}{{ getv "/deis/registry/swiftRegionName" }}{{ else }}~{{ end }}
+    container: {{ if exists "/deis/registry/swiftContainer" }}{{ getv "/deis/registry/swiftContainer" }}{{ else }}~{{ end }}
+    rootdirectory: /registry
+  {{ else }}
+  s3:
```
carmstrong: I'm surprised this works without specifying something similar to the calling format like we used to: https://github.com/deis/deis/blob/f70bc126914d48909863ee1df81a7a79749ce104/registry/templates/config.yml#L106

builder/rootfs/Dockerfile
mboersma: I think it's possible this could lead to the "simultaneous docker pull" bug, so I will investigate adding the equivalent of a `flock` guard to the relevant builder code.

registry/rootfs/etc/confd/templates/config.yml
mboersma: It does seem to work, but I plan more extensive testing on AWS to ensure Ceph stays happy with this.

bacongobbler:
hmm... will data migrations work? Most of the data in ceph will be geared towards the v1 registry. Do we have a migration in place for that or is this strictly v2?

mboersma:
> will data migrations work?

There is a v1 -> v2 migration tool, which still needs to be incorporated in this PR.

PR #4642, opened 2015-10-20
feat(controller): disable swap if there's a mem limit

### LINKED ISSUE #3205: Proposal ref(controller): disable swap usage if there is a memory limit

mboersma:
Resurrects #3214 to disable memory swap usage if the user explicitly sets app memory limits with `deis limits:set web=256M` or the like.

Now that Deis is well past Docker 1.5.0, this is viable, although to be backward-compatible the Docker version still needs to be checked so this feature can be disabled. This PR builds off #4581 to query the Docker version through docker-py. ~~Reviewers can read just the last commit 7389998720fd89fa592b25fec8ce0390883ba19d.~~ *Edit:* rebased now that #4581 is merged.

Closes #3205. ping @aledbf and @manuwell.


aledbf:
Code LGTM

PR #4643, opened 2015-10-20
[WIP]: feat(platform): end-to-end HTTPS

### LINKED ISSUE #4629: Proposal: Enable opt-in HTTPS between router and app containers
See issue #4611 

Deis currently uses "mullet TLS--" security in the front, party in the back.  Having come from healthcare, I understand all too well how this rules out the possibility of using Deis to host any workloads that deal with PHI.  This is a gap that needs to be closed.

This issue isn't meant to duplicate #4611, but rather to serve as a forum for discussing proposed designs that address this.  I'll start by throwing my proposal into the ring.  Feedback is welcome, as are alternate proposals.

Approach
-------------

I think we can pretty easily go 80/20 on this and get ourselves most of the way to where we need to be pretty quickly and then improve incrementally from there going forward.  Initially, I'd propose that we support the the following, at minimum:

* 100% opt-in on an application basis so as not to impact existing, deployed applications upon platform upgrade.
* _Option_ for end-to-end HTTPS needs to be there.  If when selecting this option, it proves infeasible to _close_ the HTTP route into the same application in the first iteration, so be it.
* Self-signed certificates are good enough, initially.  It's only the router that needs to trust the certs-- not the end user's agent (since Deis can already terminate TLS at the router using a proper cert provided by the operator, signed by a known and trusted CA).

Possible implementation
---------------------------------

I believe operators should opt-in to this on an application basis by setting an environment variable via `deis config:set`.

The following is all WRT to buildpack-based applications:

One possible implementation is to generate a self-signed certificate at build-time.  I imagine the easiest way of doing this being modifications to the slugbuilder such that after the buildpack(s) have generated the application's deployable artifacts, the generated cert and key are added to those right before compressing those artifacts into a slug (tarball).  A corresponding change to the slugrunner can start Nginx as a proxy in front of the application to terminate TLS.

Open questions and concerns
----------------------------------------

I think experimentation is the best way to gain insight into the following:

* Will application containers _all_ include a self-signed cert and Nginx?  Or just those that have opted into it?  This depends on the availability of relevant information (environment variable) at build time.  Fortunately, the environment variables for each application are _also_ inserted into etcd, so we probably do have access to this information at build-time.

* If one opts into end-to-end HTTPS, can we expose the secure port in the application containers _instead of_ the insecure port?  Or, initially, do we expose it _in addition to_ the insecure port?  Proposing we shoot for the former; settle for the latter if necessary (for the first iteration).

* How are routing and publishing affected?  Publishing may or may not be affected depending on the previous bullet.  Routing rules will be slightly different for anyone any app that opts into this.

* How does this affect health checks and liveliness checks?

* How does it work with k8s as the scheduler?  With k8s as the scheduler, publisher isn't really relevant.  Router proxies service URLs.  Those service URLs likely need to change for any app that has opted into this.

* How does this work for developers pushing their own Dockerfile or for those using `deis pull`?  My personal take is that anyone who's already exercising that level of control over their images is responsible for exposing the secure endpoint themselves.  We would just need to clearly document how the router plugs into that in such scenarios.

krancour:
Fixes #4629 

This is a WIP PR aimed at exploring one possible implementation of end-to-end HTTPS.

More commits are to follow.

cc: @helgi


krancour:
This PR set us out on a path of solving e2e HTTPS in a fashion that would have worked for all schedulers _except_ k8s.  A separate strategy would have been required for k8s.  Since we've since elected to focus on solving this _only_ for k8s, with hopes of the solution carrying over to V2, the code in this PR is not needed for any reason.  As such, I'm closing this and will open a new PR when the k8s-oriented solution is ready.

PR #4647, opened 2015-10-21
fix(units): Add a more explicit restart policy

### LINKED ISSUE #4632: Application restart policies
Right now if a deployed app fails to start, fleet will try and restart it forever. This seems wrong as it really pollutes the docker event stream and could make debugging difficult. Having a set restart policy seems like a good thing.

jchauncey:
Fixes #4632
Added two stanzas to the unit files for all deis components:
StartLimitInterval and StartLimitBurst. This tells systemd to only allow
a certain number of restarts to occur (StartLimitBurst) within the
specified interval (StartLimitInterval). If StartLimitBurst+1 retry
fails systemd will not try to start the application again. The counter
is reset after the interval.

Also added the above stanzas to the fleet template for user
applications.

### DIFF
--- a/controller/scheduler/fleet.py
+++ b/controller/scheduler/fleet.py
@@ -377,6 +377,8 @@ CONTAINER_TEMPLATE = [
     {"section": "Service", "name": "TimeoutStopSec", "value": "10"},
     {"section": "Service", "name": "RestartSec", "value": "5"},
     {"section": "Service", "name": "Restart", "value": "on-failure"},
+    {"section": "Service", "name": "StartLimitInterval", "value": "60s"},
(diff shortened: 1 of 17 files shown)


carmstrong:
Added my comment from review in https://github.com/deis/deis/issues/4632#issuecomment-149976182

mboersma:
@jchauncey do you think this could be brought over the finish line in the next week or two so it can go in the Deis v1 LTS release? I know you're busy with Deis v2 tasks, so if this isn't on your near-term roadmap we may want to close it.

jchauncey:
I say we close it. This was only useful when we were wanting to monitor v1 in a more meaningful way

PR #4648, opened 2015-10-21
ref(aws): prevent replacement of ASG in CF update

jgmize:
This builds on #4534 to also prevent the replacement of the auto scaling group, which will also replace all of the EC2 instances in the ASG.

### DIFF
--- a/contrib/aws/stack_policy.json
+++ b/contrib/aws/stack_policy.json
@@ -7,7 +7,7 @@
       "Resource" : "*",
       "Condition" : {
         "StringEquals" : {
-          "ResourceType" : ["AWS::EC2::Instance"]
+          "ResourceType" : ["AWS::EC2::Instance", "AWS::AutoScaling::AutoScalingGroup"]
         }
       }
     },


jgmize:
Unfortunately this change does not appear to be sufficient either; making a change to ASG in the CF template still causes the ASG to be deleted; I'm going to try adding "Update:Delete" to https://github.com/deis/deis/pull/4648/files#diff-efcb7b1fa718d07343c206bc34462d56L6 and test that.

jgmize:
> Unfortunately this change does not appear to be sufficient either; making a change to ASG in the CF template still causes the ASG to be deleted; I'm going to try adding "Update:Delete" to https://github.com/deis/deis/pull/4648/files#diff-efcb7b1fa718d07343c206bc34462d56L6 and test that.

Scratch that, testing error: I unintentionally ran an update with the current stack_policy.json instead of the version in this PR. The "Update:Delete" was not necessary, and should probably be avoided anyway as I believe that would prevent you from intentionally scaling down the number of EC2 instances.

PR #4651, opened 2015-10-22
ref(aws): Add SSL_CERT_ID var to term SSL at ELB

jgmize:
If this seems like an acceptable solution to support terminating SSL at the ELB instead of using the proxy protocol let me know, and I'll add docs to this PR.

### DIFF
--- a/contrib/aws/gen-json.py
+++ b/contrib/aws/gen-json.py
@@ -145,4 +145,21 @@ if VPC_ID and VPC_SUBNETS and VPC_ZONES and len(VPC_SUBNETS.split(',')) == len(V
     template['Resources']['CoreOSServerAutoScale']['Properties']['VPCZoneIdentifier'] = VPC_PRIVATE_SUBNETS.split(',')
     template['Resources']['DeisWebELB']['Properties']['Subnets'] = VPC_SUBNETS.split(',')
 
+SSL_CERT_ID = os.getenv('SSL_CERT_ID', None)
+if SSL_CERT_ID:
+    template['Resources']['DeisWebELB']['Properties']['Listeners'] = [
+        {'LoadBalancerPort': '80',
+         'Protocol': 'HTTP',
+         'InstancePort': '80',
+         'InstanceProtocol': 'HTTP'},
+        {'LoadBalancerPort': '443',
+         'Protocol': 'HTTPS',
+         'SSLCertificateId': SSL_CERT_ID,
+         'InstancePort': '80',
+         'InstanceProtocol': 'HTTP'},
+        {'LoadBalancerPort': '2222',
+         'Protocol': 'TCP',
+         'InstancePort': '2222',
+         'InstanceProtocol': 'TCP'}]
+
 print json.dumps(template)


bacongobbler:
sounds to me like a reasonable change. +1

Blystad:
Has AWS got support for websockets now on ELB? That was the origin of the proxy protocol patch.

If not, I don't see a point in adding yet another conditional/configuration option when this can be solved with the existing solution + a platform certificate.

jgmize:
> Has AWS got support for websockets now on ELB (with HTTP/HTTPS)? That was the origin of the proxy protocol patch.

> If not, I don't see a point in adding yet another conditional/configuration option when this can be solved with the existing solution + a platform certificate.

An understandable perspective, which I will address by updating this PR with docs explaining our rationale, but the main reason is that terminating SSL at the deis ngnix routers depends on SNI, which older clients do not support. There are also performance benefits to terminating SSL at the ELBs. 

> EDIT: In addition, with this, you need to remember to disable the proxyProtocol configuration option, otherwise deis-router will fail (since nginx will expect every http transmission to start with "PROXY ...")

Yes, I will be sure to include this in the docs in an update to the PR.

> EDIT2: Checked the AWS forums, still no Websocket support over HTTP/HTTPS.

Yes, and this configuration would prevent the use of websockets-- an unfortunate tradeoff that I will also document.

Blystad:
Can't you use the SSL configuration on the ELB?

jgmize:
> Can't you use the SSL configuration on the ELB?

Are you asking if we can reconfigure the ELB after provisioning? That's the way we did it for a while; I used this customization on our most recent cluster build because I was tired of tweaking things via the AWS web console.

Blystad:
Well, either post-provisioning (by using the AWS CLI, I change a few settings post-provisioning using that myself), or by modifying only the #2 argument in 

```python
template['Resources']['DeisWebELB']['Properties']['Listeners'][1]['Protocol'] = 'SSL';
template['Resources']['DeisWebELB']['Properties']['Listeners'][1]['SSLCertificateId'] = SSL_CERT_ID;
```

(I'm not sure the above syntax will work)

jgmize:
> Well, either post-provisioning (by using the AWS CLI, I change a few settings post-provisioning using that myself), or by modifying only the #2 argument in

> template['Resources']['DeisWebELB']['Properties']['Listeners'][1]['Protocol'] = 'SSL';
> template['Resources']['DeisWebELB']['Properties']['Listeners'][1]['SSLCertificateId'] = SSL_CERT_ID;

No, we need to change the `Protocol` and `InstanceProtocol` values as well, and depending on the order of the listeners seemed brittle, so I went ahead and replaced the whole list.

Blystad:
At the same time, replacing the whole list results in data duplication.

I change `Protocol` in my suggested change. I don't understand why you need to change the `InstanceProtocol` to HTTP.

jgmize:
My mistake, not sure how I missed that first line. It also hadn't occurred to me that I could use TCP for the InstanceProtocol together with HTTP or HTTPS as the ELB Protocol. I'll have to test that, because if that works it will greatly simplify things for us.

Blystad:
I think you lost me :)

What I propose is changing the following:

```json
        "Listeners": [
          {
            "InstancePort": "80",
            "InstanceProtocol": "TCP",
            "LoadBalancerPort": "80",
            "Protocol": "TCP"
          },
          {
            "InstancePort": "443",
            "InstanceProtocol": "TCP",
            "LoadBalancerPort": "443",
            "Protocol": "TCP"
          },
          {
            "InstancePort": "2222",
            "InstanceProtocol": "TCP",
            "LoadBalancerPort": "2222",
            "Protocol": "TCP"
          }
        ]
```

To
```json
        "Listeners": [
          {
            "InstancePort": "80",
            "InstanceProtocol": "TCP",
            "LoadBalancerPort": "80",
            "Protocol": "TCP"
          },
          {
            "InstancePort": "443",
            "InstanceProtocol": "TCP",
            "LoadBalancerPort": "443",
            "SSLCertificateId": SSL_CERT_ID,
            "Protocol": "SSL"
          },
          {
            "InstancePort": "2222",
            "InstanceProtocol": "TCP",
            "LoadBalancerPort": "2222",
            "Protocol": "TCP"
          }
        ]
```

This should work. I used to do that before we had proxyProtocol support, and I created a seperate ELB for a service which required websockets.

jgmize:
Ah, thanks, I misunderstood. That makes much more sense to me now. :)

mboersma:
@jgmize is this SSL_CERT_ID change still in progress?

sstarcher:
@Blystad with that configuration do you need to disable proxy protocol support?  I believe earlier this week I tried that same configuration and it did not seem to work for me.

Blystad:
Provided you do it via cloud formation it should work. If you do it afterwards you might have to reapply the configuration telling ELB to send proxy protocol packets to deis-router for requests on port 443 (I'm not sure if the UI overwrites this).

Do you have a log? Then I'd be happy to take a look.

sstarcher:
@Blystad I just verified that you sir are correct.  I'm not sure what my issue was on my previous cluster, but SSL -> TCP termination on the ELB is working great.

Blystad:
Great :)

jgmize:
> @jgmize is this SSL_CERT_ID change still in progress?

I haven't had much chance to work on this, but I hope to test out @Blystad's suggested changes on a test cluster in the next week or so and update the PR accordingly.

sstarcher:
@jgmize one thing to note that caused me a little indijestion is our backend servers were setup to redirect http -> https and were checking the X-Forwarded-Proto to see if the original request had been https.  If you're going from SSL->TCP and not from HTTPS->HTTP the ELB will not set X-Forwarded-Proto

mboersma:
> but I hope to test out @Blystad's suggested changes on a test cluster in the next week or so and update the PR

Awesome, thanks so much. I'm just trying to see what we should include in v1.12, so I'll put this on tap for the subsequent Deis release.

sstarcher:
@jgmize any interest in supporting multiple ELBs with multiple SSL certs.  If so I could stop patching my version of Deis to add 2 additional ELBs ;)

jgmize:
> @jgmize any interest in supporting multiple ELBs with multiple SSL certs. If so I could stop patching my version of Deis to add 2 additional ELBs ;)

The patch in its current form requires that you run an ELB for each additional SSL cert, because once you disable the PROXY protocol globally, you lose the ability to terminate SSL at the deis routers for any domain.

You could create the additional ELBs in cloudformation during the initial creation of the deis cluster, but you cannot add new ones with a cloudformation update without it deleting the entire autoscaling group and all EC2 instances in it-- a lesson I learned the hard way (see #4648).

Instead, we only use CF to create the the first ELB, and create the others outside of CF and attach them to the Deis ASG using the aws cli:

```sh
for ELB in $ELBS; do
    aws elb apply-security-groups-to-load-balancer \
        --load-balancer-name $ELB \
        --security-groups $DEISWEB_SG
    aws autoscaling attach-load-balancers \
        --auto-scaling-group-name $ASG \
        --load-balancer-names $ELB
done
```

sstarcher:
@jgmize you are correct, but this patch in it's current form allows for the creation of a new stack that contains 2 ELBs.  We do the same thing, but we create 3 ELBs.  1 internal for Deis and 2 different ELBs that terminate 2 separate certs.

sstarcher:
Essentially what I was thinking was changing 
+SSL_CERT_ID = os.getenv('SSL_CERT_ID', None)
to be a comma separated list of SSL certs that would create an ELB per cert.

If not I'll essentially do the same thing, but on a branch of Deis.

jgmize:
@Blystad I finally got a chance to take a deeper look at your suggested changes on this, and found that we lose the ability to get the real client IP address for the request in that configuration. We need this for geo-targetted redirects-- specifically, https://www.mozilla.org/firefox/os/ redirects to different versions based on device & language availability. It's possible to get the client IP when using the PROXY protocol, but as far as I can tell it is not with TCP, so we will need to stick with the HTTP and HTTPS protocols for the ELBs when not using PROXY. This also matches the configuration of previous versions of Deis. I still need to follow up with some docs for this PR, but I need to finish a few other tasks before I'll be able to get back to this again.

Blystad:
Regarding the IP address, yeah I understand your requirement. The original reason I got the ball rolling with the PROXY Protocol patch was that I needed to get the clients IP address & have WebSocket support (which AWS still...in 2015...does not have proper support for on HTTP-mode ELB...). Originally, before I got the patch into mainstream I ran two ELBs, one in HTTPS -> HTTP mode (for the service which required the IP address), and one in SSL -> TCP mode (for the websocket service).

Blystad:
So, I started by testing this on my existing 1.7.3 cluster, creating another ELB Listener on port 9010 for SSL -> TCP (80). That worked great, as expected. Proxy protocol was working, I got the correct SSL certificate, everything was fine.

I had to put it on TCP port 80, if I put it on TCP port 443 then nginx would complain that a plain HTTP request was sent to a HTTPS endpoint. My 1.7.3 cluster has a platform certificate.

Then, to validate my findings on the latest release, I decided to provision a new cluster with the following change:
```diff
diff --git a/contrib/aws/gen-json.py b/contrib/aws/gen-json.py
index 4a86991..ab54265 100755
--- a/contrib/aws/gen-json.py
+++ b/contrib/aws/gen-json.py
@@ -145,4 +145,9 @@ if VPC_ID and VPC_SUBNETS and VPC_ZONES and len(VPC_SUBNETS.split(',')) == len(V
     template['Resources']['CoreOSServerAutoScale']['Properties']['VPCZoneIdentifier'] = VPC_PRIVATE_SUBNETS.split(',')
     template['Resources']['DeisWebELB']['Properties']['Subnets'] = VPC_SUBNETS.split(',')
 
+SSL_CERT_ID = os.getenv('SSL_CERT_ID', None)
+if SSL_CERT_ID:
+    template['Resources']['DeisWebELB']['Properties']['Listeners'][1]['Protocol'] = 'SSL';
+    template['Resources']['DeisWebELB']['Properties']['Listeners'][1]['SSLCertificateId'] = SSL_CERT_ID;
+
 print json.dumps(template)
```

I checked the output of `gen-json.py`, and that seemed correct.

I deployed everything, added `deis pull blystad/http-server-header-dump` (a small application I wrote for checking the headers), and everything looked OK over http. Over HTTPS the page didn't load (443 -> 443), so I changed the instance port to 80 (443 -> 80) and everything worked almost as expected.

The problem I see, is that on HTTPS, these will be set incorrectly:
```
X-Forwarded-Proto: http
X-Forwarded-Ssl: off
```

(Since AWS ELB isn't 100% spec compliant with the Proxy Protocol...it doesn't tell the original PORT). I think this could be solved in deis-router.

sstarcher:
@Blystad with SSL->TCP and the lack of correct x-forwarded-proto and x-forwarded-ssl how would the Deis ELB be able to handle enforceHTTPS.

Currently my only solution for enforceHTTPS is to do one of the following.
* Disable ProxyProtocol, and do HTTPS->HTTP termination on the ELB - You lose web socket support
* SSL termination at Deis router. - I don't believe I can terminate multiple different domains

Blystad:
Good question....

With todays solution, I don't think it can...It would require modifications to `deis-router` to handle port `443` without using SSL. Something like:

[router/deis.conf#L5-L10](https://github.com/deis/deis/blob/master/router/rootfs/etc/confd/templates/deis.conf#L5-L10) - Add a "else if <some variable>" here, where we just `listen 443` normally (without SSL).
[router/nginx.conf#L247](https://github.com/deis/deis/blob/e6dd450/router/rootfs/etc/confd/templates/nginx.conf#L247) - Set `access_scheme` based on [$server_port](http://stackoverflow.com/questions/6051426/nginx-which-port-recived-the-request)

I'd strongly recommend just using either platform certificates, or application certificates.

<rant>
it's a bit annoying that AWS keeps pushing out new feature after feature, without improving their existing products...This could've been avoided if either 1) AWS ELB supported web sockets, 2) AWS ELB implemented the Proxy Protocol to spec.
</rant>

sstarcher:
@Blystad in my case I can't use the wildcard platform SSL support due to having multiple wildcard domains http://docs.deis.io/en/latest/managing_deis/ssl-endpoints/

I believe my only option is to disable proxy protocol support and do HTTPS->HTTP termination on the ELB.

jgmize:
While this PR worked for our original use case, I no longer think it makes sense to merge these changes into master for the following reasons:

1. This script will not be used at all in V2
2. This would be a new feature, so would need a compelling reason to go into the LTS release
3. This was only a partial solution to the more general problem as discussed in the comments above, and others would have needed to do further customizations anyway for situations where there is more than one additional ELB, further reducing its potential value in an LTS release.
4. We no longer need this for www.mozilla.org, due to [other changes to our SSL termination](https://twitter.com/jvehent/status/679070701425414145)

krancour:
@jgmize also, since you've noted that this wouldn't carry over to v2, I feel like its a good place to point out something I'm really excited about.  In v2 (depending on your underlying infrastructure) a single load balancer does still get provisioned in front of the router(s), _but_ a really exciting thing about v2 is that you always have the option to drop down to k8s to handle something that Deis doesn't handle out-of-the box.  In this case, you can just add additional k8s services of type `LoadBalancer` for the router and on platforms like AWS and GKE, the load balancer will be created for you.  Or if you have no reason to involvebthe router, you could actually bypass that entirely by adding services of type LoadBalancer in front of a (otherwise) deis-manged app.  Or... you could have a CloudFormation template that describes _just_ your load balancers-- which would be a very easy template to write in comparison to modifying the monolithic templates in v1.x.  I don't want to speak for everyone, but I'm pretty excited for all the new flexibility that k8s _and_ a drastically simpler provisioning and installation process buy us.

krancour:
And you could also drop to k8s to put a load balancer in front of a non-http service.

jgmize:
Thanks for pointing that out @krancour. I've been following the v2 development pretty closely, and I'm really looking forward to putting it into production once it's ready.

Issue #3751, opened 2015-05-27
Split user-data into separate files for each service


carmstrong:
See https://github.com/deis/deis/pull/3716#issuecomment-106017374

Basically, the user-data file is unruly. Adding new units or files to this is difficult to manage.

I'd like to explore creating a new Makefile target like `combine-user-data` which would simply combine all of these files together. `make discovery-url` could depend on this target before it replaces the discovery URL. 

Note that we need to be careful that we don't break the provision scripts in this. I think we should be fine as long as we still ensure users are running `make discovery-url` before running the provision scripts, since this target would also combine the user data pieces.

carmstrong:
I was thinking of a directory structure like:
```
contrib/coreos/units/stop-update-engine.service.yaml
contrib/coreos/files/run-deis-bin-preseed.yaml
...
```

croemmich:
I agree that this would make maintenance a bit easier, however the generated user-data file would still be huge, which doesn't feel like we're solving the real issue... but I don't have a better solution, so I probably can't say much :)

nathansamson:
one problem with putting too much logic in user-data is that it doesn't get updated when upgrading deis releases, unless I am mistaken....

mboersma:
> it doesn't get updated when upgrading deis

You're exactly right, and it is a big reason user-data is problematic for config management.

carmstrong:
> the generated user-data file would still be huge

True, but if the user never sees it, it probably doesn't matter how big it is ;) Making changes in the codebase is the real pain currently.

> it doesn't get updated when upgrading deis

> You're exactly right, and it is a big reason user-data is problematic for config management.

Absolutely. Just not sure there's a better solution currently. :(

croemmich:
I apply some networking and other config in my user-data so I'd still have to maintain a big file :(

kalbasit:
:+1: I do the same and maintaining the user-data is getting a bit heavy

aespinosa:
Since coreos-cloudinit accepts a URL for the location of cloudconfig, maybe we can have the 'bootstrap' user-data just contain a unit that downloads a public cloud-init (released and maintained artifact alongside w/ deis, deisctl and unit files) for bootstrapping all the things that deisctl needs to properly install deis.

krancour:
One potential problem with that is that the etcd discovery URL needs to be included in the etcd section of the userdata.  That means userdata is custom for each cluster.  We couldn't use URL that points to static userdata.  Something more complex would be required.

[...]

PR #4652, opened 2015-10-22
feat(contrib/coreos): move static user-data outside

aespinosa:
Significantly decreases user-data content to only the ones needed for
configuration.

Proposed idea for #3751


deis-admin:

contrib/coreos/user-data.example
```diff
@@ -17,266 +17,13 @@ coreos:
...
-      fi
-  - path: /opt/bin/scheduler-policy.json
-    content: |
-      {
-          "kind": "Policy",
-          "apiVersion": "v1",
-          "predicates": [{"name": "PodFitsPorts"},{"name": "PodFitsResources"},{"name": "NoDiskConflict"},{"name": "MatchNodeSelector"},{"name": "HostName"}],
-          "priorities": [{"name": "LeastRequestedPriority","weight": 1},{"name": "BalancedResourceAllocation","weight": 1},{"name": "ServiceSpreadingPriority","weight": 2},{"name": "EqualPriority","weight": 1}]
-      }
+        ExecStart=/usr/bin/coreos-cloudinit --from-url=https://raw.githubusercontent.com/aespinosa/deis/move-static-cloud-config/contrib/coreos/cloud-config
```
aespinosa: I just set this to my branch to pass the tests. This should be something canonical and managed synced with releases.

contrib/coreos/cloud-config
```diff
@@ -0,0 +1,266 @@
+#cloud-config
```
bacongobbler: Would you mind removing this file or moving it somewhere else if this is meant to be the static cloud config to be uploaded to S3?

krancour:
Jenkins, add to whitelist.

contrib/coreos/cloud-config
krancour: @bacongobbler isn't it `contrib/coreos/user-data` that gets written by `make discovery-url`?

krancour:
I like this change, mainly because it reduces the size of the artifacts that get submitted to each cloud provider during provisioning-- e.g. smaller CloudFormation templates.  That means we can put _more_ stuff in there (to make clusters more robust) without fearing the size restrictions too much.

All that being said...

I think there are some gotchas that this PR does not account for.  Take the AWS provisioning process as an example.  That process actually merges `contrib/coreos/user-data` with additional AWS-specific user-data.  You can see that [here](https://github.com/deis/deis/blob/master/contrib/aws/gen-json.py#L15-L106).  Will that continue to work?  Or will it break?

My one other concern would be with respect to workflow.  Does this make it much harder for people to customize their user-data if they have good reason to do so?

I still like this.  I just want to make sure everything is accounted for.

contrib/coreos/cloud-config
bacongobbler: Yes, which is why this file should be moved elsewhere or else `make discovery-url` will overwrite it. This file is in gitignore: https://github.com/deis/deis/blob/master/.gitignore#L68-L69

carmstrong: This file is `cloud-config`, not `user-data`.

bacongobbler: Wow, big oversight by me. Need more :coffee:! Sorry <3

carmstrong:
Agreed with @krancour - I love this in principle, but I'm thinking about the dependency on GitHub for a machine to bootstrap properly. What if GitHub is down?

aespinosa:
Hi @carmstrong 

Yes I agree that the release engineering of where this artifact should be needs some thinking. Right now user-data as-is depends on some github/ google-located artifacts. Is there a place where we consider "most reliable"? Or are we headed to the bosh/cf-style assumption of everything should be vendored on the machine doing the provisioning?

@krancour , right I haven't taken the IaaS-specific user-data being injected. That being said, I agree that managing YML-ish manifests is a hard problem. Several related tooling have come and go. Here's a few:

* https://github.com/mattray/spiceweasel
* https://github.com/cloudfoundry-incubator/spiff
* https://github.com/geofffranks/spruce

They're very specific to their own domains but I don't seem to find coreos folks thinking this is a problem with cloudinit.

carmstrong:
@aespinosa Good points. I think @krancour's point about provider-specific cloud-config is valid. I'm wondering if we should build provider-specific cloud-config files whenever we cut a Deis release, and push those somewhere public like S3. Added complexity at release time, but may be better for users.

krancour:
@carmstrong, that sounds like a maintenance burden, but I agree that it simplifies an awful lot of things.

bacongobbler:
closing as something we won't be able to maintain during the LTS release as mentioned in #3751. Thank you for the PR though :)

Issue #4491, opened 2015-09-15
Builder: Make git sha hash available during build


olalonde:
Heroku calls it `SOURCE_VERSION` (https://devcenter.heroku.com/changelog-items/630). This would be helpful for things like cache invalidation and some projects (e.g. https://github.com/drone/drone/blob/master/Makefile#L1) require access to the git hash during build phase.

olalonde:
Oh, it seems it's already available as `GIT_SHA` (https://github.com/deis/deis/pull/2089). Maybe add it to documentation? I don't recall reading about that.

mboersma:
Good point--I don't think this is documented explicitly. We would love a PR that adds this--maybe to the bottom of http://docs.deis.io/en/latest/using_deis/deploy-application/? I'm having a hard time finding a more appropriate location in the docs.

olalonde:
Wouldn't http://docs.deis.io/en/latest/using_deis/using-buildpacks/ be more appropriate since the `GIT_SHA` is not available for Dockerfile/docker pull deploys (afaik)?

olalonde:
Oh, looks like I was mistaken. The GIT_SHA environment variable is added to the app container but is not available during the actual build. Would it be possible to make it available during build phase?

technosophos:
Seems like we ought to be able to do that by adding an `-e SOURCE_VERSION=$GIT_SHA` right here:
https://github.com/deis/deis/blob/master/builder/rootfs/etc/confd/templates/builder#L98

mboersma:
Closed by #4654.

PR #4654, opened 2015-10-23
feat(builder): make SOURCE_VERSION env variable

olalonde:
Add a `SOURCE_VERSION` environment variable to slub builds which represents the git commit SHA-1 of the source being built. 

Mirrors this Heroku feature: https://devcenter.heroku.com/changelog-items/630

Fixes issue #4491

### DIFF
--- a/builder/rootfs/etc/confd/templates/builder
+++ b/builder/rootfs/etc/confd/templates/builder
@@ -85,15 +85,15 @@ if [ $CODE -ne 0 ]; then
     exit 1
 fi
 
-BUILD_OPTS=()
-BUILD_OPTS+='/usr/bin/docker'
-BUILD_OPTS+=' run -v /etc/environment_proxy:/etc/environment_proxy'
-# get application configuration
-BUILD_OPTS+=$(echo $RESPONSE | get-app-values)
-
 # if no Dockerfile is present, use slugbuilder to compile a heroku slug
 # and write out a Dockerfile to use that slug
 if [ ! -f Dockerfile ]; then
+    BUILD_OPTS=()
+    BUILD_OPTS+='/usr/bin/docker'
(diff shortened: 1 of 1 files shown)


bacongobbler:
Looks like this PR failed:

> remote: time="2015-10-23T14:18:05Z" level="fatal" msg="Error: ENV must have two arguments"

olalonde:
Oops, don't have deis setup (new laptop). Will try to fix that test.

mboersma:
@olalonde after this change I don't see `SOURCE_VERSION` inside the container:
```
$ deis run env
Running 'env'...
HOSTNAME=80db917ce87a
DEIS_APP=triple-bagpiper
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/app/bin
PWD=/app
GIT_SHA=14fb753cf3c1b740879e62852ca222426d8aa0d5
GOMAXPROCS=1
HOME=/app
SHLVL=1
PORT=5000
DEIS_RELEASE=v2
_=/usr/bin/env
```
Maybe [this line](https://github.com/deis/deis/blob/master/builder/rootfs/etc/confd/templates/builder#L127) would be a better place to make the change?

And as you noted, this wouldn't work for Dockerfile-based builds. But Deis already includes the `GIT_SHA` variable, which is equivalent to Heroku's `SOURCE_VERSION`. (Does Deis need both?)

mboersma:
I tested this with your helpful [test app](https://github.com/olalonde/deis-test) and see `SOURCE_VERSION` during the build:
```
       npm_config_version: '',
       SOURCE_VERSION: '39efe093faac0f678b31ae1c6dde4ff6f9407bdb',
       npm_config_local_address: '',
```
So mission accomplished. Thanks @olalonde!

olalonde:
Thanks :)

Issue #4657, opened 2015-10-24
bug(client/controller): deis logs replaces " with \"


olalonde:
UPDATE: made a PR to solve this: https://github.com/deis/deis/pull/4658

Actual output:

```bash
$ deis logs
2015-10-24T06:26:53UTC bitstore-test[web.2]: ::ffff:10.21.1.116 - - [24/Oct/2015:06:26:53 +0000] \"GET / HTTP/1.1\" 302 66 \"-\" \"Mozilla/5.0 (iPhone; CPU iPhone OS 8_3 like Mac OS X) AppleWebKit/600.1.4 (KHTML, like Gecko) Version/8.0 Mobile/12F70 Safari/600.1.4 (compatible; Googlebot/2.1; +http://www.google.com/bot.html)\"
2015-10-24T06:26:53UTC bitstore-test[web.3]: ::ffff:10.21.2.84 - - [24/Oct/2015:06:26:53 +0000] \"GET /docs HTTP/1.1\" 303 43 \"-\" \"Mozilla/5.0 (iPhone; CPU iPhone OS 8_3 like Mac OS X) AppleWebKit/600.1.4 (KHTML, like Gecko) Version/8.0 Mobile/12F70 Safari/600.1.4 (compatible; Googlebot/2.1; +http://www.google.com/bot.html)\"
```

Expected output:

```bash
$ deis logs
2015-10-24T06:26:53UTC bitstore-test[web.2]: ::ffff:10.21.1.116 - - [24/Oct/2015:06:26:53 +0000] "GET / HTTP/1.1" 302 66 "-" "Mozilla/5.0 (iPhone; CPU iPhone OS 8_3 like Mac OS X) AppleWebKit/600.1.4 (KHTML, like Gecko) Version/8.0 Mobile/12F70 Safari/600.1.4 (compatible; Googlebot/2.1; +http://www.google.com/bot.html)"
2015-10-24T06:26:53UTC bitstore-test[web.3]: ::ffff:10.21.2.84 - - [24/Oct/2015:06:26:53 +0000] "GET /docs HTTP/1.1" 303 43 "-" "Mozilla/5.0 (iPhone; CPU iPhone OS 8_3 like Mac OS X) AppleWebKit/600.1.4 (KHTML, like Gecko) Version/8.0 Mobile/12F70 Safari/600.1.4 (compatible; Googlebot/2.1; +http://www.google.com/bot.html)"
```

```bash
$ deisctl config logs get storageAdapterType
memory
[...]

PR #4658, opened 2015-10-24
fix(client): parse logs as JSON string

olalonde:
Fixes issue #4657.

### DIFF
--- a/client/cmd/apps.go
+++ b/client/cmd/apps.go
@@ -169,7 +169,7 @@ func AppLogs(appID string, lines int) error {
 
 // printLogs prints each log line with a color matched to its category.
 func printLogs(logs string) error {
-	for _, log := range strings.Split(strings.Trim(logs, `\n`), `\n`) {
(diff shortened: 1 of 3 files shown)


client/Makefile
```diff
@@ -1,5 +1,7 @@
 include ../includes.mk
 
+.PHONY: build
```
olalonde: not sure if there's something wrong with my environment but `make build` wouldn't work without that

Joshua-Anderson: That's a reminant from the build folder left over from the old python client.

olalonde:
Any chance to get this merged? Should I removed the `.PHONY` line?

client/Makefile
bacongobbler: let's remove this. Just remove the `client/build` folder locally and everything should work fine again. :)

bacongobbler:
@olalonde there's a server-side fix available in #5004 which fixes the underlying issue. With that fix logs are now properly formatted.

Issue #4601, opened 2015-10-12
Deis controller flipping


foxycoder:
Hi,

It seems to me that there is something wrong with my config state for the controller. It looks like it keeps receiving different values for the controller config, thus rewriting and restarting nginx. See the output below.

My platform version:

```
deisctl config platform get version
v1.11.0
```

My Controller keeps spitting out the following output:

```
Oct 12 13:53:55 ip-xxx sh[3667]: [2015-10-12 13:53:55 +0000] [172] [INFO] Booting worker with pid: 172
Oct 12 13:54:00 ip-xxx sh[3667]: [2015-10-12 13:54:00 +0000] [120] [INFO] Handling signal: hup
Oct 12 13:54:00 ip-xxx sh[3667]: [2015-10-12 13:54:00 +0000] [120] [INFO] Hang up: Master
Oct 12 13:54:00 ip-xxx sh[3667]: [2015-10-12 13:54:00 +0000] [190] [INFO] Booting worker with pid: 190
Oct 12 13:54:00 ip-xxx sh[3667]: [2015-10-12 13:54:00 +0000] [171] [INFO] Worker exiting (pid: 171)
Oct 12 13:54:00 ip-xxx sh[3667]: [2015-10-12 13:54:00 +0000] [170] [INFO] Worker exiting (pid: 170)
Oct 12 13:54:00 ip-xxx sh[3667]: [2015-10-12 13:54:00 +0000] [172] [INFO] Worker exiting (pid: 172)
Oct 12 13:54:00 ip-xxx sh[3667]: [2015-10-12 13:54:00 +0000] [173] [INFO] Worker exiting (pid: 173)
Oct 12 13:54:00 ip-xxx sh[3667]: [2015-10-12 13:54:00 +0000] [192] [INFO] Booting worker with pid: 192
Oct 12 13:54:00 ip-xxx sh[3667]: [2015-10-12 13:54:00 +0000] [174] [INFO] Worker exiting (pid: 174)
Oct 12 13:54:00 ip-xxx sh[3667]: [2015-10-12 13:54:00 +0000] [191] [INFO] Booting worker with pid: 191
Oct 12 13:54:00 ip-xxx sh[3667]: [2015-10-12 13:54:00 +0000] [193] [INFO] Booting worker with pid: 193
Oct 12 13:54:00 ip-xxx sh[3667]: [2015-10-12 13:54:00 +0000] [194] [INFO] Booting worker with pid: 194
```

[...]

Issue #4557, opened 2015-10-01
fatal: 'somerepo.git' does not appear to be a git repository


olalonde:
Getting this error when pushing to deis. I tried to restart deis-builder but same problem.

```bash
$ git push deis
fatal: 'bsync-test-green.git' does not appear to be a git repository
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

deis-builder logs:

```
Oct 01 17:50:04 ip-10-21-2-84.ec2.internal sh[311]: [info] Key accepted for user git.
Oct 01 17:50:04 ip-10-21-2-84.ec2.internal sh[311]: [info] Channel type: session
Oct 01 17:50:04 ip-10-21-2-84.ec2.internal sh[311]: Key='LANG', Value='en_US.UTF-8'
Oct 01 17:50:04 ip-10-21-2-84.ec2.internal sh[311]: Key='LC_ALL', Value='en_US.UTF-8'
Oct 01 17:50:04 ip-10-21-2-84.ec2.internal sh[311]: Key='LC_CTYPE', Value='UTF-8'
Oct 01 17:50:04 ip-10-21-2-84.ec2.internal sh[311]: [info] Checking user /deis/builder/users/olalonde
Oct 01 17:50:04 ip-10-21-2-84.ec2.internal sh[311]: [info] Found user olalonde for fingerprint 62:c7:52:a8:aa:41:1d:5c:d6:80:4c:3d:7b:45:da:30
Oct 01 17:50:04 ip-10-21-2-84.ec2.internal sh[311]: [info] Directory /home/git/bsync-test-green.git already exists.
Oct 01 17:50:04 ip-10-21-2-84.ec2.internal sh[311]: [info] git-shell -c git-receive-pack 'bsync-test-green.git'
Oct 01 17:50:04 ip-10-21-2-84.ec2.internal sh[311]: Waiting for git-receive to run.
Oct 01 17:50:04 ip-10-21-2-84.ec2.internal sh[311]: Waiting for deploy.
Oct 01 17:50:04 ip-10-21-2-84.ec2.internal sh[311]: [error] Error on command: exit status 128
Oct 01 17:50:04 ip-10-21-2-84.ec2.internal sh[311]: [error] Failed git receive: exit status 128
```

olalonde:
Fixed by removing `/home/git/bsync-test-green` in the deis-builder container. Though I'm not sure why this happened in the first place.

krancour:
[...]

Issue #3969, opened 2015-07-06
Errors in logs, possibly from store.


ngerakines:
I'm seeing a huge number of errors in the logs.

```
Jul 06 12:36:44 deis sh:  2015-07-06 16:36:44.677722 7fd2a4088700  0 <cls> cls/rgw/cls_rgw.cc:1947: ERROR: rgw_obj_remove(): cls_cxx_remove returned -2
Jul 06 12:36:44 deis sh:  2015-07-06 16:36:44.698643 7fd2a708e700  0 <cls> cls/rgw/cls_rgw.cc:1947: ERROR: rgw_obj_remove(): cls_cxx_remove returned -2
Jul 06 12:36:44 deis sh:  2015-07-06 16:36:44.716780 7fcdc54e2700  0 <cls> cls/rgw/cls_rgw.cc:1947: ERROR: rgw_obj_remove(): cls_cxx_remove returned -2
Jul 06 12:36:44 deis sh:  2015-07-06 16:36:44.737903 7fa2afd7a700  0 <cls> cls/rgw/cls_rgw.cc:1947: ERROR: rgw_obj_remove(): cls_cxx_remove returned -2
Jul 06 12:36:44 deis sh:  2015-07-06 16:36:44.764561 7f0bfbbeb700  0 <cls> cls/rgw/cls_rgw.cc:1947: ERROR: rgw_obj_remove(): cls_cxx_remove returned -2
Jul 06 12:36:44 deis sh:  2015-07-06 16:36:44.790842 7fa2ad575700  0 <cls> cls/rgw/cls_rgw.cc:1947: ERROR: rgw_obj_remove(): cls_cxx_remove returned -2
Jul 06 12:36:44 deis sh:  2015-07-06 16:36:44.806244 7feea7fad700  0 <cls> cls/rgw/cls_rgw.cc:1947: ERROR: rgw_obj_remove(): cls_cxx_remove returned -2
Jul 06 12:36:44 deis sh:  2015-07-06 16:36:44.813644 7f0bfe3f0700  0 <cls> cls/rgw/cls_rgw.cc:1947: ERROR: rgw_obj_remove(): cls_cxx_remove returned -2
Jul 06 12:36:44 deis sh:  2015-07-06 16:36:44.841653 7fa2acd74700  0 <cls> cls/rgw/cls_rgw.cc:1947: ERROR: rgw_obj_remove(): cls_cxx_remove returned -2
[...]

PR #4663, opened 2015-10-25
fix(contrib/azure): use Azure premium storage

carmstrong:
Working with Microsoft, we suspect that cluster issues are due to poor disk performance.

/cc @sgoings

### DIFF
--- a/contrib/azure/arm-template.json
+++ b/contrib/azure/arm-template.json
@@ -65,7 +65,7 @@
     "customData": {
       "type": "string",
       "metadata": {
-        "description": "Base64-encoded cloud-config.yaml file to deploy and start Fleet"
+        "description": "Base64-encoded cloud-config.yaml file to deploy and start fleet"
       }
     },
     "numberOfNodes": {
@@ -78,15 +78,26 @@
     "dockerVolumeSize": {
       "type": "int",
       "metadata": {
-        "description": "Size in GB of the Docker volume."
+        "description": "Size in GB of the Docker volume"
       },
       "defaultValue": "100"
     },
     "coreosVersion": {
       "type": "string",
       "metadata": {
-        "description": "Version of CoreOS to deploy."
+        "description": "Version of CoreOS to deploy"
       }
+    },
+    "storageAccountType": {
+      "type": "string",
+      "metadata": {
+        "description": "Storage account type"
+      },
+      "allowedValues": [
+        "Standard_LRS",
+        "Premium_LRS"
+      ],
+      "defaultValue": "Premium_LRS"
     }
   },
   "variables": {
@@ -94,7 +105,6 @@
     "subnet1Name": "Subnet-1",
     "subnet1Prefix": "10.0.0.0/24",
     "publicIPAddressType": "Dynamic",
-    "storageAccountType": "Standard_LRS",
     "imagePublisher": "CoreOS",
     "imageOffer": "CoreOS",
     "imageSKU": "Stable",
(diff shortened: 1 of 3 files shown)


carmstrong:
@helgi You're probably the best for a second LGTM

helgi:
I'll have a look - Haven't worked on Azure since ARM came out. Looks like something to verify over pizza ;) Hold please

hold

contrib/azure/arm-template.json
```diff
@@ -87,6 +87,13 @@
       "metadata": {
         "description": "Version of CoreOS to deploy."
       }
+    },
+    "storageAccountType": {
+      "type": "string",
+      "metadata": {
+        "description": "Storage account type. Can be Premium_LRS (recommended) or Standard_LRS"
+      },
+      "defaultValue": "Premium_LRS"
```
helgi: Can you add `allowedValues` (see https://azure.microsoft.com/en-us/documentation/articles/resource-group-authoring-templates/#parameters) so there is little room for mistakes.

Not sure how the view for this looks but would probably reduce the `description` part.

You may also want to add the rest of the ones from https://azure.microsoft.com/en-us/documentation/articles/resource-manager-template-storage/ - it's mostly just geo redundant and that like but won't hurt anything.

helgi:
Not sure where we should drop it, perhaps the `azure` section of deis.io but linking to https://azure.microsoft.com/en-us/documentation/articles/storage-premium-storage-preview-portal/ with the Premium recommendation and a quick reasoning would be ace

contrib/azure/arm-template.json
carmstrong: Good point on that - I thought of allowedValues after I pushed this, and wasn't sure it was worth doing. But I think it is.

carmstrong:
@helgi Did you see my changes to the "Installing Deis on Azure" docs as part of this PR? Maybe it could use more info, but I did link to premium storage and the list of regions where it's enabled.

helgi:
Nope, I was totally blind to that part! My bad :) Can I still get the `allowedValues` change in?

colrack:
I'm testing a three node cluster with azure premium storage running in westeurope.
1 load balancer, 3 DS1 VMs. One day uptime no problem. I'll let you know.
However I tried a lot of possible configurations: LRS, GRS, ZRS, all kind of vm (A series, D series etc) and every time there was a ceph issue after some time (as reported here by many users).
Is it possible to start thinking about a stateless deis cluster using azure blob storage so we can use all kind of VMs without ceph storage issues?

carmstrong:
> and every time there was a ceph issue after some time (as reported here by many users).

What sort of Ceph issue? Were the hosts going out of sync? If it's a matter of losing hosts after some time, I'm afraid that'll still require some manual intervention.

> Is it possible to start thinking about a stateless deis cluster using azure blob storage so we can use all kind of VMs without ceph storage issues?

If that's something you wanted to investigate and write documentation for, that'd be great! Azure blob store could work (as long as it's S3 compatible), and we'd still need an external Postgres instance.

colrack:
So far when I tried to deploy to Azure I encountered storage related issues like ceph mon clock skew, #3969 "cls/rgw/cls_rgw.cc:1947: ERROR: rgw_obj_remove(): cls_cxx_remove returned -2", mon down I don't know why, hosts going out of sync and maybe other. Other issues maybe not storage related like #4557 "fatal: 'somerepo.git' does not appear to be a git repository", #4601 error from controller due to multiple instances of databases launched simultaneously...
Unfortunately Azure Blob Storage is not compatible with AWS S3. I just wondered how difficult would be to integrate it. (And yes we'd still need an external Postgres instance).

PR #4669, opened 2015-10-26
ref(contrib/aws): remove disk replacements that were forgotten in #4444

helgi:
Not needed anymore due to #4444 and moving away from containerized etc2

### DIFF
--- a/contrib/aws/gen-json.py
+++ b/contrib/aws/gen-json.py
@@ -108,9 +108,6 @@ data['coreos']['units'] = new_units + data['coreos']['units']
 header = ["#cloud-config", "---"]
 dump = yaml.dump(data, default_flow_style=False)
 
-# configure etcd to use its EBS volume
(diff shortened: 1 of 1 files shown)


carmstrong:
ping @mboersma for other review

bacongobbler:
According to https://github.com/deis/deis/commit/445dbaecfadb1e2690734413cc43f30abcb13b21#commitcomment-14220568 and to https://github.com/deis/deis/blob/master/contrib/coreos/user-data.example#L14, this seems like it was still necessary. Is there somewhere else which we're missing or should we revert this? It sounds like etcd on AWS isn't using the EBS as we expected for new clusters on v1.12.

helgi:
This particular PR removed an ENV var that was no longer used, the original #4444 removed the usage of all of those. This may however need to be included in the actual `user-data` replacement in a different way.

bacongobbler:
Good point. I'll see if I can whip a PR for that

PR #4673, opened 2015-10-27
chore(router): update nginx to 1.9.6

mboersma:
See http://nginx.org/en/CHANGES

spdy is gone is favor of http2 as of 1.9.5, but it's unclear to me if that would be a breaking change since Deis didn't advertise the availability of the `spdy` protocol. ping @helgi and @aledbf for review.

### DIFF
--- a/router/rootfs/bin/build
+++ b/router/rootfs/bin/build
@@ -2,7 +2,7 @@
 
 set -eof pipefail
 
-export NGINX_VERSION=1.9.4
+export NGINX_VERSION=1.9.6
 export NAXSI_VERSION=0d53a64ed856e694fcb4038748c8cf6d5551a603
 export NDK_VERSION=0.2.19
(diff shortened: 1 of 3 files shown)


helgi:
Based on #4542 we should be good to go

carmstrong:
> Based on #4542 we should be good to go

Is that an implicit LGTM?

mboersma:
> Is that an implicit LGTM?

I thought so. :smiling_imp: But I want to wait until @aledbf has had time to field-test the new nginx as I know he will, so no rush.

carmstrong:
> so no rush.

:+1:

aledbf:
Code LGTM

mboersma:
I've tested this upgrade from v1.11.2 on an AWS cluster and discussed it with @aledbf after his testing.

Issue #4588, opened 2015-10-09
deis/contrib/gce/gce-user-data generates an invalid cloud-config


gahissy:
On `deis 1.11.0`  when running 
```bash
./contrib/gce/create-gce-user-data $(curl -s https://discovery.etcd.io/new)
``` 
the cloud-config file generated is always invalid due to a null discovery value.
This line (in `contrib/coreos/user-data.example`): 
```bash
discovery: #DISCOVERY_URL
````
is always replaced (in the output file `contrib/gce/gce-user-data`) by : 
```python
"discovery": !!null |-
      null
```
After some debugging, I noticed that the placeholder `#DISCOVERY_URL` was already replaced by `null`even before reaching the line (in `contrib/gce/create-gce-user-data`) :
```python
dump = dump.replace('#DISCOVERY_URL', url)
```
So I replaced `#DISCOVERY_URL` by `__DISCOVERY_URL__` in both `contrib/gce/create-gce-user-data` and `contrib/coreos/user-data.example` and it worked.

The problem seems to be coming from this code :
```python
configuration_coreos_template = yaml.safe_load(coreos_template)
```

PR #4680, opened 2015-10-28
fix(contrib/azure)-replace discovery url

### LINKED ISSUE #4679: Bug (contrib/azure) - unable to add new discovery URL to azure-user-data
The command `./create-azure-user-data $(curl -s https://discovery.etcd.io/new)` did not correctly parse the `user-data-template` file to replace value of DISCOVERY_URL in the `azure-user-data` file. The end result was  `discovery: null`. This was also mentioned in #4593

ritazh:
closes #4679 

Fix bug related to a null value discovery URL that was blocking deployment of Deis to Azure

### DIFF
--- a/contrib/azure/create-azure-user-data
+++ b/contrib/azure/create-azure-user-data
@@ -56,14 +56,19 @@ def main():
     azure_template = get_file("azure-user-data-template")
     coreos_template = get_file("../coreos/user-data.example")
 
-    configuration_coreos_template = yaml.safe_load(coreos_template)
+    coreos_template_w_url = None
+    with open(coreos_template.name, 'r') as file :
+      coreos_template_w_url = file.read()
+
+    coreos_template_w_url = coreos_template_w_url.replace('#DISCOVERY_URL', url)
+
+    configuration_coreos_template = yaml.safe_load(coreos_template_w_url)
     configuration_azure_template = yaml.safe_load(azure_template)
 
     configuration = combine_dicts(configuration_coreos_template,
                                   configuration_azure_template)
 
     dump = yaml.dump(configuration, default_flow_style=False)
-    dump = dump.replace('#DISCOVERY_URL', url)
 
     with azure_user_data as outfile:
         try:


deis-admin:

contrib/coreos/user-data.example
```diff
@@ -3,7 +3,7 @@
 coreos:
   etcd2:
     # generate a new token for each unique cluster from https://discovery.etcd.io/new
-    discovery: #DISCOVERY_URL
+    discovery: $DISCOVERY_URL
```
mboersma: This change will break [`make discovery-url`](https://github.com/deis/deis/blob/master/Makefile#L37), which is how  other providers get an `etcd` discovery URL.

mboersma:
I don't see how this would fix anything. `replace()` is just a python string function, so changing what it searches for from `#DISCOVERY_URL` to `$DISCOVERY_URL` is a no-op. `$DISCOVERY_URL` isn't being interpreted as a shell variable in the python code, nor in user-data, so the issue must be something different. (And this change breaks other providers as noted above.)

Could you enter a GitHub issue with details of the actual Azure failure? Maybe that would help pin it down and fix it.

sgoings:
Here's an alternative implementation: https://github.com/sgoings/deis/blob/rigger-azure/contrib/azure/create-azure-user-data#L55-L66

The problem is that the comment in YAML (#DISCOVERY_URL) is stripped out (discovery's value is empty) by yaml.safe_load on newer versions of PyYAML.

sgoings:
@mboersma she did (#4679)

mboersma:
> the comment in YAML (#DISCOVERY_URL) is stripped out

Ok, then this change would help I assume, but it needs to be made and tested against all providers.

ritazh:
@mboersma @sgoings Thanks for the feedback. What is the best way to test this against all the providers?

sgoings:
Looks bueno @ritazh. Just need a rebase + squash to 1 commit!

ritazh:
@sgoings Done!

mboersma:
It appears Deis needs a similar fix for GCE to #4588.

PR #4687, opened 2015-10-28
fix(firewall): prepend firewall chain rather than append

### LINKED ISSUE #4686: custom-firewall.sh: jump to firewall is appended rather than prepended
Currently, the custom firewall script appends a jump to `Firewall-INPUT` to the `INPUT` and `FORWARD` chains. This can lead to the following `FORWARD` chain:
```
Chain FORWARD (policy DROP 0 packets, 0 bytes)
 pkts bytes target     prot opt in     out     source               destination         
24310   31M DOCKER     all  --  *      docker0  0.0.0.0/0            0.0.0.0/0           
15664   16M ACCEPT     all  --  *      docker0  0.0.0.0/0            0.0.0.0/0            ctstate RELATED,ESTABLISHED
22853 2429K ACCEPT     all  --  docker0 !docker0  0.0.0.0/0            0.0.0.0/0           
    0     0 Firewall-INPUT  all  --  *      *       0.0.0.0/0            0.0.0.0/0           
    0     0 ACCEPT     all  --  docker0 docker0  0.0.0.0/0            0.0.0.0/0   
```
As you can see, the chain may be ACCEPTed by some docker rule without going through the firewall. This results in ports bound by docker containers (and exposed on the host) to be publically accessible.

Potential fix:
```
cat custom-firewall.sh | \
    sed 's/-A INPUT -j Firewall-INPUT/-I INPUT 1 -j Firewall-INPUT/' | \
    sed 's/-A FORWARD -j Firewall-INPUT/-I FORWARD 1 -j Firewall-INPUT/' > new-firewall.sh
```

jannispl:
fixes #4686


deis-admin:

carmstrong:
cc @aledbf for review

aledbf:
@jannispl this issue could be related to https://github.com/docker/docker/issues/15890

if you execute `systemctl list-dependencies --after iptables-restore.service` the restore service runs even before `docker.socket` so the append in the script is not wrong.
```
iptables-restore.service
● ├─system.slice
● ├─systemd-journald.socket
● └─basic.target
●   ├─-.mount
●   ├─tmp.mount
●   ├─paths.target
●   │ ├─motdgen.path
●   │ ├─systemd-ask-password-console.path
●   │ ├─systemd-ask-password-wall.path
●   │ ├─user-cloudinit@var-lib-coreos\x2dinstall-user_data.path
●   │ └─user-configdrive.path
●   ├─slices.target
●   │ ├─-.slice
●   │ ├─machine.slice
●   │ ├─system.slice
●   │ └─user.slice
●   ├─sockets.target
●   │ ├─dbus.socket
●   │ ├─docker.socket
●   │ ├─early-docker.socket
```

can you run `sudo iptables-save >ipt.txt` to get your current rules and order?

jannispl:
@aledbf I *think* the problem is that the docker iptables rules are not set by `docker.socket`, but rather by `docker.service`, which is listed in `systemctl list-dependencies --before iptables.restore.service`:
```
iptables-restore.service
● ├─iptables-store.service
● ├─network.target
● │ ├─docker.service
● │ ├─rkt-metadata.service
● │ ├─sshd.service
● │ └─systemd-resolved.service
● └─shutdown.target
```

Anyway, the other problem is that the firewall script applies the rules instantly, with docker already running. This already causes a wrong order until the next reboot. (Even though I think the wrong order persisted after reboot in my case, not sure here).

Lastly, I think it would make sense for such a custom (and in some cases essential) firewall to prepend its rules to the existing ones.

jannispl:
I realized that even if the Firewall-INPUT rule is first, docker container ports are still exposed to the public.

```
$ docker port appspark-panel_v4.web.1
80/tcp -> 0.0.0.0:32775
```
```
(local) $ curl http://appspark-panel.mycluster.com:32775
[... HTTP 200 response ...]
```

I've traced it down to this rule: `-A Firewall-INPUT -i docker0 -j ACCEPT` which I think needs to be more specific.

aledbf:
@jannispl trying to reproduce this I remembered that docker should not be using iptables at all.
Can you post the the output of `systemctl cat docker` and `cat /var/run/flannel_docker_opts.env`?

jannispl:
@aledbf 
`systemctl cat docker`:
```
# /usr/lib64/systemd/system/docker.service
[Unit]
Description=Docker Application Container Engine
Documentation=http://docs.docker.com
After=docker.socket early-docker.target network.target
Requires=docker.socket early-docker.target

[Service]
EnvironmentFile=-/run/flannel_docker_opts.env
MountFlags=slave
LimitNOFILE=1048576
LimitNPROC=1048576
ExecStart=/usr/lib/coreos/dockerd --daemon --host=fd:// $DOCKER_OPTS $DOCKER_OPT_BIP $DOCKER_OPT_MTU $DOCKER_OPT_IPMASQ

[Install]
WantedBy=multi-user.target

# /etc/systemd/system/docker.service.d/50-insecure-registry.conf
[Service]
Environment="DOCKER_OPTS=--insecure-registry 10.0.0.0/8 --insecure-registry 172.16.0.0/12 --insecure-registry 192.168.0.0/16 --insecure-registry 100.64.0.0/10"
```

Neither `/var/run/flannel_docker_opts.env` nor `/run/flannel_docker_opts.env` (as referenced in output above) exist on my nodes.

aledbf:
@jannispl ok, that explains part of the issue. What version of deis are you running? Did you upgrade your cluster from a previous version? If that's the case did you update `user-data`?

jannispl:
@aledbf As I provision my bare metal nodes I have to do extra work on `user_data`, so what I did is I took v1.11.1 `contrib/coreos/user_data.sample` and added some networking related configuration. Then installed v1.11.2 as it was released in between.
See here: https://gist.github.com/jannispl/f4499269c8b5da356f7b

aledbf:
@jannispl ok:
- if you don't have a `/var/run/flannel_docker_opts.env` file means that flannel is not running.
- if the containers are not using an IP from a subnet of 10.244.0.0/16 docker is not using flannel ^^
- the file `/var/run/flannel_docker_opts.env`  should contains something like this
```
core # more /var/run/flannel_docker_opts.env
DOCKER_OPT_BIP="--bip=10.244.XXX.0/20"
DOCKER_OPT_IPMASQ="--ip-masq=false"
DOCKER_OPT_MTU="--mtu=1450"
```

jannispl:
Okay @aledbf so what did I miss when provisioning my nodes?

aledbf:
@jannispl you have this:
```
- name: flanneld.service
      drop-ins:
      - name: 50-network-config.conf
        content: |
          [Service]
          ExecStartPre=-/usr/bin/etcdctl mk /coreos.com/network/config '{"Network": "10.244.0.0/16", "SubnetLen": 24, "SubnetMin": "10.244.0.0", "Backend": {"Type": "vxlan"}}'
```

can you add `command: start`?

```
- name: flanneld.service
  command: start
      drop-ins:
      - name: 50-network-config.conf
        content: |
          [Service]
          ExecStartPre=-/usr/bin/etcdctl mk /coreos.com/network/config '{"Network": "10.244.0.0/16", "SubnetLen": 24, "SubnetMin": "10.244.0.0", "Backend": {"Type": "vxlan"}}'
```

that will start the flannel service on boot before docker.

jannispl:
closing in favor of #4687 and #4724

PR #4688, opened 2015-10-29
fix(contrib/gce): replace discovery url properly

### LINKED ISSUE #4588: deis/contrib/gce/gce-user-data generates an invalid cloud-config

mboersma:
Replaces `#DISCOVERY_URL` explicitly to fix GCE provisioning. See #4680--thanks @ritazh!

Closes #4588.

### DIFF
--- a/contrib/azure/create-azure-user-data
+++ b/contrib/azure/create-azure-user-data
@@ -57,7 +57,7 @@ def main():
     coreos_template = get_file("../coreos/user-data.example")
 
     coreos_template_w_url = None
-    with open(coreos_template.name, 'r') as file :
+    with open(coreos_template.name, 'r') as file:
       coreos_template_w_url = file.read()
 
     coreos_template_w_url = coreos_template_w_url.replace('#DISCOVERY_URL', url)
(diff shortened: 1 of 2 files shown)


ritazh:
:+1:

PR #4671, opened 2015-10-27
chore(builder): update Docker to 1.8.3

mboersma:
This should be the final change necessary to future-proof Deis against the [disabling of the V1 API](http://blog.docker.com/2015/10/docker-hub-deprecation-1-5/) at Docker Hub on December 7.

Now that #4581 is merged, `deis pull` does not depend on the extended V1 registry-to-registry transfer API and works with images built by any version of Docker. The "last-mile" layer of ENV variables is also done by Docker now and doesn't depend on the V1 registry API. Bumping Docker up to the current 1.8.3 in deis-builder should ensure that new or upgraded Deis clusters can survive December 7.

Docker registry V2 work for deis-registry is also close to completion, but needs a migration tool and further testing: see #4641.

Refs #4636.

TODO:
- [x] add sha256sum checking to downloaded docker binary
- [x] add the equivalent of `flock` locks around builder `docker pull` and `docker build` operations

PR #4689, opened 2015-10-29
fix(buidler): Fix problem with missed git repos after builder restart

### LINKED ISSUE #3458: /home/git/project.git/hooks/pre-receive: No such file or directory
Uh. I tried to push a new version to Deis. Mid-push my computer shut down, and when I returned I could not push again because a push was in progress.

I read in the issue tracker that you could a) delete the lock file or b) restart the builder. I didn't find any lockfile, so I restarted the builder.

Now it seems like nothing works.

I get this error when pushing

```
onbjerg at AIR in ~/Projects/wallz on develop
$ git push deis develop:master
/usr/local/bin/gitreceive: line 33: /home/git/wallz.git/hooks/pre-receive: No such file or directory
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.```

### PR DESCRIPTION
**Pull request by rvadim**
Resolve #3458
Just handle situation when /home/git/app.git/ contains only cache and build dir.

### DIFF
--- a/builder/git/git.go
+++ b/builder/git/git.go
@@ -205,10 +205,10 @@ func createRepo(c cookoo.Context, repoPath, gitHome string) (bool, error) {
 		if fi.IsDir() {
 			configPath := filepath.Join(repoPath, "config")
 			if _, cerr := os.Stat(configPath); cerr == nil {
-				log.Infof(c, "Directory '%s' already exists.", repoPath)
+				log.Debugf(c, "Directory '%s' already exists.", repoPath)
 				return true, nil
 			} else {
-				log.Warnf(c, "No config file found at path `%s`; removing it and recreating.", repoPath)
+				log.Warnf(c, "No config file found at `%s`; removing it and recreating.", repoPath)
 				if err := os.RemoveAll(repoPath); err != nil {
 					return false, fmt.Errorf("Unable to remove path '%s': %s", repoPath, err)
 				}
@@ -220,7 +220,7 @@ func createRepo(c cookoo.Context, repoPath, gitHome string) (bool, error) {
 			}
 		}
 	} else if os.IsNotExist(err) {
-		log.Infof(c, "Unable to get stat for path '%s': %s .", repoPath, err)
+		log.Debugf(c, "Unable to get stat for path '%s': %s .", repoPath, err)
 	} else {
 		return false, err
 	}
--- a/builder/git/git.go
+++ b/builder/git/git.go
@@ -167,6 +167,30 @@ func plumbCommand(cmd *exec.Cmd, channel ssh.Channel, sidechannel io.Writer) *sy
 
 var createLock sync.Mutex
 
+// initRepo create a directory and init a new Git repo
+func initRepo(repoPath, gitHome string, c cookoo.Context) (bool, error) {
+	log.Infof(c, "Creating new directory at %s", repoPath)
+	// Create directory
+	if err := os.MkdirAll(repoPath, 0755); err != nil {
+		log.Warnf(c, "Failed to create repository: %s", err)
+		return false, err
+	}
+	cmd := exec.Command("git", "init", "--bare")
+	cmd.Dir = repoPath
+	if out, err := cmd.CombinedOutput(); err != nil {
+		log.Warnf(c, "git init output: %s", out)
+		return false, err
+	}
+
+	hook, err := prereceiveHook(map[string]string{"GitHome": gitHome})
+	if err != nil {
+		return true, err
+	}
+	ioutil.WriteFile(filepath.Join(repoPath, "hooks", "pre-receive"), hook, 0755)
+
+	return true, nil
+}
+
 // createRepo creates a new Git repo if it is not present already.
 //
 // Largely inspired by gitreceived from Flynn.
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by carmstrong** [comment]
Thanks for the work on this, @rvadim!

**Comment by nathansamson** [comment]
Any chance for this to be merged in for 1.12.0/ 1.12.1? This hit me again yesterday which is kinda annoying...

**Comment by carmstrong** [comment]
> Any chance for this to be merged in for 1.12.0/ 1.12.1?

I think we're waiting for @krancour et. al. to confirm this resolves #3458. Unfortunately, unlikely to make 1.12 at this point, although certainly a candidate for 1.12.1

File: builder/git/git.go [hunk 1]
```diff
@@ -177,37 +201,31 @@ func createRepo(c cookoo.Context, repoPath, gitHome string) (bool, error) {
...
+				return true, nil
+			} else {
+				log.Warnf(c, "Path '%s' have not config file, remove it and recreate.", repoPath)
+				if err := os.RemoveAll(repoPath); err != nil {
+					return false, fmt.Errorf("Unable to remove path '%s': %s", repoPath, err)
+				}
+				return initRepo(repoPath, gitHome, c)
+			}
+		} else {
+			log.Warnf(c, "Path '%s' not a directory, remove it and recreate.", repoPath)
```
krancour: Can we please change this to:

```
log.Warnf(c, "Path `%s` is not a directory; removing it and recreating.", repoPath)
```

builder/git/git.go
```diff
@@ -177,37 +201,31 @@ func createRepo(c cookoo.Context, repoPath, gitHome string) (bool, error) {
...
-			log.Warnf(c, "Failed to create repository: %s", err)
-			return false, err
+	if fi, err := os.Stat(repoPath); err == nil {
+		if fi.IsDir() {
+			configPath := filepath.Join(repoPath, "config")
+			if _, cerr := os.Stat(configPath); cerr == nil {
+				log.Infof(c, "Directory '%s' already exists.", repoPath)
+				return true, nil
+			} else {
+				log.Warnf(c, "Path '%s' have not config file, remove it and recreate.", repoPath)
```
krancour: Can we please change this to:

```
log.Warnf(c, "No config file found at path `%s`; removing it and recreating.", repoPath)
```

builder/git/git.go
```diff
@@ -177,37 +201,31 @@ func createRepo(c cookoo.Context, repoPath, gitHome string) (bool, error) {
...
-			return true, err
-		}
-		ioutil.WriteFile(filepath.Join(repoPath, "hooks", "pre-receive"), hook, 0755)
-
-		return true, nil
-	} else if err == nil {
-		return false, errors.New("Expected directory, found file.")
-	} else {
-		return false, err
+	} else if err != nil {
```
krancour: Should be using `os.IsNotExist(err)` to isolate when the file doesn't exist vs. some other unanticipated error.

builder/git/git.go
```diff
@@ -177,37 +201,31 @@ func createRepo(c cookoo.Context, repoPath, gitHome string) (bool, error) {
...
+			configPath := filepath.Join(repoPath, "config")
+			if _, cerr := os.Stat(configPath); cerr == nil {
+				log.Infof(c, "Directory '%s' already exists.", repoPath)
+				return true, nil
+			} else {
+				log.Warnf(c, "Path '%s' have not config file, remove it and recreate.", repoPath)
+				if err := os.RemoveAll(repoPath); err != nil {
+					return false, fmt.Errorf("Unable to remove path '%s': %s", repoPath, err)
+				}
+				return initRepo(repoPath, gitHome, c)
```
krancour: There are many calls to `initRepo()` in the various branches of this conditional.  Move this to the end, after the conditional, since most paths need to result in this.  Paths that do not can just return early.

krancour:
Overall, this looks like a nice improvement.  See some of my comments in-line requesting minor cleanup.  I still need to test this manually, but I would say this looks very promising.

krancour:
I cannot seem to reproduce #3458 today on Vagrant or AWS.  I have tried v1.11.2 and v1.10.0.  Without reproducing, it's for me to say this has solved anything-- altho the code looks promising.  @rvadim has this definitely resolved the issues that you were seeing in your cluster?

krancour:
@rvadim typo in commit message also needs to be fixed.  Change "buidler" to "builder".  (Want to keep commit history grepable by component.)

krancour:
I've spent all day on this and one thing I have discovered is that I can reproduce this using version 1.11.2 or prior, but I can't reproduce it with the code that's currently in the master branch-- and that's code that does _not_ include this PR.  I strongly believe that something else that went in recently inadvertently resolved #3458 altho I am not yet sure what.  cc @technosophos @mboersma

krancour:
Using `git bisect`, it would seem that #4671 inadvertently fixed the issue.  I'm not at all sure why.

rvadim:
Since apply this patch to builder, no more requests with "unable to deploy, git ..." from our cluster users. (v1.10.0 and v1.11.2)

krancour:
@rvadim would you mind trying using the builder from master and let me know what you find?  Using what's currently in master (i.e. 1.12.0-dev) I don't see any issues.

rvadim:
In fact, this part of builder is risky, in my opinion, we need to fix it in any way.
@krancour ok, i will try to check version of builder from master. It takes 2 - 4 hours

krancour:
@rvadim, we appreciate all the efforts.  2 - 4 hours seems high.  Just in case you were not aware... you don't need to build and deploy the whole platform from source.  If you have a cluster going already, you can build and deploy just the builder component with:

```
$ make -C builder build deploy
```

rvadim:
At this moment i can't reproduce this on master branch builder.

krancour:
@rvadim thanks for checking.  I'm going to move the discussion back to #3458 as there were more participants there.  I'm curious to see what results the rest of them have with the builder from master.

krancour:
@rvadim since there's been no luck in you or me or anyone else identifying the root cause and despite thinking that #4671 inadvertently fixed this, I think this PR (which treats the symptom, really) still contains some positive improvements and a nice fail-safe in case the problem isn't already quite as fixed as we think/hope it is.  I do want to merge this...

If you would be so kind, please squash the two commits (and reword to fix spelling of "buidler" to "builder") and this then gets my LGTM.

Really appreciate your hard work on this issue.  I owe you a :beer:

bacongobbler:
no kill like overkill :+1:

rvadim:
Done. I fixed typo and recheck all cases with last changes. Thank you.

krancour:
@rvadim thank you very much!

builder/git/git.go
```diff
@@ -177,37 +201,30 @@ func createRepo(c cookoo.Context, repoPath, gitHome string) (bool, error) {
...
-			log.Warnf(c, "git init output: %s", out)
-			return false, err
-		}
-
-		hook, err := prereceiveHook(map[string]string{"GitHome": gitHome})
-		if err != nil {
-			return true, err
+	if fi, err := os.Stat(repoPath); err == nil {
+		if fi.IsDir() {
+			configPath := filepath.Join(repoPath, "config")
```
bacongobbler: wouldn't config be in `filepath.Join(repoPath, ".git", "config")`?

bacongobbler: Ah never mind, I see that's already added to the path [here](https://github.com/deis/deis/blob/29b6c0b2aab51739cd2a128f7fd5dbbb6b305809/builder/git/git.go#L93).

builder/git/git.go
```diff
@@ -177,37 +201,30 @@ func createRepo(c cookoo.Context, repoPath, gitHome string) (bool, error) {
...
+			if _, cerr := os.Stat(configPath); cerr == nil {
+				log.Infof(c, "Directory '%s' already exists.", repoPath)
+				return true, nil
+			} else {
+				log.Warnf(c, "No config file found at path `%s`; removing it and recreating.", repoPath)
```
bacongobbler: we should print configPath here instead

builder/git/git.go
```diff
@@ -177,37 +201,30 @@ func createRepo(c cookoo.Context, repoPath, gitHome string) (bool, error) {
...
+				return false, fmt.Errorf("Unable to remove path '%s': %s", repoPath, err)
+			}
 		}
-		ioutil.WriteFile(filepath.Join(repoPath, "hooks", "pre-receive"), hook, 0755)
-
-		return true, nil
-	} else if err == nil {
-		return false, errors.New("Expected directory, found file.")
+	} else if os.IsNotExist(err) {
+		log.Infof(c, "Unable to get stat for path '%s': %s .", repoPath, err)
```
bacongobbler: Let's put this one at `log.Debugf()`. It's not an error (we're going to create this directory), but it's still useful information for debugging.

builder/git/git.go
krancour: Partially agree.  Changing that would muddy the second half of the statement-- the "removing it and recreating" part.  Maybe better wording would be:

```
log.Warnf(c, "No config file found in repo at `%s`; removing it and recreating.", repoPath)
```

builder/git/git.go
krancour: ack

krancour:
@rvadim can you make the two small changes mentioned above?  And then I think this would be good to go.

builder/git/git.go
```diff
@@ -177,37 +201,30 @@ func createRepo(c cookoo.Context, repoPath, gitHome string) (bool, error) {
...
-		}
-
-		hook, err := prereceiveHook(map[string]string{"GitHome": gitHome})
-		if err != nil {
-			return true, err
+	if fi, err := os.Stat(repoPath); err == nil {
+		if fi.IsDir() {
+			configPath := filepath.Join(repoPath, "config")
+			if _, cerr := os.Stat(configPath); cerr == nil {
+				log.Infof(c, "Directory '%s' already exists.", repoPath)
```
krancour: This should be `log.Debugf` instead.  Creating a new dir (as on line 172) is good info to log.  The fact that we _don't_ need to create a new dir is generally extraneous info that one would only want to see during debugging, I imagine.

rvadim:
Fixed

krancour:
Ping @bacongobbler.  Can you give this a second glance when you have a chance?

krancour:
@rvadim thanks!  We're just about there.

bacongobbler:
conference internet is best internet

PR #4699, opened 2015-10-31
fix(contrib/linode): fix discovery url issues in deployment scripts

croemmich:
Fix discovery url replacement and detection caused by user-data changes.

### DIFF
--- a/contrib/linode/apply-firewall.py
+++ b/contrib/linode/apply-firewall.py
@@ -46,7 +46,8 @@ def get_discovery_url_from_user_data():
     try:
         current_dir = os.path.dirname(__file__)
         user_data_file = file(os.path.abspath(os.path.join(current_dir, name)), 'r')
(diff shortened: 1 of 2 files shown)


mboersma:
Thanks so much for patching up Linode after the apparently breaking `#DISCOVERY_URL` change. :disappointed_relieved:  For reference, here is how this was fixed for [Azure](https://github.com/deis/deis/pull/4680) and [GCE](https://github.com/deis/deis/pull/4688).

croemmich:
You bet!

Is there an advantage to `with open(file.name, 'r') as file` as was done for Azure and GCE?

mboersma:
`with open(file)` just ensures that the file is closed explicitly even if an exception occurs under the `with` block. But this should also happen when the file object is garbage-collected, or when the python interpreter exits. So it's good general practice but not a big deal for a read-only file handle in a run-then-exit script like this I'd say.

croemmich:
Good to know. Thanks!

PR #4700, opened 2015-10-31
fix(contrib/linode): fix issues with DHCP on Linode

croemmich:
Disable DHCP and mark the connection as critical on Linode. This has completely removed the network drops I was experiencing.

### DIFF
--- a/contrib/linode/linode-user-data-template.yaml
+++ b/contrib/linode/linode-user-data-template.yaml
@@ -5,18 +5,23 @@ coreos:
     metadata: name=%H
   units:
   - name: 00-eth0.network
-    runtime: false
+    runtime: true
     content: |
       [Match]
       Name=eth0
 
       [Network]
+      DHCP=ipv4
       Address=$public_ipv4/24
       Address=$private_ipv4/17
       Gateway=$gateway.1
       DNS=72.14.179.5
       DNS=8.8.8.8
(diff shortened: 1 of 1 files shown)


mboersma:
So `UseDNS=false` seems to mean that the resolver will ignore any nameserver entries returned by the DHCP response and stick with the hardcoded `DNS=` entries above it...correct? I had a hard time finding much info about this user-data stanza and its settings, so any further explanation would be helpful.

Regardless, I trust you've tested this and it helped your Linode deploys of Deis. Thanks a ton for keeping Deis contrib/ healthy @croemmich!

croemmich:
That's my understanding as well, but as you said, documentation is limited. For some reason the entries in the unit were being set up correctly on the initial run, but on the next lease they would get wiped out. `UseDNS=false` seems to have fixed the issue.

The most important part of this commit is `CriticalConnection=true`. I was having a very strange issue in which my ip was being dropped by the interface because the valid_lft on the address was not being updated even though DHCP leases were being issued successfully. See coreos/docs#377 for more info.

I've been running this configuration for about 90 days with success. (Which probably means it's working for all of the deployments on Linode :wink:)

carmstrong:
> I've been running this configuration for about 90 days with success.

Good enough for me. Thanks @croemmich!

Issue #4708, opened 2015-11-03
deis-database won't upgrade to postgresql 9.4.4


mboersma:
Upgrading from v1.11.2 to v1.12.0-rc2 on AWS failed on deis-database:
```console
$ deisctl journal database
...

Nov 03 03:28:52 ip-10-21-2-50.us-west-2.compute.internal sh[7781]: ok
Nov 03 03:28:52 ip-10-21-2-50.us-west-2.compute.internal sh[7781]: No usable system locales were found.
Nov 03 03:28:52 ip-10-21-2-50.us-west-2.compute.internal sh[7781]: Use the option "--debug" to see details.
Nov 03 03:28:52 ip-10-21-2-50.us-west-2.compute.internal sh[7781]: creating conversions ... ok
Nov 03 03:28:52 ip-10-21-2-50.us-west-2.compute.internal sh[7781]: creating dictionaries ... ok
Nov 03 03:28:52 ip-10-21-2-50.us-west-2.compute.internal sh[7781]: setting privileges on built-in objects ... ok
Nov 03 03:28:52 ip-10-21-2-50.us-west-2.compute.internal sh[7781]: creating information schema ... ok
Nov 03 03:28:52 ip-10-21-2-50.us-west-2.compute.internal sh[7781]: loading PL/pgSQL server-side language ... ok
Nov 03 03:28:52 ip-10-21-2-50.us-west-2.compute.internal sh[7781]: vacuuming database template1 ... ok
Nov 03 03:28:52 ip-10-21-2-50.us-west-2.compute.internal sh[7781]: copying template1 to template0 ... ok
Nov 03 03:28:52 ip-10-21-2-50.us-west-2.compute.internal sh[7781]: copying template1 to postgres ... ok
Nov 03 03:28:53 ip-10-21-2-50.us-west-2.compute.internal sh[7781]: syncing data to disk ...
Nov 03 03:28:53 ip-10-21-2-50.us-west-2.compute.internal sh[7781]: WARNING: enabling "trust" authentication for local connections
Nov 03 03:28:53 ip-10-21-2-50.us-west-2.compute.internal sh[7781]: You can change this by editing pg_hba.conf or using the option -A, or
Nov 03 03:28:53 ip-10-21-2-50.us-west-2.compute.internal sh[7781]: --auth-local and --auth-host, the next time you run initdb.
Nov 03 03:28:53 ip-10-21-2-50.us-west-2.compute.internal sh[7781]: ok
[...]

PR #4709, opened 2015-11-03
fix(database): revert to alpine:3.1 for PostgreSQL

mboersma:
wal-e failed to restore from version 9.3.7 in alpine:3.1 to version 9.4.4 in alpine:3.2.

Refs #4708.

### DIFF
--- a/database/Dockerfile
+++ b/database/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:3.2
+FROM alpine:3.1
 
 # install common packages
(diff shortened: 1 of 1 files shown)


carmstrong:
Yeah, I think in order to accomplish this, we'll need to [perform an upgrade](http://www.postgresql.org/docs/9.4/static/upgrading.html) of the data and then kick off a fresh backup. That will require work and testing that we don't really need to accomplish right now. Keeping on 9.3.x is just fine.

PR #4720, opened 2015-11-05
fix(create_bucket): check for existence of None

### LINKED ISSUE #4713: Registry tries creating already existing bucket v.12.1
Just upgraded to v.12.1 and the registry throws S3 errors. Previous similar errors in v.11 have been resolved with the fix mentioned in #4710.

We are using a cephless cluster on AWS.
Deis tries to create bucket even if it already owns it. 

```
Nov 05 08:17:32 ip-10-21-2-170.eu-west-1.compute.internal sh[30590]: registry: creating bucket
Nov 05 08:17:32 ip-10-21-2-170.eu-west-1.compute.internal sh[30590]: Traceback (most recent call last):
Nov 05 08:17:32 ip-10-21-2-170.eu-west-1.compute.internal sh[30590]: File "/app/bin/create_bucket", line 21, in <module>
Nov 05 08:17:32 ip-10-21-2-170.eu-west-1.compute.internal sh[30590]: conn.create_bucket(name)
Nov 05 08:17:32 ip-10-21-2-170.eu-west-1.compute.internal sh[30590]: File "/usr/lib/python2.7/site-packages/boto/s3/connection.py", line 616, in create_bucket
Nov 05 08:17:32 ip-10-21-2-170.eu-west-1.compute.internal sh[30590]: response.status, response.reason, body)
Nov 05 08:17:32 ip-10-21-2-170.eu-west-1.compute.internal sh[30590]: boto.exception.S3CreateError: S3CreateError: 409 Conflict
Nov 05 08:17:32 ip-10-21-2-170.eu-west-1.compute.internal sh[30590]: <?xml version="1.0" encoding="UTF-8"?>
Nov 05 08:17:32 ip-10-21-2-170.eu-west-1.compute.internal sh[30590]: <Error><Code>BucketAlreadyOwnedByYou</Code><Message>Your previous request to create the named bucket succeeded and you already own it.</Message><BucketName>dreicloud-deis-khan</BucketName><RequestId>381338F14C6E9FB1</RequestId><HostId>RyvNTqiBMN7KDhjOQcp7u8WaCBwEG3AlyZ80yx6EUECb54ICRIu+d7wLjDIx6MUKQFLkb53M83Y=</HostId></Error>
```

bacongobbler:
There are small implications between `if not x` and `if x is None` in
Python:

 - `if not x` checks if the value is false (empty list, `None`, `False`)
 - `if x is None` checks if x is equal to the literal value `None`.

conn.lookup returns `None` if no bucket exists. This code is explicitly
stated in their docs as the proper way to check if a bucket does not
exist.

See: http://boto.readthedocs.org/en/latest/s3_tut.html

I'm not sure how the original issue is caused, but this will likely
resolve it.

closes #4713 

ping @lukasmartinelli and @afterthought. Would you mind verifying that this patch works for you?

### DIFF
--- a/database/templates/create_bucket
+++ b/database/templates/create_bucket
@@ -12,5 +12,5 @@ conn = boto.connect_s3(
     calling_format=OrdinaryCallingFormat())
 name = sys.argv[1]
 
-if not conn.lookup(name):
+if conn.lookup(name) is None:
     conn.create_bucket(name)
--- a/registry/templates/create_bucket
+++ b/registry/templates/create_bucket
@@ -22,5 +22,5 @@ name = '{{ getv "/deis/registry/s3bucket" }}'
 name = '{{ getv "/deis/registry/bucketName" }}'
 {{ end }}
 
-if not conn.lookup(name):
+if conn.lookup(name) is None:
     conn.create_bucket(name)


obmarg:
I can't see how this change would fix anything.

> if not x checks if the value is false (empty list, None, False)

> conn.lookup returns None if no bucket exists.

So the only way `if not conn.lookup()` wouldn't work as expected would be if `conn.lookup()` returned something equivalent to False (but not None) on success.  It _should_ return a Bucket object if the lookup is sucesful, which should not be equivalent to False.  I checked:

```
In [17]: b = conn.lookup('a-bucket-that-exists')

In [18]: bool(b)
Out[18]: True
```

bacongobbler:
@obmarg I think the next step is figuring out what's causing the root issue in #4713. Regardless, this technically is an improvement if this is the "recommended" way of checking for presence of a bucket.

Issue #4718, opened 2015-11-05
Flannel doesn't start after reboot


afterthought:
https://github.com/deis/deis/commit/445dbaecfadb1e2690734413cc43f30abcb13b21#diff-be438a7ce6a159afe476e6cb80c59eddR53

I had been using "set" instead of "mk" in my unit. I was updating my units to match the official Deis files and when I changed this to "mk" flannel wouldn't boot because this key was already set. I might be missing something, because I would think others would have run into this very fast. Might just be something unique in my setup.

aledbf:
@afterthought in the flanneld.service is missing `command: start`

afterthought:
I added command:start as well. However, before I added that it was still starting (maybe because of a dependency?) and it kept failing over and over on the line creating the etcd key.

mboersma:
This fix for `flannel.service` is in [Deis v1.12.1](https://github.com/deis/deis/releases/tag/v1.12.1).

PR #4721, opened 2015-11-05
fix(user-data): always start flannel on boot

aledbf:


krancour:
Fixes #4718

carmstrong:
@aledbf Does this also replace #4687 and fix #4686?

aledbf:
@carmstrong I think it will. Docker will start with the flag `--ip-masq=false` (defined in `/var/run/flannel_docker_opts.env`)

carmstrong:
closes #4687 
fixes #4686

contrib/coreos/user-data.example
```diff
@@ -48,11 +48,12 @@ coreos:
           [Service]
           Environment="DOCKER_OPTS=--insecure-registry 10.0.0.0/8 --insecure-registry 172.16.0.0/12 --insecure-registry 192.168.0.0/16 --insecure-registry 100.64.0.0/10"
     - name: flanneld.service
+      command: start
       drop-ins:
       - name: 50-network-config.conf
         content: |
           [Service]
-          ExecStartPre=/usr/bin/etcdctl mk /coreos.com/network/config '{"Network": "10.244.0.0/16", "SubnetLen": 24, "SubnetMin": "10.244.0.0", "Backend": {"Type": "vxlan"}}'
+          ExecStartPre=-/usr/bin/etcdctl mk /coreos.com/network/config '{"Network": "10.244.0.0/16", "SubnetLen": 24, "SubnetMin": "10.244.0.0", "Backend": {"Type": "vxlan"}}'
```
carmstrong: Are we assuming that if this fails, the network already exists? It could fail for other reasons, like etcd being unavailable.

aledbf: @carmstrong this drop-in just sets the value. The default `flaneld.service` contains this `After=etcd.service etcd2.service early-docker.service` which means that etcd is running

carmstrong: Ok. And we're okay if it fails because that means the key's already there?

PR #4725, opened 2015-11-06
fix(user-data): fix flanneld configuration setup

### LINKED ISSUE #4718: Flannel doesn't start after reboot

jannispl:
fixes #4718
replaces #4721


deis-admin:

krancour:
This is identical to the PR it was intended to replace.  The other may have come into that identical state after this one was submitted, but it already has two lgtms, so I am closing this one in favor of the other.

PR #4729, opened 2015-11-07
fix(registry): fix create_bucket s3 compatability

obmarg:
The create_bucket script in the registry has some incompatibilities with
s3. In 1.12 deis is using the boto `lookup` function to determine
whether a bucket exists or not.  The boto s3 connection is created using
an OrdinaryCallingFormat, for ceph compatibility.  Unfortunately, S3
returns a 301 when lookup is called with OrdinaryCallingFormat, which
causes the lookup function to return None.

Prior to 1.12 the boto `get_all_buckets` function was used to determine
whether a bucket existed or not, which does not seem to have this
problem.

I've fixed it by updating the create_bucket script to not pass an
OrdinaryCallingFormat if the `/deis/registry/s3bucket` setting is
present, which fixes this issue.  I've also omitted the
`is_secure=False` setting, which would cause boto to use s3 over http.

Additionally, if `/deis/store/gateway/accessKey` is not present (which
should be the case in a stateless setup) I'm using the s3 access keys
configured for the registry.

This fixes #4713 (and maybe #4710 and #4688).  #4710 was labelled as a
documentation bug, but I'm not sure it is - seems like the stateless
registry should just read the s3 configuration from a single place
rather than relying on the `/deis/store/gateway/accessKey` config that
is usually set by the store-gateway.

PR #4730, opened 2015-11-07
Don't require deis store.

jwaldrip:
This breaks if you try to boot a cluster in stateless mode. This check prevents it from trying to connect to the old store.


deis-admin:

registry/templates/create_bucket
```diff
@@ -22,5 +23,6 @@ name = '{{ getv "/deis/registry/s3bucket" }}'
 name = '{{ getv "/deis/registry/bucketName" }}'
 {{ end }}
 
-if conn.lookup(name) is None:
+if not conn.lookup(name):
```
helgi: cf12b888ffc180144e13c455c9d4de8522f942af just changed this from `not` to `is None` - Any reason to revert that?

registry/templates/create_bucket
```diff
@@ -1,3 +1,4 @@
+{{ if exists "/deis/store/gateway" }}
```
helgi: Doing this means you can't have a bucket created for `stateless` clusters that use `s3` for baking - This particular script is a little confused and needs love to make it work for both scenarios. #4729 cleans up a lot of that, does it achieve what you are looking for?

jwaldrip: It would, but in this case, the assumption in this script is that we are using the deis-store, which we are not.

registry/templates/create_bucket
jwaldrip: This is from the last tag. I didn't merge in. But I can update it.

jwaldrip:
Seems to also be solved with #4729

PR #4732, opened 2015-11-08
ref(user-data): disable iptables in docker

### LINKED ISSUE #4686: custom-firewall.sh: jump to firewall is appended rather than prepended

aledbf:
Flannel provides the required rule to access internet from the containers (`--ip-mask=true`) and a dedicated ip address for all the containers running in the node. This means that is not required the creation of any iptable rules to forward traffic

closes #4686

### DIFF
--- a/contrib/coreos/user-data.example
+++ b/contrib/coreos/user-data.example
@@ -48,10 +48,10 @@ coreos:
           [Unit]
           Requires=flanneld.service
           After=flanneld.service
-      - name: 50-insecure-registry.conf
+      - name: 50-custom-options.conf
         content: |
           [Service]
-          Environment="DOCKER_OPTS=--insecure-registry 10.0.0.0/8 --insecure-registry 172.16.0.0/12 --insecure-registry 192.168.0.0/16 --insecure-registry 100.64.0.0/10"
+          Environment="DOCKER_OPTS=--iptables=false --insecure-registry 10.0.0.0/8 --insecure-registry 172.16.0.0/12 --insecure-registry 192.168.0.0/16 --insecure-registry 100.64.0.0/10"
     - name: flanneld.service
       command: start
       drop-ins:


carmstrong:
Could we somehow add something to `DOCKER_OPTS` rather than overridding the ExecStart? In this scenario, we'll have to update the drop-in unit anytime the Docker commandline changes

aledbf:
> Could we somehow add something to DOCKER_OPTS rather than overridding the ExecStart? I

no because is not possible to reference the previous DOCKER_OPTS  value (systemd fault).
EnvironmentFile is the alternative for that but the all current docker command variables are being used so is not a real solution either.

carmstrong:
@aledbf Could you try rebasing this? It looks like we're running into #4795 which I don't really understand.

krancour:
FYI, I'm doing some testing with this today.

krancour:
@carmstrong do you know how to reliably reproduce the docker networking issues that this hopefully addresses?  It happens when containers crash.  I'm not quite sure how to force a container to crash.

bacongobbler:
@krancour just make the CMD cause the container to exit non-zero:

```
FROM debian:jessie
CMD ["/bin/exit", "2"]
```

I'd imagine that *should* work for testing purposes.

carmstrong:
We may need to rebase this again after merging #4800

krancour:
Doesn't Jenkins test PRs against master?

bacongobbler:
it does, which is why this would need to be rebased so that the tests will pass :)

krancour:
@sgoings has educated me that "against master" doesn't mean in our current Jenkins setup what it means in Travis for v2.  So I understand the need for the rebase now @bacongobbler

carmstrong:
@aledbf If you could rebase this one more time, we should be good to go now.

krancour:
Integration tests are repeatedly failing on this:

```
Using ELB deis-test-DeisWebE-W5ZHCSDH0TMM
waiting for record to sync: <Status:PENDING>
<Status:INSYNC>
waiting for wildcard domain to become available...ok
Waiting for ELB to see an instance in service...

>>> Running integration suite with Python Client <<<

make -C tests/ test-full
make[1]: Entering directory `/home/jenkins/workspace/test-ec2/src/github.com/deis/deis/tests'
gofmt -l dockercli etcdutils mock utils apps_test.go auth_test.go builds_test.go config_test.go domains_test.go healthcheck_test.go integration_test.go keys_test.go limits_test.go main.go perms_test.go ps_test.go releases_test.go smoke_test.go tags_test.go users_test.go
godep go vet github.com/deis/deis/tests github.com/deis/deis/tests/dockercli github.com/deis/deis/tests/etcdutils github.com/deis/deis/tests/mock github.com/deis/deis/tests/utils
golint ./...
dockercli/dockercli.go:128:1: exported function CreateVolume should have comment or be unexported
dockercli/dockercli.go:160:1: exported function WaitForLine should have comment or be unexported
GOMAXPROCS=1 godep go test -tags integration -v -run TestGlobal
=== RUN   TestGlobal
auth:register http://deis.test-6e2cb66ed0.deis.works --username=test --password=asdf1234 --email=test@test.co.nz
Error: http://deis.test-6e2cb66ed0.deis.works does not appear to be a valid Deis controller.
error at command wait
--- FAIL: TestGlobal (1.45s)
	itutils.go:268: exit status 1
FAIL
exit status 1
FAIL	github.com/deis/deis/tests	1.450s
godep: go exit status 1
make[1]: Leaving directory `/home/jenkins/workspace/test-ec2/src/github.com/deis/deis/tests'
```

krancour:
I kicked off a build of master as a sanity check and it's _not_ failing in this spot.

I do believe this issue might be related to the PR.

@aledbf any thoughts?

krancour:
Ok.  The plot thickens.  Comparing the logs from multiple failed executions of the CI job against this PR, we always get the same error, but not always at the same point.  It's always in the integration tests, but the first few sometimes pass, while other times they do not.

krancour:
Here's an unrelated PR failing with the same issue... so I am no longer thinking this PR is to blame.  (Still, I want to see it pass CI before merging in case there are _other_ issues lurking.)

https://ci.deis.io/job/test-ec2/3735/consoleFull

krancour:
Or maybe not... the error is occurring at a catch-all for many failed integration tests (any where failure wasn't expected):

https://github.com/deis/deis/blob/master/tests/utils/itutils.go#L268

mboersma:
Jenkins, test this please.

carmstrong:
I would really like to get this passing and part of the LTS release. Docker iptables rules continue to bite us.

krancour:
@carmstrong, I spent days trying to figure out why this is failing CI and couldn't.  I have tested the change manually without difficulty.

bacongobbler:
Unfortunately this consistently fails on https://github.com/deis/deis/pull/4732#issuecomment-165263930 even after I rebased on master with #4880 so I think this is a legitimate failure.

krancour:
@mboersma we're not going to follow through with this one?

krancour:
Nevermind.  I didn't notice initially that it was superseded by #4992.  Ignore me.

PR #4735, opened 2015-11-09
ref(contrib): expose COREOS ENV vars via export in utils.sh. Add DEIS_RELEASE

helgi:
Without this COREOS_VERSION and COREOS_CHANNEL aren't picked up by sourcing or external scripts

### DIFF
--- a/contrib/utils.sh
+++ b/contrib/utils.sh
@@ -12,5 +12,6 @@ function echo_green {
   echo -e "\033[0;32m$1\033[0m"
 }
 
-COREOS_CHANNEL=${COREOS_CHANNEL:-stable}
-COREOS_VERSION=${COREOS_VERSION:-766.5.0}
+export COREOS_CHANNEL=${COREOS_CHANNEL:-stable}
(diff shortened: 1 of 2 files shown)


carmstrong:
If we're adding the release version to a new file, we should add the file to `bumpver` in https://github.com/deis/deis/blob/master/docs/roadmap/releases.rst#patch-release

helgi:
Done

carmstrong:
One (hopefully) final thing - the version in master should be `1.13.0-dev`: https://github.com/deis/deis/commit/d4673ea3754703b5383687a8356013e87cde6243

helgi:
Ah hmm bummer that may not work for my purpose - The reason I use this is to pull the latest stable of deisctl into a bastion host. Could just move to using the non-versioned curl URL instead of relying on that one. Will still bump the version in `utils.sh`

carmstrong:
LGTM now, but @sgoings should probably re-review and issue a LGTM 2.

carmstrong:
:ship:

PR #4737, opened 2015-11-10
chore(MAINTAINERS): don't enumerate maintainers

carmstrong:
This steals the language for contributing maintainers from the Helm
project. We have many contributing maintainers who aren't enumerated
here.

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -66,12 +66,7 @@ conversations where the contributing maintainer is knowledgeable.
 
 The duties of a contributing maintainer are very similar to those of a core maintainer, but they are limited to areas of the Deis project where the contributing maintainer is knowledgeable.
 
-The current contributing maintainers of Deis:
-* Johannes Würbach - <johannes.wuerbach@googlemail.com> ([@johanneswuerbach](https://github.com/johanneswuerbach))
-* Manuel de Brito Fontes - <alejandro.debrito@soficom.cl> ([@aledbf](https://github.com/aledbf))
-* Seth Goings - <sgoings@engineyard.com> ([@sgoings](https://github.com/sgoings))
-* Sivaram Mothiki - <smothiki@engineyard.com> ([@smothiki](https://github.com/smothiki))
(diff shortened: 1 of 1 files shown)


carmstrong:
cc @gabrtv @mboersma

MAINTAINERS.md
```diff
@@ -66,12 +66,7 @@ conversations where the contributing maintainer is knowledgeable.
 
 The duties of a contributing maintainer are very similar to those of a core maintainer, but they are limited to areas of the Deis project where the contributing maintainer is knowledgeable.
 
-The current contributing maintainers of Deis:
-* Johannes Würbach - <johannes.wuerbach@googlemail.com> ([@johanneswuerbach](https://github.com/johanneswuerbach))
-* Manuel de Brito Fontes - <alejandro.debrito@soficom.cl> ([@aledbf](https://github.com/aledbf))
-* Seth Goings - <sgoings@engineyard.com> ([@sgoings](https://github.com/sgoings))
-* Sivaram Mothiki - <smothiki@engineyard.com> ([@smothiki](https://github.com/smothiki))
-* Wael Nasreddine - <waelnasreddine@gmail.com> ([@kalbasit](https://github.com/kalbasit))
+Contributing maintainers are defined in practice as those who have write access to the Deis repository. All maintainers can review pull requests and add LGTM labels as appropriate.
```
bacongobbler: A small suggestion: could we add a link to https://github.com/orgs/deis/people here so others can see who has write access to the Deis organization?

Joshua-Anderson: Note that this will only show people who have chosen to publicize their membership (which is hidden by default), and it won't show who is read-only or what repositories they have write access to. That said, it's probably better than not giving a list at all.

carmstrong: Two issues with this: as @Joshua-Anderson said, it only shows people who are public. It also shows people who don't have write access to the repos. Some are simply members of org who aren not part of the maintainers group.

mboersma: Since there isn't a clean way to link to the exact set of maintainers, I'm ok with leaving this as-is.

bacongobbler: +1

PR #4745, opened 2015-11-12
fix(registry): disallow dots in s3 buckets

helgi:
OrdinaryCallingFormat is important for any bucket with dot in it, no matter the region. Fixes bug introduced by #4729

### DIFF
--- a/docs/customizing_deis/registry_settings.rst
+++ b/docs/customizing_deis/registry_settings.rst
@@ -22,7 +22,7 @@ The following etcd keys are set by the registry component, typically in its /bin
 ===========================              =================================================================================
 setting                                  description
 ===========================              =================================================================================
-/deis/registry/bucketName                store component bucket used for registry image layers (default: registry)
+/deis/registry/bucketName                store component bucket used for registry image layers. Periods are not allowed in the name (default: registry)
 /deis/registry/host                      IP address of the host running registry
 /deis/registry/port                      port used by the registry service (default: 5000)
 /deis/registry/protocol                  protocol for registry (default: http)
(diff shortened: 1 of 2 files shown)


obmarg:
I get that not supporting periods in bucket names is bad, but will this change not just re-introduce #4713?

OrdinaryCallingFormat does not work well with `conn.lookup`:

    In [22]: conn = boto.connect_s3(is_secure=True)
    
    In [23]: conn.get_bucket('rp-deis')
    Out[23]: <Bucket: rp-deis>

    In [24]: conn = boto.connect_s3(is_secure=True, calling_format=OrdinaryCallingFormat())

    In [25]: conn.get_bucket('rp-deis')

In this case `rp-deis` is an S3 bucket created in the AWS admin.  Interestingly, buckets created by `boto.connect_s3` seem to have no such problem:

    In [28]: conn.create_bucket('test-85756356')
    Out[28]: <Bucket: test-85756356>

    In [30]: conn.get_bucket('test-85756356')
    Out[30]: <Bucket: test-85756356>

I guess for those of us with manually created s3 buckets, this will cause `lookup` to fail again, and then `create_bucket` to fall over.

obmarg:
The get_bucket failure happens with s3 repositories with names containing periods as well:
   
    In [35]: conn.get_bucket('test.with.dots')
    S3ResponseError: S3ResponseError: 301 Moved Permanently

helgi:
That's very special - Wonder if boto has any issues open about the lack of 301 following

helgi:
https://github.com/boto/boto/issues/2836 and https://github.com/boto/boto/issues/443 touch on this problem and the ongoing problem people are having with this

carmstrong:
I'd *love* to solve this in boto. That would require shipping a fix and then updating wal-e to use an updated boto.

helgi:
Given the ongoing discussions and attempts by the community to fix boto then am not holding out a whole lot of hope to have this fixed in any reasonable timeframe :(

We may be best off not using `OrdinaryCallingFormat` for non-storage connections *and* disallow periods in bucket names via our own scripts.

carmstrong:
> We may be best off not using OrdinaryCallingFormat for non-storage connections and disallow periods in bucket names via our own scripts.

:+1: although I imagine periods in bucket names is a rare occurrence.

helgi:
Yeah, thinking more as a guard - Better to throw an exception or similar with a good error message.

Wondering if we even need that fix for `database` given it only talks to S3 if the user specifically points to the intended S3 hostname, which (it seems) gets past this overall issue if I am reading all the issues correctly.

PR #4757, opened 2015-11-18
[WIP]: feat(platform): e2e https for v1.x + k8s scheduler

### LINKED ISSUE #4611: Encrypt the router<->container link
Deis should support encrypting the router<->container hop.

My company's services must be HIPAA compliant. That means that we can't have any unencrypted personal health information (PHI) ever hitting the wire. Currently we're using Deis for internal tools with a strict no-PHI policy, but we'd like to open up usage to teams that do handle PHI.

Unfortunately, the mullet VIP (TLS in the front, party in the back) that Deis uses by default makes this impossible.

Given the current architecture of Deis, I think the only two options are:

1. Terminate TLS in the router, but make a new TLS connection to a container, which requires:
    1. Add support for setting `https` for `proxy_pass`'s protocol, depending on an app's config
    2. Allow configuring a self-signed cert for the routers to trust OR let Deis generate and inject private keys into containers, which Deis is pre-configured to trust
2. Only Terminate TLS in the container, which requires:
    1. Add support for setting `https` for `proxy_pass`'s protocol, depending on an app's config
    2. Configure routing affinity to allow TLS sessions to be set up and used correctly

Are there any plans for encrypting the final hop in the works?

Note that in #deis @Joshua-Anderson suggested "It could be interesting v2 feature. The k8s proxy could rencrypt traffic and then expose the keys to app containers."

krancour:
Fixes #4611 and #4629, but _only_ with k8s as the scheduler.

This is currently lacking documentation, but I wanted to start getting this going with CI-- hence the WIP.

If anyone wants to try this, enabling requires two things:

1. You must have provided a wildcard cert for the platform.  See [here](http://docs.deis.io/en/latest/managing_deis/ssl-endpoints/#installing-ssl-on-the-deis-routers).  This cert is re-used for the link between the routers and the application pods _if..._

2. You must opt into this on an app-by-app basis.  Because extensive modifications to the router are not in scope now (but may be for v2), the only reasonable way to accomplish this is by keeping all routing at layer 7 using SNI.  This requires terminating SSL at the router and re-initiating it for the upstream connection to the application.  This imposes a minor performance overhead that not all operators will want-- thus this remains opt in.  Opt in like so:

```
[/my/app/directory]$ deis config:set E2E_HTTPS=true
```


krancour:
Ping @mboersma ... note that this introduces a new "component."  It's a Docker image used for SSL-terminating "sidecar" containers in each pod (if e2e HTTPS is enabled, that is).  It's not anything that needs to be scheduled or anything-- controller just adds the sidecar where applicable.

I'm mentioning this in case this might have any bearing on the release process, although I _think_ I've got that covered in this PR... I'm not positive.

krancour:
Closing this.  This won't land in V1.x.  The exercise will inform the e2e implementation in V2.

Issue #4746, opened 2015-11-12
Need to block client request for change configs of app twice


rvadim:
```sh
$ deis info
=== forecast Processes
--- web:
web.1 up (v15)

$ deis config
=== forecast Config
$ deis config:set BLABLA=1
Creating config... o..^C

$ deis info
=== forecast Processes
--- web:
web.1 up (v15)
web.1 created (v16)

$ deis config:set BLABLA=2
Creating config... .o.^C
$ deis config:set BLABLA=3
Creating config... .o.^C
$ deis info
=== forecast Processes
--- web:
web.1 up (v15)
web.1 down (v16)
web.1 created (v17)
web.1 created (v17)
web.1 created (v18)
web.1 created (v18)
web.1 created (v18)
web.1 created (v18)

$ deis info
=== forecast Processes
--- web:
web.1 up (v18)
web.1 up (v18)
web.1 up (v18)
web.1 up (v18)
```
In simple lightweight app this operation just spawn many process with one instance and one version. With hardcore big java app (big image) it makes registry slow down and make starting application 1 - 3 hours. Also no one can deploy or start any new containers due this configuring will be resolved.

We really need to create lock on any operations in controller with containers.

rvadim:
Can be fixed by: #4758

PR #4758, opened 2015-11-18
fix(controller): don't allow user to call deis config multipl times

rvadim:
Create item in etcd, which will block ability to call config more then one time until configuration will be applied.
Fixing: #4746


controller/api/locking.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+class AlreadyLocked(Exception):
+    pass
+
+
+def lock(name, ttl=3600):
+    key = LOCK_PATH_TEMPLATE.format(name=name)
+    value = 'locked'
+    client = get_etcd_client()
```
bacongobbler: I'm not sure that we need to use etcd as our locking mechanism. Why not just use a [Lock](https://docs.python.org/3/library/asyncio-sync.html#asyncio.Lock) instead?

arschles: presumably there will be many controller api servers running at once, so has to be a distributed lock

bacongobbler: Currently we don't support multiple api servers running at the same time. etcd has burned us in the past with dropped connections, etcd timeout failures etc. so I would prefer a simpler locking mechanism.

arschles: oh, in that case definite +1 on the in-memory lock

rvadim: There is no asyncio in python2.
```
 nse deis-controller
bash-4.3# ps
PID   USER     TIME   COMMAND
    1 root       2:34 bash /app/bin/boot
  120 deis       2:17 {gunicorn} /usr/bin/python /usr/bin/gunicorn -c deis/gconf.py deis.wsgi
  133 root      13:54 confd -node 10.21.12.124:4001 --confdir /app --log-level error --interval 5
 5050 deis       2:29 {gunicorn} /usr/bin/python /usr/bin/gunicorn -c deis/gconf.py deis.wsgi
 5051 deis       2:39 {gunicorn} /usr/bin/python /usr/bin/gunicorn -c deis/gconf.py deis.wsgi
 5052 deis       2:19 {gunicorn} /usr/bin/python /usr/bin/gunicorn -c deis/gconf.py deis.wsgi
 5053 deis       2:41 {gunicorn} /usr/bin/python /usr/bin/gunicorn -c deis/gconf.py deis.wsgi
 5054 deis       2:33 {gunicorn} /usr/bin/python /usr/bin/gunicorn -c deis/gconf.py deis.wsgi
 5055 deis       2:29 {gunicorn} /usr/bin/python /usr/bin/gunicorn -c deis/gconf.py deis.wsgi
 5056 deis       2:43 {gunicorn} /usr/bin/python /usr/bin/gunicorn -c deis/gconf.py deis.wsgi
 5057 deis       2:30 {gunicorn} /usr/bin/python /usr/bin/gunicorn -c deis/gconf.py deis.wsgi
 5058 deis       2:48 {gunicorn} /usr/bin/python /usr/bin/gunicorn -c deis/gconf.py deis.wsgi
24977 root       0:00 sleep 10
24978 root       0:00 bash
24983 root       0:00 ps
bash-4.3# /usr/bin/python --version
Python 2.7.10
```
Im spending some time to find any in memory locking mechanism for python 2.7 and get nothing, if you find any, i will rewrite this to in memory lock.
Also i try to make locking through fcntl fs lock, but there are some problems with this realization due to in one thread sync and multi thread sync.

krancour: > Currently we don't support multiple api servers running at the same time.

But we would presumably like to change that in v2.  I think it's worth finding the right way to do a distributed lock.

arschles: then we're back to etcd or postgres (I'm positing that we don't want to add extra infrastructure here). I'm happy to prototype something if we decide to go do this. maybe something for post-beta

mboersma: We use a python `flock` elsewhere for a similar purpose. When controller runs, gunicorn spawns multiple Python interpreters, so no memory-based locking solution will work.

krancour:
When complete, this feature should be ported forward to deis/workflow as well.

controller/api/locking.py
bacongobbler: I'd be more comfortable with a file lock, in that case. Less chance of there being outages/instability with this code.

mboersma:
> this feature should be ported forward to deis/workflow...

See deis/controller#594.

PR #4763, opened 2015-11-23
Enable placement options should not effect legacy tag placement.

jwaldrip:
Legacy tag based placement now gets messed up unless we enable placement options on the platform. I don't believe this was supposed to be a breaking change.

### DIFF
--- a/controller/scheduler/fleet.py
+++ b/controller/scheduler/fleet.py
@@ -150,10 +150,13 @@ class FleetHTTPClient(AbstractSchedulerClient):
             f['value'] = f['value'].format(**l)
         # prepare tags only if one was provided
         tags = kwargs.get('tags', {})
-        tagset = ' '.join(['"{}={}"'.format(k, v) for k, v in tags.viewitems()])
+        unit_tags = tags.viewitems()
         if settings.ENABLE_PLACEMENT_OPTIONS in ['true', 'True', 'TRUE', '1']:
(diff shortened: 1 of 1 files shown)


deis-admin:

krancour:
Nice catch on the bug, however, the PR introduces a new problem...

The code in the PR could result in two `[X-Fleet]` sections in the generated unit file.

The right way to fix this would be:

1. Get the tags
2. If `settings.ENABLE_PLACEMENT_OPTIONS` is `True` then add `dataPlane=true` to the tags
3. Add `[X-Fleet]` section only if tags >= 1

jwaldrip:
@krancour should be all set now.

krancour:
Thanks!

carmstrong:
@jwaldrip Would you mind squashing your commits into one titled something like `fix(controller): fix legacy tags`?

krancour:
Good catch @carmstrong.  Had not noticed commit messages.  @jwaldrip, for reference, all our commits follow these conventions: http://docs.deis.io/en/latest/contributing/standards/#commit-style

jwaldrip:
@carmstrong should be good to go.

carmstrong:
Jenkins, add to whitelist.

carmstrong:
Once Jenkins passes, we should be good to go. Thanks @jwaldrip!

PR #4771, opened 2015-11-30
chore(deisctl): update stateless warning message

olalonde:
`deis logs` works using the in memory logger component

### DIFF
--- a/deisctl/cmd/platform.go
+++ b/deisctl/cmd/platform.go
@@ -19,9 +19,9 @@ func InstallPlatform(b backend.Backend, cb config.Backend, checkKeys func(config
 	}
 
 	if stateless {
-		fmt.Println("Warning: With a stateless control plane, `deis logs` will be unavailable.")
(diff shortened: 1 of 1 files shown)


krancour:
Nice catch.  I'm wondering if the line about components needing to be configured for use with external storage already describes the situation accurately.  Logs can be drained to the likes of Papertrail or Splunk for longer term archiving.  Especially considering that in-memory log storage isn't yet a default, this line is doubly true-- some configuration is required.

I think the best bet is to only remove the line about `desis logs` not working, and that would be sufficient without adding anything further.

olalonde:
Agreed. Is that better?

> Warning: With a stateless control plane, some components require manual configuration.
> See the official Deis documentation for details on running a stateless control plane.
> http://docs.deis.io/en/latest/managing_deis/running-deis-without-ceph/

krancour:
Looks great!  Thanks for this!

carmstrong:
Kicked off another test run. Once it passes, we should be good to merge. Thanks, @olalonde!

PR #4787, opened 2015-12-10
fix(controller): ensure keys are saved before return

lslaslo:
We noticed when using the DPL tool in Travis that keys were
not immediately available for use, and an artificial wait was
needed. This change should ensure that keys have been saved
and are available when keys:add is finished executing.


deis-admin:

carmstrong:
One downside of moving this block is that the keys will no longer automatically be published on a restart of the controller, where the controller walks through all records and re-saves them to ensure they exist in etcd - see #4394.

aj-may:
@carmstrong are you sure about that?  Since he has moved the logic into the save method of the model, it should still be called on re-save.

carmstrong:
> Since he has moved the logic into the save method of the model, it should still be called on re-save.

Ahh, maybe I misread the changeset.

mboersma:
Jenkins, add to whitelist.

bacongobbler:
we should probably do the same for all of the django signal calls if this is the precedence we want to set.

aj-may:
@bacongobbler is there any reason that we need be using post-save signals for any of the models? If not @lslaslo and myself can look into making that change.  Thoughts @carmstrong?

bacongobbler:
there was no technical reason AFAIK. We've been using post-save and post-delete signals since [0.1.0](https://github.com/deis/deis/blob/v0.1.0/api/models.py#L838-L840). Back then it made sense since those were tasks that was not tied to a specific model, but now the etcd key and the model saved in the database are practically the same.

mboersma:
Jenkins, test this please.

lslaslo:
@bacongobbler @carmstrong @aj-may , instead of these particular changes, do you think the django pre_save signal would work, in lieu of post_save? https://docs.djangoproject.com/en/1.9/ref/signals/#django.db.models.signals.post_save

bacongobbler:
a pre_save is used before the actual database transaction occurs. Since it's an operation that we cannot roll back if the `save()` call fails, I'd be worried about rollback failures. Basically here's my thinking on how I'd imagine a pre_save signal would fail:

 - create an app called `foo`
 - etcd is up, db is down
 - etcd key is created, but db is down so we don't save to the database
 - try again after db is up
 - etcd set fails because key already exists

I'm leaning more towards adding it directly into the `save()` operation, though it looks like we'll have to mock out calls to etcd in order for the tests to pass.

aj-may:
@bacongobbler @mboersma @lslaslo, even if we decide to augment the save method, unless we incorporate rollback logic into the method, we are in the same situation.  In addition, with the current method using post save signals, we have the exact opposite possibility.  We can end up writing to the db, but not writing to etcd.  Would it be appropriate to simply switch to pre-save signals to satisfy this issue, but create another issue to implement a rollback upon failure?

bacongobbler:
> In addition, with the current method using post save signals, we have the exact opposite possibility. We can end up writing to the db, but not writing to etcd.

At the very least, if that situation occurs then we can still technically re-run `deis create` and get the desired state we want in the database. For the most part etcd is the afterthought when it comes to api-related stuff, since we're just using it mostly for service discovery purposes.

I feel like if this is unsolvable without some tradebacks then we should keep it the way things are and just document that etcd-related updates takes x seconds to be updated. If you're looking to figure out if keys have been added to a user, the best single point of truth would be the database. Ultimately we should've made builder just read from the database instead of writing to postgres *as well* as etcd, but what can ya do.

bacongobbler:
Closing this bug for v1. In v2 this issue is resolved because we are intending to completely remove etcd, so this sort of problem won't occur. For v1, the current implementation is probably the best solution in terms of stability. Unfortunately that means etcd is not always up to date, but if you want up-to-date changes on the state of the controller, it's recommended to connect to the database instead as that's guaranteed to be updated when a record is modified, created or destroyed. Closing!

Issue #4795, opened 2015-12-12
mock-store error: mock_s3/main.py: patch does not apply


carmstrong:
```
(33/33) Installing python-dev (2.7.10-r1)
Executing busybox-1.23.2-r0.trigger
OK: 227 MiB in 62 packages
Cloning into '/app/mock-s3'...
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100   424  100   424    0     0   1725      0 --:--:-- --:--:-- --:--:--  1744
error: patch failed: mock_s3/main.py:65
error: mock_s3/main.py: patch does not apply
The command '/bin/sh -c DOCKER_BUILD=true /tmp/build.sh' returned a non-zero code: 1
make[2]: *** [mock-store] Error 1
make[2]: Leaving directory `/home/jenkins/workspace/test-ec2/src/github.com/deis/deis/tests'
make[1]: *** [test-functional] Error 2
make[1]: Leaving directory `/home/jenkins/workspace/test-ec2/src/github.com/deis/deis/database'
make: *** [test-functional] Error 2
```

This can be seen here: https://ci.deis.io/job/test-ec2/3719/consoleFull#682192135502066c60b2236c09cc9bf8

carmstrong:
Closed by #4800.

PR #4800, opened 2015-12-15
fix(tests): use and patch known "good" version of mock-s3

krancour:
This removes the download and application of a gisted diff during building of the mock-store test fixture.  Downloading this file has long resulted in intermittent 503s, and as of recently, have broken the build consistently because the diff no longer applies.

Even if this passes CI, we should still circle back to verify it does not affect #3128

### DIFF
--- a/tests/fixtures/mock-store/Dockerfile
+++ b/tests/fixtures/mock-store/Dockerfile
@@ -8,5 +8,6 @@ EXPOSE 8888
 CMD ["/app/bin/boot"]
 ADD bin/boot /app/bin/boot
 
+ADD mock-s3-patch.diff /tmp/mock-s3-patch.diff
 ADD build.sh /tmp/build.sh
 RUN DOCKER_BUILD=true /tmp/build.sh
--- a/tests/fixtures/mock-store/build.sh
+++ b/tests/fixtures/mock-store/build.sh
@@ -23,10 +23,11 @@ apk add --update-cache \
 curl -sSL -o /usr/local/bin/etcdctl https://s3-us-west-2.amazonaws.com/get-deis/etcdctl-v0.4.9 \
   && chmod +x /usr/local/bin/etcdctl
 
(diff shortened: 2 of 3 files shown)


krancour:
Worth noting... because this clones https://github.com/jserver/mock-s3 without specifying a tag or sha, the builds are non-deterministic and subject to failure anytime that library is updated.  Idk if this is worth fixing right now.  @carmstrong ?

krancour:
fwiw, this is what is holding up #4732

carmstrong:
Instead of removing this patch, an arguably safer fix would be to pin the clone to a known-working SHA of mock-s3. That would prevent us from reverting logic that potentially fixed something with wal-e, and future-proof Deis 1.x.

krancour:
I thought about that.  If you're saying the same thing, it wasn't that bad an idea.  I'll do that.

carmstrong:
> it wasn't that bad an idea

:+1: Not ideal, but seemingly the safest.

krancour:
@carmstrong altho there is still the problem of intermitent 503s when fetching the gisted diff.  I could include that locally, but that that point, what's the point?  git is designed to manage such shenanigans for us.  Maybe I should fork jserver/mock-s3 as deis/mock-s3 and apply the patch there?

carmstrong:
> Maybe I should fork jserver/mock-s3 as deis/mock-s3 and apply the patch there?

Can we add the patch into the component itself and apply locally at build-time? I'd rather not fork something we have no intention of maintaining long-term.

krancour:
Yes.  That was an option i was alluding to.  I don't think I was clear.  I just feel like it's odd... git is _for_ managing diffs elegantly.  It feels unnatural to check a diff _into_ a git repo.  But I do see your point.

krancour:
@carmstrong fixed in the manner you suggested.

carmstrong:
Closes #4795

Issue #4828, opened 2016-01-04
(router) vhost_traffic_status zone running out of memory


daanemanz:
During a period of heavy load on our platform, the routers started spitting out error messages in the journal:
```
Jan 04 05:57:05 <redacted> sh[8859]: [INFO] - 2016/01/04 05:57:05 [crit] 1054#0: ngx_slab_alloc() failed: no memory in vhost_traffic_status_zone "vhost_traffic_status"
Jan 04 05:57:05 <redacted> sh[8859]: [INFO] - 2016/01/04 05:57:05 [error] 1054#0: *50207015 shm_add_upstream failed while logging request, [...]
```

One should be able to increase the size of the `vhost_traffic_status` zone's shared memory according to the [VTS module's documentation](https://github.com/vozlt/nginx-module-vts#vhost_traffic_status_zone) as it is set to only `1m` by default.

PR #4829, opened 2016-01-04
feat(router) make vhost_traffic_status_zone configurable

daanemanz:
This change makes the size in the `vhost_traffic_status_zone` directive
configurable. See also https://github.com/deis/deis/issues/4828.


krancour:
fwiw, I'm not planning to port this forward to v2 since v2 deliberately has dropped the status page due to security concerns.  As a result, this module isn't used in v2.

PR #4845, opened 2016-01-08
feat(contrib/linode): update firewall to allow for cluster expansion.

thealmightygrant:
Adds a parameter to the Linode firewall setup script in order to open ports 2379,2380 when adding nodes to an existing cluster so that etcd can be properly initialized. Without this (after the firewall has been setup on the CoreOS cluster), there is no way to easily add nodes to an existing cluster.


thealmightygrant:
After looking at this a bit, I see that leaving these ports open is necessary for adding members, but when I add members, I have to leave them open (even after CoreOS has been fully installed). I think this has something to do with the additional members being added as proxies in etcd. I'll close this PR and resubmit once I have some more concrete information.

PR #4849, opened 2016-01-12
feat(contrib/linode): allow for cluster expansion and standardize scripts

thealmightygrant:
1. I added the option to open ports 2379 and 2380 during cluster expansion. This is useful due to [the new nodes being added as proxies within etcd when using a public discovery URL that has reached it's size limit](https://coreos.com/etcd/docs/latest/clustering.html#public-etcd-discovery-service).
2. I standardized and normalized the python functions used during CoreOS provisioning (i.e. Linode API calls and logging).
3. I added multiple methods of providing IPs when applying a firewall to the cluster. Namely, IPs can now be provided via a Linode display-group, manually, or using the discovery URL. This is useful particularly when running this command outside of the CoreOS cluster, but also when adding nodes to an existing cluster because (as noted above) the new nodes will be added as proxies within etcd and will not be included in the information provided by the public discovery URL.
4. Updated the docs accordingly. 

None of these are breaking changes. Please let me know if you need any changes or have any questions, and I will be happy to oblige.

### DIFF
--- a/contrib/linode/apply-firewall.py
+++ b/contrib/linode/apply-firewall.py
@@ -56,7 +56,7 @@ def validate_ip_address(ip):
     return True if re.match('([0-9]{1,3}\.){3}[0-9]{1,3}', ip) else False
 
 
-def get_firewall_contents(node_ips, private=False):
+def get_firewall_contents(node_ips, private=False, adding_new_nodes=False):
     rules_template_text = """*filter
 :INPUT DROP [0:0]
 :FORWARD DROP [0:0]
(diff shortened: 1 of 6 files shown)


bacongobbler:
ping @croemmich. Any opinions on this change? :)

croemmich:
@bacongobbler I haven't had a chance to test it, but it looks sane. I especially like the display-group firewall provisioning. Excited to have another LInodian contributing!

carmstrong:
Awesome stuff, @thealmightygrant!

PR #4851, opened 2016-01-13
fix(controller): legacy fix should modify dict instead of dict view

thealmightygrant:
When trying to modify the unit_tags, the following error occurs:
`TypeError: dict_items object does not support item assignment`.
This is because it is a view and not a dict. You have to modify
the dict directly in order for this to work.

### DIFF
--- a/controller/scheduler/fleet.py
+++ b/controller/scheduler/fleet.py
@@ -152,7 +152,7 @@ class FleetHTTPClient(AbstractSchedulerClient):
         tags = kwargs.get('tags', {})
         unit_tags = tags.viewitems()
         if settings.ENABLE_PLACEMENT_OPTIONS in ['true', 'True', 'TRUE', '1']:
-            unit_tags['dataPlane'] = 'true'
+            tags['dataPlane'] = 'true'
         if unit_tags:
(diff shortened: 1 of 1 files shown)


controller/scheduler/fleet.py
```diff
@@ -152,7 +152,7 @@ def _create_container(self, name, image, command, unit, **kwargs):
         tags = kwargs.get('tags', {})
         unit_tags = tags.viewitems()
```
bacongobbler: we should also remove this if it's unused code

thealmightygrant: Actually, it's used. When the `tags` variable is updated the "view" `unit_tags` is also updated, and that is used at https://github.com/deis/deis/pull/4851/files#diff-43358db3bfa05551b90bcf90c48dfe2cR157

bacongobbler:
@thealmightygrant would you mind rebasing off master? The test failure was unrelated, but is fixed up in master so your PR should go green now. Thanks!

thealmightygrant:
Should be good to go now.

bacongobbler:
I just hit this on master, so I can confirm this is a bug (sorry for the delay btw, we're all working on v2 at the moment) :)

bacongobbler:
thanks @thealmightygrant!

PR #4859, opened 2016-01-20
chore(*): bump CoreOS to 835.10.0

carmstrong:
https://coreos.com/releases/#835.10.0


bacongobbler:
looks like a flake8 issue. ping @helgi

helgi:
Oh yeah, it's not the only failure but I did see the same thing in deis/workflow - solution is to move the `and` statements to the previous line but nothing else

mboersma:
The linting error is fixed in #4861, so after that is merged we can rebase this and it should be good.

carmstrong:
CoreOS 835.11.0 is out. Will open a new PR once #4861 is merged to fix the style errors.

PR #4861, opened 2016-01-22
ref(controller): remove scheduler previews

### LINKED ISSUE #4860: remove scheduler tech previews for LTS release
The experiment is finally over, and we are migrating Deis on top of kubernetes for Deis v2! We have had some users now still trying the tech previews in v1, which was never meant to be used in production nor is in any working state. For the LTS release, we should remove these tech previews so users won't be shooting themselves in the foot.

bacongobbler:
closes #4860


carmstrong:
I'd like @helgi to have eyes on this too

mboersma:
This also effectively fixes a `flake8` linting error that somehow appeared in master and has been failing other PR builds. :+1:

carmstrong:
We'll also need to remove all the documentation around choosing a scheduler - specifically, http://docs.deis.io/en/latest/customizing_deis/choosing-a-scheduler/

bacongobbler:
done!

carmstrong:
Looks like we also mention `schedulerModule` in http://docs.deis.io/en/latest/customizing_deis/controller_settings/, so that'll have to be removed too.

bacongobbler:
thank you for catching my doc blunders. I've removed that line now.

mboersma:
There is also `deisctl` support tied to installing and removing the scheduler preview units, the fleet units files themselves (e.g. `deisctl/units/deis-swarm-node.service`), and the `mesos/` and `swarm/` directories. I think those should be removed as well for the next release, unless I am missing some backward-compatibility use case. (And if so, that could be a separate PR IMHO.)

bacongobbler:
600 more LOC removed this round. thanks @mboersma

bacongobbler:
correction: 3,500 lines after removing the `mesos` and `swarm` components as well.

bacongobbler:
pinging @helgi and @mboersma for another look-see. Also @mboersma it looks like you LGTM'd this PR twice so I'll remove one. :)

PR #3096, opened 2015-02-15
WebSockets on EC2

johanneswuerbach:
This PR is a proof-of-concept providing websocket support for deis on ec2 without further configuration.

To enable websockets and remote IPs, I used the proxy protocol as already proposed by @Blystad. The major difference is, that I forward 80 and 443 directly via TCP and do not terminate SSL at the ELB. SSL termination can now done by deis itself at the router (like on GCE or DO).

This approach has multiple advantages:
* apps still know the forward protocol
* spdy support for EC2
* paves the way for multi TLD SSL on EC2
* Simplifies deis as EC2 routing now works more like on GCE or DO

Alternative for https://github.com/deis/deis/pull/3095 #2717
When this concept is accepted, I'll merge the work from @Blystad into this PR and ensure that every commit is attributed correctly.

I tested this manually and this sockets work now and my IP is displayed correctly:
```
git clone https://github.com/heroku-examples/ruby-ws-test.git
cd ruby-ws-test
deis create
git push deis master
deis open
```

PR #4863, opened 2016-01-22
ref(contrib): do not enable proxyProtocol on AWS

### LINKED ISSUE #4400: Force SSL behind AWS ELB results in a redirect loop.
`deisctl config router set enforceHTTPS=true`

Results in redirect loop on Cephless 1.9 install.

bacongobbler:
closes #4400


carmstrong:
ping @blystad and @johanneswuerbach - proxy protocol support was first attempted in #2717 and then implemented in #3096. I want to ensure we're fully aware of the implications of disabling it.

Blystad:
Ping acknowledged (thanks!).

I've briefly read into the issue...We redirect port 80 -> 80 and 443 -> 443...couldn't we simply get if it's HTTP or HTTPS from the port?

Something like:
```
if incoming port = 443
   type = HTTPS
else
   type = HTTP
```

johanneswuerbach:
Thanks for the ping :-) Disabling proxy protocol will basically limit you to the features exposed by elb, no websocket, no elb, no custom ssl. Did somebody use that, I can't tell.

Blystad:
We used websockets, which was my motivation for contributing this. It would be great to get rid of it...but AWS seems to want to add service #1001 rather than improve their existing services :(

bacongobbler:
To be clear the motivation is to keep this a part of the router, but disable it by default. There have been issues with SSL redirection with proxy protocol enabled because certain headers typically populated by the ELB like `http_x_forwarded_proto` were not being sent, which is what SSL redirection relies on.

If you still want websocket support you'll just have to run `deisctl config router set proxy protocol=1` to enable it and everything is all honkey-dory. If you're upgrading from an older cluster this setting was already enabled so the upgrade path is not broken with this change either. This just turns it off by default so users who are aware of this feature can enable it, but it won't hurt others working with other nginx modules. Makes sense? :)

bacongobbler:
To elaborate, the TCP listener in the ELB does not send those HTTP headers other modules were expecting, not proxy protocol itself. By changing the listener type to TCP we lost those headers, but nobody really noticed until 6 releases later :)

Blystad:
Right, but can't we assign the headers based on the incoming port?

It's not just running deisctl blablabla, you also have to run various AWS
commands to enable proxy protocol on the ELB.

This PR feels like it tries to solve the symptom rather than the problem.

krancour:
I'd like to experiment with this a bit before we consider merging.

bacongobbler:
I'm doing exactly that. I want to check what headers are available and what are not when proxy protocol is enabled/disabled.

krancour:
@bacongobbler per comments in #4400 (which I have now closed), I don't think this solves anything.  I would move that we close this PR as well.

PR #4864, opened 2016-01-22
feat(includes.mk): add Go 1.5 vendor support

### LINKED ISSUE #4476: Go 1.5 Vendoring
We need to switch Deis to use GO15VENDOREXPERIMENT while we are in the development cycle. The Go developers have already committed to carrying this on into Go 1.6 (where it will be the default).

Because of this, we should also modify Makefiles to stop using godep.

bacongobbler:
closes #4476


bacongobbler:
d'oh:

```
package github.com/deis/deis/builder/tests (test)
	imports github.com/deis/deis/tests/dockercli
	imports github.com/docker/docker/api/client
	imports github.com/docker/docker/api
	imports github.com/docker/docker/engine
	imports github.com/docker/docker/utils
	imports github.com/docker/docker/pkg/archive
	imports github.com/docker/docker/vendor/src/code.google.com/p/go/src/pkg/archive/tar: must be imported as src/code.google.com/p/go/src/pkg/archive/tar
```

Sounds like http://google-opensource.blogspot.ca/2015/03/farewell-to-google-code.html finally happened. I recall reading something on Hacker News this morning about this.

arschles:
@bacongobbler yup, it happened :(

krancour:
See my comment on #4874 re: switching to glide.  imo, we need to do that for LTS.  If not, then we wouldn't need this change either.

bacongobbler:
closing this for the reasons stated above as well as some failures, as I cannot seem to migrate Godep over to the Go 1.5 vendor experiment due to #4878.

PR #4866, opened 2016-01-22
docs(managing_deis): add workaround for cephless cluster

### LINKED ISSUE #4568: Deis Registry miss /deis/store/gateway key while use S3 as storage
I've got the error: "confd[641]: ERROR 100: Key not found (/deis/store) [366456]"

One option is to remove   "/deis/store/gateway" from registry/conf.d/config.toml
and from registry/conf.d/create_bucket.toml

Or a line like: store_gateway: {{if exists "/deis/store/gateway"}}{{ getv "/deis/store/gateway" }}{{ else }}""{{ end }} to config.yml

bacongobbler:
The registry will fail to template the config if /deis/store/gateway
is not set. This workaround works for users running Deis v1 without
Ceph.

closes #4568

### DIFF
--- a/docs/managing_deis/running-deis-without-ceph.rst
+++ b/docs/managing_deis/running-deis-without-ceph.rst
@@ -41,7 +41,7 @@ ships these to the :ref:`logger` component.  By default, the logger writes the
 logs to a distributed Ceph filesystem. These logs can then be fetched by the
 :ref:`controller` component via HTTP.
 
-In a Ceph-less clutser, the Logger component should be configured, instead, to
+In a Ceph-less cluster, the Logger component should be configured, instead, to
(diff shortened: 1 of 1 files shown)


carmstrong:
ping @krancour - I'd like his approval here

krancour:
This is a blast from the past.  When I was running cephless clusters, I used to provide a whole bunch of `/deis/store/gateway/...` configuration even though Ceph wasn't being used.  That always annoyed me.  It's interesting to know now that I could have circumvented all that with this little workaround.

This LGTM, but did either of the other reviewers actually test-drive this?  If not, I'll be happy to.

carmstrong:
> I'll be happy to.

Please do! :+1:

krancour:
Ok.  Got hands on with the documented workaround.  LGTM3.  Merge away.

PR #4876, opened 2016-01-26
fix(logspout): Truncate lines too big for a single UDP packet

### LINKED ISSUE #2046: logspout message too long error
After running the deis-logspout component for 2 days I see this error in the journal:
```console
Oct 06 11:56:25 nodo-1 sh[17045]: 2014/10/06 14:56:25 syslog: write udp 192.168.2.201:514: message too long
```
(occurred 2 times in those 2 days)

krancour:
Fixes #2046

### DIFF
--- a/logspout/logspout.go
+++ b/logspout/logspout.go
@@ -20,6 +20,11 @@ import (
 	"golang.org/x/net/websocket"
 )
 
+const (
+	MAX_UDP_MSG_BYTES = 65507
+	MAX_TCP_MSG_BYTES = 1048576
+)
+
 var debugMode bool
 
 func debug(v ...interface{}) {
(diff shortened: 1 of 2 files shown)


bacongobbler:
IRT "requires-manual-testing", would it be possible to just write an integration test for this? Essentially this is how we tested it before: https://github.com/deis/deis/blob/master/tests/apps_test.go#L124-L125

We could just change 1024 to be MAX_UDP_MSG_BYTES + 1.

krancour:
@bacongobbler we could, except that doing so won't actually help us assert that long log lines don't crash logspout anymore.

I suppose what we could do is do what you have suggested _plus_ an immediate `deis logs`.  We can look for the truncated output-- which would be absent if logspout _had_ crashed.

bacongobbler:
> I suppose what we could do is do what you have suggested plus an immediate `deis logs`

Sorry! I assumed that code did so already, so I thought that was implied. I was just doing a flyby review :) Checking for truncated output as well as perhaps a `deisctl list` to ensure the logger is still up would probably be a safe fix.

krancour:
@bacongobbler, I updated the test.  See the comment in it to understand the strategy.

bacongobbler:
close enough. +1

krancour:
This test seems to timeout now.  Working on fixing that.

krancour:
fwiw, tests timed out because the fleet unit produced in response to `deis apps:run echo <heinously long string>` has an `ExecStart=` that exceeds a limit of 2048 characters.  This lead to parsing errors and the task would run and re-run infinitely while the client hung waiting.

I've countered this with a somewhat more elaborate test strategy which is full documented in the comments accompanying the test code.

krancour:
@bacongobbler I am rescinding your initial LGTM1 and requesting a re-review because of the revised testing strategy.

logspout/logspout.go
```diff
@@ -65,32 +70,42 @@ func syslogStreamer(target Target, types []string, logstream chan *Log) {
...
-			conn = tcpconn
 		} else if strings.EqualFold(target.Protocol, "udp") {
+			// Truncate the message if it's too long to fit in a single UDP packet.
+			// Get the bytes first.  If the string has non-UTF8 chars, the number of
+			// bytes might exceed the number of characters and it would be good to
+			// know that up front.
+			dataBytes := []byte(data)
+			if len(dataBytes) > MAX_UDP_MSG_BYTES {
+				// Truncate the bytes and add ellipses.
+				dataBytes = append(dataBytes[:MAX_UDP_MSG_BYTES-3], "..."...)
```
mboersma:
```golang
"..."...
```
I can't decide if this part is clever or evil. :imp:

arschles: I don't think you need the second `...`. Here's an example: http://play.golang.org/p/98lLjaz5Vf

logspout/logspout.go
```diff
@@ -65,32 +70,42 @@ func syslogStreamer(target Target, types []string, logstream chan *Log) {
...
+			pid,
+			data)
+
 		if strings.EqualFold(target.Protocol, "tcp") {
 			addr, err := net.ResolveTCPAddr("tcp", target.Addr)
 			assert(err, "syslog")
-			tcpconn, err := net.DialTCP("tcp", nil, addr)
+			conn, err := net.DialTCP("tcp", nil, addr)
+			assert(err, "syslog")
+			assert(conn.SetWriteBuffer(MAX_TCP_MSG_BYTES), "syslog")
```
arschles: assuming it's ok to leave these `asserts` for LTS?

arschles:
@krancour left one question and one nitpick comment. Otherwise LGTM

logspout/logspout.go
krancour: @arschles it differs from your example a bit.  In your example, you're appending strings to a slice of strings.  Here, I am appending a string to a slice of bytes.  There's an implicit conversion that happens.

@mboersma as to whether it is clever or evil-- it appears unusual, for sure, but this particular use is documented in the go docs:

https://golang.org/pkg/builtin/#append

logspout/logspout.go
mboersma: [`assert`](https://golang.org/doc/faq#assertions) does seem like an anti-pattern in Go, but the code in question was already using it heavily so this doesn't really change existing error handling.

logspout/logspout.go
mboersma: It's just the unfortunate coincidence of the actual string to be appended matching the operator, so it's initially confusing to read. But it makes sense. :+1:

logspout/logspout.go
krancour: In this case, "assert" is actually an error handing function defined here:

https://github.com/krancour/deis/blob/a0da307f4fec1492b5883e823bd01b14d4d6a232/logspout/logspout.go#L36

(I didn't write that.)

logspout/logspout.go
krancour: > it's initially confusing to read

No argument there.  I did feel weird writing it.

PR #4877, opened 2016-01-26
fix(router): Fix issues establishing real end-client IP

### LINKED ISSUE #4869: Deis Whitelist breaks under specific configuration
In reference to #4400 (comment)

* enforceHTTPS=true
* enforceWhitelist=true
* proxyProtocol=false
* SSL termination at AWS ELB
* DEIS_WHITELIST fails to block connections

krancour:
Fixes #4869 

When PROXY protocol is not in use, this change will attempt to get the end client's IP form the `X-Forwarded-For` HTTP header.  If that header doesn't exist, perhaps because there is no load balancer or proxy in play, as in the case of Vagrant, for instance, there is no harm.

I have confirmed that this fixes #4869 and allows the use of whitelists _without_ PROXY protocol.  This is extremely useful to anyone terminating SSL at the load balancer.

This change should be ported forward to the Deis v2 router.

__EDIT:__ Also, this PR fixes some additional bugs and redundancies relating to PROXY protocol and/or the `X-Forwarded-For` header and the proper determination and use of the real end-client IP.

For instance, in the file `deis.conf`, the strategy used for determining whether PROXY protocol is to be used or not was inconsistent with the strategy used for the same purpose in `nginx.conf`.  As a result, `/deis/router/proxyProtocol=false` in etcd actually left PROXY protocol _enabled_ in `deis.conf`.

Another for instance is that there was a _lot_ of conditional logic that was completely redundant as long as the real end-client IP has been established properly.  The `real_ip` module overwrites the Nginx variable `$remote_addr` (whether from PROXY protocol or the `X-Forwarded-For` HTTP header).  At no point _after_ that should we care what the source of the `$remote_addr` value was.  It can simply be used.  This applies both to logs and to the setting of the `X-Forwarded-For` HTTP header for requests to the upstream backend.

### DIFF
--- a/router/rootfs/etc/confd/templates/deis.conf
+++ b/router/rootfs/etc/confd/templates/deis.conf
@@ -1,9 +1,10 @@
+{{ $useProxyProtocol := or (getv "/deis/router/proxyProtocol") "false" }}
 server_name_in_redirect off;
 port_in_redirect off;
-listen 80{{ if exists "/deis/router/proxyProtocol" }} proxy_protocol{{ end }};
+listen 80{{ if ne $useProxyProtocol "false" }} proxy_protocol{{ end }};
 
 {{ if exists "/deis/router/sslCert" }}
-listen 443 ssl http2{{ if exists "/deis/router/proxyProtocol" }} proxy_protocol{{ end }};
+listen 443 ssl http2{{ if ne $useProxyProtocol "false" }} proxy_protocol{{ end }};
 ssl_certificate /etc/ssl/deis.cert;
 ssl_certificate_key /etc/ssl/deis.key;
 include ssl.conf;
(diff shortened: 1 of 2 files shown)


carmstrong:
Looks good, but I have not manually tested this.

krancour:
@bacongobbler and @carmstrong I'm rescinding your LGTMs because I've amended this PR.

I noticed that the conditional logic around deciding on a log format is not at all necessary when real_ip is configured correctly.  I fixed that.  I have tested this in AWS with and without PROXY proto and I get the real end-client IP in the logs in both cases.

krancour:
Don't re-review yet.  There are actually some more redundant things in here.

krancour:
@carmstrong @bacongobbler this can be re-reviewed now.

bacongobbler:
can confirm that connections are blocked now. LGTM!

PR #4887, opened 2016-01-28
chore(*): remove k8s scheduler code

### LINKED ISSUE #4882: remove k8s-specific templating in deis/router
refs #4860 

See https://github.com/deis/deis/blob/master/router/rootfs/etc/confd/templates/nginx.conf#L377-L379 for an example.

bacongobbler:
This code is now considered dead for v1.13.0 LTS, so let's remove it.

ref #4860 
closes #4882

### DIFF
--- a/contrib/coreos/user-data.example
+++ b/contrib/coreos/user-data.example
@@ -277,31 +277,3 @@ write_files:
           sleep 1 && echo .;
         done;
       exit $?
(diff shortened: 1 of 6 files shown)


carmstrong:
I was skeptical that router's only use of registry was for k8s, but it does seem like that is the case. Code LGTM.

mboersma:
Less == more. Well done @bacongobbler.

PR #4923, opened 2016-02-18
fix(client): only delete local ~/.deis/client.json if cancelling logged in user

### LINKED ISSUE #4896: [client] `deis auth:cancel --username=someuser` removes ~/.deis/client.json
When calling `deis auth:cancel --username=someuser` to remove a user different than the one which is currently logged in, the deis client removes `~/.deis/client.json`. This is only the expected behaviour when   removing the currently logged in user (e.g. without the --username argument).

helgi:
Fixes #4896

### DIFF
--- a/client/cmd/auth.go
+++ b/client/cmd/auth.go
@@ -247,7 +247,7 @@ func Cancel(username string, password string, yes bool) error {
 	}
 
 	// If user targets themselves, logout.
-	if username != "" || c.Username == username {
+	if username == "" || c.Username == username {
 		if err := client.Delete(); err != nil {
 			return err
 		}


carmstrong:
Failure seems to be relevant:
```
auth:register http://deis.test-d838656bca.deis.works --username=cancel-test --password=test --email=test@test.co.nz
Registered cancel-test
Logged in as cancel-test
ok
auth:login http://deis.test-d838656bca.deis.works --username=test --password=asdf1234
Logged in as test
ok
auth:cancel --username=cancel-test --yes
Account cancelled
ok
auth:whoami
You are test at http://deis.test-d838656bca.deis.works
--- FAIL: TestAuth (11.66s)
	itutils.go:262: <nil>
FAIL
exit status 1
FAIL	github.com/deis/deis/tests	11.660s
godep: go exit status 1
make[1]: Leaving directory `/home/jenkins/workspace/test-ec2/src/github.com/deis/deis/tests'
```

bacongobbler:
indeed. I tried restarting the job a few times to verify but this does look like a legit failure.

carmstrong:
@Joshua-Anderson Any chance you have an idea as to what's happening here?

Joshua-Anderson:
[This line](https://github.com/deis/deis/blob/master/tests/auth_test.go#L53) was checking that the admin is logged out after deleting another user. Apparently this was the behavior of the old python client and the bug that is fixed in this PR happened to persist it in the golang rewrite. You could change this line to make sure the command *doesn't* fail now.

bacongobbler:
rebased and PR'd the changes at #5010. Closing!

PR #4939, opened 2016-02-28
fix(client): simplify URL prefixing

mboersma:
Go's net/url package returns malformed urls when a port is specified to `url.Parse()`.

For example, given "kiddie-armchair.192.168.64.2.xip.io:31258", `deis open` would call this `AppOpen` func and open a browser to "http:31258". The code appears to intend just to ensure a URL scheme prefix, so string manipulation should be adequate.

See deis/workflow#426.

### DIFF
--- a/client/cmd/apps.go
+++ b/client/cmd/apps.go
@@ -2,7 +2,6 @@ package cmd
 
 import (
 	"fmt"
-	"net/url"
 	"os"
 	"strings"
 	"time"
@@ -142,15 +141,12 @@ func AppOpen(appID string) error {
 		return err
 	}
 
-	u, err := url.Parse(app.URL)
-
-	if err != nil {
-		return err
+	u := app.URL
+	if !(strings.HasPrefix(u, "http://") || strings.HasPrefix(u, "https://")) {
+		u = "http://" + u
 	}
 
-	u.Scheme = "http"
-
-	return webbrowser.Webbrowser(u.String())
+	return webbrowser.Webbrowser(u)
 }
 
 // AppLogs returns the logs from an app.


Joshua-Anderson:
Yeah. The intention was only to set the protocol to `http`, since there is no way to detect if ssl is in use and the router might automatically redirect if `forceHTTPS` is set.

Why are you listening on a non-standard port anyway?
Has this issue already been reported in the go issue tracker?

`url.Parse` is also used in register/login/init-session code as well, so if the controller is listening on a non-standard port, likely given that the apps aren't listening on it, the cli wouldn't be able to communicate with the controller.

Joshua-Anderson:
Are you sure that this is related to `url.Parse()`? I'm unable to reproduce with the [golang playground](https://play.golang.org/p/6Y0Cpjx0Lg).

Are you sure this isn't a problem with the webbrowser package?

Joshua-Anderson:
Ah. It is a problem with url.Parse()

If you [set `url.Scheme`](https://play.golang.org/p/JtcJVlvtR_) for some reason it clears the host :frowning:

mboersma:
> Why are you listening on a non-standard port anyway?

Testing in kube-solo or on vagrant you get

Joshua-Anderson:
Ok. I figured this out:

A scheme is not required to use `//` after it. Therefore, go parses anything prior to the first `:` as the scheme.

So `kiddie-armchair.192.168.64.2.xip.io` is parsed as the scheme, and `31258` as the host. This is probably a wontfix by golang as they are complaint with the spec.

@mboersma For this particular location, you could just prepend `http://`, since the api never returns a scheme. However, if we want to support the controller running on a non-standard port as well, we will need to factor `url.Parse` out of the entire codebase.

mboersma:
> need to factor `url.Parse` out

So is this fix reasonable @Joshua-Anderson?

Joshua-Anderson:
The api will never return a scheme from the app URL, so you can just prepend `http://` to the returned URL.

The thing that confuses me is that we use url.Parse in a dozen or so places to communicate with the controller, so I can't understand why app:open wouldn't work but they would, since we do the exact same thing to set a scheme there.

mboersma:
> just prepend http:// to the returned URL

That approach would seem to preclude https:// URLs, which is why I checked for a schema first. But if the API will always return a simple host:port as you mentioned, then how can we know whether SSL is being used?

Also @Joshua-Anderson, I see your point about `url.Parse` being in multiple places. I always test with a custom port (using [kube-solo](https://github.com/TheNewNormal/kube-solo-osx)) but this and deis/workflow#426 are the only issues I've encountered so far.

bacongobbler:
> then how can we know whether SSL is being used?

I guess we really can't tell for sure beforehand unless we initiated a connection to the controller. The only thing we could reasonably do is check if the port # is 443.

bacongobbler:
We could also enforce users to provide a scheme when running `deis login` to mitigate this risk to users who modify their client.json file.

PR #4944, opened 2016-02-29
chore(buildpacks): update heroku-buildpack-go to v31

mboersma:
Adds Go 1.6 support.
See https://github.com/heroku/heroku-buildpack-go/compare/v30...v31


bacongobbler:
wouldn't the end-to-end tests cover this since the default app is example-go? :)

mboersma:
> wouldn't the end-to-end tests cover this...?

You're right. Sometimes we get lucky.

PR #4945, opened 2016-02-29
chore(buildpacks): update heroku-buildpack-scala to v66

mboersma:
Adds SBT 0.13.11 support.
See https://github.com/heroku/heroku-buildpack-scala/compare/v65...v66


krancour:
I'm manually testing this.

krancour:
Tested with the example-scala app and all checks out.  I'm not sure why this is showing up as needing a re-base, but it is.

PR #4959, opened 2016-03-10
feat(contrib): monitor router events

aespinosa:
A utility to watch the /deis/services/** space for events related to confd updating
the nginx configuration.


carmstrong:
Should this be documented somewhere? If it's merged into Deis itself, I think the project takes on the maintenance burden. One option is for it to live at `aespinosa/deis` and link to it in [contrib/README.md](https://github.com/deis/deis/blob/master/contrib/README.md).

mboersma:
Deis LTS is in maintenance mode, so new features like this aren't likely to be merged since as @carmstrong pointed out, it effectively increases the support burden. I'm going to close this since it's been sitting a few months, but please re-open it if the suggestion above is amenable.

PR #4970, opened 2016-03-21
chore(buildpacks): update heroku-buildpack-php to v97

mboersma:
See https://github.com/heroku/heroku-buildpack-php/compare/v94...v97


mboersma:
I tested this with [example-php](https://github.com/deis/example-php).

PR #4971, opened 2016-03-21
chore(buildpacks): update heroku-buildpack-grails to v20

mboersma:
See https://github.com/heroku/heroku-buildpack-grails/compare/v19...v20


mboersma:
I tested this with Stackato's [hello-grails](https://github.com/Stackato-Apps/hello-grails).

PR #4972, opened 2016-03-21
chore(buildpacks): update heroku-buildpack-java to v44

mboersma:
See https://github.com/heroku/heroku-buildpack-java/compare/v43...v44


mboersma:
I tested this with [example-java-jetty](https://github.com/deis/example-java-jetty).

PR #4973, opened 2016-03-21
chore(buildpacks): update heroku-buildpack-scala to v67

mboersma:
See https://github.com/heroku/heroku-buildpack-scala/compare/v66...v67


mboersma:
I tested this with [example-scala](https://github.com/deis/example-scala).

PR #4974, opened 2016-03-21
chore(buildpacks): update heroku-buildpack-ruby to v145

mboersma:
See https://github.com/heroku/heroku-buildpack-ruby/compare/v144...v145


mboersma:
I tested this with [example-ruby-sinatra](https://github.com/deis/example-ruby-sinatra).

PR #4975, opened 2016-03-21
chore(buildpacks): update heroku-buildpack-multi to v1.0.0

mboersma:
This is the same as the previous SHA 26fa21a, but tidier.


mboersma:
I tested this with dzuelke's [heroku-multipack-nodejs-php-example](https://github.com/dzuelke/heroku-multipack-nodejs-php-example), which also tested #4970 and #4976.

PR #4976, opened 2016-03-21
chore(buildpacks): update heroku-buildpack-nodejs to v89

mboersma:
See https://github.com/heroku/heroku-buildpack-nodejs/compare/v88...v89


mboersma:
I tested this with [example-nodejs-express](https://github.com/deis/example-nodejs-express).

bacongobbler:
thanks @mboersma! I'll also confirm tomorrow morning.

PR #4977, opened 2016-03-22
chore(buildpacks): update heroku-buildpack-python to v78

mboersma:
See https://github.com/heroku/heroku-buildpack-python/compare/v77...v78


krancour:
@mboersma I've done this before and now I can't remember how...

How can I test with a custom build of the slugbuilder based on this branch?

krancour:
Will a custom build of builder bring this along with it?

mboersma:
> Will a custom build of builder bring this along with it?

@krancour yes--`make -C builder deploy` should do it.

krancour:
@mboersma I'll check out all the buildpack updates as soon as I have a window.

mboersma:
@krancour thanks--no rush. It would be nice to get these updates in for the LTS release, but not strictly required.

mboersma:
I tested this with [example-python-django](https://github.com/deis/example-python-django) and [example-python-flask](https://github.com/deis/example-python-flask). 

See also deis/example-python-django#6.

PR #4981, opened 2016-03-22
fix(auth): return a 409 if a user is cancelled that has apps

### LINKED ISSUE #4909: [controller] When deleting a user, its apps are deleted from the database but keep running
When deleting a user, all the apps it owns are deleted from the controller's database. However, the apps (fleet services and docker containers) keep running. Afterwards, when re-creating an app with the same name as one of the apps that was deleted, `deis logs` will show the logs of the deleted app of the same name.

To reproduce:

- deis register http://controller.com
- deis create someapp
- deis auth:cancel
- deis login (with another admin user)

Notice that `deis apps` doesn't list the app that was deleted (`someapp`) but `fleetctl list-units | grep someapp` show that the app is still running and `http://someapp.controller.com` is still up.

deis v1.12.2

mboersma:
Closes #4909.

Backported from deis/controller#391. Thanks @helgi!

### DIFF
--- a/client/cmd/auth.go
+++ b/client/cmd/auth.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net/url"
 	"os"
+	"reflect"
 	"strings"
 	"syscall"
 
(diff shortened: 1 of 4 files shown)


mboersma:
Works in my testing:
```console
$ deis auth:cancel
Please log in again in order to cancel this account
username: matt
password: 
Logged in as matt
cancel account matt at http://deis.local3.deisapp.com? (y/N): y
Error: 
409 CONFLICT
detail: matt still has applications assigned. Delete or transfer ownership

$ deis destroy -a sanest-ironwood --confirm=sanest-ironwood
Destroying sanest-ironwood...
done in 4s
$ deis auth:cancel
Please log in again in order to cancel this account
username: matt
password: 
Logged in as matt
cancel account matt at http://deis.local3.deisapp.com? (y/N): y
Account cancelled
```

PR #4985, opened 2016-03-23
fix(store-volume): workaround kernel bug for CephFS mount

### LINKED ISSUE #4937: bug(ceph): do not upgrade to 4.3+ linux kernels
This is just a warning in case a stable CoreOS upgrade. 
http://tracker.ceph.com/issues/14360

carmstrong:
The latest CoreOS stable release breaks Deis because of a kernel bug relating to CephFS.

See http://tracker.ceph.com/issues/14360

closes #4937
refs https://github.com/coreos/bugs/issues/1092

### DIFF
--- a/deisctl/units/deis-store-volume.service
+++ b/deisctl/units/deis-store-volume.service
@@ -5,7 +5,7 @@ Description=deis-store-volume
 EnvironmentFile=/etc/environment
 ExecStartPre=/usr/bin/mkdir -p /var/lib/deis/store
 ExecStartPre=/bin/sh -c "echo waiting for store-monitor... && until etcdctl get /deis/store/monSetupComplete >/dev/null 2>&1; do sleep 2; done"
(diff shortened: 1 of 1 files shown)


carmstrong:
This is the advertised fix, but I haven't yet had a chance to test due to being on slow WiFi. @aledbf if you are able to test this replacement unit, that would be great!

stuszynski:
@carmstrong It's working for me. Thanks!

mboersma:
Nice work. @carmstrong @aledbf if this is a good fix, shouldn't Deis v1 also move to CoreOS stable 899.13.0? It obviously can't be prevented everywhere thanks to different provisioning schemes, so I think we roll with it.

aledbf:
LGTM. 
tested with latest stable and alpha

carmstrong:
Closing in favor of #4987 which includes this fix.

PR #4987, opened 2016-03-23
chore(*): bump CoreOS to 899.13.0

### LINKED ISSUE #4937: bug(ceph): do not upgrade to 4.3+ linux kernels

mboersma:
Includes #4985.

Closes #4937.
Refs coreos/bugs#1092

### DIFF
--- a/contrib/azure/parameters.json
+++ b/contrib/azure/parameters.json
@@ -27,7 +27,7 @@
     "value": 100
   },
   "coreosVersion": {
-    "value": "835.13.0"
+    "value": "899.13.0"
   },
   "storageAccountType": {
     "value": "Premium_LRS"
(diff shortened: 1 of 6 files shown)


carmstrong:
One concern I have with the fix in #4985 is that it seems to be a workaround for SELinux interaction... so I want to see if it's upgrade-compatible for folks upgrading to the LTS release.

carmstrong:
closes #4937

mboersma:
This needs to be tested on an upgrade from v1.12.3.

mboersma:
I've tested this on an upgrade from v1.12.3 without any issues.

carmstrong:
> I've tested this on an upgrade from v1.12.3 without any issues.

:+1:

jgmize:
After running CoreOS 899.13.0 + Deis 1.12.3 on a 5-node test cluster for a few days and doing some heavy load testing, we're now running that combo on 2 production clusters, with no problems so far.

carmstrong:
> we're now running that combo on 2 production clusters, with no problems so far.

:+1:

PR #4995, opened 2016-03-24
chore(*): update base to alpine:3.3

### LINKED ISSUE #4904: Update all components to use the same version of Alpine Linux
Our components are using a combination of Alpine Linux versions. We should consider getting everything on the same version for consistency and so we fully understand our exposure for vulnerabilities.

The components are the following versions:

3.1
- database
- registry

3.2 
- builder
- controller
- logger
- logspout
- publisher
- router
- mock-store
- test-etcd

Note that Alpine 3.3 is the latest release, so we could upgrade everything to 3.3 as well.

mboersma:
Updates Docker base images to `alpine:3.3` to benefit from current code and fixes.

Closes #4904.

TODO:
- [x] upgrade builder ~~(compile errors currently)~~
- [x] upgrade registry ~~(compile errors currently)~~
- [x] upgrade database or document why it isn't (new version of postgres uses a different data dir, so an upgrade script would be needed) **Will create a separate PR when this succeeds**

### DIFF
--- a/registry/build.sh
+++ b/registry/build.sh
@@ -11,7 +11,7 @@ if [[ -z $DOCKER_BUILD ]]; then
 fi
 
 # install required packages (copied from dotcloud/docker-registry Dockerfile)
-apk add --update-cache \
+apk add --no-cache \
   build-base \
   git \
   openssl-dev \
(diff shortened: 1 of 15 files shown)


registry/build.sh
```diff
@@ -22,16 +22,13 @@ apk add --update-cache \
...
-# workaround to python > 2.7.8 SSL issues
-pip install --disable-pip-version-check --no-cache-dir pyopenssl ndg-httpsclient pyasn1
+curl -sSL https://bootstrap.pypa.io/get-pip.py | python - pip==8.1.1
 
 # create a registry user
 adduser -D -s /bin/bash registry
 
 # add the docker registry source from github
-git clone -b new-repository-import-master --single-branch https://github.com/deis/docker-registry /docker-registry && \
+git clone -b deis-v1-lts --single-branch https://github.com/deis/docker-registry /docker-registry && \
```
bacongobbler: for other maintainer's reference, this contains commit https://github.com/deis/docker-registry/commit/c37b9ddd3bade4ce8420fcbcb74a370ceeefee16.

mboersma:
I'm still investigating changes to make deis/database compatible with `alpine:3.3`, but it's not straightforward. If that bears fruit, I'll create a separate PR, but consider this one is complete as-is.

PR #4998, opened 2016-03-29
fix(router): Allow for comma-delimited X-Forwarded-Proto

### LINKED ISSUE #4997: X-Forwarded-Proto can be comma-delimited, but router doesn't account for that
This issue supersedes #4504

**TL;DR:** I'm pretty sure we should account for the possibility of multi-valued `X-Forwarded-Proto` headers.

#4504 could not be reproduced for the longest time.  @gerred and I have finally managed to do so, and in the course of doing so, uncovered the root cause.

First a sordid little tale of service discovery:  In Deis v1.x, there is no mechanism for applications (in a micro-service architecture, perhaps) deployed to Deis to discover one another and interact directly.  So in rare cases, some users have deployed apps that talk to one another by routing traffic from one, out to the internet and back into the cluster through the load balancer and the router.  It's not ideal, but it's a workaround real users have used.

Back to our regularly scheduled bug report... Consider the following (actual) scenario: a service proxy deployed to Deis is a facade over numerous other back-end services.  It's implemented using [hapi](http://hapijs.com/).  In this example, hapi is configured to set `X-Forwarded-For` and other, related headers on the outbound HTTPS request.  These headers were already set by the Deis router on the request that was inbound to the service proxy.  hapi copies these headers over from the inbound request to the outbound request, _but also amends them_.  Where the incoming `X-Forwarded-Proto` header may have read `https`, the outbound `X-Forwarded-Proto` header will read `https,https`.

The `X-Forwarded-Proto` header, like it's better known brother `X-Forwarded-For`, is a de-facto header and is not formally described anywhere.  A brief google search has revealed however, that apart from hapi, numerous other frameworks accept or rely on the possibility that the `X-Forwarded-Proto` header might contain a comma-delimited list.  The Deis router is not one of these...

The Deis router, internally, derives a variable named `$access_scheme` from the `X-Forwarded-Proto` header, as depicted here:

```
# trust http_x_forwarded_proto headers correctly indicate ssl offloading
map $http_x_forwarded_proto $access_scheme {
    default $http_x_forwarded_proto;
    ''      $scheme;
}
```

The ultimate intention of this code is to allow the `$access_scheme` variable to assume the value `https`, even if the protocol used for the inbound request was HTTP, _as long as it appears that SSL offloading has already been performed_ (e.g. by the load balancer).

If it is possible for `X-Forwarded-Proto` to bear comma-delimited values, it is also possible for the `$access_scheme` variable to obtain a value such as `https,https`.

When Deis router's `enforceHTTPS` feature is enabled, the `$access_scheme` variable is evaluated in determining if/when a redirect to an insecure URL's secure equivalent should be performed:

```
if ($access_scheme != "https") {
    return 301 https://$host$request_uri;
}
```

In a case such as this, the value `https,https` != `https`, and as a result a redirect is issued... but this will _potentially_ initiate a series of infinite redirects.  In the specific example given, hapi was configured to follow no more than five redirects, so after the sixth, hapi quit and returned a `502` response to the router, which propagated that response back to the client.

Correcting this bug may call for some careful thought.  Some questions:

1. Should SSL offloading be considered to have occurred (the `$access_scheme` variable thereby receiving a value of `https`) if `https` occurs _anywhere_ in the comma-delimited value of the `X-Forwarded-Proto` header?  Or only if the proximate forwarded protocol was `https`?

2. Is the Deis router also guilty of _setting_ this header incorrectly on outbound requests to back-end services?  For instance, if SSL were offloaded by the load balancer, and the request inbound to the router therefore already had the `X-Forwarded-Proto` header with a value of `https`, should the router merely copy that header to the outbound request as it does today?  Or should it amend it as hapi does?  i.e. Should the outbound request have the `X-Forwarded-Proto` header set to `https,https`.  (fwiw, I believe `https,http` would be _incorrect_ since the request was treated by the router _as if_ it were secure.  This seems to be how hapi deals with this as well.)  If we fix this, do we need to / should we _also_ change how we write other `X-Forwarded-*` headers?  `X-Forwarded-For`, for instance, can be comma-delimited as well.

3. Is there an edge case where a request arrives via HTTPS, with an `X-Forwarded-Proto` header of `http`?  Such a case is possible if a client (proxy of some sort) has different opinions than us on how `X-Forwarded-Proto` should be used.  This will trip up our _existing_ logic, because such a request will be treated as insecure, despite having arrived over a secure channel... and that will _also_ trigger the infinite loop described by this issue.

4. Are there technical constraints to correcting this?  Nginx configuration isn't an imperative language, so short of resorting to the Lua module, there are limited options for tackling this.

Whatever is done, the change should also be ported forward to the Deis v2 Router.

krancour:
Fixes #4997

This will _also_ have the actual protocol used trump whatever the `X-Forwarded-Proto` header says about whether the request is a secure one or not.

### DIFF
--- a/router/rootfs/etc/confd/templates/nginx.conf
+++ b/router/rootfs/etc/confd/templates/nginx.conf
@@ -65,10 +65,19 @@ http {
         ''      close;
     }
 
-    # trust http_x_forwarded_proto headers correctly indicate ssl offloading
-    map $http_x_forwarded_proto $access_scheme {
-      default $http_x_forwarded_proto;
-      ''      $scheme;
+    # The next two maps work together to determine the $access_scheme:
+    # 1. Determine if SSL may have been offloaded by the load balancer, in such cases, an HTTP request should be
+    # treated as if it were HTTPs.
+    map $http_x_forwarded_proto $tmp_access_scheme {
+      default $scheme;               # if X-Forwarded-Proto header is empty, $tmp_access_scheme will be the actual protocol used
+      "~^(.*, ?)?http$" "http";      # account for the possibility of a comma-delimited X-Forwarded-Proto header value
+      "~^(.*, ?)?https$" "https";    # account for the possibility of a comma-delimited X-Forwarded-Proto header value
+    }
+    # 2. If the request is an HTTPS request, upgrade $access_scheme to https, regardless of what the X-Forwarded-Proto
+    # header might say.
+    map $scheme $access_scheme {
+      default $tmp_access_scheme;
+      "https" "https";
     }
 
     ## HSTS instructs the browser to replace all HTTP links with HTTPS links for this domain until maxAge seconds from now


bacongobbler:
Is there an application/cluster setup that would be best for testing this?

krancour:
@bacongobbler I'll stand up a cluster with a wildcard cert and everything... I also have example applications courtesy of @gerred that demonstrate the bug.  You can see how those apps behave before and after the fix.  I'll ping you when it's ready.

PR #5008, opened 2016-04-03
fix(controller): prevent overlapping config:set operations

### LINKED ISSUE #4746: Need to block client request for change configs of app twice

mboersma:
Uses a file lock to ensure that multiple gunicorn processes can be coordinated in the container. If a `deis config:set` is in progress, another client attempting `config:set` will try to obtain the lock for 5 seconds, then return an error:
```
$ deis config:set POWERED_BY=Second
Creating config... Error: 
409 CONFLICT
error: Config changes already in progress.
[Errno 11] Resource temporarily unavailable
```

Closes #4746.
Closes #4758.

### DIFF
--- a/controller/api/views.py
+++ b/controller/api/views.py
@@ -1,6 +1,10 @@
 """
 RESTful view classes for presenting Deis API objects.
 """
+
+import os.path
+import tempfile
+
 from django.conf import settings
 from django.core.exceptions import ValidationError
 from django.contrib.auth.models import User
@@ -15,6 +19,7 @@ from rest_framework.permissions import IsAuthenticated
 from rest_framework.response import Response
 from rest_framework.viewsets import GenericViewSet
 from rest_framework.authtoken.models import Token
+from simpleflock import SimpleFlock
 
 from api import authentication, models, permissions, serializers, viewsets
 
@@ -276,6 +281,18 @@ class ConfigViewSet(ReleasableViewSet):
     """A viewset for interacting with Config objects."""
     model = models.Config
     serializer_class = serializers.ConfigSerializer
+
+    def create(self, request, **kwargs):
+        # Guard against overlapping config changes, using a filesystem lock so that
+        # multiple controller processes can be coordinated.
+        # Use a tempfile such as "/tmp/violet-valkyrie-config".
+        lockfile = os.path.join(tempfile.gettempdir(), kwargs['id'] + '-config')
+        try:
+            with SimpleFlock(lockfile, timeout=5):
+                return super(ConfigViewSet, self).create(request, **kwargs)
+        except IOError as err:
+            msg = "Config changes already in progress.\n{}".format(err)
+            return Response(status=status.HTTP_409_CONFLICT, data={'error': msg})
 
     def post_save(self, config):
         release = config.app.release_set.latest()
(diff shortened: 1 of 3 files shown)


bacongobbler:
Code LGTM, but pending official LGTM :tm: until some unit tests are added in.

Would the unit test be as easy to create the file then try to create config?

mboersma:
> Would the unit test be as easy to create the file then try to create config?

That's what I'm planning to do, because after some thought I'm having a hard time thinking of how to coordinate two threaded clients in a reliable way for tests.

mboersma:
This is ready for review now. I added a unit test to cover this 409 case, but let me know if you can think of better ways to test this.

I tried to move the `SimpleFlock` logic to the `Config.save()` method, but the behavior reverted: somehow the operations get serialized so the last `config:set` wins, which is essentially the current behavior we're trying to fix here. Leaving it attached to the `ConfigSetView.create()` method achieves the desired behavior. I'll put more thought into it for deis/controller#594 since it will also require different unit tests. I appreciate the goal of keeping logic in models rather than views, but for v1 LTS I'm hoping this fix is adequate.

bacongobbler:
I'm always happy to ship working code over perfect code, especially if we're sun-setting most of this.

controller/api/views.py
```diff
@@ -277,6 +282,17 @@ class ConfigViewSet(ReleasableViewSet):
     model = models.Config
     serializer_class = serializers.ConfigSerializer
 
+    def create(self, request, **kwargs):
+        # Guard against overlapping config changes, using a filesystem lock so that
+        # multiple controller processes can be coordinated.
+        lockfile = os.path.join(tempfile.gettempdir(), kwargs['id'])  # e.g., /tmp/violet-valkyrie
```
helgi: given it is config related, or more precisely build / deploy then shouldn't we append that so we can also have other locks for diff operations under the same namespace?

mboersma: Good idea. How granular should that be? I'm thinking `/tmp/violet-valkyrie-config` will be good enough, rather than `/tmp/violet-valkyrie-config-create`.

helgi: Yeah, your suggestion is granular enough

PR #5010, opened 2016-04-04
fix(client): only delete local ~/.deis/client.json if cancelling logged in user

### LINKED ISSUE #4896: [client] `deis auth:cancel --username=someuser` removes ~/.deis/client.json

bacongobbler:
fixes #4896 
closes #4923

### DIFF
--- a/client/cmd/auth.go
+++ b/client/cmd/auth.go
@@ -247,7 +247,7 @@ func Cancel(username string, password string, yes bool) error {
 	}
 
 	// If user targets themselves, logout.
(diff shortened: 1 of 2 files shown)


bacongobbler:
thanks @Joshua-Anderson!

PR #5013, opened 2016-04-05
ref(client): improve client error message

bacongobbler:
This PR improves the resultant client error message they get when the returned response is
incorrect. This is especially important for determining the error message

Old message when using a v1 client with a v2 cluster:

```
http://deis.10.245.1.3.xip.io does not appear to be a valid Deis controller. Make sure that the
Controller URI is correct and the server is running.
```

New message:

```
http://deis.10.245.1.3.xip.io/v1/ does not appear to be a valid Deis controller. Make sure that the
Controller URI is correct and the server is running.
```

Notice the appended /v1/ path, which users can assume this is a v1 client.

### DIFF
--- a/client/controller/client/http.go
+++ b/client/controller/client/http.go
@@ -157,8 +157,6 @@ func CheckConnection(client *http.Client, controllerURL url.URL) error {
 	errorMessage := `%s does not appear to be a valid Deis controller.
 Make sure that the Controller URI is correct and the server is running.`
 
-	baseURL := controllerURL.String()
-
 	controllerURL.Path = "/v1/"
 
 	req, err := http.NewRequest("GET", controllerURL.String(), bytes.NewBuffer(nil))
@@ -171,13 +169,13 @@ Make sure that the Controller URI is correct and the server is running.`
 	res, err := client.Do(req)
 
 	if err != nil {
-		fmt.Printf(errorMessage+"\n", baseURL)
+		fmt.Printf(errorMessage+"\n", controllerURL.String())
 		return err
 	}
 	defer res.Body.Close()
 
 	if res.StatusCode != 401 {
-		return fmt.Errorf(errorMessage, baseURL)
+		return fmt.Errorf(errorMessage, controllerURL.String())
 	}
 
 	checkAPICompatibility(res.Header.Get("DEIS_API_VERSION"))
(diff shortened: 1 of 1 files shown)


mboersma:
Good thinking! I tested this change locally:
```console
$ make -C client/ build
$ deis login http://not.real
http://not.real/v1/ does not appear to be a valid Deis controller.
Error: Get http://not.real/v1/: dial tcp: lookup not.real on 8.8.8.8:53: no such host
```

mboersma:
Should this also be ported forward for [Deis v2](https://github.com/deis/workflow-cli)?

bacongobbler:
I had a branch available but github wouldn't let me PR it lol. Thanks for porting the fix!

carmstrong:
Could we even take it a step farther and emit a very specific "You must upgrade your client" error when the client detects "/v2" in the controller URL? I'm not sure it's obvious to users who aren't all that familiar with Deis that a `v2

carmstrong:
Could we take it a step further and emit a specific "You must upgrade your client" error when detecting "v2" in the controller URL with a v1 client? I don't think that including the full controller URL with `/v2` is obvious to the average Deis user -- it could just be an API version or somesuch that is inconsequential.

mboersma:
> Could we take it a step further...

The client doesn't know that a parallel `/v2/` endpoint does exist, only that `/v1/` didn't. And I think this message could be triggered by conditions other than a Deis version mismatch, so I would argue this is good enough. At least there is a way to tell now what endpoint `deis` was trying.

mboersma:
Or were you suggesting the CLI have additional probing code to see if there is specifically a version mismatch?

bacongobbler:
I don't think we should do a version probe or else we're gonna have to do that for every major release going forward. However, @slack's recommendation at https://github.com/deis/workflow-cli/pull/34#issuecomment-206439428 would be a great middleground.

carmstrong:
> Or were you suggesting the CLI have additional probing code to see if there is specifically a version mismatch?

I was thinking that it'd be nice to do a major version mismatch check.



> I don't think we should do a version probe or else we're gonna have to do that for every major release going forward.

Agreed.

bacongobbler:
@carmstrong we actually [do a check](https://github.com/bacongobbler/deis/blob/better-client-error/client/controller/client/http.go#L182) for the DEIS_API_VERSION header. The problem is that the v2 server does not respond to /v1/ requests so we get a 404 with no DEIS_API_VERSION header :'(

bacongobbler:
Considering that https://ci.deis.io/job/test-ec2/4030/ passed and 4031 is only a error message change (which we already changed the message before and it passed in 4030), I think this is safe enough to merge to get the LTS release out the door today. If something breaks I owe @mboersma a :beer: :)

PR #5022, opened 2016-04-27
imp(publisher): concurrent polling

laurrentt:
I've noticed that my apps could take a very long time before getting published to the router. After investigation, I found that it was caused by the cumulation of all the HEALTHCHECK_INITIAL_DELAY.

For exemple, let's say all my apps have an initial delay of 1 minute. If I have 100 apps, it will take 1 minute to publish the first one, 2 minutes for the second, and so on. The last one will be published only after 100 minutes.

This PR fixes the issue


mboersma:
This looks like a great improvement--as usual, thanks @laurrentt!

PR #5023, opened 2016-04-28
chore(roadmap): update for May planning meeting

slack:


mboersma:
This looks great to me @slack. Deis v1 LTS testing is temporarily broken until I sort out some issues on the new CI nodes, so disregard the failure here: you can see in the logs that the `make -C docs build test` passed.

slack:
Ah cool. Thanks for the heads up!

PR #5024, opened 2016-04-29
chore(buildpacks): update heroku-buildpack-nodejs to v90

mboersma:
See https://github.com/heroku/heroku-buildpack-nodejs/compare/v89...v90


mboersma:
I tested this against Deis v1 master with [example-nodejs-express](https://github.com/deis/example-nodejs-express).

PR #5025, opened 2016-05-02
chore(buildpacks): update heroku-buildpack-php to v102

mboersma:
See https://github.com/heroku/heroku-buildpack-php/compare/v97...v102


mboersma:
I tested this against Deis v1 master with [example-php](https://github.com/deis/example-php).

PR #5026, opened 2016-05-02
chore(buildpacks): update heroku-buildpack-ruby to v146

mboersma:
See https://github.com/heroku/heroku-buildpack-nodejs/compare/v145...v146


mboersma:
I tested this against Deis v1 master with [example-ruby-sinatra](https://github.com/deis/example-ruby-sinatra).

PR #5032, opened 2016-05-05
docs(badge): added code-beat badge

chaitanyaenr:
contributes to https://github.com/deis/builder/issues/278, https://github.com/deis/builder/issues/218, https://github.com/deis/builder/issues/277


mboersma:
Should we merge this change at all? Currently it shows that codebeat gives Deis v1 a "B." [codebeat badge](https://codebeat.co/projects/github-com-deis-deis)

That's fine (actually that's pretty good given the disparate nature of the big code base and its varied test coverage).

But Deis v1 LTS is in maintenance mode now, so it's very unlikely anyone will invest the time into improving the codebeat grade. So it will stay a "B" forever most likely, which doesn't look great.

I'm all for more metrics and for open disclosure of code quality. But I feel like these grade badges are an incentive to raise coverage and quality, but that's not happening in this maintenance codebase. So I guess I'd prefer not to put this in.

Other opinions?

PR #5034, opened 2016-05-11
chore(planning): updates for May planning

slack:


docs/roadmap/planning.rst
```diff
@@ -31,13 +31,24 @@ Release Planning Meetings are open to the public with access coordinated via the
 Notes from past meetings are below, along with links to a recording of the entire meeting on YouTube.
 
 
-May 2016
+June 2016
 ~~~~~~~~
```
mboersma: This needs another tilde `~` I think. ReStructuredText is strict about matching in titles.

slack: Ahh, didn't know that. Updated and pushed.

Issue #5035, opened 2016-05-11
Docs: clarify AWS instructions on upgrading


rstacruz:
In the [Upgrading Deis](http://docs.deis.io/en/latest/managing_deis/upgrading-deis/) page, it has this to say about AWS:

> On AWS, Deis enables the PROXY protocol by default. If an in-place upgrade is required, run deisctl config router set proxyProtocol=1, enable PROXY protocol for ports 80 and 443 on the ELB, add a TCP 443:443 listener, and change existing targets and health checks from HTTP to TCP.

I think it should be clarified a little further.

- Do you really need to enable PROXY protocol? Shouldn't this have already been enabled for you when you first installed Deis? If not, how do you enable it? As far as I know, it's not a straight-forward process; there aren't controls for this in the AWS web console.

- `deisctl config router set proxyProtocol=1` probably needs to be ran *before* the uprgade, so this note should be higher up the page.

- It says to _"change existing targets and health checks from HTTP to TCP."_, but not exactly what to configure here (like what port.

I almost messed up an upgrade here because ELB thought my nodes are dead; I changed the health check to TCP and it thought my nodes are dead, haha.

For reference, it looks like this in the AWS web console:

![image](https://cloud.githubusercontent.com/assets/74385/15188135/da60da3c-17d6-11e6-888f-f0f354cfd5c3.png)

bacongobbler:
> Shouldn't this have already been enabled for you when you first installed Deis? If not, how do you enable it?

Yes. At the the end of the AWS install script, deisctl makes a call to [enable proxy protocol](https://github.com/deis/deis/blob/master/contrib/aws/provision-aws-cluster.sh#L174) for you. This section of the docs only applies to users upgrading from v1.5.0 and earlier, since v1.6.0 enables proxy protocol by default.

If you have a cluster newer than v1.5.0, this section of the docs does not apply.

mboersma:
Closed by #5036. Thanks @rstacruz!

PR #5036, opened 2016-05-11
docs(upgrading): clarify AWS-specific instructions

rstacruz:
I've clarified the AWS notes to highlight that updating proxy protocol support will only be necessary when upgrading from Deis versions that are older than v1.6.

I've also changed the ELB health check note to recommend to not change it—your instances will simply return anyway, you just need to wait. I'm not sure if this is the best way to go, but I've tried *changing health checks from HTTP to TCP* (as noted in the current documentation), but I wasn't sure what port to set it to; I tried setting it to port 80 and it marked my instances as `OutOfService` anyway.

Also see #5035.


rstacruz:
On second thought, not changing ELB health checks a bad idea! it will make your production websites unavailable for a brief period (not good!).

PR #5038, opened 2016-05-12
chore(buildpacks): update heroku-buildpack-go to v41

mboersma:
See https://github.com/heroku/heroku-buildpack-go/compare/v34...v41


mboersma:
I tested this with [example-go](https://github.com/deis/example-go) on Vagrant with 1.13.1+.

PR #5039, opened 2016-05-17
docs(index): integrate v2 header into index

bacongobbler:


sgoings:
I’m concerned that since it’s not a part of a warning message with different styling it’s now hidden enough people won’t see it...

sgoings:
@bacongobbler and I discussed this and since we want to direct people to use v2 over v1, this awkward (but effective) header seems to do the trick quite nicely.

PR #5042, opened 2016-06-01
chore(*): bump CoreOS to 1010.5.0

mboersma:
See https://coreos.com/releases/#1010.5.0

### DIFF
--- a/contrib/azure/parameters.json
+++ b/contrib/azure/parameters.json
@@ -27,7 +27,7 @@
     "value": 100
   },
   "coreosVersion": {
-    "value": "899.17.0"
+    "value": "1010.5.0"
   },
   "storageAccountType": {
     "value": "Premium_LRS"
--- a/contrib/linode/provision-linode-cluster.py
+++ b/contrib/linode/provision-linode-cluster.py
@@ -357,7 +357,7 @@ def main():
                                   help='Node data center id. Use list-data-centers to find the id.')
     provision_parser.add_argument('--cloud-config', required=False, default='linode-user-data.yaml', type=file, dest='cloud_config',
                                   help='CoreOS cloud config user-data file')
-    provision_parser.add_argument('--coreos-version', required=False, default='899.17.0', dest='coreos_version',
+    provision_parser.add_argument('--coreos-version', required=False, default='1010.5.0', dest='coreos_version',
                                   help='CoreOS version number to install')
     provision_parser.add_argument('--coreos-channel', required=False, default='stable', dest='coreos_channel',
                                   help='CoreOS channel to install from')
--- a/contrib/utils.sh
+++ b/contrib/utils.sh
@@ -13,5 +13,5 @@ function echo_green {
 }
 
 export COREOS_CHANNEL=${COREOS_CHANNEL:-stable}
-export COREOS_VERSION=${COREOS_VERSION:-899.17.0}
+export COREOS_VERSION=${COREOS_VERSION:-1010.5.0}
 export DEIS_RELEASE=1.13.1
--- a/controller/bin/boot
+++ b/controller/bin/boot
@@ -86,6 +86,11 @@ if addgroup -g "$(stat -c "%g" /var/run/docker.sock)" docker; then
 	addgroup deis docker
 fi
 
+# allow deis user permission to fleet
+if addgroup -g "$(stat -c "%g" /var/run/fleet.sock)" fleet; then
+    addgroup deis fleet
+fi
+
 # run an idempotent database migration
 sudo -E -u deis ./manage.py syncdb --migrate --noinput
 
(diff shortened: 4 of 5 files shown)


mboersma:
This may need an update to docker-py or somesuch. It's currently failing on the first app push with `failed to retrieve configuration`.

laurrentt:
I also get a `raise CannotSendRequest()` whenever the controller tries to make a request to fleet. I'm investigating.

bacongobbler:
keeping @arschles' LGTM, but I'm removing mine until I can actually manually test this as well.

laurrentt:
Can you add this:

```
# allow deis user permission to fleet
if addgroup -g "$(stat -c "%g" /var/run/fleet.sock)" fleet; then
	addgroup deis fleet
fi
```

to controller/bin/boot ? It seems to fix the fleet issue.

mboersma:
Thanks @laurrentt! I was just starting to debug why controller couldn't talk to fleet any longer...

I tested this change locally on vagrant. 👍

aledbf:
@mboersma just in case I'm getting [this](https://github.com/coreos/bugs/issues/1301) error with 1010.5.0  if overlayfs is used (in some nodes)

bacongobbler:
thanks for pointing this out, @aledbf. I'll see if I can reproduce it in my testing environment to see if this is a showstopper and we should wait for 1080.0.0 to get on the stable channel.

bacongobbler:
I was not able to reproduce given the following diff for [example-dockerfile-python](https://github.com/deis/example-dockerfile-python), which is based on alpine:

```
diff --git a/Dockerfile b/Dockerfile
index 1c632ba..644fc63 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,6 @@
 FROM gliderlabs/alpine:3.4
-
+MAINTAINER matt fisher <me@bacongobbler.com>
+RUN mkdir -p /tmp/foo
 RUN apk-install python
 ADD . /app
 WORKDIR /app
```

@aledbf do you have a way to reliably repro this bug?

aledbf:
@bacongobbler no but I see this with images that share some intermediate layer

aledbf:
@bacongobbler just in case I reinstalled the cluster with `899.17.0`.

mboersma:
I also tested this by pushing several Dockerfile and buildpack apps on CoreOS 1010.5.0, but I assume the bug isn't that easy to reproduce: worked fine.

mboersma:
@aledbf is this bug easy or difficult to trigger? Is there a workaround?

Given that @bacongobbler and I couldn't reproduce it with initial testing, I'm trying to assess if the severity of coreos/bugs#1301 is enough to affect the [v1.13.2](https://github.com/deis/deis/releases/tag/v1.13.2) maintenance release. Options are:
- release v1.13.2 as-is (essentially same as master), knowing this bug may occur
- revert this PR and ship v1.13.2 with CoreOS 899.17.0
- delay a v1.13.2 release until there is a resolution for coreos/bugs#1301

aledbf:
> Is there a workaround?

No. The only solution (mentioned in a docker issue) is to not use overlay if SELinux is enabled

> is this bug easy or difficult to trigger?

For me after 5 minutes with 1010.5.0 is impossible to start containers in some nodes (apps using buildpacks). Now maybe the issue it is not easy to reproduce with a low number of containers (I have ~3800)

I mentioned this issue here because is not possible to return to 1.9.1 after the upgrade to 1.10.3 which is a huge issue in a production cluster.

bacongobbler:
> Options are:

My vote is to revert this PR to ship v1.13.2, given that [there are a metric ton of good fixes](https://github.com/deis/deis/compare/v1.13.1) in this release. OTOH, coreos/bugs#1301 has been given P1 priority and should hopefully be fixed soon, but we're not sure how long it'll take until said fix will land in the stable channel.

laurrentt:
I've been running this code in staging for the last 2 weeks and in prod for a week and personally, I won't not revert it. So just to be clear, those who use Deis pull are not impacted, right ?

mboersma:
The issue is fixed in https://github.com/coreos/coreos-overlay/pull/2006, but I don't know when that will make it to stable.

laurrentt:
@mboersma Sorry for my confusing post 😕 . Here's my question: Does this CoreOS issue only affects the Deis builder (the only instance of docker inside docker I know in the platform) ? Thank you 😄

mboersma:
> Does this only affect the Deis builder?

I don't know for sure. My assumption was it could also be triggered by the Deis controller component, since it mounts `docker.sock` inside the container. Sorry I wasn't more specific--maybe @aledbf can explain where it was triggered for him?

It seems safer to revert CoreOS to the "known good" version until this bug is resolved--see #5053. Does that make sense to you @laurrentt? You should be able to stay on CoreOS 1010.5.0 if you prefer with the next Deis patch release, if the bug wasn't affecting you.

aledbf:
> Does this CoreOS issue only affects the Deis builder

No, I saw issues with sysdig and `gcr.io/google_containers` images with layers in commons

aledbf:
@laurrentt as @mboersma said if the bug is not affecting you just continue using CoreOS 1010.5.0

laurrentt:
Thanks @mboersma and @aledbf for the precisions. I'm keeping my fingers crossed that it won't happen in my environments and will update my CoreOS as soon as the patch is out.

(Not sure if crossing fingers qualify as best practise though... 😂 but I have faith)

PR #5057, opened 2016-06-20
feat(router): Add server_tokens option

Crispy1975:
Adding the options to turn server tokens on/off and the ability to prevent click-jacking with the `X-Frame-Options` header being present.

### DIFF
--- a/docs/customizing_deis/router_settings.rst
+++ b/docs/customizing_deis/router_settings.rst
@@ -70,6 +70,7 @@ setting                                      description
 /deis/router/maxWorkerConnections            maximum number of simultaneous connections that can be opened by a worker process (default: 768)
 /deis/router/serverNameHashMaxSize           nginx server_names_hash_max_size setting (default: 512)
 /deis/router/serverNameHashBucketSize        nginx server_names_hash_bucket_size (default: 64)
+/deis/router/serverTokens                    nginx server_tokens setting (default: not set)
 /deis/router/sslCert                         cluster-wide SSL certificate
 /deis/router/sslCiphers                      cluster-wide enabled SSL ciphers
 /deis/router/sslKey                          cluster-wide SSL private key
--- a/router/rootfs/etc/confd/templates/nginx.conf
+++ b/router/rootfs/etc/confd/templates/nginx.conf
@@ -12,6 +12,9 @@ events {
 
 
 http {
+    # Server signature toggle (default is on)
+    server_tokens {{ or (getv "/deis/router/serverTokens") "on" }};
+
     # basic settings
     vhost_traffic_status_zone shared:vhost_traffic_status:{{ or (getv "/deis/router/trafficStatusZoneSize") "1m" }};
 


deis-admin:

krancour:
This doc should be updated as part of this PR: https://github.com/deis/deis/blob/master/docs/customizing_deis/router_settings.rst#settings-used-by-router

Crispy1975:
Thanks, will sort that out later.

Crispy1975:
@krancour docs updated.

krancour:
@Crispy1975, thanks for the doc updates. This is looking good. As I consider this further, one nagging question keeps popping into my head, and that is one of whether the value of the `X-Frame-Options` header is better configured per Deis application rather than router-wide. By contrast, disabling server tokens router-wide makes sense to me because that seems like something an ops person would _want_ to enforce cluster-wide if policy dictates. But when it comes to `X-Frame-Options`, I could very easily see application-specific values being necessary. While one application might require `DENY` for security reasons, another, might (let's say a SPA) might require `SAMEORIGIN`, and yet a third (let's say an ad server) might require that this not be restricted.

Obviously, making this configurable per application is a great deal more work, but I also want to make sure we're not hastily adding a feature that impairs overall flexibility.

The possibility also exists that, since we've gotten by without any explicit support for `X-Frame-Options` up until now, that we should continue to do so. I do believe that applications that must set a particular value in that header for one reason or another can easily do so themselves, and the value should pass through in the proxied response.

@Crispy1975, any thoughts on this? I'd like to hear some fellow maintainers weigh in on this as well.

Crispy1975:
@krancour we have actually had this question come up. We have a couple of containers that have nginx inside of them running AngularJS apps and have the `X-Frame-Options` header set inside making it in effect app specific. I actually agree that it would be nice to have app specific. How that would be achieved I am not sure. `etcd` values that correlate with app names then outputting the relevant nginx settings in the `http` blocks? Say `/deis/router/xFrameOptions/<app>` as the key?

I fear that setting this on the deis client side might be beyond me as I am not a Go coder at this time.

bacongobbler:
> I fear that setting this on the deis client side might be beyond me

any keys added via `deis config:set` are available through etcd, as shown [here](https://github.com/deis/deis/blob/9d5feab5fbd5f3dcb518ce81ea65ca1310d74421/router/rootfs/etc/confd/templates/nginx.conf#L228). If you add a key like `/deis/config/appname/deis_x_frame_options` then users can set it via `deis config:set DEIS_X_FRAME_OPTIONS`. :)

Crispy1975:
Thanks @bacongobbler will take a look at that...

krancour:
Personally, I am not in favor of driving this off of environment variables. I've said this many times before in relation to other issues, but environment variables set via `deis config:set` create env vars "owned" by the application. As such, those should affect the behavior of the application itself and _not_ the platform. And in fact, there were places in v1.x where env vars were misused out of convenience and those have been (or are in the process of) being corrected now in v2.x. I cannot in good conscience endorse further abuse of that pattern with the knowledge that it's just one more thing to fix later.

If already this can be handled by the application itself, why is it necessary at all for the router to support this?  I would vote that this PR be pared down to just the parts having to do with disabling server tokens.

Crispy1975:
@krancour I can do that no problem as that would suit our current requirements at the moment, however only two of our apps have nginx inside the containers. The rest are HapiJS/Node.js scaled up so no nginx there to be able to set the `X-Frame-Options` on. I suppose Hapi could be configured to send that header in return also...

krancour:
@Crispy1975 I have to believe _most_ (albeit, probably not all) application frameworks can accommodate this.

Crispy1975:
@krancour worth keeping the existing global `xFrameOptions` setting for those that do want to go universal? Perhaps adding a note to the docs?

If not will remove.

krancour:
@Crispy1975, imo, I'd not include that option.

My reasoning is that, in general, I think the router should handle the most general concerns and the things that applications cannot or should not reasonably be expected to account for themselves-- gzip, SSL termination, anything having to do with proxying the requests / responses (e.g. determining real end user ip). The thing that all of these have in common is that if the router, as a middle-man, doesn't account for them, they don't work (or at the very least aren't easily accomplished by most application frameworks owing to an expectation that the web server handles those concerns). That's not the case with setting an app-specific response header.

If any other maintainer disagrees, I'm happy to hear another perspective on it.

Crispy1975:
OK, happy with that. Unless there are any disagreements I will drop the `X-Frame-Options` setting and just leave the tokens stuff.

Crispy1975:
@krancour `X-Frame-Options` stuff removed, just server tokens left.

krancour:
@Crispy1975 can you squash the commits please and make sure the commit message adheres to the documented [commit style](https://deis.com/docs/workflow/contributing/submitting-a-pull-request/#commit-style). After that, LGTM.  Thanks!

Crispy1975:
@krancour that should be done as request now.

Crispy1975:
Was this ok now @krancour ?

krancour:
@Crispy1975, sorry for the delay. LGTM.

krancour:
And thank you for the work on this.

Crispy1975:
:+1:

PR #5068, opened 2016-07-19
chore(*): bump CoreOS to 1068.8.0

mboersma:
See https://coreos.com/releases/#1068.8.0

PR #5071, opened 2016-07-20
Use official docker/dind image

nathansamson:
This seems to have resolved the problems with newer CoreOS versions


nathansamson:
Note I tried this with 2 buildpack applications and did not see any problems after a combined 5 deploys

When using the default image it would error out after one or 2 pushes max.

I only tested buildpacks and not docker build (as I don't use this feature myself)

mboersma:
Awesome, thanks @nathansamson. I was considering this change to fix up #5068, although I first started by updating the `wrapdocker` script itself from jpetazzo's master (didn't work).

mboersma:
```console
$ docker run -it --privileged dbuilder:v1.13.2 docker --version
Docker version 1.8.3, build f4bf5c7
$ docker run -it docker:1.8-dind docker --version
Docker version 1.8.3, build f4bf5c7
```

nathansamson:
Could be closed in favor of #5074

nathansamson:
Closing this as I believe this is not essential to the fix of the problem. Even if so, it currently is included in #5074  as well

PR #5072, opened 2016-07-20
feat(client) Sort listings

nathansamson:
This sorts listing of

* domain names
* applications
* process types

Main motivation is consistency of output. Eg when reinstalling a cluster from scratch (lets say migrate V1 to V2) this makes it easier to ran these commands and see where the differences are between the clusters.


nathansamson:
Added some more types as suggested in deis/workflow-cli#144

I "skipped" certs and tags because I believe they are sorted in the PrettyPrints method but I did not verify this

Unless I am mistaken these should sort all the listings which weren't sorted before

Joshua-Anderson:
Thanks, looks good! However, there is a gofmt error that needs to be fixed up in `cmd/keys.go`. You can just run `gofmt -w cmd/keys.go` to format the file for you.

nathansamson:
Fixed formatting. Ready for review again

nathansamson:
Refactored as per @bacongobbler  comments

nathansamson:
@Joshua-Anderson code is completely different now, please re review

Thanks

bacongobbler:
Would you mind adding unit tests for these changes?:)

nathansamson:
@bacongobbler I should not have refactored; then I did not have to write those pesky unit tests ;)

But fair point, on it already

nathansamson:
@bacongobbler  Can you give me guidance on how to write those tests? (as in what to test for exactly, and where to add the tests)

I am asking as nothing in controller/api/ has any tests yet (although admittedly none of them defines any functions) and for example the method ByTpe has currently no tests either.

Also it feels rather useless to write tests for eg `func (a Apps) Len()`  as it such a trivial method....

Joshua-Anderson:
@nathansamson I think @bacongobbler is looking for tests that show the array of structs being sorted correctly.

nathansamson:
@bacongobbler @Joshua-Anderson I've added an example for apps, is this something you expect?

If so I will apply for the others as well

bacongobbler:
precisely. :)

nathansamson:
Finished more unit tests and rebased.

Ready for review now

Joshua-Anderson:
Thank you so much for all your work! Looks awesome!

nathansamson:
Just to improve my own life :)

Issue #4146, opened 2015-07-29
Potential concurrency issue: unable to bind to 10.164.164.159:6789: (98) Address already in use


carmstrong:
```
ceph-mon: set fsid to 57ec8385-38d7-4b6c-be17-ac82119c8528
ceph-mon: created monfs at /var/lib/ceph/mon/ceph-ci.deis.io for mon.ci.deis.io
2015-07-29 23:08:57.761971 7f0dd561e8c0  0 ceph version 0.94.2 (5fb85614ca8f354284c713a2f9c610860720bbf3), process ceph-mon, pid 1
2015-07-29 23:08:57.794402 7f0dd561e8c0  0 starting mon.ci.deis.io rank 0 at 10.164.164.159:6789/0 mon_data /var/lib/ceph/mon/ceph-ci.deis.io fsid 57ec8385-38d7-4b6c-be17-ac82119c8528
2015-07-29 23:08:57.794484 7f0dd561e8c0 -1 accepter.accepter.bind unable to bind to 10.164.164.159:6789: (98) Address already in use
2015-07-29 23:08:57.794512 7f0dd561e8c0 -1 accepter.accepter.bind was unable to bind. Trying again in 5 seconds 
2015-07-29 23:09:02.794658 7f0dd561e8c0 -1 accepter.accepter.bind unable to bind to 10.164.164.159:6789: (98) Address already in use
2015-07-29 23:09:02.794704 7f0dd561e8c0 -1 accepter.accepter.bind was unable to bind. Trying again in 5 seconds 
2015-07-29 23:09:07.794845 7f0dd561e8c0 -1 accepter.accepter.bind unable to bind to 10.164.164.159:6789: (98) Address already in use
2015-07-29 23:09:07.794890 7f0dd561e8c0 -1 accepter.accepter.bind was unable to bind after 3 attempts: (98) Address already in use
2015-07-29 23:09:07.795950 7f0dd561e8c0 -1 unable to bind monitor to 10.164.164.159:6789/0
```

mboersma:
This is not evident in Deis v1 CI any more.

nathansamson:
isn't it? see https://ci.deis.io/job/test-ec2/4116/console

PR #5074, opened 2016-07-21
chore(*): bump CoreOS to 1068.8.0

nathansamson:
Another try explicitely mention the cgroup driver

See https://github.com/docker/docker/issues/21444 and https://github.com/docker/docker/issues/21678

tested manually and no problems detected but that proved to be not too reliable previously.

PS: sorry for the "spamming" and overloading the CI system but this really seems to be a tricky issue to test otherwise....


nathansamson:
Can someone restart this build; as I am not sure that one successful build proves it works now....

mboersma:
> Can someone restart this build

Actually, if the build succeeds, Jenkins' naginator plugin doesn't show the "Retry" link, and I don't know how to force a new CI run except by `git commit --amend && git push -f`. 😄 I tried resending the GH push hook as well, but Jenkins knew it had seen it before. @vdice any ideas?

@nathansamson I would be happy to do some manual testing on this as well. I agree that we should proceed carefully. Great work though--thanks again.

nathansamson:
Restarted the build myself as suggested.

Today was a public holiday, but tomorrow its back to work for us so normally I would manually test this is well a few times.

nathansamson:
Pretty sure it failed because of CI issues and not the issue at hand, so starting test again...

nathansamson:
Again no gain:

> remote: [8] System error: mkdir /sys/system.slice: operation not permitted
>To ssh://git@deis.test-8e3e7e5c06.deis.works:2222/buildsample.git
> ! [remote rejected] master -> master (pre-receive hook declined)
> error: failed to push some refs to 'ssh://git@deis.test-8e3e7e5c06.deis.works:2222/buildsample.git'
> error at command wait

I am wondering if enabling the cgroup driver on the host might be a solution, or the systemd driver explicetly both on the host and the client

nathansamson:
See also https://github.com/coreos/bugs/issues/1435

vdice:
@mboersma if there is no 'Retry' option, yes, the advised recourse is to amend the commit and force push... That being said, I can help look into adding the 'Retry' option to the job in question, if you haven't done so already?

nathansamson:
@mboersma any idea why the tests keep failing? It doesn't look like anything to do with my latest change, but I don't understand the tests enough to be sure...

Joshua-Anderson:
@nathansamson I restarted the CI build on the chance that's it a sporadic failure (the CI can be flakey with as many moving parts as deis has).

nathansamson:
@Joshua-Anderson I know it can be a bit unstable, but it seems very flaky right now...

Thing is my change actually did not actually do anything (did not check correctly in the first place)

Joshua-Anderson:
Yeah, the same error came up again, so this doesn't look like a sporadic failure. I'm just mystified as you as to the cause though.

nathansamson:
I am giving it one more try, actually applying what I wanted to apply in the first place.

(who thought environment entries in dropin files would have less precedence than the ones in the main file, seems like defeating the point of dropin files...)

nathansamson:
Apparantly not the first time this particular error happened (see #4146)

Any possibility something is wrong on the test cluster and somehow it needs a reset somewhere? (I would not expect there to be any shared state but you never know....)

nathansamson:
The intermittent failures in functional tests seem to have gone for now.

The build succeeded again, I rebased my code (proper commits) in the hope that a second run will actually work as well (although we had a successful run before but that did last long)

nathansamson:
2 success builds in a row...

I think we found the solution, but I guess more manual testing is necessary

nathansamson:
For good measure I started the build a third time

nathansamson:
3 out of 3 success cases.

Anyone not convinced this probably fixes the issue?
In all honesty I even think we can drop the usage of the official docker image, but I would not do so because
A) This currently works, no need to go through another few test cases just because
B) It does not do any harm, and actually makes us much closer to upstream which is always a plus to me

I will probably move my production cluster to this updated cloudconfig file tomorrow and will be doing deploys all day long just to see it keeps working

@mboersma If you can manually test this as well that would be greatly appreciated

nathansamson:
I currently have provisioned a new cluster with the updated cloud config and CoreOS 1068.8.0, but without the new builder image (still the vanilla 1.13.2 image).

Tomorrow I am supposed to do a number of deploys so I will see if I run into any trouble. If I do I will try the same with the updated image as well to see if the combination is needed to make it work

mboersma:
> @mboersma If you can manually test this as well that would be greatly appreciated

Definitely--I'll get on it later today.

nathansamson:
I can report that I deployed about 20 times the past 2 days over several applications without any problem.
This is with the vanilla image but with CoreOS 1068.8.0 and the updated cloud config script. 

As previously mentioned the changes inside builder are not required so it seems.

builder/docker/docker.go
```diff
@@ -72,6 +72,8 @@ func Start(c cookoo.Context, p *cookoo.Params) (interface{}, cookoo.Interrupt) {
 		"192.168.0.0/16",
 		"--insecure-registry",
 		"100.64.0.0/10",
+                "--exec-opt",
```
bacongobbler: can you use tabs instead of spaces here?

bacongobbler:
I'm also a bit concerned about shipping a patch to change the driver from systemd to cgroupfs. Personally I'd rather CoreOS try to resolve this upstream rather than force everyone to use the cgoupfs driver.

Can you also remove the "rebuild" commit? Seems like a commit squash is in order here. :)

builder/docker/docker.go
nathansamson: Done.

Note that this change probably was unneeded anyway, but I left it in to be explicit. If docker decides to changes their default it wont't be impacted (although obviously docker in docker does not get upgraded automatically so this probably won't happen)

nathansamson:
Fixed up and rebased.

Yes I feel what you are saying about changing something global just to workaround a bug.
If you want to wait for CoreOS to fix this bug that is certainly acceptable and not entirely illogical

I sadly could not wait for CoreOS to fix this as I was directly impacted (and with DigitalOcean there is no way to force an older CoreOS version)

Note that I am not sure CoreOS/docker is going to fix this bug, except for changing their default as well in which case (the effect) of this patch will be applied anyway

Note that I feel other people who are deploying new V1 clusters on DO right now will also be impacted; and at least now there is a branch published which they can use

nathansamson:
@bacongobbler @mboersma 
While I certainly understand the hesitation to merge this I want to confirm that after more than a week of usage it did not bring up a single  (new) issue in my cluster.

I also believe a decision should be made rather soon to either merge this; or indeed close this and just wait for a proper fix (eg upstream CoreOS and/or docker)

mboersma:
> after more than a week of usage...

Good to hear. I'm ok with merging this, but I meant to forge upstream and ensure CoreOS' doesn't already plan a fix--and I haven't done that yet. :-( We will get off the fence on this one ASAP, thanks for the nudge @nathansamson.

bacongobbler:
It's been 2.5 weeks with no release, so I think it's time to merge given @nathansamson's testing as well.

PR #5075, opened 2016-07-23
feat(contrib): Add script to restart all apps.

nathansamson:
This is usefull after a reinstall of a cluster with stateless storage.
When the new cluster is started it will still remember the requested
scale of all applications, but all containers will be in "destroyed"
state.
This means that a `deis ps:restart` does not work and so manual
rescaling needs to be done.
This is a tedious and error-prone task if you have many applications,
so this little script takes care of that.

### DIFF
--- a/contrib/util/reset-ps-all-apps.sh
+++ b/contrib/util/reset-ps-all-apps.sh
@@ -1,0 +1,12 @@
+apps=$(curl -H "Authorization: token $DEIS_TOKEN" http://$DEIS_SERVER/v1/apps  | jq -r '.results | map(.id) | join(" ")')
+
+for app in $apps; do
+  echo "Resetting containers of $app"
+
+  originalscale=$(curl -H "Authorization: token $DEIS_TOKEN" http://$DEIS_SERVER/v1/apps/$app/containers/ 2>/dev/null | jq -r '(.results) | [group_by(.type)[] | max_by(.num)] | [map(.type), map(.num)] | transpose | map([.[0], .[1] | tostring] | join("=")) | join(" ")')
+  zeroscale=$(curl -H "Authorization: token $DEIS_TOKEN" http://$DEIS_SERVER/v1/apps/$app/containers/  2>/dev/null | jq -r '(.results) | unique_by(.type) | map([.type, "0"] | join("=")) | join(" ")')
+
+  deis ps:scale $zeroscale -a $app
+  deis ps:scale $originalscale -a $app
+  echo
+done


nathansamson:
I am not sure if something similar is needed in the V2 world, in which case the script needs to be modified a little / lot as the API has changed quite a bit in this area

contrib/util/reset-ps-all-apps.sh
```diff
@@ -0,0 +1,12 @@
+apps=$(curl -H "Authorization: token $DEIS_TOKEN" http://$DEIS_SERVER/v1/apps  | jq -r '.results | map(.id) | join(" ")')
```
bacongobbler: Isn't this just a call to `deis apps:list`?

contrib/util/reset-ps-all-apps.sh
```diff
@@ -0,0 +1,12 @@
...
+
+for app in $apps; do
+  echo "Resetting containers of $app"
+
+  originalscale=$(curl -H "Authorization: token $DEIS_TOKEN" http://$DEIS_SERVER/v1/apps/$app/containers/ 2>/dev/null | jq -r '(.results) | [group_by(.type)[] | max_by(.num)] | [map(.type), map(.num)] | transpose | map([.[0], .[1] | tostring] | join("=")) | join(" ")')
```
bacongobbler: why not call `deis ps:list` here instead?

bacongobbler:
conventiently, you could write a [CLI plugin](http://docs.deis.io/en/latest/customizing_deis/cli-plugins/) for this. For example: 

```
$ cat /bin/deis-reboot
[add logic here]
$ chmod 755 /bin/deis-reboot
$ deis reboot
```

bacongobbler:
there's a command called `deis ps:restart` that will reboot containers for you, FYI.

contrib/util/reset-ps-all-apps.sh
nathansamson: Yes except for the pesky header it includes. I know I could get rid of that with a simple grep -v but I needed to use jq anyways

contrib/util/reset-ps-all-apps.sh
nathansamson: Again that format is not suitable to change to "typeA=N typeB=N". Probably doable with a combination of sed, awk and special magic, but not something doable with my capabilities.

nathansamson:
@bacongobbler See my explanation in commit message why reboot would not work

Unless you feel that is an actual bug that needs to be fixed I and others might need this script for V1->V1 migrations in the future. (eg when coreOS updates and you, I feel its actually easier to setup a new cluster let it boot from shared database and move over my load balancer after everything is alright)

I will probably write a similar script for V1->V2 migrations but then creating apps + config (including domains and certificates) instead of a simple restart

bacongobbler:
Hmm, if they're in destroyed state then yes that is a bug. Is that when you restart just the controller or the entire cluster? Fwiw v2 fixed that problem where you can restart the controller and the pod state is always as it should be.

nathansamson:
This is when I rreinstall the entire cluster = just after a deisctl install platform-stateless with a database configuration which already has applications but where fleet / etcd / docker has never seen my containers before on this cluster.

I generally do this when updating CoreOS and/or when I need to change my cloud config file.
I can reprovision the cluster with only a few manual operations now, and it takes less than an hour. After the new cluster is up and running I just change the floating IP of the load balancers and everything is on the new system right away.

I am btw not sure if this bug is really solvable
A) I prefer containers not to be booted up automatically on a . If you have process types that should only run maximum once(eg crontab like setups), you don't want them to run on both the old and the new cluster automatically.

B) Not restarting containers in the destroyed state also makes sense as this generally happens during a deploy where destroyed is the state of the old release as far as I know. If you would run a restart at the wrong time you might end up with a restarted container with the old version

contrib/util/reset-ps-all-apps.sh
bacongobbler: Personally I'd rather `deis apps:list | grep -v "==="` because then the client can properly handle/display server errors.

bacongobbler:
looks fine to me considering this is in `contrib`, but I suspect this is just working around an existing bug as it should be as simple as calling `deis ps:restart` on all apps. Anyways, LGTM.

PR #5081, opened 2016-08-09
chore(*): update base to alpine:3.4

mboersma:
Updates Docker base images to `alpine:3.4` to benefit from current code and fixes.

Leaves database intact, since upgrading to a new version of postgresql isn't a smooth upgrade path.

Mea culpa: I'm having problems with Go builds locally so I'm tossing this to Jenkins so I can test with the images it builds.

### DIFF
--- a/controller/Dockerfile
+++ b/controller/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:3.3
+FROM alpine:3.4
 
 # install common packages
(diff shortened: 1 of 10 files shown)


bacongobbler:
LGTM assuming CI passes

mboersma:
`openssl` apparently isn't installed by way of `openldap` now in alpine 3.4, so I've added that `apk` package where it is needed.

PR #5096, opened 2016-09-23
fix(models): prune old containers before deploy

mboersma:
In some extreme cases (such as a gunicorn worker timeout), Container objects can be orphaned and eventually accumulate. This cleans up any such stale Containers before starting a deploy operation.


bacongobbler:
Note for anyone else reviewing: this does not require to be ported over to v2 as the Container model has been removed. We now just fetch pod information from the k8s API and deploy using deployments, so this logic is gone.

PR #5099, opened 2016-09-29
fix(client): do not add top level command to cmdArgs

smt116:
## Background

We have a private PaaS with multiple Deis clusters (staging + two productions at the moment). We have also a tool that allows us to manage all clusters from a single web panel. It also connects other APIs such as servers provider and DNS provider. As the result we can perform actions that touches many parts of the system (for example create DNS records for all Deis workers and make CNAME to use with `DEISCTL_TUNNEL` or create a database and assign `DATABASE_URL` for a given application). We have also created plugins to extend Deis CLI. For example there is a support for database (+backups) management and easy SSH access for our developers.

## Problem

We have `deis backups:[create|show|restore|list]` commands and we want to show usage help if user tries `deis backups`. There is a problem in such case because Deis CLI invokes `deis-backups backups`. Without additional support on the plugin side, it will throw an invalid command error.

## Solution

We can solve it on the plugin side to handle unexpected arguments, but it feels not okay. Instead, Deis CLI should skip top level command (`backups` in this case) in `cmdArgv`. It already does that when used as `deis backups:list` or even `deis backups:`.

## Code snippets

The output contains `binary` and `cmdArgv` values using the following patch:

```diff
diff --git a/client/deis.go b/client/deis.go
index a256756..1377817 100644
--- a/client/deis.go
+++ b/client/deis.go
@@ -137,6 +137,10 @@ Use 'git push deis master' to deploy to an application.

                cmdArgv = append(cmdArgv, argv[1:]...)

+               fmt.Fprintf(os.Stdout, "binary: %v\n", binary)
+               fmt.Fprintf(os.Stdout, "cmdArgv: %v\n", cmdArgv)
+               return 1
+
                err = syscall.Exec(binary, cmdArgv, env)
                if err != nil {
                        parser.PrintUsage()
```

### Before

```bash
client λ ./deis backups
binary: /Users/smefju/.rbenv/shims/deis-backups
cmdArgv: [deis-backups backups]

client λ ./deis backups:list
binary: /Users/smefju/.rbenv/shims/deis-backups
cmdArgv: [deis-backups list]

client λ ./deis backups:list:details
binary: /Users/smefju/.rbenv/shims/deis-backups
cmdArgv: [deis-backups list:details]

client λ ./deis backups:list --details
binary: /Users/smefju/.rbenv/shims/deis-backups
cmdArgv: [deis-backups list --details]
```

### After

```bash
client λ ./deis backups
binary: /Users/smefju/.rbenv/shims/deis-backups
cmdArgv: [deis-backups]

client λ ./deis backups:list
binary: /Users/smefju/.rbenv/shims/deis-backups
cmdArgv: [deis-backups list]

client λ ./deis backups:list:details
binary: /Users/smefju/.rbenv/shims/deis-backups
cmdArgv: [deis-backups list:details]

client λ ./deis backups:list --details
binary: /Users/smefju/.rbenv/shims/deis-backups
cmdArgv: [deis-backups list --details]
```

### Difference

Difference is only in the first command's output.

## ToDo

* [x] skip top level command in `cmdArgv`
* [x] add specs

----

I don't have experience in GO language so feel free to propose a better solution. Tests are missing. I couldn't found any valuable materials about how to test `syscall.Exec` properly in such case. It would be great if you could drop some advice or add missing tests.

### DIFF
--- a/client/deis.go
+++ b/client/deis.go
@@ -11,6 +11,8 @@ import (
 	docopt "github.com/docopt/docopt-go"
 )
 
+const extensionPrefix = "deis-"
+
 // main exits with the return value of Command(os.Args[1:]), deferring all logic to
 // a func we can test.
 func main() {
@@ -119,23 +121,14 @@ Use 'git push deis master' to deploy to an application.
 		return 0
 	default:
 		env := os.Environ()
-		extCmd := "deis-" + command
 
-		binary, err := exec.LookPath(extCmd)
+		binary, err := exec.LookPath(extensionPrefix + command)
 		if err != nil {
 			parser.PrintUsage()
 			return 1
 		}
 
-		cmdArgv := []string{extCmd}
-
-		cmdSplit := strings.Split(argv[0], command+":")
-
-		if len(cmdSplit) > 1 {
-			argv[0] = cmdSplit[1]
-		}
-
-		cmdArgv = append(cmdArgv, argv...)
+		cmdArgv := prepareCmdArgs(command, argv)
 
 		err = syscall.Exec(binary, cmdArgv, env)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


bacongobbler:
Nice patch! Would you also be interested in [porting this to v2](https://github.com/deis/workflow-cli/blob/a6cd7e6a4cc73c1999e879cd3e6f6918abedf9c4/deis.go#L169-L173)? If not I can do that, no big deal :)

I wonder if we should refactor this part out in the v2 client and see if we can wrap some unit tests around this logic.

smt116:
@bacongobbler sure, I will port this fix to V2. I will also think about extracting the code to separate function to make it testable.

smt116:
I have added tests to the commit (the same as in https://github.com/deis/workflow-cli/pull/255/files).

bacongobbler:
@mboersma would you mind taking a look at this when you have time?

PR #5105, opened 2016-11-03
docs(layout): Update notice with URL for published v2 docs

krancour:
Previously this was a link to the doc source. A Container
Days conference attendee just pointed out that it would
be more convenient to be linked to the published docs
instead. I think he's right.


bacongobbler:
merging since this is just a docs fix. :+1:

